PR #41, opened 2026-04-04
feat: per-thread isolated working directories

qijie850:
## Summary

Each Discord thread now gets its own working directory under `<working_dir>/sessions/<thread_id>/`, preventing file conflicts when multiple threads run concurrently.

Closes #38

## Changes

**`src/acp/pool.rs`** (only file changed):

1. **`get_or_create`**: creates `<working_dir>/sessions/<thread_id>/` before spawning the agent, passes it as the `cwd` for both `AcpConnection::spawn` and `session_new`
2. **`cleanup_idle`**: removes the per-thread directory when a stale session is reaped

## Behavior

- **Default isolation**: all new sessions get isolated directories automatically, no config flag needed
- **Backward compatible**: existing single-session setups work as before (just with an extra subdirectory level)
- **Cleanup on reap**: when `cleanup_idle` removes a stale session, it also deletes the working directory

## Testing

Tested with `claude-agent-acp` on EC2 (Ubuntu 24.04, t3.large):
- Multiple concurrent threads each get their own workspace ✅
- Files in one thread don't affect another ✅
- Directories cleaned up after TTL expiry ✅

PR #42, opened 2026-04-04
feat: /close slash command to manually terminate a thread session

### LINKED ISSUE #40: feat: /close command to manually terminate a thread session
## Problem

Currently the only way to free a session slot is to wait for TTL expiry or restart agent-broker. Users have no way to manually end a session when they're done with a thread.

This becomes important when `session_ttl_hours` is set high (e.g. 7 days) to preserve context — sessions accumulate and can hit `max_sessions`.

## Proposed Solution

Add a `/close` (and `/end` alias) command that users can type inside a thread to:

1. **Kill the ACP child process** (free RAM)
2. **Remove the session mapping** (if session persistence is implemented, ref #38)
3. **Clean up the working directory** (if per-thread dirs are implemented, ref #38)
4. **Reply with confirmation** in the thread

## Implementation

In `discord.rs`, intercept the command before passing to the ACP agent:

```rust
// In message handler, before get_or_create:
if prompt.trim().eq_ignore_ascii_case("/close") || prompt.trim().eq_ignore_ascii_case("/end") {
    if in_thread {
        let thread_key = msg.channel_id.get().to_string();
        pool.close_session(&thread_key).await?;
        msg.channel_id.say(&ctx.http, "✅ Session closed.").await?;
    }
    return;
}
```

In `pool.rs`, add a `close_session` method:

```rust
pub async fn close_session(&self, thread_id: &str) -> Result<()> {
    let mut conns = self.connections.write().await;
    if conns.remove(thread_id).is_some() {
        info!(thread_id, "closed active session");
    }
    // Optionally clean up working directory and persisted mapping
    Ok(())
}
```

## Additional Considerations

- Could also support `/status` to show active session count / current thread info
- The command should only work inside threads (not in the main channel)
- Related: #38 (per-thread working directories — cleanup on close)

We've tested this locally and it works well. Happy to submit a PR.

qijie850:
## Summary

Adds a Discord slash command `/close` that lets users manually terminate a coding session inside a thread.

Closes #40
Depends on #41 (per-thread working directories)

## Changes

**`src/discord.rs`**:
- Register `/close` as a global slash command on `ready`
- Handle `InteractionCreate` events for the `/close` command
- Validates it's called inside a thread before closing

**`src/acp/pool.rs`**:
- Add `close_session(thread_id)` method that kills the ACP process and removes the per-thread working directory

## Design Decision: Slash Command vs Text Matching

Per @thepagent's feedback on #40, this uses **Discord slash commands** (`InteractionCreate` event) instead of text matching to avoid shadowing agent-side `/close` commands:

```
Text message "/close"  ──> forwarded to ACP agent (agent's command)
Slash command /close   ──> handled by broker (Discord interaction)
```

Zero collision between broker commands and agent commands.

## Behavior

- `/close` in a thread → kills session + cleans workdir → "\u2705 Session closed."
- `/close` with no active session → "\u2139\ufe0f No active session in this thread."
- `/close` outside a thread → "\u26a0\ufe0f Only works inside a thread."

## Testing

Compiled and verified against serenity 0.12 on EC2 (Ubuntu 24.04). Slash command registration and interaction handling tested locally.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@
 config.toml
 *.swp
 .DS_Store
+sessions/
--- a/src/acp/pool.rs
+++ b/src/acp/pool.rs
@@ -55,6 +55,15 @@ impl SessionPool {
             }
             warn!(thread_id, "stale connection, rebuilding");
             conns.remove(thread_id);
+            // Clear the old session directory so the replacement agent starts clean
+            let stale_dir = self.thread_dir(thread_id);
+            drop(conns);
+            match tokio::fs::remove_dir_all(&stale_dir).await {
+                Ok(_) => {}
+                Err(e) if e.kind() == std::io::ErrorKind::NotFound => {}
+                Err(e) => warn!(thread_id, error = %e, "failed to remove stale session directory"),
+            }
+            conns = self.connections.write().await;
         }
 
         if conns.len() >= self.max_sessions {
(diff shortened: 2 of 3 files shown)


thepagent (commented):
Good work using Discord slash commands — exactly the approach we discussed in #40. A few items:

**Must fix:**

1. **Overlapping diff with #41** — This PR includes the same `pool.rs` per-thread dir changes from #41. Please rebase onto #41's branch (or wait for #41 to merge, then rebase). Otherwise we'll get merge conflicts.

2. **Carry over #41 review fixes** — The `pool.rs` changes here have the same issues flagged in #41 review:
   - Path traversal hardening (`thread_id` validation)
   - `shutdown()` directory cleanup
   - sync `exists()` → use `tokio::fs` or just handle `NotFound`
   - Extract `thread_dir()` helper (now duplicated 3 times)

3. **Make response ephemeral** — `/close` response is visible to everyone. Add `.ephemeral(true)` so only the caller sees it:
```rust
CreateInteractionResponseMessage::new()
    .content(response_content)
    .ephemeral(true)
```

4. **Global command registration on every restart** — `create_global_command` in `ready` runs on every bot restart. Global commands have rate limits and take up to 1 hour to propagate. Consider:
   - Use **guild commands** instead (instant propagation, registered per-server)
   - Or gate with a one-time check

**Looks good otherwise:**

- Lock scope in `close_session` is well done (release write lock before async fs cleanup)
- Thread detection via `thread_metadata.is_some()` is correct
- Clean separation between slash command handling and text message flow

Suggested merge order: land #41 first → rebase this PR → address feedback → merge.

chaodu-agent:
Closing this PR — the core pain point (session slot exhaustion) has been addressed by the auto-eviction mechanism added in the 0.8.0 pool rewrite.

<details>
<summary>Context</summary>

When this PR was opened, hitting `max_sessions` returned an error with no recovery path. The only options were waiting for TTL expiry or restarting the broker.

Current main now has:
- **Auto-eviction**: when the pool is full, the oldest idle session is automatically suspended to make room
- **Auto-resume**: when a user returns to an evicted thread, `session/load` restores the conversation context
- **`/cancel`**: cancels the current in-flight turn without killing the session

This covers the original issue #40 use case. The remaining value of `/close` (explicit no-resume semantics + immediate workdir cleanup) is a nice-to-have rather than a must-have.

Additionally, this PR depends on the original PR #41 which has been rebased and redesigned — the `pool.rs` changes here are no longer compatible with current main.

</details>

@qijie850 — thanks for the contribution! Feel free to reopen issue #40 if you find that explicit session termination is still needed in practice. 🙏

chaodu-agent:
## 🔴 Stale — Needs full rewrite against current main

The `/close` slash command concept is valid, but the implementation targets a significantly older codebase:
- Pool architecture mismatch (`connections` vs `PoolState` with `active`/`suspended`/`cancel_handles`)
- Thread key format mismatch (raw ID vs `discord:{id}`)
- Slash command registration pattern changed (shared `commands` vec)

If still wanted, recommend rewriting against current `main` — easier than rebasing given the scope of changes.

🟢 **Worth preserving**: Path traversal validation and ephemeral response pattern.

---
*Reviewed by 超渡法師 🪬*

PR #57, opened 2026-04-05
feat: lightweight management API for session observability and lifecycle control

### LINKED ISSUE #39: feat: lightweight management API for session observability and lifecycle control
## Problem

agent-broker currently has no control plane. Once running, there is no way to:

- Query how many sessions are active, what directories they're working in, or how long they've been idle
- Terminate a single session without restarting the entire process (killing all sessions)
- Expose a health endpoint for orchestrator probes (Kubernetes `livenessProbe` / `readinessProbe`)

This is a gap for any production deployment. The project ships with Helm charts and K8s manifests, but `deployment.yaml` has no health probe configured because there's nothing to probe. If the Discord gateway disconnects or the process enters a bad state, the pod won't be restarted.

For multi-session deployments, there's no visibility into resource usage per session. The only management option is `kill 1` / container restart, which tears down every active session — unacceptable when some sessions have long-running work in progress.

## Proposed Solution

Add an optional lightweight HTTP listener (no framework dependency) configured via a new `[management]` section in `config.toml`:

```toml
[management]
enabled = true
bind = "0.0.0.0:8090"
# Optional: bearer token for non-localhost deployments
# auth_token = "${MANAGEMENT_TOKEN}"
```

### MVP Endpoints

**`GET /healthz`** — Orchestrator probe target

```json
{ "status": "ok", "uptime_seconds": 3600, "discord_connected": true }
```

**`GET /sessions`** — Session inventory

```json
{
  "active_sessions": 2,
  "max_sessions": 10,
  "sessions": [
    {
      "thread_id": "1234567890",
      "alive": true,
      "idle_seconds": 45
    }
  ]
}
```

**`DELETE /sessions/{thread_id}`** — Graceful single-session teardown (child process killed via `kill_on_drop`)

```json
{ "killed": "1234567890" }
```

**`DELETE /sessions`** — Terminate all sessions (equivalent to pool shutdown without stopping the broker)

### Why this approach

**Why not Discord slash commands?** Slash commands mix the control plane with the data plane. Management operations (kill a session, check health) are infrastructure concerns, not user-facing features. They should be accessible to orchestrators and scripts, not just Discord users.

**Why not axum/actix?** The project's philosophy is minimal dependencies and a single binary. A raw `tokio::net::TcpListener` with basic HTTP parsing keeps the dependency tree unchanged while covering the MVP. If the management API grows beyond these endpoints (e.g., Prometheus `/metrics`), migrating to a lightweight framework would make sense as a follow-up.

**Why config opt-in?** Not every deployment needs a management port. Bare-metal single-user setups may prefer zero attack surface. Defaulting to `enabled = false` keeps backward compatibility.

### Future extensions (out of scope for MVP)

- `GET /metrics` — Prometheus exposition format (session count gauge, prompt latency histogram)
- Webhook notifications on session create/destroy
- `POST /sessions` — Programmatic session creation (for bot-to-broker orchestration)

thepagent:
## Summary

Add an optional lightweight HTTP management server (zero new dependencies) for session observability and lifecycle control.

### Endpoints

| Method | Path | Description |
|--------|------|-------------|
| `GET` | `/healthz` | Uptime + discord connection status (K8s probe target) |
| `GET` | `/sessions` | List active sessions with idle time and alive status |
| `DELETE` | `/sessions/{thread_id}` | Graceful single-session teardown |
| `DELETE` | `/sessions` | Terminate all sessions |

### Config

Opt-in via `[management]` section in `config.toml`:

```toml
[management]
enabled = true
bind = "0.0.0.0:8090"
```

Defaults to `enabled = false` — no behavior change for existing deployments.

### Changes

- **`src/config.rs`** — `ManagementConfig` struct (enabled, bind)
- **`src/acp/pool.rs`** — `list_sessions()`, `remove_session()`, `remove_all_sessions()`, `max_sessions()`
- **`src/discord.rs`** — `discord_connected` `AtomicBool` set on `ready()`
- **`src/management.rs`** — Raw TCP HTTP server (`tokio::net::TcpListener`, no framework)
- **`src/main.rs`** — Wire management server, shared state
- **`config.toml.example`** — `[management]` section
- **`k8s/deployment.yaml`** — `livenessProbe` + `readinessProbe` on port 8090
- **Helm chart** — `management.enabled`/`management.bind` in values, configmap, deployment template

### Design decisions

- **No framework deps** — raw `TcpListener` keeps the dependency tree unchanged
- **Config opt-in** — zero attack surface for deployments that don't need it
- **`AtomicBool` for discord status** — lightweight, lock-free signaling from serenity handler

Closes #39

### DIFF
--- a/charts/agent-broker/templates/configmap.yaml
+++ b/charts/agent-broker/templates/configmap.yaml
@@ -22,6 +22,10 @@ data:
     max_sessions = {{ .Values.pool.maxSessions }}
     session_ttl_hours = {{ .Values.pool.sessionTtlHours }}
 
+    [management]
+    enabled = {{ .Values.management.enabled }}
+    bind = "{{ .Values.management.bind }}"
+
     [reactions]
     enabled = {{ .Values.reactions.enabled }}
     remove_after_reply = {{ .Values.reactions.removeAfterReply }}
--- a/charts/agent-broker/templates/deployment.yaml
+++ b/charts/agent-broker/templates/deployment.yaml
@@ -44,6 +44,24 @@ spec:
           resources:
             {{- toYaml . | nindent 12 }}
           {{- end }}
+          {{- if .Values.management.enabled }}
+          ports:
+            - name: management
+              containerPort: 8090
+              protocol: TCP
+          livenessProbe:
+            httpGet:
+              path: /healthz
+              port: management
+            initialDelaySeconds: 10
+            periodSeconds: 15
+          readinessProbe:
+            httpGet:
+              path: /healthz
+              port: management
+            initialDelaySeconds: 5
+            periodSeconds: 10
+          {{- end }}
           volumeMounts:
             - name: config
               mountPath: /etc/agent-broker
--- a/charts/agent-broker/values.yaml
+++ b/charts/agent-broker/values.yaml
@@ -34,6 +34,10 @@ pool:
   maxSessions: 10
   sessionTtlHours: 24
 
+management:
+  enabled: false
+  bind: "0.0.0.0:8090"
+
 reactions:
   enabled: true
   removeAfterReply: false
--- a/config.toml.example
+++ b/config.toml.example
@@ -29,6 +29,10 @@ working_dir = "/home/agent"
 max_sessions = 10
 session_ttl_hours = 24
 
+[management]
+enabled = false
+bind = "0.0.0.0:8090"
+
 [reactions]
 enabled = true
 remove_after_reply = false
--- a/k8s/deployment.yaml
+++ b/k8s/deployment.yaml
@@ -28,6 +28,22 @@ spec:
                   key: discord-bot-token
             - name: HOME
               value: /home/agent
+          ports:
+            - name: management
+              containerPort: 8090
+              protocol: TCP
+          livenessProbe:
+            httpGet:
+              path: /healthz
+              port: management
+            initialDelaySeconds: 10
+            periodSeconds: 15
+          readinessProbe:
+            httpGet:
+              path: /healthz
+              port: management
+            initialDelaySeconds: 5
+            periodSeconds: 10
           volumeMounts:
             - name: config
               mountPath: /etc/agent-broker
--- a/src/acp/pool.rs
+++ b/src/acp/pool.rs
@@ -80,6 +80,33 @@ impl SessionPool {
         f(conn).await
     }
 
+    pub fn max_sessions(&self) -> usize {
+        self.max_sessions
+    }
+
+    pub async fn list_sessions(&self) -> Vec<(String, bool, u64)> {
+        let conns = self.connections.read().await;
+        conns
+            .iter()
+            .map(|(id, c)| {
+                let idle = c.last_active.elapsed().as_secs();
+                (id.clone(), c.alive(), idle)
+            })
+            .collect()
+    }
+
+    pub async fn remove_session(&self, thread_id: &str) -> bool {
+        let mut conns = self.connections.write().await;
+        conns.remove(thread_id).is_some()
+    }
+
+    pub async fn remove_all_sessions(&self) -> usize {
+        let mut conns = self.connections.write().await;
+        let count = conns.len();
+        conns.clear();
+        count
+    }
+
     pub async fn cleanup_idle(&self, ttl_secs: u64) {
         let cutoff = Instant::now() - std::time::Duration::from_secs(ttl_secs);
         let mut conns = self.connections.write().await;
--- a/src/config.rs
+++ b/src/config.rs
@@ -11,7 +11,25 @@ pub struct Config {
     pub pool: PoolConfig,
     #[serde(default)]
     pub reactions: ReactionsConfig,
+    #[serde(default)]
+    pub management: ManagementConfig,
 }
+
+#[derive(Debug, Deserialize)]
+pub struct ManagementConfig {
+    #[serde(default)]
+    pub enabled: bool,
+    #[serde(default = "default_management_bind")]
+    pub bind: String,
+}
+
+impl Default for ManagementConfig {
+    fn default() -> Self {
+        Self { enabled: false, bind: default_management_bind() }
+    }
+}
+
+fn default_management_bind() -> String { "0.0.0.0:8090".into() }
 
 #[derive(Debug, Deserialize)]
 pub struct DiscordConfig {
--- a/src/discord.rs
+++ b/src/discord.rs
@@ -8,6 +8,7 @@ use serenity::model::gateway::Ready;
 use serenity::model::id::{ChannelId, MessageId};
 use serenity::prelude::*;
 use std::collections::HashSet;
+use std::sync::atomic::{AtomicBool, Ordering};
 use std::sync::Arc;
 use tokio::sync::watch;
 use tracing::{error, info};
@@ -16,6 +17,7 @@ pub struct Handler {
     pub pool: Arc<SessionPool>,
     pub allowed_channels: HashSet<u64>,
     pub reactions_config: ReactionsConfig,
+    pub discord_connected: Arc<AtomicBool>,
 }
 
 #[async_trait]
@@ -152,6 +154,7 @@ impl EventHandler for Handler {
     }
 
     async fn ready(&self, _ctx: Context, ready: Ready) {
+        self.discord_connected.store(true, Ordering::Relaxed);
         info!(user = %ready.user.name, "discord bot connected");
     }
 }
--- a/src/main.rs
+++ b/src/main.rs
@@ -2,12 +2,15 @@ mod acp;
 mod config;
 mod discord;
 mod format;
+mod management;
 mod reactions;
 
 use serenity::prelude::*;
 use std::collections::HashSet;
 use std::path::PathBuf;
+use std::sync::atomic::AtomicBool;
 use std::sync::Arc;
+use tokio::time::Instant;
 use tracing::info;
 
 #[tokio::main]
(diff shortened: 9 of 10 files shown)


ruan330:
Thanks for pushing this forward — we've been running a very similar always-on management API in production (bare metal + Docker) for the past few days, so wanted to share what we've hit. Happy to rebase onto this design once it lands.

**1. `/healthz` as both liveness and readiness will flap.**

The serenity Discord gateway reconnects its WebSocket periodically (roughly every ~30 minutes, sooner under packet loss). During the reconnect, `discord_connected` briefly flips to `false`. If K8s uses `/healthz` as **liveness** and expects `discord_connected: true`, the kubelet will restart the pod every reconnect cycle — which then drops every active session.

Our Docker healthcheck runs into the same class of issue for a different reason (`/status` returns 200 even when Discord is disconnected, so we can't distinguish "process alive but idle" from "process alive and serving").

Suggested split:
- `/livez` — process alive + HTTP server responsive. Never checks Discord. Liveness probe target.
- `/readyz` — `discord_connected == true` AND pool not exhausted. Readiness probe target.

The current PR's `livenessProbe` + `readinessProbe` both pointing at `/healthz` would lead to restart storms under flaky networks. We'd recommend different paths in `k8s/deployment.yaml`.

**2. `DELETE /sessions` (bulk) — is the use case worth the risk?**

Genuine question: what's the production scenario? We've never needed it. Our `/kill/{thread_id}` is used in two cases:
- `flutter run` dev servers that pass `alive()` but block streaming → per-thread kill, user unaffected
- Occasional stuck ACP state during debugging → specific thread

Bulk delete would wipe every active user conversation indiscriminately. If someone hits `DELETE /sessions` by accident (`curl`, browser extension, typo), there's no recovery — Discord threads lose their agent state and users get silent sessions.

If there's a real use case (e.g. before broker upgrade), a `?confirm=true` query param or an `admin_token` gate would dramatically reduce blast radius. Or drop it entirely and rely on individual deletes — we've never missed it.

**3. `GET /sessions` — worth reserving a `cwd` field.**

Our `/sessions` returns an extra `cwd` field per session (working directory the agent was spawned with). In production debugging, the most common question is "which project is this stuck thread in?" — `thread_id + alive + idle_seconds` alone requires cross-referencing Discord channel metadata.

This is currently fork-specific (our `[cwd:]` directive), but once §2a lands `{base_working_dir}/{thread_id}/`, every session will have a meaningful `cwd` to surface. Reserving the field now would save a schema version bump later.

**4. Default bind + auth.**

`bind = "0.0.0.0:8090"` is an opinionated choice — perfect for bare metal in a trusted network (which is our setup), but surprising for a K8s deployment without NetworkPolicy.

Options we'd consider, ordered by disruption:
- Default to `127.0.0.1:8090` (backward incompatible for existing always-on users, so probably not)
- Keep `0.0.0.0:8090` but document the assumption in `config.toml.example`
- Optional `[management] auth_token = "${MGMT_TOKEN}"` — skip validation when empty, require `Authorization: Bearer` when set. Two lines of diff, keeps the simple-deployment story.

We don't feel strongly; the token option is what we'd add if this were being shipped to users we don't control.

---

None of these are blockers — the core approach (raw TCP, opt-in, zero deps) matches what we independently arrived at, and we'd love to drop our inline implementation in favor of this once it lands. Let me know if any of these would be useful as follow-up PRs vs folded in here.

chaodu-agent:
## Community Triage Review — 超渡法師 + 普渡法師

**Verdict: Changes required before merge**

The feature is needed — main has no management API, no K8s health probes, and no session observability. The core design (raw TCP, opt-in, zero deps) is solid and independently validated by production experience from other deployments. However, there are compile-breaking and production-breaking issues that need to be addressed first.

### 🔴 Suggested Changes

<details>
<summary><b>[1] pool.rs field name conflict — needs rebase (compile error)</b></summary>

The diff references `self.connections.read().await`, but main uses `self.state: RwLock<PoolState>`. This PR was written against an older pool.rs structure. Direct merge will fail to compile. Needs rebase and update to match current `self.state` API.
</details>

<details>
<summary><b>[2] k8s/deployment.yaml adds probes unconditionally (restart loop when management disabled)</b></summary>

The Helm chart correctly wraps probes in `{{- if .Values.management.enabled }}`, but the raw `k8s/deployment.yaml` adds liveness/readiness probes unconditionally. When `management.enabled = false` (the default), port 8090 has no listener → probes fail forever → pod restart loop.

Fix: wrap in the same conditional, or remove probes from the raw manifest and only include them in the Helm chart.
</details>

<details>
<summary><b>[3] /healthz as both liveness and readiness — restart storm on Discord reconnect</b></summary>

As @ruan330 noted from production experience: serenity's Discord gateway reconnects its WebSocket roughly every ~30 minutes. During reconnect, `discord_connected` briefly flips to `false`. If kubelet uses `/healthz` as liveness probe, it will restart the pod — killing all active sessions. This is independently validated by both OpenClaw and Hermes Agent architectures (see PR #302 research comment).

Suggested split:
- `/livez` — process alive + HTTP responsive, never checks Discord → liveness target
- `/readyz` — `discord_connected == true` → readiness target (only affects traffic routing, never triggers pod restart)
</details>

### 🟡 NIT

<details>
<summary><b>[4] DELETE /sessions bulk delete — no blast-radius protection</b></summary>

A single mistyped `curl -X DELETE` wipes every active session with no recovery. As @ruan330 noted, the production use case is almost always per-thread kill. Suggested: add `?confirm=true` query param, or drop bulk delete entirely and keep only per-thread `DELETE /sessions/{thread_id}`.
</details>

<details>
<summary><b>[5] 0.0.0.0:8090 bind + no auth — undocumented security assumption</b></summary>

With `0.0.0.0`, any pod in the same namespace can call `DELETE /sessions`. In a cluster without NetworkPolicy this is a footgun. Suggested: document the security assumption in `config.toml.example`, or add optional `auth_token` config (OpenClaw pattern: skip validation on loopback, require `Authorization: Bearer` when token is set — minimal diff).
</details>

<details>
<summary><b>[6] remove_session / remove_all_sessions bypass graceful teardown</b></summary>

Both methods do a raw `conns.clear()` / `conns.remove()`, relying on Arc Drop to kill the process group. If the Arc has other holders (e.g. a task currently handling a request), the process won't die immediately. Acceptable for now, but worth a comment marking the assumption.
</details>

<details>
<summary><b>[7] read_line has no size limit</b></summary>

The HTTP request parser uses `buf_reader.read_line(&mut request_line)` without a size cap. A malformed or malicious request could exhaust memory. Add a reasonable cap (e.g. 8 KB) before the `splitn` parse.
</details>

### ✅ What's good

- Zero new dependencies (raw `TcpListener`) — matches project philosophy
- `enabled = false` default — zero attack surface for existing deployments
- `AtomicBool` for Discord status is the right call (lock-free, only needs eventual visibility)
- Helm chart conditional is correct; the Helm path is production-ready
- Error handling in `serve()` loop is solid: bind failure returns, accept error continues
- @ruan330's production feedback validates the core approach independently

### Summary

| # | Issue | Severity | Status |
|---|---|---|---|
| 1 | pool.rs field name conflict | 🔴 compile error | needs rebase |
| 2 | k8s/deployment.yaml unconditional probes | 🔴 restart loop | needs fix |
| 3 | `/healthz` as liveness + readiness | 🔴 restart storm | split to `/livez` + `/readyz` |
| 4 | Bulk DELETE no protection | 🟡 blast radius | add `?confirm=true` or drop |
| 5 | `0.0.0.0` no auth docs | 🟡 security | add comment or optional token |
| 6 | Graceful teardown bypassed | 🟡 assumption | add code comment |
| 7 | `read_line` no size limit | 🟡 robustness | cap at 8 KB |

thepagent:
@ruan330  feel free to continue. Let me know when you are ready for review.

chaodu-agent:
## 四法師 Collaborative Triage Review — PR #57 (2026-05-01)

**Verdict: CHANGES_REQUESTED** — Feature is net-new and valuable (no management API exists on `main`), but the PR is based on an extremely old codebase and needs a complete rewrite against current `main`.

<details>
<summary>🟢 INFO — What looks good</summary>

- **Feature is still needed**: Current `main` has no health endpoint, no session listing, no management API. K8s probes currently rely on `pgrep` which requires `procps`.
- **Design is sound**: opt-in via `[management]` config, zero new dependencies (raw `TcpListener`), `AtomicBool` for discord status.
- **Endpoints are well-chosen**: `/healthz` (K8s probes), `/sessions` (observability), `DELETE /sessions/{id}` (lifecycle control).
- **P1 label** — this was flagged as high priority.

</details>

<details>
<summary>🔴 SUGGESTED CHANGES — Blocking</summary>

### 1. Stale branch — complete rewrite needed

The PR references `charts/agent-broker/` (renamed to `charts/openab/` months ago). The codebase has changed fundamentally since this PR was opened:

- **Pool architecture**: `main` now uses `SessionEntry` struct, `RwLock<PoolState>` with `sessions` HashMap, session persistence (`mapping_path`/`save_mapping`), cancel handles. PR uses the original `connections: RwLock<HashMap<String, AcpConnection>>`.
- **Multi-adapter**: `main` has `ChatAdapter` trait, `AdapterRouter`, Slack adapter, LINE adapter. PR is Discord-only.
- **Signal handling**: `main` has `shutdown_signal()`, cron, markdown, gateway modules.
- **Helm chart**: completely restructured (multi-agent, per-agent config, RBAC, PDB, etc.).

### 2. Rewrite scope

The management API concept should be reimplemented from scratch on current `main`:
- `list_sessions()` needs to work with `PoolState.sessions` (not `connections`)
- Health endpoint should reflect multi-adapter status (Discord + Slack + LINE), not just Discord
- Helm chart integration needs the current `charts/openab/` structure
- Consider whether `axum` (already a dependency via LINE adapter #521) should replace raw `TcpListener`

### 3. Missing Discord Discussion URL

PR description has no Discord Discussion URL — required per contribution guidelines.

</details>

<details>
<summary>📋 Recommendation</summary>

**Option A (preferred)**: Close this PR, open a fresh issue capturing the management API requirements, and implement against current `main`. The original design (endpoints, opt-in config) is still valid — only the implementation needs to be redone.

**Option B**: Contributor rewrites the entire PR against current `main`. Given the scope of changes needed, this is effectively a new PR.

</details>

---

Reviewed by 超渡法師 on behalf of the 四法師 triage team.

<@1493128125402320996> <@1496097857940361326> <@1496553369442189472> — 法師們，PR #57 management API，net-new feature 但基於極舊 codebase，我建議 close + 重開。請補充你們的觀點。

Issue #76, opened 2026-04-06
fix: notification loop assumes ordered events, bounded prompts, and managed session lifecycle — none hold in production


ruan330:
## Problem

The notification loop in `stream_prompt` (`discord.rs`) is built on three assumptions that don't hold in production with long-running ACP backends like Claude Code:

### Assumption 1: ACP events arrive in order

The loop breaks when it sees a response with `id`:

```rust
while let Some(notification) = rx.recv().await {
    if notification.id.is_some() {
        break;  // assumes all text chunks arrived before this
    }
}
```

**Reality**: `end_turn` (the response with `id`) sometimes arrives *before* the final `agent_message_chunk` notifications. When this happens, the loop breaks early and `text_buf` is empty — the user sees "(no response)" even though the agent did respond.

Evidence from production logs:
```
06:31:39  end_turn id=40, totalTokens=0     ← response arrives
06:31:56  agent_message_chunk: "先"          ← text arrives AFTER response
06:31:57  agent_message_chunk: "更新實作..."  ← more text, already discarded
```

### Assumption 2: Prompts always complete

The loop has no exit condition other than receiving the response `id` or the channel closing:

```rust
while let Some(notification) = rx.recv().await {
    // no timeout, no liveness check
}
```

**Reality**: Agents run long tool calls (build commands, test suites) that produce no ACP notifications for minutes. If the tool call never completes (e.g., `flutter run` is a long-lived app server), the loop blocks forever. Combined with the global pool write lock (#58), this freezes the entire broker.

Evidence: a single `flutter run` command blocked all sessions for 7+ hours until manual restart.

### Assumption 3: Session lifecycle is self-managing

There is no cleanup of Discord threads when the broker restarts. After a restart, the in-memory session pool is empty, but Discord threads from previous sessions remain active. Users typing in these stale threads create new sessions with no conversation context, silently consuming session pool slots.

With `max_sessions = 5`, three stale threads + two new threads = pool exhausted.

## These are one problem, not three

All three stem from the same architectural gap: **the notification loop has no resilience against real-world conditions** — event ordering violations, unbounded blocking, and lifecycle mismatches.

Fixing them individually produces three localized patches that don't address the underlying fragility. Fixing them together produces a robust notification loop that handles production realities.

## Reference implementation

We encountered all three issues in our production deployment (multi-session, Claude Code backend, long-running Flutter builds) and submitted a PR with our fixes:

[...]

PR #77, opened 2026-04-06
fix: robust notification loop — per-connection locking, alive check, drain window, and startup cleanup

### LINKED ISSUE #58: fix: pool write lock held during entire prompt streaming causes cross-session deadlock

ruan330:
## Problem

The notification loop in `stream_prompt` is built on three assumptions that don't hold in production with long-running ACP backends like Claude Code (#76):

1. **ACP events arrive in order** — `end_turn` sometimes arrives before the final `agent_message_chunk`, causing empty responses
2. **Prompts always complete** — long tool calls (build commands, test suites) block the loop forever, and with the global pool write lock, freeze the entire broker
3. **Session lifecycle is self-managing** — after restart, stale Discord threads create zombie sessions that consume pool slots

These are three symptoms of one root cause: the notification loop has no resilience against real-world conditions.

## Solution

A single cohesive fix addressing all three, not three separate patches:

### Per-connection locking (pool.rs)

`with_connection()` holds a global write lock for the entire streaming duration. One busy session blocks all others (#58).

Replace `HashMap<String, AcpConnection>` with `HashMap<String, Arc<Mutex<AcpConnection>>>`. Pool `RwLock` is now only held for brief lookups. Each connection has its own `Mutex`.

```rust
// Before: global write lock held for minutes
pool.with_connection(thread_key, |conn| { /* streaming */ })

let conn_arc = pool.get_connection(thread_key).await?;
let mut conn = conn_arc.lock().await;  // only locks THIS connection
```

### Alive check + hard timeout (discord.rs)

Replace bare `while let rx.recv()` with `tokio::select!`:

- Every 30s: check if agent process is alive. Dead → break immediately
- After 30 minutes: hard timeout safety net. Alive but stuck → break

Correctly handles long tool calls (process alive → wait) without false timeouts, while preventing infinite blocking.

### Drain window (discord.rs)

After receiving `end_turn`, drain the notification channel for 200ms to capture late-arriving text chunks:

```rust
if notification.id.is_some() {
    let drain_until = Instant::now() + Duration::from_millis(200);
    while let Ok(Some(n)) = timeout_at(drain_until, rx.recv()).await {
        if let Some(AcpEvent::Text(t)) = classify_notification(&n) {
            text_buf.push_str(&t);
        }
    }
    break;
}
```

### Empty response fallback (discord.rs)

If `text_buf` is empty after draining but tool activity was recorded, compose a fallback from tool lines instead of showing "(no response)".

### Startup thread cleanup (discord.rs, ready handler)

On startup, archive all active threads in allowed channels created by this bot. Prevents stale threads from spawning zombie sessions after restart.

## Tradeoffs

| Decision | Cost | Why |
|----------|------|-----|
| 200ms drain window | Adds 200ms to every prompt completion | Small; avoids losing entire responses |
| 30-min hard timeout | Very long tasks get interrupted | Safety net; should not trigger in normal use |
| Auto-archive on startup | Can't resume old threads | Old threads have no session context anyway |
| `Arc<Mutex>` vs checkout pattern | Slightly more memory per connection | Correct abstraction; enables future prompt queueing via `try_lock()` |

## Testing

Tested with `claude-agent-acp` backend, concurrent sessions:

- Two threads running simultaneously — no cross-session blocking ✅
- Long tool call (flutter build, 5+ min) — other sessions respond normally ✅
- Process crash — detected within 30s, session released ✅
- Stale threads after restart — auto-archived on startup ✅
- ACP event ordering violation — text captured via drain window ✅

## Changes

| File | Lines | What |
|------|-------|------|
| `pool.rs` | +54/-43 | `Arc<Mutex<AcpConnection>>`, `get_connection()`, remove `with_connection()` |
| `discord.rs` | +145/-90 | Per-connection lock, alive check, hard timeout, drain window, fallback, startup cleanup |

Supersedes #59. Closes #58. Ref #76.

### DIFF
--- a/src/acp/pool.rs
+++ b/src/acp/pool.rs
@@ -2,12 +2,13 @@ use crate::acp::connection::AcpConnection;
 use crate::config::AgentConfig;
 use anyhow::{anyhow, Result};
 use std::collections::HashMap;
-use tokio::sync::RwLock;
+use std::sync::Arc;
+use tokio::sync::{Mutex, RwLock};
 use tokio::time::Instant;
 use tracing::{info, warn};
 
 pub struct SessionPool {
-    connections: RwLock<HashMap<String, AcpConnection>>,
+    connections: RwLock<HashMap<String, Arc<Mutex<AcpConnection>>>>,
     config: AgentConfig,
     max_sessions: usize,
 }
@@ -22,24 +23,27 @@ impl SessionPool {
     }
 
     pub async fn get_or_create(&self, thread_id: &str) -> Result<()> {
-        // Check if alive connection exists
+        // Check if alive connection exists (read lock only)
         {
             let conns = self.connections.read().await;
-            if let Some(conn) = conns.get(thread_id) {
+            if let Some(conn_arc) = conns.get(thread_id) {
+                let conn = conn_arc.lock().await;
                 if conn.alive() {
                     return Ok(());
                 }
             }
         }
 
-        // Need to create or rebuild
+        // Need to create or rebuild (write lock)
         let mut conns = self.connections.write().await;
 
         // Double-check after acquiring write lock
-        if let Some(conn) = conns.get(thread_id) {
+        if let Some(conn_arc) = conns.get(thread_id) {
+            let conn = conn_arc.lock().await;
             if conn.alive() {
                 return Ok(());
             }
+            drop(conn);
             warn!(thread_id, "stale connection, rebuilding");
             conns.remove(thread_id);
         }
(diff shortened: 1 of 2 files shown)


ruan330:
Closing this in favor of 3 smaller, focused PRs — on reflection this one bundled too many concerns for a clean review. Sorry for the churn.

The split plan:

1. **Per-connection `Arc<Mutex<AcpConnection>>`** — just the pool.rs architecture change + discord.rs call sites. This implements the locking approach discussed in RFC #78 §2b, and closes #58 / supersedes #59.
2. **Notification loop resilience** — drain window + empty-response fallback (`end_turn` can arrive before the last `agent_message_chunk` via `tokio::select!`). Fixes #76, which we have a production repro for. ~20-30 lines.
3. **Alive check + hard timeout** — defensive safety net. 30s alive check + 30min hard timeout via `tokio::select!`.

**Dropped from original scope:** startup thread cleanup. We reversed that decision in our fork — archiving active threads on restart loses user context, and with per-thread state persistence (related to RFC #78 Phase 2) there's no longer a correctness reason to do it.

I'll send them in order, starting with #1 soon. Happy to discuss any of these beforehand if you'd prefer a different approach.

Issue #93, opened 2026-04-07
feat: Slack adapter — bring agent-broker to Slack workspaces


chaodu-agent:
## Summary

Add a Slack adapter to agent-broker, allowing teams to interact with ACP-compatible agents (Kiro CLI, Claude Code, Codex, Gemini) directly from Slack channels and threads.

## Motivation

- agent-broker currently supports Discord only (via `serenity`), with a Telegram adapter proposed in #86
- Many engineering teams use Slack as their primary communication tool
- Supporting Slack would significantly expand the user base and enable enterprise adoption

## Proposed Architecture

```
┌──────────────┐  Slack Events API  ┌──────────────┐  ACP stdio    ┌──────────────┐
│   Slack      │◄──────────────────►│ agent-broker │──────────────►│  coding CLI  │
│   User       │                    │  (Rust)      │◄── JSON-RPC ──│  (acp mode)  │
└──────────────┘                    └──────────────┘               └──────────────┘
```

### Adapter Abstraction

To support multiple platforms cleanly, agent-broker should introduce a `ChatAdapter` trait:

```rust
#[async_trait]
trait ChatAdapter {
    async fn start(&self, pool: Arc<SessionPool>) -> Result<()>;
    async fn send_message(&self, channel: &str, content: &str) -> Result<()>;
    async fn edit_message(&self, channel: &str, msg_id: &str, content: &str) -> Result<()>;
    async fn create_thread(&self, channel: &str, title: &str) -> Result<String>;
    async fn add_reaction(&self, channel: &str, msg_id: &str, emoji: &str) -> Result<()>;
}
```

Discord, Telegram (#86), and Slack would each implement this trait. The session pool and ACP layer remain unchanged.

### Slack-Specific Design

| Feature | Slack Implementation |
|---------|---------------------|
| Trigger | `@mention` in allowed channels |
| Threading | Slack native threads (reply_broadcast optional) |
| Streaming | Edit message in thread every ~1.5s (same as Discord) |
| Reactions | Slack emoji reactions for status (👀→🤔→🔥→✅) |
| Auth | Slack Bot Token + App-Level Token (Socket Mode) |
| Channel allowlist | Slack channel IDs in `allowed_channels` |

### Config Example

```toml
[slack]
bot_token = "${SLACK_BOT_TOKEN}"
app_token = "${SLACK_APP_TOKEN}"       # for Socket Mode
allowed_channels = ["C1234567890"]

[agent]
command = "kiro-cli"
[...]

PR #86, opened 2026-04-06
feat: Telegram adapter — session lifecycle, memory compaction & crash recovery

henrieopenclaw:
Supersedes #60. Full Telegram adapter replacing Discord, plus team/personal chat modes.

## Architecture

```mermaid
graph LR
    U(["👤 Telegram User"]) <-->|Bot API| AB["agent-broker (Rust)"]
    AB <-->|"ACP stdio JSON-RPC"| CLI["kiro-cli / claude / codex / gemini"]
```

## Chat Modes

| | Personal (default) | Team |
|---|---|---|
| Any message in `#general` | Auto-creates topic, replies | Silent buffer 👀 |
| `@mention` in `#general` | Auto-creates topic, replies | Replies in-place |
| `!kiro` in `#general` | Auto-creates topic | Creates topic (restricted to `topic_creator_id`) |
| Plain message in topic | Replies directly | Silent buffer 👀 |
| `@mention` in topic | Replies directly | Replies directly |

## Session Lifecycle

```mermaid
flowchart TD
    M([User message]) --> GC[get_or_create]
    GC -->|alive| USE[use existing session]
    GC -->|dead/new| SP[spawn CLI]
    SP --> SN[session/new or session/load]
    SN --> PC{pending_context?}
    PC -->|yes| PP[prepend summary]
    PC -->|no| ST
    PP --> ST[stream_prompt]
    ST --> PD[prompt_done]
    PD --> CI[cleanup_idle every 15min]
    CI --> IT{idle > TTL?}
    IT -->|yes| CO[compact → evict → notify]
    CO --> SS[summary stored, injected on resume]
```

### Idle Eviction

```mermaid
flowchart TD
    A([User idle for 2 hours]) --> B[cleanup_idle detects idle > TTL]
    B --> C[Send compaction prompt to live kiro process]
    C --> D[Receive summary → store in summaries]
    D --> E[Kill kiro process, notify user ⏱]
    E --> F([User sends next message])
    F --> G[Spawn new kiro → inject summary]
    G --> H[Kiro remembers previous conversation ✅]
```

### Crash Recovery

```mermaid
flowchart TD
    A([User sends message]) --> B[record_user_message → VecDeque max 20]
    B --> C{... kiro crashes ...}
    C --> D[cleanup_idle detects !alive]
    D --> E[Attempt compaction → fails]
    E --> F[Fallback: VecDeque → Recent messages before crash]
    F --> G[Store in summaries]
    G --> H([User sends next message])
[...]

PR #94, opened 2026-04-07
docs: ADR for Multi-Platform Adapter Architecture

chaodu-agent:
## Summary

Architecture Decision Record for the multi-platform adapter architecture, reformatted from RFC 002 into the standard ADR format under `docs/adr/`.

## What this covers

- **ChatAdapter trait**: platform-agnostic abstraction for Discord, Slack, Telegram, and future adapters
- **AdapterRouter**: shared routing logic extracted from platform-specific code
- **Simultaneous multi-platform (Contract B)**: single instance serving Discord + Slack concurrently, validated by #259
- **Platform-agnostic references**: `ChannelRef`, `MessageRef`, `SenderContext` supporting two threading models
- **Config design**: TOML-based per-platform sections with shared pool
- **Security**: secure-by-default allowlists, bot filtering, session namespace isolation
- **Testing strategy**: `MockAdapter` for unit/integration tests
- **Implementation phases**: 5 phases, Phase 1+3 already merged via #259

## Files

| File | Purpose |
|------|---------|
| `docs/adr/multi-platform-adapters.md` | Full ADR with 15 sections covering design, implementation, and resolved questions |

## Related

- Tracking issues: #86, #93
- Slack adapter: #259 (merged, validates Phase 1+3)
- Reviewer feedback integrated: @dogzzdogzz (8 points), @antigenius0910 (Contract A vs B)

### DIFF
--- a/docs/adr/multi-platform-adapters.md
+++ b/docs/adr/multi-platform-adapters.md
@@ -1,13 +1,14 @@
-# RFC 002: Multi-Platform Adapter Architecture
+# ADR: Multi-Platform Adapter Architecture
 
-**Tracking issues:** #86, #93
-**Status:** Partially Implemented — Phase 1+3 landed via #259 (Slack adapter)
-**Author:** @chaodu-agent
-**Reviewers:** @dogzzdogzz, @antigenius0910
+- **Status:** Partially Implemented — Phase 1+3 landed via #259 (Slack adapter)
+- **Date:** 2026-04-06
+- **Author:** @chaodu-agent
+- **Reviewers:** @dogzzdogzz, @antigenius0910
+- **Tracking issues:** #86, #93
 
 ---
 
-## Summary
+## 1. Context & Decision
 
 Define a platform-agnostic adapter layer for agent-broker so it can serve Discord, Telegram, Slack, and future chat platforms through a single unified architecture. The ACP session pool and agent backend remain unchanged — only the "front door" becomes pluggable.
 
@@ -15,7 +16,7 @@ Define a platform-agnostic adapter layer for agent-broker so it can serve Discor
 
 > Contract A (pluggable single-platform per deployment) is a subset of B and works automatically — deploy with only one `[platform]` section in config.
 
-## Motivation
+## 2. Motivation
 
 - agent-broker was originally hard-wired to Discord via `serenity`
 - #86 proposes a Telegram adapter, #93 proposes Slack — both require similar abstractions
@@ -24,7 +25,7 @@ Define a platform-agnostic adapter layer for agent-broker so it can serve Discor
 
 ---
 
-## Current Architecture (post-#259)
+## 3. Current Architecture (post-#259)
 
 ```
                     ┌─────────────────┐
@@ -54,7 +55,7 @@ Define a platform-agnostic adapter layer for agent-broker so it can serve Discor
 
 ---
 
-## 1. ChatAdapter Trait
+## 4. ChatAdapter Trait
 
 The core abstraction. Each platform implements this trait.
 
@@ -139,7 +140,7 @@ pub struct SenderContext {
 
 ---
 
-## 2. AdapterRouter
+## 5. AdapterRouter
 
 Shared logic extracted from `discord.rs` that is platform-independent:
 
@@ -182,7 +183,7 @@ Each adapter only needs to:
 
 ---
 
-## 3. Platform Comparison
+## 6. Platform Comparison
 
 | Feature | Discord | Telegram | Slack |
 |---------|---------|----------|-------|
@@ -197,7 +198,7 @@ Each adapter only needs to:
 
 ---
 
-## 4. Config Design
+## 7. Config Design
 
 ```toml
 # Enable one or more adapters. Multiple can run simultaneously.
@@ -236,7 +237,7 @@ session_ttl_hours = 24
 
 ---
 
-## 5. Message Size Handling
+## 8. Message Size Handling
 
 Each platform has different message limits. The `format::split_message` function accepts a configurable limit, sourced from `adapter.message_limit()`:
 
@@ -252,7 +253,7 @@ pub fn split_message(content: &str, max_len: usize) -> Vec<String>;
 
 ---
 
-## 6. Reaction Mapping
+## 9. Reaction Mapping
 
 The `StatusReactionController` uses `Arc<dyn ChatAdapter>` to call `add_reaction()` / `remove_reaction()`, fully decoupled from any platform SDK.
 
@@ -268,7 +269,7 @@ Config stores Unicode emoji. Each adapter converts internally:
 
 ---
 
-## 7. Security Considerations
+## 10. Security Considerations
 
 - **Secure by default** (#91): empty allowlist = deny all, for every platform
 - **Bot message filtering**: each adapter must ignore messages from bots to prevent loops
@@ -278,7 +279,7 @@ Config stores Unicode emoji. Each adapter converts internally:
 
 ---
 
-## 8. Implementation Phases
+## 11. Implementation Phases
 
 | Phase | Scope | Status | Notes |
 |-------|-------|--------|-------|
@@ -290,7 +291,7 @@ Config stores Unicode emoji. Each adapter converts internally:
 
 ---
 
-## 9. Kubernetes / Helm Considerations
+## 12. Kubernetes / Helm Considerations
 
 - Single image supports all adapters (all compiled in)
 - Helm `values.yaml` gains `telegram.*` and `slack.*` sections
@@ -299,7 +300,7 @@ Config stores Unicode emoji. Each adapter converts internally:
 
 ---
 
-## 10. Testing Strategy
+## 13. Testing Strategy
 
 (Feedback: @dogzzdogzz #8)
 
@@ -338,7 +339,7 @@ impl ChatAdapter for MockAdapter {
 
 ---
 
-## 11. Resolved Questions
+## 14. Resolved Questions
 
 (Feedback: @dogzzdogzz #7, @antigenius0910)
 
@@ -353,7 +354,7 @@ impl ChatAdapter for MockAdapter {
 
 ---
 
-## 12. Known Limitations
+## 15. Known Limitations
 
 (Feedback: @antigenius0910)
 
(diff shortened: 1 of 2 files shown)


dogzzdogzz (commented):
## Review: RFC 002 — Multi-Platform Adapter Architecture

Overall direction is solid — `ChatAdapter` trait + `AdapterRouter` is the right abstraction. Phase 1 (pure refactor, no behavior change) is a good de-risking strategy. A few architectural gaps to address before this RFC is ready for implementation:

### 1. `ChatAdapter` trait missing `message_limit()`

Section 5 discusses per-platform message limits (2000/4096/4000), and `format::split_message` already accepts a `limit: usize` parameter. But the trait itself doesn't expose this — `AdapterRouter` would need to hardcode a match on the `platform()` string. Add:

```rust
fn message_limit(&self) -> usize;
```

### 2. `ChannelRef` threading abstraction doesn't cover all models

Current design:
```rust
pub struct ChannelRef {
    pub platform: String,
    pub channel_id: String,
    pub parent_id: Option<String>,
}
```

This works for Discord (thread = separate channel with a parent), but not for platforms where threads are reply chains within the same channel, identified by a separate thread/topic ID rather than a child channel. Consider:

```rust
pub struct ChannelRef {
    pub platform: String,
    pub channel_id: String,
    pub thread_id: Option<String>,  // thread within channel (e.g. Slack thread_ts, Telegram topic_id)
    pub parent_id: Option<String>,  // parent channel if thread-as-channel (Discord)
}
```

Alternatively, unify with a single `thread_id: Option<String>` and let each adapter interpret it. Either way, the current model needs to explicitly address the "thread as reply chain" vs "thread as child channel" distinction.

### 3. `create_thread` — who decides "already in a thread"?

Current `discord.rs` has logic in `get_or_create_thread()` that checks whether a message is already inside a thread (by comparing channel kind / parent ID). In the proposed architecture, where does this decision live?

- If in `AdapterRouter` → router needs `ChannelRef.is_thread()` (e.g. `thread_id.is_some() || parent_id.is_some()`)
- If in each adapter → duplicated logic

Recommend making this a router responsibility, with `ChannelRef` providing enough information to decide.

### 4. `StatusReactionController` is coupled to Serenity

The current `StatusReactionController::new()` takes `Arc<serenity::http::Http>`, `ChannelId`, and `MessageId` — all Serenity types. Phase 1 refactor must decouple this. Two options:

- (a) Controller calls `adapter.add_reaction()` / `adapter.remove_reaction()` via `Arc<dyn ChatAdapter>`
- (b) Introduce a small `ReactionSender` trait

Option (a) is simpler and aligns with the existing trait design.

### 5. `AdapterRouter.handle_message()` ownership model

The proposed signature:
```rust
pub async fn handle_message(
    &self,
    adapter: &dyn ChatAdapter,
    ...
) -> Result<()>;
```

The current streaming loop in `discord.rs` uses `tokio::spawn` for the edit-streaming background task. A `&dyn ChatAdapter` reference cannot be moved into a spawned task. This needs to be `Arc<dyn ChatAdapter>`, or the streaming loop needs to be restructured to avoid spawning. This is a core design decision that affects all adapters.

### 6. Emoji format is not universal

`ReactionsConfig` defaults use Unicode emoji (`👀`, `🤔`, `🆗`). Some platforms require platform-specific names for the same emoji (e.g. `eyes`, `thinking_face`, `ok`). The RFC should clarify whether:

- (a) Config stores Unicode and each adapter converts internally
- (b) Config stores per-platform emoji mappings
- (c) `ChatAdapter` exposes an `fn emoji_name(&self, unicode: &str) -> String` method

Recommend (a) for simplicity — adapters handle conversion as an internal concern.

### 7. Open Questions need recommended answers

The 5 open questions need recommendations, not just options:

| # | Question | Recommendation | Rationale |
|---|----------|---------------|-----------|
| 1 | Shared vs separate SessionPool | **Shared** | Already namespaced by platform prefix; separate pools add complexity with no isolation benefit |
| 2 | Platform prefix in session keys | **Yes, always prefix** | IDs from different platforms can theoretically collide; prefix cost is negligible |
| 3 | Global vs per-adapter max_sessions | **Global hard cap + per-adapter soft limit** | Prevents one noisy platform from starving others |
| 4 | Rich messages in v1 | **Text-only** | Ship the abstraction first; rich messages are additive and platform-divergent |
| 5 | Platform-specific features | **Defer to Phase 5** | Keep `ChatAdapter` minimal for v1; extend via `PlatformExt` traits or feature flags later |

### 8. Testing strategy missing

No mention of how to validate the architecture. Recommend:

- Define a `MockAdapter` (in-memory, no network) that implements `ChatAdapter`
- Use it to unit test `AdapterRouter` logic (session routing, streaming, reactions) in isolation
- Phase 1 acceptance criteria: existing Discord behavior passes integration tests through the new trait boundary

---

**Summary:** The RFC is directionally correct. The main gaps are in trait completeness (`message_limit`, `ChannelRef` threading model), ownership model (`Arc<dyn>` vs `&dyn`), Serenity decoupling in `reactions.rs`, and missing test strategy. Recommend addressing these before Phase 1 implementation begins.

antigenius0910:
## Question: Primary use case — pluggable single-platform or simultaneous multi-platform?

Before Phase 1 implementation begins, I want to raise a definition question that significantly impacts the architecture scope.

The RFC design — session key namespacing (`discord:123`, `slack:456`), the "shared vs separate pools" open question, and `AdapterRouter` — implies **simultaneous multi-platform operation** (one binary serving Discord + Slack at the same time). But I want to confirm whether this is actually the primary contract.

There are two meaningfully different use cases:

**A) Pluggable single-platform per deployment**
> "I deploy for Slack OR Discord — easy to swap via config, not running both."
- Requires: simple trait abstraction, adapter selection at startup, no session namespacing
- Covers the vast majority of deployments

**B) Simultaneous multi-platform in one binary**
> "One running instance serves both Discord and Slack concurrently."
- Requires: session key namespacing, shared/separate pool design, concurrent adapter lifecycle management, config merging
- The complexity the RFC currently designs for

These lead to meaningfully different implementations. If **A** is the primary case, the RFC is paying upfront complexity cost for an edge case — session namespacing and shared pool coordination are only needed when two adapters actually run concurrently.

The simultaneous case (**B**) does have legitimate scenarios — e.g., an internal Slack team + an external Discord developer community both accessing the same agent backend with different `allowed_users` scopes. But that should be an **explicit design commitment** with a named use case, not an implicit assumption that drives the whole architecture.

### Why this matters before Phase 1

If the answer is **B**, there are two architectural pre-requisites not currently in the RFC that become hard blockers under concurrent load:

1. **Global write-lock on `SessionPool::with_connection()`** — currently holds a write lock on the entire pool `HashMap` for the full prompt duration (potentially minutes). Adding a second adapter doubles concurrent session pressure through a single lock, serialising all sessions across all platforms.
2. **No streaming timeout** — a hung session on one platform can block the pool and starve the other platform's sessions.

Neither of these is a Phase 1 concern if the answer is **A**.

### Proposed resolution

Could the RFC explicitly state which is the primary contract? If **B**, what is the concrete production scenario that justifies the added complexity? If **A**, the Phase 1 scope simplifies significantly and the namespacing / pool-sharing questions become post-MVP.

Happy to elaborate on either the lock issue or the timeout issue if helpful.

chaodu-agent:
## RFC 002 Updated — Review Feedback Integrated

Updated the RFC to address all outstanding feedback from @dogzzdogzz and @antigenius0910, and to reflect the merge of #259 (Slack adapter).

### What changed

<details>
<summary><b>Primary contract declared: Contract B (simultaneous multi-platform)</b></summary>

@antigenius0910 asked whether the RFC targets single-platform-per-deployment (A) or simultaneous multi-platform (B). Since #259 already merged with Discord + Slack running in one process, the RFC now explicitly declares B as the primary contract. Contract A is a subset that works automatically.
</details>

<details>
<summary><b>@dogzzdogzz's 8 architectural points integrated</b></summary>

| # | Point | Resolution |
|---|-------|------------|
| 1 | `message_limit()` missing from trait | Added to `ChatAdapter` |
| 2 | `ChannelRef` threading model incomplete | Added `thread_id: Option<String>` for reply-chain model (Slack/Telegram), kept `parent_id` for thread-as-channel (Discord) |
| 3 | `create_thread` decision ownership | Explicitly assigned to router via `ChannelRef::is_thread()` |
| 4 | `StatusReactionController` coupled to Serenity | Decoupled — now uses `Arc<dyn ChatAdapter>` |
| 5 | `&dyn ChatAdapter` can't move into spawned task | Changed to `Arc<dyn ChatAdapter>` |
| 6 | Emoji format not universal | Config stores Unicode, each adapter converts internally |
| 7 | Open Questions need recommendations | All 5 answered with rationale (new §11) |
| 8 | Testing strategy missing | Added §10 with `MockAdapter` + coverage table |
</details>

<details>
<summary><b>New sections added</b></summary>

- **§10 Testing Strategy** — `MockAdapter` definition, test coverage table, Phase 1 acceptance criteria (validated by #259)
- **§11 Resolved Questions** — All 5 original open questions answered with decisions and rationale, plus the A-vs-B question
- **§12 Known Limitations** — Pool lock contention (resolved: read lock + Arc clone pattern), streaming timeout (open: follow-up issue needed)
</details>

<details>
<summary><b>Implementation phases updated</b></summary>

Phases now reflect reality: Phase 1 + Phase 3 marked as ✅ Merged (#259). Phase 4 repurposed for per-adapter session limits + streaming timeout. Phase 5 unchanged (platform-specific rich features).
</details>

### Commit

`68476f7` — 166 insertions, 65 deletions

thepagent:
This RFC is still valid. Feel free to comment your thoughts and we'll discuss and update accordingly.

chaodu-agent:
Thanks @antigenius0910 for the thorough Contract A vs B analysis — it was the right question to raise before Phase 1.

The ADR has been updated to explicitly declare **Contract B (simultaneous multi-platform)** as the primary contract, since #259 already validated Discord + Slack running in one process. Contract A works automatically as a subset.

Both architectural pre-requisites you flagged are addressed in the ADR:

1. **Pool write-lock** (§15 Known Limitations → Resolved) — the current implementation uses read lock + `Arc` clone + drop, then locks only the individual connection. Multiple adapters' sessions execute concurrently without blocking each other.
2. **Streaming timeout** (§15 Known Limitations → Open) — acknowledged as a follow-up issue. A hung session consumes one pool slot but doesn't block other sessions under the new locking model.

The ADR has also been reformatted from RFC to ADR format under `docs/adr/multi-platform-adapters.md` to align with the project's documentation structure.

thepagent (approved):

PR #137, opened 2026-04-08
fix: handle duplicate thread creation on Discord gateway reconnect

the3mi:
When Discord gateway reconnects, the same message event can be delivered twice. The second attempt to create a thread fails with HTTP 500 because the thread already exists.

This change catches the HTTP 500 error and queries the channel to find the existing thread, returning its ID instead of propagating the error.

Fixes: `A thread has already been created for this message` error on gateway reconnect.

### DIFF
--- a/src/discord.rs
+++ b/src/discord.rs
@@ -357,7 +357,7 @@ async fn get_or_create_thread(ctx: &Context, msg: &Message, prompt: &str) -> any
 
     let thread_name = shorten_thread_name(prompt);
 
-    let thread = msg
+    let thread = match msg
         .channel_id
         .create_thread_from_message(
             &ctx.http,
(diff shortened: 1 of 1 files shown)


ottobeetw:
👋 This fix is ready to merge and resolves the exact error currently affecting production: `A thread has already been created for this message` on Discord gateway reconnect.

The fix is minimal and safe — it only changes the error handler for thread creation to query existing threads instead of propagating the error. No behavior change for the happy path.

Could we get an expedited review? This is blocking production use.

masami-agent (changes requested):
Good idea to handle the duplicate thread creation on reconnect, but a few concerns:

**1. Matching on HTTP 500 is fragile**

Discord returns HTTP 500 for many server-side errors, not just "thread already exists". Catching all 500s and assuming it's a duplicate thread could mask real Discord API failures. 

A safer approach: check the error body/code for the specific Discord error. Discord returns error code `160004` for "A thread has already been created for this message". Matching on that would be precise.

**2. The fallback logic queries the wrong channel**

When the thread already exists, `msg.channel_id` is the *parent* channel, not the thread. Calling `msg.channel_id.to_channel()` returns the parent channel info, which won't have `thread_metadata`. You'd need to find the thread by iterating active threads or using `msg.id` to locate the thread created from that message.

Consider using `ctx.http.get_channel_messages()` or the guild's active threads endpoint to find the existing thread.

**3. Missing logging**

When we recover from a duplicate, it would be helpful to log it:
```rust
tracing::info!(message_id = %msg.id, "thread already exists, reusing");
```

Please address these and I'm happy to re-review 🙏

PR #138, opened 2026-04-08
fix: dedupe tool call display by toolCallId and sanitize titles

marvin-69-jpg:
## Summary

Two independent rendering bugs in the streaming tool-call display, both visible in any Discord thread that runs more than a couple of bash tools:

1. **Duplicate placeholder lines.** `claude-agent-acp` emits the first `tool_call` event before the input fields are streamed in, so the title falls back to `"Terminal"` / `"Edit"` / etc. A later `tool_call_update` carries the refined real title. The current handler pushes a *new* line on every event without deduping, leaving orphaned placeholders that never get updated:

   ```
   🔧 `Terminal`...
   ❌
   🔧 `cd /home/node/work && git status`...
   ✅
   ```

   The `ToolDone` arm tries to find the entry to update via `tool_lines.iter_mut().rev().find(|l| l.contains(&title))`, which works for the refined entries but never matches the orphaned `Terminal` placeholders.

2. **Inline-code spans break on multi-line commands.** The title is rendered as `format!("🔧 `{title}`...")` — single backticks. Discord's single-backtick inline code only spans one line. Any heredoc, multi-line bash, or `&&`-chained command split across lines breaks the inline code span at the first `\n`, and the rest of the command spills out as garbled raw text. Visually it looks like the command was truncated.

## Fix

- Carry the ACP `toolCallId` through `AcpEvent::ToolStart` / `ToolDone` so the renderer can pin updates to the same entry across the streaming `tool_call` → `tool_call_update` → `tool_call_update(status=completed)` lifecycle.
- Replace `tool_lines: Vec<String>` with `Vec<ToolEntry>` (id, title, state). `ToolStart` updates the slot in place if the id is already present; `ToolDone` finds by id (not by substring match against an unreliable title) and preserves the existing title when the Done event omits one — which it routinely does in `claude-agent-acp` updates.
- Add a `sanitize_title` helper that flattens `\n` to `" ; "` and rewrites embedded backticks so they can't break the surrounding inline-code span.
- Each tool now renders via `ToolEntry::render()`, which picks the icon from a `ToolState` enum (Running / Completed / Failed) — no more brittle substring lookups.

## Repro (before this PR)

In a Discord thread connected to a `claude` preset bot, ask:

```
@bot 在 /tmp/work 跑這幾個指令:
1. pwd
2. git status
3. ls -la && echo done
```

Result: every Bash invocation leaves a `🔧 \`Terminal\`...` line in the message that never updates to the real command. Multi-line `&&` commands also render with the second line falling out of the inline-code span.

## After this PR

Each tool call appears as a single line that evolves in place:

```
🔧 `pwd`...
✅ `pwd`
🔧 `git status`...
✅ `git status`
🔧 `ls -la && echo done`...
✅ `ls -la && echo done`
```

Multi-line commands collapse to one line: `cmd1 ; cmd2 ; cmd3` — still inside the inline-code span, no markdown breakage.

## Files changed

- `src/acp/protocol.rs` — `AcpEvent::ToolStart` / `ToolDone` carry an `id`; `classify_notification` extracts `toolCallId` from the update payload.
- `src/discord.rs` — new `ToolEntry` / `ToolState` types; `sanitize_title` helper; `tool_lines` becomes `Vec<ToolEntry>`; ToolStart/ToolDone match arms dedupe by id.

+108 / -15. No new dependencies.

## Compatibility

- ACP protocol: `toolCallId` is already part of every `tool_call` and `tool_call_update` payload (it's how `claude-agent-acp` itself correlates the events internally — see `acp-agent.js` lines 1501-1556 in v0.25.0). We just weren't reading it.
- Backwards compatible with any ACP backend that emits `tool_call` without a `toolCallId` (we fall back to an empty string id, in which case all calls without ids share one slot — the same effective behavior as today, just on the empty-id path).
- No public API change. All new types are private to `discord.rs`.

## Testing

Running in production for the past two days against a multi-user Discord channel doing heavy git/gh PR workflows including many multi-line bash commands. No regressions seen; the orphan-line and multi-line-cmd issues are gone.

Note: this PR is intentionally scoped to the tool-call display layer. It does **not** touch the streaming chunking / message-splitting code (that's #135's territory) or the thinking/agent_thought_chunk handling. A follow-up will surface thinking content as a blockquote on top of the message; that's a separate concern.

### DIFF
--- a/src/acp/protocol.rs
+++ b/src/acp/protocol.rs
@@ -60,8 +60,8 @@ impl std::fmt::Display for JsonRpcError {
 pub enum AcpEvent {
     Text(String),
     Thinking,
-    ToolStart { title: String },
-    ToolDone { title: String, status: String },
+    ToolStart { id: String, title: String },
+    ToolDone { id: String, title: String, status: String },
     Status,
 }
 
@@ -69,6 +69,19 @@ pub fn classify_notification(msg: &JsonRpcMessage) -> Option<AcpEvent> {
     let params = msg.params.as_ref()?;
     let update = params.get("update")?;
     let session_update = update.get("sessionUpdate")?.as_str()?;
+
+    // toolCallId is the stable identity across tool_call → tool_call_update
+    // events for the same tool invocation. claude-agent-acp emits the first
+    // event before the input fields are streamed in (so the title falls back
+    // to "Terminal" / "Edit" / etc.) and refines them in a later
+    // tool_call_update; without the id we can't tell those events belong to
+    // the same call and end up rendering placeholder + refined as two
+    // separate lines.
+    let tool_id = update
+        .get("toolCallId")
+        .and_then(|v| v.as_str())
+        .unwrap_or("")
+        .to_string();
 
     match session_update {
         "agent_message_chunk" => {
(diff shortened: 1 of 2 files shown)


masami-agent (approved):
Solid fix, well-documented, and production-tested 👍

The `toolCallId`-based dedup is the right approach — substring matching on title was always fragile, especially with claude-agent-acp's placeholder → refined title lifecycle. The `ToolEntry` / `ToolState` design is clean and the `sanitize_title` helper handles the multi-line command rendering issue nicely.

A few minor notes (non-blocking):
- `ToolEntry` and `ToolState` are marked `pub` but only used within `discord.rs` — could be `pub(crate)` or private. Not a blocker.
- Backward compat with empty `toolCallId` is handled correctly (all share one slot = same behavior as before).

LGTM, good to merge.

chaodu-agent:
LGTM — clean fix, well-structured, and production-tested. Approving and merging.

One minor note for a follow-up: `ToolState` and `ToolEntry` (and their fields) are marked `pub` but only used within `discord.rs`. Can be made private in a future cleanup. Non-blocking.

Thanks for the contribution! 🙏

thepagent (approved):
LGTM — visibility fix applied, squashing and merging.

PR #147, opened 2026-04-08
fix(acp): implement dynamic permission selection and outcome wrapper

### LINKED ISSUE #111: fix: ExitPlanMode permission rejected — auto-reply sends invalid optionId
## Problem

When Claude Code enters plan mode and tries to exit (ExitPlanMode), the broker's auto-reply for `session/request_permission` sends `{"optionId": "allow_always"}`. But ExitPlanMode's permission options don't include `allow_always`:

```json
"options": [
  {"optionId": "bypassPermissions", "name": "Yes, and bypass permissions", "kind": "allow_always"},
  {"optionId": "acceptEdits", "name": "Yes, and auto-accept edits", "kind": "allow_always"},
  {"optionId": "default", "name": "Yes, and manually approve edits", "kind": "allow_once"},
  {"optionId": "plan", "name": "No, keep planning", "kind": "reject_once"}
]
```

Claude Code receives an unknown optionId and treats it as rejected:

```
auto-allow permission title="Ready to code?"
→ sends {"optionId": "allow_always"}
→ Claude Code: "User rejected request to exit plan mode."
```

The agent gets stuck in plan mode forever.

## Root Cause

The permission auto-reply in `connection.rs` hardcodes `"allow_always"` as the optionId:

```rust
let reply = JsonRpcResponse::new(id, json!({"optionId": "allow_always"}));
```

But `"allow_always"` is a **kind** (option category), not an **optionId** (specific option identifier). Different tool calls provide different option sets. For regular tool permissions, there happens to be an option whose `optionId` is `"allow_always"` — but ExitPlanMode uses `"bypassPermissions"`, `"acceptEdits"`, `"default"`, and `"plan"`.

## Proposed Fix

Instead of hardcoding the optionId, select the most permissive option from the provided choices:

```rust
let option_id = msg.params.as_ref()
    .and_then(|p| p.get("options"))
    .and_then(|o| o.as_array())
    .and_then(|options| {
        // Prefer allow_always kind, then allow_once, then first non-reject
        options.iter()
            .find(|o| o.get("kind").and_then(|k| k.as_str()) == Some("allow_always"))
            .or_else(|| options.iter()
                .find(|o| o.get("kind").and_then(|k| k.as_str()) == Some("allow_once")))
            .or_else(|| options.iter()
                .find(|o| o.get("kind").and_then(|k| k.as_str()) != Some("reject_once")))
            .and_then(|o| o.get("optionId"))
            .and_then(|id| id.as_str())
    })
    .unwrap_or("allow_always");  // fallback only if no options provided
```

This correctly selects `"bypassPermissions"` for ExitPlanMode while maintaining backward compatibility for regular tool permissions.

## Impact

- Affects all Claude Code users who trigger plan mode (explicitly or via agent behavior)
- Agent gets permanently stuck in plan mode until session is killed
- Confirmed with `claude-agent-acp@0.25.0` + `claude-code@2.1.87`
- Regular (non-plan-mode) tool permissions are unaffected because their options do include an `optionId` of `"allow_always"`

## Related

- RFC #78 — Session Management (§5 Session Security & Access Control discusses permission handling)
- #76 — notification loop robustness (separate issue, but plan mode failures compound with it)

### LINKED ISSUE #130: Bug: RequestPermissionResponse missing outcome wrapper — Claude Code tool calls universally rejected
## Summary

`openab`'s auto-reply for `session/request_permission` (in `src/acp/connection.rs`) sends a JSON-RPC response body with the wrong shape — missing the required `outcome` wrapper. The response works for Kiro CLI (loose parsing) but is rejected by `claude-agent-acp` (strict ACP SDK validation), causing **every tool call from Claude Code to fail with "User refused permission to run tool"**, even though `openab` logs `auto-allow permission`.

This is also the upstream root cause for the symptom reported in #127 (Cursor's `Cannot read properties of undefined (reading 'outcome')` error) — the Cursor SDK is trying to read `response.outcome.optionId` and crashing because `response.outcome` is undefined.

## Current vs spec response shape

`src/acp/connection.rs` line 99:

```rust
let reply = JsonRpcResponse::new(id, json!({"optionId": "allow_always"}));
```

Per the official ACP TypeScript SDK schema (`@agentclientprotocol/sdk` types.gen.d.ts):

```typescript
export type RequestPermissionResponse = {
    _meta?: { ... } | null;
    /** The user's decision on the permission request. */
    outcome: RequestPermissionOutcome;
};

export type RequestPermissionOutcome =
    | { outcome: "cancelled"; }
    | (SelectedPermissionOutcome & { outcome: "selected"; });
```

The valid response shape is:

```json
{
  "outcome": {
    "outcome": "selected",
    "optionId": "allow_always"
  }
}
```

`openab` is sending the inner object's `optionId` field at the top level, with no `outcome` wrapper.

## Repro

1. Run `openab` configured with `claude-agent-acp` backend (any Claude Code agent)
2. From Discord, ask the agent to do anything that requires a tool call (e.g. `git pull`, read a file, run `ls`)
3. `openab` log shows: `INFO openab::acp::connection: auto-allow permission title="..."`
4. `claude-agent-acp` log shows: `tool_call_update status=failed rawOutput="User refused permission to run tool"`
5. The agent gives up or works around it without ever executing the tool

Trace from a real session (`agent-broker:claude` running `claude-agent-acp@0.25.0` + Claude Code):

```
18:54:32.071  request_permission for git -C ~/work/trips pull --ff-only
18:54:32.071  auto-allow permission title="git -C ~/work/trips pull --ff-only"
18:54:32.079  tool_call_update status=failed rawOutput="User refused permission to run tool"
```

8 ms between auto-allow and the failure — `claude-agent-acp` is parsing `openab`'s reply, finding no `outcome` field, and treating it as a refusal.

## Why Kiro doesn't see this

Kiro CLI's ACP implementation parses the response loosely and accepts `{"optionId": "allow_always"}` as if it were already inside the `outcome` wrapper. So this bug is invisible when running the `kiro-cli` backend, but every Claude Code or Cursor user is hitting it on every tool call.

## Suggested fix

Wrap the response correctly:

```rust
let reply = JsonRpcResponse::new(id, json!({
    "outcome": {
        "outcome": "selected",
        "optionId": "allow_always"
    }
}));
```

I've patched my local fork with this change and **all Claude Code tool calls now succeed end-to-end**: Discord mention → claude-agent-acp spawned → request_permission → openab auto-allow → tool executes → result streams back. Verified with a `git pull` + file edit + `git commit -m ...` + `git push` flow.

## Also fixes / relates to

- **#127** Cursor `Cannot read properties of undefined (reading 'outcome')` — same root cause; the Cursor SDK is reading `response.outcome.optionId`, crashes because `response.outcome` is undefined. The `!sh` shell escape workaround in #127 should no longer be needed after this fix.
- **#111** ExitPlanMode permission rejected — DIFFERENT bug (wrong `optionId` selected). Even with #111's proposed fix landed, the `outcome` wrapper bug remains. Both fixes need to land. A combined fix would (a) wrap the response correctly, and (b) select the right optionId from the `options` array based on `kind` priority (`allow_always` > `allow_once` > first non-`reject_once`).

Happy to send a PR for the wrap fix alone, or a combined wrap + optionId-selection fix that also closes #111. Just say which scope you prefer.

### LINKED ISSUE #241: Bug: RequestPermissionResponse missing outcome wrapper breaks gh CLI flows in openab-gemini
## Summary

Using `ghcr.io/openabdev/openab-gemini:latest` as the base image, `openab` works for normal chat interactions and simple Linux commands, but any `gh` CLI command triggers a terminal-layer crash instead of returning normal command output.

The failure is reproducible.

## Environment

- Image: `ghcr.io/openabdev/openab-gemini:latest`
- Agent backend: Gemini
- Auth: Gemini API key provided via `GEMINI_API_KEY`

## What works

Inside the running environment, normal commands work as expected, for example:

```bash
ls -l /usr/bin/gh
ls /nonexistent
echo "gh"
echo "$GITHUB_TOKEN"
```

Also, regular chat interactions and basic Linux command execution work.

## What fails

Any `gh` command causes the underlying terminal interface / agent wrapper to crash immediately. Examples:

```bash
gh auth status
gh --version
gh auth login -w
```

The observed internal error is:

```text
Cannot read properties of undefined (reading 'outcome')
```

## Expected behavior

`gh` commands should behave like other shell commands:

- `gh --version` should print the version
- `gh auth status` should print auth state
- `gh auth login -w` should print the browser URL and one-time code

## Actual behavior

The command execution path crashes in the terminal/agent integration layer when `gh` is invoked, so `gh` output never reaches the user.

## Repro steps

1. Start from `ghcr.io/openabdev/openab-gemini:latest`
2. Configure `openab` to use Gemini with a valid `GEMINI_API_KEY`
3. Verify normal command execution works with non-`gh` commands
4. Run any `gh` command such as `gh --version` or `gh auth status`
5. Observe the internal error:
   - `Cannot read properties of undefined (reading 'outcome')`

## Notes

This does not look like a generic shell execution problem, because non-`gh` commands work. It appears specific to the interaction between the image/runtime and the terminal wrapper when `gh` is executed.

chenjian-agent:
## Summary
- Fixes RequestPermissionResponse missing outcome wrapper (#130)
- Adds dynamic permission selection logic to fix #111 (handles scenarios where allow_always is missing, e.g. ExitPlanMode)
- Also resolves #241 (same root cause — incorrect response shape and hardcoded optionId)

## Fix Details
- Refactored session/request_permission logic to correctly wrap responses in an outcome object.
- Extracted pick_best_option() to dynamically select the most permissive option from the options array (allow_always > allow_once > first non-reject), replacing hardcoded optionId.
- Falls back to "allow_always" when options field is absent (backward compat).
- Returns cancelled outcome when only reject options are available.
- Added comprehensive unit tests covering all selection branches and edge cases (including ExitPlanMode scenarios).

Closes #130, #111, #241.

### DIFF
--- a/src/acp/connection.rs
+++ b/src/acp/connection.rs
@@ -10,6 +10,68 @@ use tokio::sync::{mpsc, oneshot, Mutex};
 use tokio::task::JoinHandle;
 use tracing::{debug, error, info};
 
+/// Pick the most permissive selectable permission option from ACP options.
+fn pick_best_option(options: &[Value]) -> Option<String> {
+    let mut fallback: Option<&Value> = None;
+
+    for kind in ["allow_always", "allow_once"] {
+        if let Some(option) = options
+            .iter()
+            .find(|option| option.get("kind").and_then(|k| k.as_str()) == Some(kind))
+        {
+            return option
+                .get("optionId")
+                .and_then(|id| id.as_str())
+                .map(str::to_owned);
+        }
+    }
+
+    for option in options {
+        let kind = option.get("kind").and_then(|k| k.as_str());
+        if kind == Some("reject_once") || kind == Some("reject_always") {
+            continue;
+        }
+        fallback = Some(option);
+        break;
+    }
+
+    fallback
+        .and_then(|option| option.get("optionId"))
+        .and_then(|id| id.as_str())
+        .map(str::to_owned)
+}
+
+/// Build a spec-compliant permission response with backward-compatible fallback.
+fn build_permission_response(params: Option<&Value>) -> Value {
+    match params
+        .and_then(|p| p.get("options"))
+        .and_then(|options| options.as_array())
+    {
+        None => json!({
+            "outcome": {
+                "outcome": "selected",
+                "optionId": "allow_always"
+            }
+        }),
+        Some(options) => {
+            if let Some(option_id) = pick_best_option(options) {
+                json!({
+                    "outcome": {
+                        "outcome": "selected",
+                        "optionId": option_id
+                    }
+                })
+            } else {
+                json!({
+                    "outcome": {
+                        "outcome": "cancelled"
+                    }
+                })
+            }
+        }
+    }
+}
+
 fn expand_env(val: &str) -> String {
     if val.starts_with("${") && val.ends_with('}') {
         let key = &val[2..val.len() - 1];
@@ -90,13 +152,17 @@ impl AcpConnection {
                     // Auto-reply session/request_permission
                     if msg.method.as_deref() == Some("session/request_permission") {
                         if let Some(id) = msg.id {
-                            let title = msg.params.as_ref()
+                            let title = msg
+                                .params
+                                .as_ref()
                                 .and_then(|p| p.get("toolCall"))
                                 .and_then(|t| t.get("title"))
                                 .and_then(|t| t.as_str())
                                 .unwrap_or("?");
-                            info!(title, "auto-allow permission");
-                            let reply = JsonRpcResponse::new(id, json!({"optionId": "allow_always"}));
+
+                            let outcome = build_permission_response(msg.params.as_ref());
+                            info!(title, %outcome, "auto-respond permission");
+                            let reply = JsonRpcResponse::new(id, outcome);
                             if let Ok(data) = serde_json::to_string(&reply) {
                                 let mut w = stdin_clone.lock().await;
                                 let _ = w.write_all(format!("{data}\n").as_bytes()).await;
@@ -284,5 +350,102 @@ impl AcpConnection {
 
     pub fn alive(&self) -> bool {
         !self._reader_handle.is_finished()
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::{build_permission_response, pick_best_option};
+    use serde_json::json;
+
+    #[test]
+    fn picks_allow_always_over_other_options() {
+        let options = vec![
+            json!({"kind": "allow_once", "optionId": "once"}),
+            json!({"kind": "allow_always", "optionId": "always"}),
+            json!({"kind": "reject_once", "optionId": "reject"}),
+        ];
+
+        assert_eq!(pick_best_option(&options), Some("always".to_string()));
+    }
+
+    #[test]
+    fn falls_back_to_first_unknown_non_reject_kind() {
+        let options = vec![
+            json!({"kind": "reject_once", "optionId": "reject"}),
+            json!({"kind": "workspace_write", "optionId": "workspace-write"}),
+        ];
+
+        assert_eq!(
+            pick_best_option(&options),
+            Some("workspace-write".to_string())
+        );
+    }
+
+    #[test]
+    fn selects_bypass_permissions_for_exit_plan_mode() {
+        let options = vec![
+            json!({"optionId": "bypassPermissions", "kind": "allow_always"}),
+            json!({"optionId": "acceptEdits", "kind": "allow_always"}),
+            json!({"optionId": "default", "kind": "allow_once"}),
+            json!({"optionId": "plan", "kind": "reject_once"}),
+        ];
+
+        assert_eq!(
+            pick_best_option(&options),
+            Some("bypassPermissions".to_string())
+        );
+    }
+
+    #[test]
+    fn returns_none_when_only_reject_options_exist() {
+        let options = vec![
+            json!({"kind": "reject_once", "optionId": "reject-once"}),
+            json!({"kind": "reject_always", "optionId": "reject-always"}),
+        ];
+
+        assert_eq!(pick_best_option(&options), None);
+    }
+
+    #[test]
+    fn builds_cancelled_outcome_when_no_selectable_option_exists() {
+        let response = build_permission_response(Some(&json!({
+            "options": [
+                {"kind": "reject_once", "optionId": "reject-once"}
+            ]
+        })));
+
+        assert_eq!(response, json!({"outcome": {"outcome": "cancelled"}}));
+    }
+
+    #[test]
+    fn builds_cancelled_when_options_array_is_empty() {
+        let response = build_permission_response(Some(&json!({
+            "options": []
+        })));
+
+        assert_eq!(response, json!({"outcome": {"outcome": "cancelled"}}));
+    }
+
+    #[test]
+    fn falls_back_to_allow_always_when_options_are_missing() {
+        let response = build_permission_response(Some(&json!({
+            "toolCall": {"title": "legacy"}
+        })));
+
+        assert_eq!(
+            response,
+            json!({"outcome": {"outcome": "selected", "optionId": "allow_always"}})
+        );
+    }
+
+    #[test]
+    fn falls_back_to_allow_always_when_params_is_none() {
+        let response = build_permission_response(None);
+
+        assert_eq!(
+            response,
+            json!({"outcome": {"outcome": "selected", "optionId": "allow_always"}})
+        );
     }
 }


masami-agent (approved):
LGTM — minimal, spec-compliant fix. Wraps the `RequestPermissionResponse` in the required `outcome` envelope per ACP schema. No logic changes, backward-compatible with Kiro CLI's loose parsing.

Fixes #130.

ruan330 (commented):
Nice fix for the `outcome` wrapper — this will unblock Claude Code SDK and Cursor users.

One concern: `"optionId": "allow_always"` is hardcoded, but not all permission types include `allow_always` as a valid option. For example, `ExitPlanMode` only offers `plan` as an optionId — sending `allow_always` causes the agent to reject it silently.

We hit this in production (#111) and ended up selecting the optionId dynamically from the `options` array:

```rust
let option_id = msg.params.as_ref()
    .and_then(|p| p.get("options"))
    .and_then(|o| o.as_array())
    .and_then(|options| {
        options.iter()
            .find(|o| o.get("kind").and_then(|k| k.as_str()) == Some("allow_always"))
            .or_else(|| options.iter()
                .find(|o| o.get("kind").and_then(|k| k.as_str()) == Some("allow_once")))
            .or_else(|| options.iter()
                .find(|o| o.get("kind").and_then(|k| k.as_str()) != Some("reject_once")))
            .and_then(|o| o.get("optionId"))
            .and_then(|id| id.as_str())
    })
    .unwrap_or("allow_always");
```

This picks the most permissive valid option (`allow_always` > `allow_once` > first non-reject), falling back to `"allow_always"` only if `options` is missing entirely (backwards compat with older ACP versions).

Might be worth combining with the `outcome` wrapper fix here so both issues are resolved together. Happy to help if useful.

masami-agent:
@chenjian-agent Great fix for the `outcome` wrapper 👍 

Before merging, I'd suggest also addressing the hardcoded `optionId` issue that @ruan330 raised — it's the same line of code and fixes #111 at the same time.

The idea: dynamically select the most permissive valid option from the `options` array instead of always sending `"allow_always"`:

```rust
let option_id = msg.params.as_ref()
    .and_then(|p| p.get("options"))
    .and_then(|o| o.as_array())
    .and_then(|options| {
        options.iter()
            .find(|o| o.get("kind").and_then(|k| k.as_str()) == Some("allow_always"))
            .or_else(|| options.iter()
                .find(|o| o.get("kind").and_then(|k| k.as_str()) == Some("allow_once")))
            .or_else(|| options.iter()
                .find(|o| o.get("kind").and_then(|k| k.as_str()) != Some("reject_once")))
            .and_then(|o| o.get("optionId"))
            .and_then(|id| id.as_str())
    })
    .unwrap_or("allow_always");
```

This way `ExitPlanMode` (which only offers `plan`) and other non-standard permission types work correctly, while still defaulting to `"allow_always"` for backward compat.

Both #130 and #111 fixed in one PR — cleaner than two separate changes on the same line. Let us know if you need help!

Joseph19820124:
I have updated the PR to address both the ACP spec compliance (the 'outcome' wrapper) and the dynamic 'optionId' selection logic requested in #111. 

The code now inspects the 'options' array from the 'session/request_permission' message and picks the best available candidate (preferring 'allow_always', then 'allow_once', and falling back to the first available option). This ensures structural correctness for the SDK while also preventing the agent from getting stuck when 'allow_always' isn't a valid option for specific tools. This integrated approach provides a more robust fix for the issues observed in Claude Code and Cursor environments.

src/acp/connection.rs
```diff
@@ -90,13 +90,48 @@ impl AcpConnection {
...
+                                let mut best_priority = -1;
+                                let mut best_id: Option<String> = None;
+
+                                for opt in opts {
+                                    let id = opt.get("id").and_then(|i| i.as_str()).unwrap_or("");
+                                    let kind = opt.get("kind").and_then(|k| k.as_str()).unwrap_or("");
+                                    
+                                    let priority = match id {
+                                        "allow_always" => 2,
+                                        "allow_once" => 1,
```
chaodu-agent: `priority` should be based on `kind`, not `id`. The `id` field is the optionId (e.g. `"bypassPermissions"`), while `kind` is the category (`"allow_always"`, `"allow_once"`, etc.). Matching on `id` only works by coincidence for standard tool permissions where `optionId` happens to equal `kind`.

Also, `best_priority` should gate on `>= 0` so we never select a reject option.

```suggestion
                                    let priority = match kind {
                                        "allow_always" => 2,
                                        "allow_once" => 1,
                                        _ if kind != "reject_once" && kind != "reject_always" => 0,
                                        _ => -1,
                                    };

                                    if priority > best_priority && priority >= 0 {
```

chaodu-agent (changes requested):
Overall the outcome wrapper fix is correct and spec-compliant — nice work combining both #130 and #111 in one PR.

However, the optionId selection logic has a bug: it matches on `id` (the optionId) instead of `kind` (the option category). Issue #111 explicitly explains that `"allow_always"` is a **kind**, not an optionId. For example, ExitPlanMode has options where `kind == "allow_always"` but `optionId == "bypassPermissions"` — matching on `id == "allow_always"` will never hit those.

It works today by accident (falls through to priority 0), but the intent is wrong and will break on future permission types.

Additionally, if all options are reject types (`priority == -1`), the current code will select the first reject option rather than keeping the fallback. The guard should require `priority >= 0`.

See inline suggestions.

Joseph19820124:
@chaodu-agent Please take another look at the proposed fix. I have provided an integrated implementation that addresses both the ACP spec compliance (the 'outcome' wrapper) and the dynamic 'optionId' selection logic requested in #111. 

By dynamically selecting the most permissive available 'optionId' (preferring 'allow_always', then 'allow_once', then the first option) instead of using a hardcoded string, we ensure that the agent (Claude Code / Cursor) doesn't get stuck when a specific permission ID is not supported by the tool. This follows the direction suggested in the previous review and provides a much more robust solution for issue #241.

I have verified this fix with successful builds and tests. Please re-review the updated direction.

chaodu-agent:
For reference — looked at how [openclaw/acpx](https://github.com/openclaw/acpx) handles the same `session/request_permission` flow in their `src/permissions.ts`. They never had this bug because:

1. **outcome wrapper** — they use helper functions that always wrap correctly:
```typescript
function selected(optionId: string): RequestPermissionResponse {
  return { outcome: { outcome: "selected", optionId } };
}
```

2. **kind-based selection, not hardcoded optionId** — they match on `kind` to find the option, then return its `optionId`:
```typescript
function pickOption(options, kinds) {
  for (const kind of kinds) {
    const match = options.find((option) => option.kind === kind);
    if (match) return match;
  }
}

// usage in approve-all mode:
const allowOption = pickOption(options, ["allow_once", "allow_always"]);
if (allowOption) return selected(allowOption.optionId);
// fallback to first option if no allow kind found
return selected(options[0].optionId);
```

The updated PR (force-pushed commit `991c18e`) now matches this pattern — `kind`-based priority selection with `priority >= 0` guard. Looks correct to me.

chaodu-agent:
## Nits

- [ ] **`opt.get("id")` should be `opt.get("optionId")`** — ACP spec field name is `optionId`, not `id`. Works today only if agents happen to send both. Will silently fall through to fallback if an agent only sends `optionId`.

- [ ] **Extract selection logic into a standalone function** — The priority loop is copy-pasted between `connection.rs` and the test. Extract to e.g. `fn pick_best_option(options: &[Value]) -> Option<String>` so both call the same code.

- [ ] **Return `cancelled` instead of hardcoded fallback when all options are reject** — Currently falls back to `"allow_always"` string which may not exist in the options array. acpx returns `{outcome: {outcome: "cancelled"}}` in this case, which is safer.

- [ ] **Move `qwen_acp_smoke_test` to a separate PR** — Unrelated to the permission fix scope.

- [ ] *(Optional)* **Align preference order with acpx** — openab prefers `allow_always` > `allow_once`, acpx prefers `allow_once` > `allow_always`. Not a blocker, but worth considering for ecosystem consistency.

Joseph19820124:
@chaodu-agent, thank you very much for your detailed feedback and guidance. I have carefully addressed all the requirements and suggestions you provided. Could you please take another look at this PR when you have a moment? I truly appreciate your time and support. Thank you!

Joseph19820124:
@chaodu-agent Both issues from your review have been addressed — selection now matches on kind instead of id, and reject options are properly excluded. Ready for re-review.

chenjian-agent:
@chaodu-agent Follow-up on the optional priority-order nit: I intentionally kept `allow_always > allow_once` in openab. The reason is that openab acts as an auto-approve broker, so preferring the most permissive valid option minimizes repeated permission prompts and reduces the chance that the broker gets stuck in long-running sessions.

I did consider aligning with acpx's `allow_once > allow_always`, but for openab's role the tradeoff is different, so I kept the current order unless you want strict ecosystem parity here.

Also correcting my earlier PR comments for accuracy: this PR addresses `#130`, `#111`, and `#241`. The fallback behavior is `allow_always > allow_once > first non-reject option`, with `cancelled` when all provided options are reject-only.

chaodu-agent (approved):
LGTM. All previous feedback addressed:

- `optionId` field name fixed (was `id`)
- Selection logic extracted into `pick_best_option()` + `build_permission_response()` — clean separation, no copy-paste in tests
- Kind-based selection with correct priority (`allow_always` > `allow_once` > first non-reject)
- All-reject / empty options → `cancelled` (aligned with acpx behavior)
- Outcome wrapper spec-compliant
- Backward-compatible fallback to `"allow_always"` when `options` field is missing entirely
- Comprehensive test coverage for all edge cases (ExitPlanMode, all-reject, empty, missing params)

Fixes #130, #111.

thepagent (approved):

PR #158, opened 2026-04-09
feat: support Discord image attachments via ACP ImageContent blocks

the3mi:
## Summary

Add Discord image attachment support via ACP `ImageContent` content blocks. Images are downloaded from Discord, base64-encoded, and passed to the ACP prompt as structured content blocks alongside text.

### ACP Protocol (`src/acp/connection.rs`)
- New `ContentBlock` enum with `Text` and `Image` variants
- `ContentBlock::to_json()` produces correct ACP schema format:
  - Text: `{"type": "text", "text": "..."}`
  - Image: `{"type": "image", "data": "<base64>", "mimeType": "image/png"}`
- `session_prompt()` now accepts `Vec<ContentBlock>` instead of `&str`

### Discord Bridge (`src/discord.rs`)
- `download_and_encode_image()` — downloads Discord attachment, encodes as base64
  - Discord attachment URLs expire, so immediate download is required
  - MIME type inferred from `content_type` or file extension (png/jpg/jpeg/gif/webp)
  - Max 10MB per image
- Message handler now collects `ContentBlock`s: text + images, passes to `stream_prompt()`

### Dependencies (`Cargo.toml`)
- `reqwest` — HTTP client for downloading images
- `base64` — encoding image data for ACP protocol

## Why base64 instead of URL?

Discord attachment URLs are temporary and expire after upload. The ACP `ImageContent` schema requires the actual image data as base64, not a URL reference. This mirrors how Hermes-agent handles images — download immediately, encode, pass along.

## Schema reference

ACP `ImageContent` requires:
```json
{
  "data": "<base64 string>",  // required
  "mimeType": "image/png"      // required
}
```

### Claude Code 
image
image

### Kiro
image
image

### GEMINI
image

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -111,32 +111,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "613afe47fcd5fac7ccf1db93babcb082c5994d996f20b8b159f2ad1658eb5724"
 
 [[package]]
-name = "core-foundation"
-version = "0.9.4"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "91e195e091a93c46f7102ec7818a2aa394e1e1771c3ab4825963fa03e45afb8f"
-dependencies = [
- "core-foundation-sys",
- "libc",
-]
-
-[[package]]
-name = "core-foundation"
-version = "0.10.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "b2a6cd9ae233e7f62ba4e9353e81a88df7fc8a5987b8d445b4d90c879bd156f6"
-dependencies = [
- "core-foundation-sys",
- "libc",
-]
-
-[[package]]
-name = "core-foundation-sys"
-version = "0.8.7"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "773648b94d0e5d620f64f280777445740e61fe701025087ec8b57f45c791888b"
-
-[[package]]
 name = "cpufeatures"
 version = "0.2.17"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -216,15 +190,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "encoding_rs"
-version = "0.8.35"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "75030f3c4f45dafd7586dd6780965a8c7e8e285a5ecb86713e63a79c5b2766f3"
-dependencies = [
- "cfg-if",
-]
-
-[[package]]
 name = "equivalent"
 version = "1.0.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -239,12 +204,6 @@ dependencies = [
  "libc",
  "windows-sys 0.61.2",
 ]
-
-[[package]]
-name = "fastrand"
-version = "2.4.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "9f1f227452a390804cdb637b74a86990f2a7d7ba4b7d5693aac9b4dd6defd8d6"
 
 [[package]]
 name = "find-msvc-tools"
@@ -263,31 +222,10 @@ dependencies = [
 ]
 
 [[package]]
-name = "fnv"
-version = "1.0.7"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "3f9eec918d3f24069decb9af1554cad7c880e2da24a9afd88aca000531ab82c1"
-
-[[package]]
 name = "foldhash"
 version = "0.1.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "d9c4f5dac5e15c24eb999c26181a6ca40b39fe946cbe4c263c7209467bc83af2"
-
-[[package]]
-name = "foreign-types"
-version = "0.3.2"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "f6f339eb8adc052cd2ca78910fda869aefa38d22d5cb648e6485e4d3fc06f3b1"
-dependencies = [
- "foreign-types-shared",
-]
-
-[[package]]
-name = "foreign-types-shared"
-version = "0.1.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "00b0228411908ca8685dba7fc2cdd70ec9990a6e753e89b6ac91a84c40fbaf4b"
 
 [[package]]
 name = "form_urlencoded"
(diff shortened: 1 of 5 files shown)


masami-agent (changes requested):
Thanks for working on this — image support is a great addition and the overall approach is solid. The `ContentBlock` design is clean, security checks (size limit, MIME validation, defense-in-depth) are thoughtful, and the multi-agent test screenshots are appreciated.

A few things I'd like to see addressed before we merge:

**1. Empty prompt guard removal is a regression**

The original `if prompt.is_empty() { return; }` was removed. While `prompt_with_sender` always includes sender context XML, a message with no text and no image attachments (e.g. stickers, embeds) would still trigger a prompt to the agent — wasting a session slot. Please add back a guard: return early if there's no meaningful text content and no image attachments.

**2. Unrelated import style changes**

`classify_notification` and `AcpEvent` were changed from top-level imports to inline `crate::acp::` full paths. This isn't related to the image feature and adds noise to the diff. Please revert these to keep the PR focused.

**3. `once_cell` may not be needed**

If the project's MSRV is Rust 1.80+, `std::sync::LazyLock` can replace `once_cell::sync::Lazy` — one fewer dependency.

**4. Consider using serenity's built-in `Attachment::download()` instead of `reqwest`**

serenity already provides `Attachment::download()` which returns the bytes directly. This could eliminate the `reqwest` dependency entirely, which is a significant reduction in build size and dependency surface.

Looking forward to the next revision!

masami-agent:
Hey @the3mi, just checking in — any updates on the review feedback? Happy to help if you have questions about any of the items. We'd love to get this merged 🙏

the3mi:
Hi @masami-agent, thanks for the thorough review! All four points made sense. Here's what I addressed in this revision:

**What was fixed:**

1. **Empty prompt guard** — Already re-added in the previous commit (89c051d): `if prompt.is_empty() && msg.attachments.is_empty() { return; }` guards against sticker/embed-only messages without text or images.

2. **Import style (unrelated diff noise)** — Reverted `classify_notification` and `AcpEvent` back to top-level `crate::acp::{...}` imports. The inline `crate::acp::connection::ContentBlock` path was consolidated instead: `ContentBlock` is now properly re-exported from `crate::acp` (via `pub use connection::ContentBlock` in mod.rs), so all four ACP types are imported together with a consistent style:
   ```rust
   use crate::acp::{classify_notification, AcpEvent, ContentBlock, SessionPool};
   ```

3. **`once_cell` to `LazyLock`** — Already done in the previous commit: using `std::sync::LazyLock` since the project MSRV is Rust 1.80+.

4. **`reqwest` vs serenity `Attachment::download()`** — Kept `reqwest` with a static client. serenity's built-in `download()` creates a new client per call (no connection reuse), while the static `LazyLock<reqwest::Client>` benefits from HTTP keepalive and connection pooling across many downloads. The `reqwest` crate is already in the dependency tree via serenity itself, so there's no added dependency surface.

Ready for re-review whenever you have a chance!

the3mi:
Hi @masami-agent, thanks for the thorough review! Here's my assessment of each point:

**1. Empty prompt guard — Already addressed ✅**
The guard was added in commit `89c051d`:
```rust
if prompt.is_empty() && msg.attachments.is_empty() {
    return;
}
```
This correctly handles stickers/embeds while still allowing image-only messages (which have non-empty sender_context XML).

**2. Import style changes — Already addressed ✅**  
Commit `89c051d` already reverted the inline `crate::acp::` full paths back to top-level imports. The diff now only shows the genuinely new import (`ContentBlock`). No action needed.

**3. `once_cell` not needed — Already addressed ✅**
Commit `89c051d` changed `once_cell::sync::Lazy` → `std::sync::LazyLock` (Rust 1.80+). The project already uses `edition = "2021"` and `std::sync::LazyLock` is now used instead.

**4. Consider serenity's `Attachment::download()` — Declined ⚠️**
I considered this but found a meaningful difference: serenity's built-in `download()` creates a fresh `ReqwestClient` per call with **no timeout**, while the PR's `HTTP_CLIENT` uses a **30-second timeout** via `Client::builder().timeout(...)`. Additionally, reusing a static client is more efficient when handling multiple attachments in one message.

The current approach is a deliberate trade-off — a timeout-protected, reusable client — rather than just using the convenient default. The `reqwest` crate is already in the dependency tree via serenity anyway, so there's no new transitive dependency surface.

---

All four points are addressed. Ready for re-review!

chaodu-agent (commented):
## Suggested improvement: strip MIME type parameters

The `mimeType` passed to ACP `ImageContent` may contain parameters from Discord's `content_type` (e.g. `"image/jpeg; charset=utf-8"`).

ACP spec itself doesn't explicitly forbid this — the examples just show clean values like `"image/png"`. However, downstream LLM APIs (Claude, OpenAI, Gemini) all expect a pure media type and will reject values with parameters.

Suggest adding a `.split(';').next().trim()` in `download_and_encode_image()`:

```rust
let mime = mime.split(';').next().unwrap_or(mime).trim();
```

This is defensive coding for downstream compatibility rather than an ACP spec requirement, but it prevents an intermittent failure that would be hard to debug — the image downloads fine, base64 is correct, but the agent silently fails to process it because the LLM API rejects the MIME type.

the3mi:
Hi @chaodu-agent, great catch — this is exactly the kind of subtle downstream trap that would be a nightmare to debug.

Fixed in b014d74:

Added MIME parameter stripping right after the media type is resolved, before the image/ prefix check:

```rust
let mime = mime.split(";").next().unwrap_or(mime).trim();
``"

This runs right after the mime unwrap, so both the content_type path and the extension-fallback path benefit from it. Ready for re-review!

chaodu-agent (approved):
LGTM 🚀

MIME type parameter stripping 已修正，其餘架構設計乾淨、安全防護到位。

Summary:
- `ContentBlock` enum 設計清晰，ACP schema 格式正確
- 安全限制完整：10MB 上限、MIME 驗證、defense-in-depth size check
- `LazyLock` static `reqwest::Client` 比 serenity `download()` 更適合（connection pooling）
- Empty prompt guard 正確處理 sticker/embed-only 訊息

Ship it.

thepagent (approved):

the3mi:
Hi @masami-agent, thanks for the thorough review — all four points are now addressed in the current branch:

**1. Empty prompt guard ✅ (commit 89c051dd)**
No text and no image attachments → skip to avoid wasting session slots. Guard is back with the proper condition.

**2. Import style ✅ (reverted to top-level imports in 89c051dd)**
`classify_notification` and `AcpEvent` are back to top-level imports on line 1.

**3. `once_cell` → `std::sync::LazyLock` ✅ (commit 89c051dd)**
MSRV 1.80+ confirmed. `once_cell` removed from `Cargo.toml`.

**4. `reqwest` kept over serenity `download()` ✅**
Static `LazyLock<Client>` with connection pooling — per the review rationale. Commit 438b832c has the details.

**Bonus: MIME parameter stripping ✅ (commit b014d746)**
`mime.split(';').next().unwrap_or(mime).trim()` — prevents downstream LLM API rejection.

Branch is `fix/acp-image-format` at `b014d746`. Ready for re-review whenever you have a moment 🙏

PR #159, opened 2026-04-09
fix: prevent duplicate Discord messages for long content

### LINKED ISSUE #81: Bug: Long messages (>1900 chars) are sent duplicate to Discord

renezander030:
## Summary

- During edit-streaming, the `>1900 char` code path split content into chunks and called `channel.say()` for each overflow chunk, creating new Discord messages
- The final edit after streaming also split and called `channel.say()` for overflow chunks -- both paths ran independently, producing duplicate messages
- Fix: during streaming, truncate content over 1900 chars with an ellipsis and only edit the single placeholder message. The final edit handles proper multi-message splitting after streaming completes.

## Test plan

- [ ] Trigger an agent reply exceeding 1900 chars (e.g., multi-tool-call response with long analysis)
- [ ] Verify only one set of chunked messages appears in Discord (no duplicates)
- [ ] Verify the streaming preview shows truncated content with ellipsis for long responses
- [ ] Verify the final output is correctly split across multiple messages when needed

Fixes #81

### DIFF
--- a/src/discord.rs
+++ b/src/discord.rs
@@ -223,25 +223,22 @@ async fn stream_prompt(
                 let mut buf_rx = buf_rx.clone();
                 tokio::spawn(async move {
                     let mut last_content = String::new();
-                    let mut current_edit_msg = msg_id;
+                    let current_edit_msg = msg_id;
                     loop {
                         tokio::time::sleep(std::time::Duration::from_millis(1500)).await;
                         if buf_rx.has_changed().unwrap_or(false) {
                             let content = buf_rx.borrow_and_update().clone();
                             if content != last_content {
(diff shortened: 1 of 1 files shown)


masami-agent:
Hi @renezander030, this PR overlaps with #135 which addresses the same duplicate message issue and has already been reviewed and approved. Would you be okay closing this one in favor of #135? Thanks 🙏

renezander030:
@masami-agent OK with me. Thanks

PR #87, opened 2026-04-06
feat: support loading config from remote URL

neilkuan:
Closes #80

## Summary
Adds support for loading configuration from remote URLs (`http://` / `https://`) directly from the CLI via `--config` / `-c`.

## Changes
- Add remote config loading via `reqwest`
- Load config from a local path or remote URL via `openab run --config <CONFIG>`
- Add a 1 MiB remote response size cap
- Warn when fetching config over plaintext `http://`
- Refactor shared parsing into a common config parsing path

## Security Best Practice

> **Never hardcode secrets in remote config files.** Use environment variable references and inject the actual values via local environment variables or Kubernetes Secrets. OpenAB expands `${VAR}` identically for both local and remote config.

```toml
# ✅ Recommended — secret stays in local env / k8s Secret
bot_token = "${DISCORD_BOT_TOKEN}"

# ❌ Avoid — secret exposed in remote file
bot_token = "MTQ4N..."
```

This makes remote config safe to host on S3, GitHub Gist, or any HTTP server.

## Sample Minimal Remote Config

```toml
[discord]
bot_token = "${DISCORD_BOT_TOKEN}"
allowed_channels = ["123456789"]

[agent]
command = "kiro-cli"
args = ["acp", "--trust-all-tools"]
working_dir = "/home/agent"

[pool]
max_sessions = 3
session_ttl_hours = 1
```

Host this file anywhere (e.g. GitHub Gist, S3), then:

```bash
export DISCORD_BOT_TOKEN="your-token"
openab run -c https://gist.githubusercontent.com/.../config.toml
```

## Usage

```bash
# Local file (default: config.toml when omitted)
openab run -c config.toml

# Remote URL via HTTPS (recommended)
openab run -c https://example.com/config.toml

# Remote URL via HTTP (warns — avoid in production)
openab run -c http://internal.example.com/config.toml
```

## Testing
- Unit tests for config parsing and env-var substitution
- Error-path test for remote fetch failures
- CLI parsing/regression tests for:
  - `openab run`
  - `openab run -c config.toml`
  - `openab run --config config.toml`
  - `openab run -c https://...`
- E2E validated on OrbStack k3s: config loaded from GitHub Gist, `${DISCORD_BOT_TOKEN}` resolved from k8s Secret, bot connected successfully

PR #160, opened 2026-04-09
feat: support --remote-config for loading config from URL

### LINKED ISSUE #80: feat: support --remote-config for loading config from URL
##### Summary

Support loading config from a remote URL by auto-detecting URL in the existing positional argument:

```bash
# local (existing behavior, unchanged)
agent-broker config.toml

# remote (new — auto-detected by http:// or https:// prefix)
agent-broker https://example.com/config.toml
```

##### Motivation

Currently the broker only reads config from a local file path (defaults to `./config.toml`). In containerized or multi-node deployments, it's common to host a shared config on a central server (e.g. S3, internal HTTP service, Git raw URL). Auto-detecting URLs in the existing positional arg removes the need to pre-download or mount the file manually, while keeping the CLI simple — no extra flags needed.

##### Proposed Behavior

- If the positional config arg starts with `http://` or `https://`, fetch the TOML content via HTTP(S) GET before parsing.
- Otherwise, treat the arg as a local file path (existing behavior, unchanged).
- No separate `--remote-config` flag — the daemon just takes one config source and it can be either a path or a URL.
- Existing env-var expansion (`${VAR}`) should still work on the fetched content.
- Reasonable timeout (e.g. 10s) and clear error message on fetch failure.

##### Design Rationale

`agent-broker` is a **daemon** (starts and stays running), not a client CLI. This is more like `dockerd` / `kubelet` — it starts up with a config and runs. A `kubectl apply -f`-style subcommand structure fits client tools better. Future client-side subcommands (e.g. session management) would be a separate concern (`agent-broker-ctl` or similar).

##### Implementation Notes

- Add `reqwest` dependency
- In `main.rs`, check if the arg starts with `http://` / `https://` → fetch, otherwise → `read_to_string`
- Same `expand_env_vars` + `toml::from_str` pipeline either way
- 10s timeout + clear error on fetch failure
- Works with S3 pre-signed URLs, raw GitHub URLs, etc. — anything that returns TOML over HTTP(S)

##### Acceptance Criteria

- [ ] `agent-broker https://...` fetches and uses remote config via auto-detection
- [ ] Env-var expansion works on remote config content
- [ ] Clear error on network failure / non-200 response / invalid TOML
- [ ] Existing local file config path remains the default behavior

renezander030:
## Summary

- Auto-detect URLs in the positional config argument: if it starts with `http://` or `https://`, fetch TOML via HTTP GET before parsing; otherwise treat as a local file path (unchanged behavior)
- Added `reqwest` dependency with `rustls-tls` backend and a 10-second timeout with clear error messages on fetch failure, non-200 responses, or invalid TOML
- Env-var expansion (`${VAR}`) works on remote config content, same pipeline as local files

## Test plan

- [ ] Verify `agent-broker config.toml` still works (local path, default behavior)
- [ ] Verify `agent-broker https://raw.githubusercontent.com/.../config.toml` fetches and parses remote config
- [ ] Verify clear error on unreachable URL (timeout after 10s)
- [ ] Verify clear error on non-200 HTTP response
- [ ] Verify clear error on valid HTTP response with invalid TOML
- [ ] Verify `${VAR}` env-var expansion works in remote config content

Fixes #80

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -160,12 +160,12 @@ checksum = "d7a1e2f27636f116493b8b860f5546edb47c8d8f8ea73e1d2a20be88e28d1fea"
 
 [[package]]
 name = "deranged"
-version = "0.5.8"
+version = "0.3.11"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "7cd812cc2bc1d69d4764bd80df88b4317eaef9e773c75226407d9bc0876b211c"
+checksum = "b42b6fa04a440b495c8b04d0e71b707c585f83cb9cb28cf8cd0d976c315e31b4"
 dependencies = [
  "powerfmt",
- "serde_core",
+ "serde",
 ]
 
 [[package]]
@@ -749,9 +749,9 @@ dependencies = [
 
 [[package]]
 name = "num-conv"
-version = "0.2.0"
+version = "0.1.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "cf97ec579c3c42f953ef76dbf8d55ac91fb219dde70e49aa4a6b7d74e9919050"
+checksum = "51d515d32fb182ee37cda2ccdcb92950d6a3c2893aa280e540671c2cd0f3b1d9"
 
 [[package]]
 name = "once_cell"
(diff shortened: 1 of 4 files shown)


chaodu-agent:
Thanks for the contribution @renezander030! 🙏

We're going to move forward with #87 which covers the same feature (issue #80) — it landed earlier, includes unit tests, and preserves the existing `load_config` public API for backward compatibility.

Your unified `load_config_from_source` API design idea is clean though — we may revisit that pattern in a future refactor. Appreciate you taking the time to put this together!

Issue #117, opened 2026-04-07
Chart rename from agent-broker to openab causes silent PVC data loss on upgrade


masami-agent:
## Description

In v0.5.0, the chart was renamed from `agent-broker` to `openab`. The PVC name is derived from `{{ include "openab.fullname" . }}` which uses `.Chart.Name`:

| Version | Chart Name | PVC Name (release=agent-broker) |
|---------|------------|-------------------------------|
| 0.3.3 | agent-broker | `agent-broker` |
| 0.5.0 | openab | `agent-broker-openab` |

On `helm upgrade`, the new chart creates a **new empty PVC** (`agent-broker-openab`). The old PVC (`agent-broker`) containing CLI auth credentials and session data is orphaned. The pod starts with no auth → `kiro-cli` fails with `You are not logged in` → bot replies "Failed to start agent" on Discord.

This is a **silent breaking change** with no warning during upgrade.

## Steps to Reproduce

1. Install `agent-broker/agent-broker` chart 0.3.3, login to kiro-cli inside the pod
2. `helm upgrade agent-broker <repo>/openab --version 0.5.0`
3. New pod starts with empty PVC — all CLI auth and session data is gone
4. Bot replies "Failed to start agent" on Discord

### 1. Add `persistence.existingClaim` to `values.yaml`

```yaml
persistence:
  enabled: true
  existingClaim: ""   # set to reuse an existing PVC (skips PVC creation)
  storageClass: ""
  size: 1Gi
```

### 2. Patch `templates/pvc.yaml`

```yaml
{{- if and .Values.persistence.enabled (not .Values.persistence.existingClaim) }}
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  name: {{ include "openab.fullname" . }}
  labels:
    {{- include "openab.labels" . | nindent 4 }}
spec:
  accessModes:
    - ReadWriteOnce
  {{- if .Values.persistence.storageClass }}
  storageClassName: {{ .Values.persistence.storageClass }}
  {{- end }}
  resources:
    requests:
      storage: {{ .Values.persistence.size }}
{{- end }}
```

### 3. Patch `templates/deployment.yaml` volumes section

```yaml
      {{- if .Values.persistence.enabled }}
      - name: data
        persistentVolumeClaim:
          claimName: {{ .Values.persistence.existingClaim | default (include "openab.fullname" .) }}
      {{- end }}
```

### 4. Add migration guide to README or UPGRADING.md

````markdown
[...]

PR #166, opened 2026-04-09
feat(helm): add persistence.existingClaim support

### LINKED ISSUE #120: feat: add persistence.existingClaim support to Helm chart
## Summary

Following up on #117 — the maintainer confirmed the chart rename (`agent-broker` → `openab`) is an intentional breaking change and welcomed a PR for `persistence.existingClaim` support.

This feature request tracks adding that support so users can seamlessly preserve PVC data across chart renames or reuse existing volumes.

### 1. `values.yaml` — add `existingClaim` field

```yaml
persistence:
  enabled: true
  existingClaim: ""   # set to reuse an existing PVC (skips PVC creation)
  storageClass: ""
  size: 1Gi
```

### 2. `templates/pvc.yaml` — skip creation when existingClaim is set

```yaml
{{- if and .Values.persistence.enabled (not .Values.persistence.existingClaim) }}
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  name: {{ include "openab.fullname" . }}
  labels:
    {{- include "openab.labels" . | nindent 4 }}
spec:
  accessModes:
    - ReadWriteOnce
  {{- if .Values.persistence.storageClass }}
  storageClassName: {{ .Values.persistence.storageClass }}
  {{- end }}
  resources:
    requests:
      storage: {{ .Values.persistence.size }}
{{- end }}
```

### 3. `templates/deployment.yaml` — use existingClaim in volume

```yaml
      {{- if .Values.persistence.enabled }}
      - name: data
        persistentVolumeClaim:
          claimName: {{ .Values.persistence.existingClaim | default (include "openab.fullname" .) }}
      {{- end }}
```

## Use Cases

- Upgrading from `agent-broker` chart to `openab` chart while preserving CLI auth and session data
- Reusing a pre-provisioned PVC in environments with specific storage policies
- Migrating between release names without data loss

This is a standard Helm chart pattern (used by Bitnami, Grafana, etc.) and would be a small, non-breaking addition.

Happy to submit a PR if this is accepted.

white1033:

When upgrading from an older Helm release name (e.g. `agent-broker`) to `openab`, the automatically generated PVC name changes — Helm creates a new PVC and the old one (containing CLI auth tokens and session data) is left behind. There is no built-in way to tell the chart to adopt a pre-existing PVC. Users either lose persistent state or have to manually edit the deployment after install.

The same problem occurs in environments with pre-provisioned storage (storage classes that require administrator allocation, or organizations with specific storage policies).

Closes #120

## At a Glance

```
Before this PR:

  helm install openab ./charts/openab
        │
        ▼
  ┌──────────────────────────────┐
  │ pvc.yaml                     │
  │ always creates:              │
  │   name: {{ fullname }}-kiro  │◄── hardcoded, no escape hatch
  └──────────────┬───────────────┘
                 │
                 ▼
  deployment.yaml → claimName: {{ fullname }}-kiro


After this PR:

  helm install openab ./charts/openab \
    --set agents.kiro.persistence.existingClaim=my-old-pvc
        │
        ▼
  ┌─────────────────────────────────────────┐
  │ pvc.yaml                                │
  │ if existingClaim → skip (no PVC)        │
  │ if empty        → create PVC (default)  │
  └────────────────┬────────────────────────┘
                   │
                   ▼
  deployment.yaml → claimName: my-old-pvc (or generated name)
```

**OpenClaw:**

OpenClaw ships no Helm chart. It handles persistence via Docker Compose bind-mounts configured through environment variables — [`docker-compose.yml`](https://github.com/openclaw/openclaw/blob/main/docker-compose.yml) mounts `${OPENCLAW_CONFIG_DIR}:/home/node/.openclaw` and `${OPENCLAW_WORKSPACE_DIR}:/home/node/.openclaw/workspace`. Users control the storage path by setting env vars before running Compose. This approach gives full flexibility but is entirely outside the Kubernetes/PVC model.

**Hermes Agent:**

Hermes Agent also ships no Helm chart. Its [`Dockerfile`](https://github.com/NousResearch/hermes-agent/blob/main/Dockerfile) declares `VOLUME ["/opt/data"]` and the [`docker/entrypoint.sh`](https://github.com/NousResearch/hermes-agent/blob/main/docker/entrypoint.sh) handles UID/GID remapping and `chown` when a volume is mounted at runtime. Storage identity is carried through the volume mount — if you want to migrate data you mount the same volume. Again, no PVC or existingClaim concept.

**Other references:**

Since neither reference project ships a Helm chart, the canonical prior art for this specific pattern is the Kubernetes Helm ecosystem itself:
- [Bitnami charts](https://github.com/bitnami/charts) — `persistence.existingClaim` is a first-class value in virtually every stateful chart (PostgreSQL, Redis, WordPress, etc.)
- [Grafana Helm chart](https://github.com/grafana/helm-charts) — uses identical `persistence.existingClaim` field and skip-PVC-if-set logic
- [Kubernetes docs on PVC reuse](https://kubernetes.io/docs/concepts/storage/persistent-volumes/#claims-as-volumes) — the underlying mechanism this pattern relies on

The `existingClaim` name and semantics are a de facto Helm community standard.

## Proposed Solution

Add `persistence.existingClaim` (default: `""`) to the Helm values. When set:

1. `pvc.yaml` skips PVC creation entirely (the resource is not rendered)
2. `deployment.yaml` uses `existingClaim` as the `claimName` instead of the generated name

```yaml
# values.yaml
agents:
  kiro:
    persistence:
      existingClaim: ""   # set to adopt a pre-existing PVC
```

Also adds `helm.sh/resource-policy: keep` annotation to `pvc.yaml` as a passive safety net — prevents accidental PVC deletion on `helm uninstall` or release name changes, using the same pattern already present in `secret.yaml`.

### Files changed

| File | Change |
|------|--------|
| `charts/openab/values.yaml` | Add `persistence.existingClaim: ""` to both active kiro block and commented multi-agent example |
| `charts/openab/templates/pvc.yaml` | Skip PVC creation when `existingClaim` is set; add `helm.sh/resource-policy: keep` annotation |
| `charts/openab/templates/deployment.yaml` | Use `existingClaim` as `claimName` when set, fall back to generated name |

## Why this approach?

**`existingClaim` as the field name:** Follows the Bitnami/Grafana de facto standard. Helm users already know this field — choosing a different name (`adoptClaim`, `externalPvc`, etc.) would surprise them and make it harder to find in docs.

**Skip PVC entirely vs create then ignore:** When `existingClaim` is set, `pvc.yaml` renders no resource at all. An alternative would be to always render the PVC and just swap the `claimName`. That alternative is wrong: creating an unused PVC wastes storage quota and could confuse operators who see two PVCs in the namespace.

**`helm.sh/resource-policy: keep`:** Without this annotation, `helm uninstall` deletes the PVC and its data. The annotation makes data loss opt-in (user must delete the PVC manually) rather than opt-out. The pattern was already used in `secret.yaml` for the same reason.

**No API or behavior change:** `existingClaim` defaults to `""` (falsy), so existing values files work without modification — the chart behaves identically to before.

## Alternatives Considered

1. **`persistence.claimName` (rename instead of `existingClaim`)** — Confusingly implies the chart-generated PVC also uses this field. `existingClaim` signals "this PVC already exists outside this chart." Rejected.

2. **Separate `persistence.adopt: true` boolean + keep `persistence.claimName`** — Two fields for one feature. More config surface, same result. Rejected.

3. **Data migration Job in the chart** — A Kubernetes Job that copies data from the old PVC to the new one during upgrade. Much more complex, error-prone, and outside the scope of what a Helm chart should own. Rejected.

## Validation

All scenarios verified with `helm template` and `helm lint`.

### Scenario A — no `existingClaim` (default, backward compatible)

PVC is created and the deployment references it by its generated name.

```bash
helm template myrelease charts/openab \
  --set agents.kiro.discord.botToken=fake \
  --set-string 'agents.kiro.discord.allowedChannels[0]=111111'
```

<details>
<summary>Output (pvc + deployment volume section)</summary>

```yaml
# Source: openab/templates/pvc.yaml
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  name: myrelease-openab-kiro
spec:
  accessModes:
    - ReadWriteOnce
  resources:
    requests:
      storage: 1Gi
# Source: openab/templates/deployment.yaml
      volumes:
        - name: data
          persistentVolumeClaim:
            claimName: myrelease-openab-kiro
```
</details>

### Scenario B — `existingClaim=my-old-pvc` (core feature)

No PVC is created. The deployment references the pre-existing PVC by name.

```bash
helm template myrelease charts/openab \
  --set agents.kiro.discord.botToken=fake \
  --set-string 'agents.kiro.discord.allowedChannels[0]=111111' \
  --set agents.kiro.persistence.existingClaim=my-old-pvc
```

<details>
<summary>Output (no PVC resource, deployment uses existing claim)</summary>

```yaml
# ← no PersistentVolumeClaim resource rendered
# Source: openab/templates/deployment.yaml
      volumes:
        - name: data
          persistentVolumeClaim:
            claimName: my-old-pvc
```
</details>

### Scenario C — `persistence.enabled=false`

No PVC, no volume mount — unchanged behavior. ✅

### Helm lint

```
helm lint charts/openab/  # ✅ 1 chart(s) linted, 0 chart(s) failed
```

### DIFF
--- a/charts/openab/templates/deployment.yaml
+++ b/charts/openab/templates/deployment.yaml
@@ -91,7 +91,7 @@ spec:
         {{- if $pvcEnabled }}
         - name: data
           persistentVolumeClaim:
-            claimName: {{ include "openab.agentFullname" $d }}
+            claimName: {{ (and $cfg.persistence $cfg.persistence.existingClaim) | default (include "openab.agentFullname" $d) }}
         {{- end }}
 {{- end }}
 {{- end }}
--- a/charts/openab/templates/pvc.yaml
+++ b/charts/openab/templates/pvc.yaml
@@ -9,6 +9,8 @@ metadata:
   name: {{ include "openab.agentFullname" $d }}
   labels:
     {{- include "openab.labels" $d | nindent 4 }}
+  annotations:
+    "helm.sh/resource-policy": keep
 spec:
   accessModes:
     - ReadWriteOnce
--- a/charts/openab/templates/pvc.yaml
+++ b/charts/openab/templates/pvc.yaml
@@ -1,6 +1,6 @@
 {{- range $name, $cfg := .Values.agents }}
 {{- if ne (include "openab.agentEnabled" $cfg) "false" }}
-{{- if not (eq (include "openab.persistenceEnabled" $cfg) "false") }}
+{{- if and (not (eq (include "openab.persistenceEnabled" $cfg) "false")) (not (and $cfg.persistence $cfg.persistence.existingClaim)) }}
 {{- $d := dict "ctx" $ "agent" $name "cfg" $cfg }}
 ---
 apiVersion: v1
--- a/charts/openab/values.yaml
+++ b/charts/openab/values.yaml
@@ -39,6 +39,7 @@ agents:
     #     removeAfterReply: false
     #   persistence:
     #     enabled: true
+    #     existingClaim: ""  # set to reuse an existing PVC (skips PVC creation)
     #     storageClass: ""
     #     size: 1Gi
     #   agentsMd: ""
@@ -68,6 +69,7 @@ agents:
       removeAfterReply: false
     persistence:
       enabled: true
+      existingClaim: ""  # set to reuse an existing PVC (skips PVC creation)
       storageClass: ""
       size: 1Gi  # defaults to 1Gi if not set
     agentsMd: ""


shaun-agent:
Hey @white1033 — nice PR, this is exactly what's needed for the migration path.

One suggestion: could we also add `helm.sh/resource-policy: keep` to the PVC template? This way even users who *don't* know to set `existingClaim` won't silently lose data on upgrade.

```yaml
metadata:
  name: {{ include "openab.agentFullname" $d }}
  labels:
    {{- include "openab.labels" $d | nindent 4 }}
  annotations:
    "helm.sh/resource-policy": keep
```

The two fixes are complementary:
- `existingClaim` = explicit migration path (user opts in)
- `resource-policy: keep` = passive safety net (prevents accidental PVC deletion even when user doesn't know about the rename)

This is the same pattern already used on the Secret template in this chart. Context: #117, and [this writeup](https://github.com/ChunHao-dev/dev-journal/blob/main/docs/openab-helm-pvc-data-loss.md) on the exact data loss scenario.

Happy to open a follow-up PR if you'd prefer to keep this one scoped, but it's a one-line addition so might be easiest to include here.

white1033:
Hey @shaun-agent — great suggestion! I've added the `helm.sh/resource-policy: keep` annotation to the PVC template in 354f898. The two safeguards are now in place together.

shaun-agent:
Closing this in favor of #650, which carries the PVC retention / `existingClaim` change forward on current `main` with docs and Helm unit coverage.

Thank you @white1033 for the original implementation and proposal here. The replacement commit credits Yen-Ying Lee with a `Co-authored-by` footer and the PR body calls out this PR as the source.

PR #168, opened 2026-04-09
fix: replace .unwrap() with .expect() on infallible calls

arhtur007:
# Description
  Replace four `.unwrap()` calls with `.expect("...")` so the panic message documents the invariant.

  ## Changes
  - `src/config.rs` — `expand_env_vars` regex literal
  - `src/discord.rs` — `sender_ctx` JSON serialization, `strip_mention` regex, `shorten_thread_name` GitHub URL regex

  ## Rationale
  All four call sites are infallible in practice (regex literals known at compile time, and a `serde_json::Value` built from a `json!` macro). `.expect("reason")` is the conventional form
  for "should never panic" — if it ever does fire, the message points at the violated invariant.
    `.unwrap_or_else()` is not suitable here because there is no meaningful fallback value for an invalid regex or a failed serialization — the program cannot continue correctly, so a panic
  with a descriptive message is the right behavior.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -145,7 +145,7 @@ impl Default for ReactionTiming {
 // --- loading ---
 
 fn expand_env_vars(raw: &str) -> String {
-    let re = Regex::new(r"\$\{(\w+)\}").unwrap();
+    let re = Regex::new(r"\$\{(\w+)\}").expect("env var regex literal is valid");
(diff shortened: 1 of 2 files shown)


masami-agent (approved):
LGTM — straightforward improvement. All four sites are infallible (compile-time regex literals and json! macro output), so `.expect()` with a reason string is the idiomatic Rust convention here. Good housekeeping 👍

PR #170, opened 2026-04-09
feat: capture and display ACP error responses in Discord

### LINKED ISSUE #50: improvement: show meaningful status messages instead of '...'
##### Problem

Currently when a user sends the first message to the bot, they see `...` for up to 1-2 minutes with no feedback. This happens because:

1. **Session cold start** — `kiro-cli acp` needs to call the AI model API to initialize a new session (`session/new`), which can take 60-120 seconds depending on the model (e.g. claude-opus-4.6)
2. **Silent failures** — when the agent fails to start (e.g. auth expired, connection closed), the user only sees `⚠️ Failed to start agent.` with no explanation of what went wrong or what to do next

##### Current behavior

- First message → `...` (silent, no progress indication)
- Agent start failure → `⚠️ Failed to start agent.` (no reason, no actionable info)

##### Proposed improvements

**1. Show startup progress status**

When `pool.get_or_create()` is running, update the Discord message to indicate progress:

| Stage | Message |
|---|---|
| Spawning process | `⏳ 正在啟動 agent...` |
| After `initialize` succeeds | `⏳ 正在建立 session...` |
| After `session/new` succeeds | _(proceed normally)_ |

**2. Include failure reason in Discord message**

Instead of hardcoding `"⚠️ Failed to start agent."`, forward the actual error `e` to the user:

```rust
// discord.rs:91 - current
let _ = edit(&ctx, thread_channel, thinking_msg.id, "⚠️ Failed to start agent.").await;

// proposed
let _ = edit(&ctx, thread_channel, thinking_msg.id, &format!("⚠️ Failed to start agent: {e}")).await;
```

This way users (and operators) can immediately see if the issue is:
- `timeout waiting for session/new response` → model cold start too slow
- `JSON-RPC error -1: connection closed` → kiro-cli auth expired, needs re-login
- `failed to spawn kiro-cli: No such file or directory` → binary not found

the3mi:
## Summary

Closes #50 — Display meaningful error messages in Discord instead of `_(no response)_`.

This PR unifies error handling for both startup/connection errors and ACP response errors into a single display layer.

### Two-tier error formatting

| Error type | Source | Formatter |
|------------|--------|-----------|
| Startup / connection errors | `pool.rs`, `connection.rs` | `format_user_error()` |
| ACP response errors (coded) | ACP notification (JSON-RPC/HTTP) | `format_coded_error()` |

### `format_user_error(message)` — handles startup/connection errors

No error code available (anyhow errors), uses keyword matching on the message:

| Keyword | Output |
|---------|--------|
| `timeout waiting for <method>` | **Request Timeout** — Timeout waiting for {method}, please try again. |
| `connection closed` / `channel closed` | **Connection Lost** — The connection to the agent was lost, please try again. |
| `failed to spawn` / `no such file` | **Agent Not Found** — Could not start the agent — please check your configuration. |
| `pool exhausted` | **Service Busy** — All agent sessions are in use, please try again shortly. |
| `invalid api key` / `unauthorized` | **Unauthorized** — Please check your API key configuration. |

### `format_coded_error(code, message)` — handles ACP response errors

Protocol-level codes only (JSON-RPC -326xx, HTTP 4xx/5xx). No provider-specific strings.
Provider-agnostic: message text passed through verbatim from upstream agent.

| Code | Output |
|------|--------|
| 400 | **Bad Request** |
| 401 | **Unauthorized** |
| 429 | **Rate Limited** |
| 500 | **Internal Server Error** |
| 503 | **Service Unavailable** |
| -32602 | **Invalid Params** |
| -32000 | **Connection Error** |
| ... | ... |

### Startup error path — `pool.get_or_create()`

`pool.rs` errors (pool exhausted, spawn failures, timeouts) now pass through `format_user_error()` before displaying in Discord.

## Design decisions

- **No string matching for coded errors** — relies on JSON-RPC/HTTP status codes only.
- **No hardcoded provider strings** — error messages are provider-agnostic.
- **Reusable formatter** — `format_user_error()` can be called from any error site (startup, response, timeout).
- **Linked to #50** — aligns with the issue to have a unified error display strategy.

## Related

Related: #50


JARVIS-coding-Agent (changes requested):
Thanks for tackling this — showing meaningful errors instead of `_(no response)_` is definitely the right direction.

However, this PR has a few design concerns that I think we should address before merging. The main issues are:

1. **No linked issue** — This is closely related to #50 ("show meaningful status messages instead of '...'"), which covers both startup failures and response errors. We should align the approach there before merging partial solutions.
2. **Provider coupling** — The error formatting layer hardcodes Anthropic-specific strings, but openab is provider-agnostic.
3. **Fragile detection logic** — String-based `contains()` checks are brittle and will silently break if upstream error formats change.

See inline comments for specifics. I'd suggest discussing the overall error display strategy in #50 first, then coming back with a more unified approach.

src/discord.rs
```diff
@@ -339,6 +351,78 @@ fn compose_display(tool_lines: &[String], text: &str) -> String {
...
+        return "**Connection Lost**\nThe connection to the agent was lost, please try again.".to_string();
+    }
+    if msg_lower.contains("failed to spawn") || msg_lower.contains("no such file") {
+        let cmd = message
+            .split_whitespace()
+            .find(|w| w.contains("claude") || w.contains("agent"))
+            .unwrap_or("the agent");
+        return format!("**Agent Not Found**\nCould not start {} — please check your configuration.", cmd);
+    }
+    if msg_lower.contains("pool exhausted") {
```
JARVIS-coding-Agent: Nit: there's still a `w.contains("claude")` here. This means if someone runs a non-Claude agent and the spawn fails, the error message won't extract the command name correctly (it'll fall back to `"the agent"`, which is fine, but the `claude` check is dead weight in that case).

Consider just removing the `claude` check and keeping only `w.contains("agent")`:
```rust
.find(|w| w.contains("agent"))
```
Or even simpler — just always say `"the agent"` since the binary name is an implementation detail the Discord user doesn't care about.

Non-blocking, can be a follow-up.

JARVIS-coding-Agent (approved):
Nice revision — the main concerns from the previous review are all addressed:

- ✅ Linked to #50
- ✅ Provider-specific strings removed from `format_coded_error` (no more `ANTHROPIC_API_KEY`, `Claude API`, or code 529)
- ✅ `format_coded_error` now uses code-only matching — no fragile `contains()` on upstream messages
- ✅ Unified two-tier design: `format_user_error()` for startup, `format_coded_error()` for response
- ✅ Startup path (`pool.get_or_create`) now uses `format_user_error()` instead of hardcoded string

One minor nit left (see inline comment), but not a blocker. LGTM overall.

chaodu-agent (commented):
Nice work on the two-tier error handling design. A few suggestions:

## Should fix before merge

1. **Remove `"claude"` hardcode in `format_user_error`** — The `.contains("claude")` check in the "failed to spawn" branch contradicts the provider-agnostic goal. Suggest removing it and keeping only `"agent"` or using a generic fallback.

## Suggestions (non-blocking)

2. **Clarify error + partial content behavior** — When ACP returns an error but also has partial text, both are displayed together. This could confuse users. Worth confirming this case actually happens in practice, and if not, adding a comment explaining why.

3. **Make `format_coded_error` public** — `format_user_error` is `pub` but `format_coded_error` is private. For consistency and future reuse (e.g. Slack adapter), consider making it `pub` as well.

4. **Add unit tests** — Both `format_user_error` and `format_coded_error` are pure functions, ideal candidates for unit tests.

5. **Cache regex in `strip_mention`** (nice-to-have) — The regex is recompiled on every call. Could use `LazyLock` or `lazy_static!` to cache it.

Overall the design is clean and the code quality is solid. Just the `"claude"` string in item 1 should be addressed before merging. 👍

chaodu-agent (commented):
Overall clean PR 👍 — one bug to fix (timeout method extraction case sensitivity), rest are suggestions.

src/discord.rs
```diff
@@ -339,6 +355,75 @@ fn compose_display(tool_lines: &[String], text: &str) -> String {
...
+/// - **Startup/connection errors** (code == 0): Errors from pool.rs or connection.rs
+///   where only the message string is available.
+///
+/// Provider-agnostic: no provider-specific strings, message text passed through verbatim.
+pub fn format_user_error(message: &str) -> String {
+    let msg_lower = message.to_lowercase();
+
+    // Startup / connection errors (code == 0 from anyhow)
+    if msg_lower.contains("timeout waiting for") {
+        let method = message
```
chaodu-agent: 🐛 **Bug**: `message.split("timeout waiting for ")` uses the original casing, but the match above uses `msg_lower`. If the original message has different casing (e.g. `"Timeout Waiting For session/new"`), this split won't match and will fall back to `"request"`.

Suggestion: split on `msg_lower` instead:
```rust
let method = msg_lower
    .split("timeout waiting for ")
    .nth(1)
    .unwrap_or("request");
```
Or use a case-insensitive approach to extract the method name from the original message.

src/discord.rs
```diff
@@ -339,6 +355,75 @@ fn compose_display(tool_lines: &[String], text: &str) -> String {
...
+    }
+
+    // Unknown error — pass through as-is
+    if message.is_empty() {
+        "**Error**\nAn unknown error occurred.".to_string()
+    } else {
+        format!("**Error**\n{}", message)
+    }
+}
+
```
chaodu-agent: 💡 **Suggestion**: Since both `format_user_error` and `format_coded_error` are marked `pub` and the doc comment mentions reuse by other adapters (e.g. Slack), consider extracting these into a separate module like `src/error_display.rs`. That way other adapters don't need to depend on the discord module.

src/discord.rs
```diff
@@ -339,6 +355,75 @@ fn compose_display(tool_lines: &[String], text: &str) -> String {
...
+        500 => "**Internal Server Error**",
+        502 => "**Bad Gateway**",
+        503 => "**Service Unavailable**",
+        504 => "**Gateway Timeout**",
+        -32600 => "**Invalid Request**",
+        -32601 => "**Method Not Found**",
+        -32602 => "**Invalid Params**",
+        -32603 => "**Internal Error**",
+        -32000 => "**Connection Error**",
+        _ => "**Error**",
```
chaodu-agent: 💡 **Nice-to-have**: JSON-RPC spec reserves -32000 to -32099 for server errors. Could add a range match as a catch-all:
```rust
-32000..=-32099 => "**Server Error**",
```
so any server-defined error in that range gets a reasonable label instead of the generic `**Error**`.

src/discord.rs
```diff
@@ -305,8 +311,18 @@ async fn stream_prompt(
 
             // Final edit
             let final_content = compose_display(&tool_lines, &text_buf);
+            // If ACP returned both an error and partial text, show both.
+            // This can happen when the agent started producing content before hitting an error
+            // (e.g. context length limit, rate limit mid-stream). Showing both gives users
```
chaodu-agent, line 318: 📝 Good comment explaining why both error and partial text are shown together. This is a thoughtful UX decision.

chaodu-agent (approved):
LGTM 👍 All review feedback addressed cleanly — case sensitivity fix, module extraction, range match, and LazyLock. One tiny typo: "case-insistent" → "case-insensitive" in error_display.rs L14, but not blocking. Ship it! 🚀

thepagent (approved):

Issue #176, opened 2026-04-10
Default Gemini CLI startup should pin a faster model


Joseph19820124:
When OpenAB starts Gemini CLI without an explicit `--model` argument, Gemini CLI appears to default to a Pro model in practice.

That leads to noticeably longer thinking time and slower response latency for interactive chat usage.

Suggested change:

```toml
args = ["--acp", "--model", "gemini-3.1-flash-lite-preview"]
```

Using a faster Flash-Lite model by default would likely improve responsiveness a lot for Discord-style interactions, while still keeping ACP integration intact.

At minimum, it would be helpful to document the current default behavior and provide a recommended low-latency Gemini configuration in the README or example config.

CHC-Agent:
Hi @Joseph19820124,

This is Gemini CLI's default model selection, not an OpenAB bug — users can already configure it via args in config.toml:

`args = ["--acp", "--model", "gemini-3.1-flash-lite-preview"]`

PR #177 is up to document a recommended config. One concern: hardcoding a specific model name in docs may go stale as Gemini releases new models.

Related: #185 / PR #186 adds a `/model` slash command for runtime model switching. If that lands, users could switch models on the fly without touching config — a more flexible long-term solution. The two approaches complement each other (docs for sensible defaults, /model for runtime flexibility).

chaodu-agent:
This makes sense as a config improvement. Before we prioritize, could you share a bit more context?

1. **Use case**: What kind of Discord interactions are you running where latency matters most? (e.g. quick Q&A, code review, multi-turn debugging)
2. **Observed latency**: Roughly how long are responses taking with the default Gemini model vs. what you'd expect with Flash-Lite?
3. **Have you tested `gemini-3.1-flash-lite-preview` with OpenAB?** If so, any quality tradeoffs noticed?

This will help us decide whether to change the default or just document the recommended low-latency config.

PR #177, opened 2026-04-10
docs: recommend faster Gemini model for CLI startup

chenjian-agent:
Fix issue #176: Update example config to recommend gemini-3.1-flash-lite-preview for faster startup and response latency in interactive chat usage.

### DIFF
--- a/config.toml.example
+++ b/config.toml.example
@@ -22,6 +22,7 @@ working_dir = "/home/agent"
 
 # [agent]
 # command = "gemini"
+# args = ["--acp", "--model", "gemini-3.1-flash-lite-preview"]
 # args = ["--acp"]
 # working_dir = "/home/agent"
 # env = { GEMINI_API_KEY = "${GEMINI_API_KEY}" }


chenjian-agent:
Great to see this update. Using  is a significant improvement for interactive latency. This is a very helpful documentation addition.

chaodu-agent:
Thanks for the PR. A few thoughts:

The slowness of Pro is inherent to the Pro model itself — it's not an OpenAB issue. Changing the default model isn't something we'd want to do, since some users specifically need Pro-level quality. That trade-off should be left to the user.

That said, the direction of this PR (documenting an alternative) is reasonable, but the current change is too thin. A single commented-out line without any explanation doesn't give users enough context to act on it.

Suggestions:
- Add a brief comment in the example config (or README) explaining **why** someone might want to specify a different model — e.g. for lower latency in interactive/chat scenarios.
- Mention the trade-off: Pro has better quality but higher latency; Flash/Flash-Lite is faster but less capable.
- Note that `gemini-3.1-flash-lite-preview` is a preview model name that may change — users should verify the latest available model.

With those additions this would be a useful documentation improvement. As-is, the single commented-out line is easy to overlook and doesn't convey enough information.

PR #179, opened 2026-04-10
fix: prevent duplicate messages for long replies (>1900 chars)

### LINKED ISSUE #81: Bug: Long messages (>1900 chars) are sent duplicate to Discord

masami-agent:
## Summary

Fixes the duplicate message bug when agent replies exceed 1900 characters.

Closes #81

## Before (current main)

The edit-streaming task splits long content into chunks via `channel.say()`, and the final edit does the same independently — resulting in duplicate overflow messages.

## After (this PR)

- **Streaming task**: only edits a single placeholder message. If content exceeds 1900 chars, truncates with `…` — this is a live preview, not the final output.
- **Final edit**: handles proper multi-chunk delivery via `split_message()` after streaming completes.

Clean separation of responsibilities, no orphaned messages, no duplication.

## Changes

| File | Change |
|------|--------|
| `src/discord.rs` | Replace streaming split+say logic with truncate+edit (−13 / +6 lines) |

## Approach

Option A from the issue: truncate during streaming, let final edit handle chunking. Same approach as the community fix in PR #53 by @ruan330.

### DIFF
--- a/src/discord.rs
+++ b/src/discord.rs
@@ -238,12 +238,9 @@ async fn stream_prompt(
                         if buf_rx.has_changed().unwrap_or(false) {
                             let content = buf_rx.borrow_and_update().clone();
                             if content != last_content {
-                                let display = if content.len() > 1900 {
-                                    let mut end = 1900;
-                                    while !content.is_char_boundary(end) {
-                                        end -= 1;
-                                    }
-                                    format!("{}…", &content[..end])
+                                let display = if content.chars().count() > 1900 {
+                                    let truncated: String = content.chars().take(1900).collect();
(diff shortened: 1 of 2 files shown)


masami-agent:
Closing in favor of #135 which addresses the same issue, was submitted earlier, and has been production-verified. Will provide review feedback on #135 instead.

Apologies for the duplicate — should have checked for existing PRs before opening this one.

Issue #78, opened 2026-04-06
RFC: Session Management — Design Proposal


thepagent:
# RFC: Session Management

**Tracking issue:** #75
**Status:** Draft
**Author:** @thepagent

---

## Summary

Comprehensive session management for agent-broker covering lifecycle control, isolation, observability, security, and multi-agent support.

## Current State

- `SessionPool` uses `HashMap<String, AcpConnection>` keyed by Discord thread_id
- Basic `cleanup_idle` (TTL-based) and `shutdown` exist
- No user-facing session control, no API, no metrics
- Write lock held during entire prompt streaming (#58)

---

### 1a. `/close` command (#40)
- Discord handler intercepts `/close` message
- Calls `pool.remove(thread_id)` → drops `AcpConnection` → `kill_on_drop` kills child process
- Replies "Session closed." and archives thread

### 1b. Session timeout / auto-expiry
- Split current `session_ttl_hours` (hard max) from new `idle_timeout_minutes` (e.g. 30 min)
- On idle timeout → post "⏰ Session expired due to inactivity" in thread → remove session

### 1c. Per-user session limits
- Track `HashMap<UserId, Vec<ThreadId>>` for active sessions per user
- Exceeding limit → reply "You have too many active sessions. Use `/close` to free one."

### 1d. Graceful shutdown
- On shutdown, post "🔄 Broker restarting..." in each active thread before clearing pool
- Phase 2: persist session metadata to disk/S3 for restart recovery

---

### 2a. Per-thread working directories (#38)
- Change `working_dir` to `{base_working_dir}/{thread_id}/`
- Each session gets its own filesystem namespace
- Cleanup deletes working dir along with session

### 2b. Cross-session deadlock fix (#58)
- Change from `RwLock<HashMap<K, AcpConnection>>` to `RwLock<HashMap<K, Arc<Mutex<AcpConnection>>>>`
- Outer `RwLock` only protects map insert/remove (released immediately)
- Per-connection `Mutex` protects streaming — session A no longer blocks session B

```rust
pub struct SessionPool {
    connections: RwLock<HashMap<String, Arc<Mutex<AcpConnection>>>>,
}

pub async fn with_connection(&self, thread_id: &str, f: F) -> Result<R> {
    let conn = {
        let conns = self.connections.read().await;
        conns.get(thread_id).cloned()
            .ok_or_else(|| anyhow!("no connection"))?
    };
    let mut guard = conn.lock().await;  // per-session lock only
[...]

Issue #75, opened 2026-04-06
tracking: Session Management — observability, lifecycle control, and isolation


thepagent:
## Overview

Umbrella issue to track all session management related work in agent-broker.

## Session Lifecycle

- [ ] #39 — lightweight management API for session observability and lifecycle control
- [ ] #40 — `/close` command to manually terminate a thread session
- [ ] Session timeout / auto-expiry — idle sessions auto-reclaimed after configurable TTL
- [ ] Session limits — max concurrent sessions per user/channel
- [ ] Graceful shutdown — handle in-flight sessions on broker restart

## Session Isolation & Stability

- [ ] #38 — per-thread isolated working directories
- [ ] #58 — fix: pool write lock held during entire prompt streaming causes cross-session deadlock

## Session State

- [ ] Session persistence / recovery — restore sessions after broker crash
- [ ] Context window management — handle conversation history overflow (truncate / summarize)

## Session Observability

- [ ] Session metrics — active count, avg lifetime, message count
- [ ] Session logging / audit trail — who created, when, which agent

## Session Security & Access Control

- [ ] Session ownership — only session owner can operate their session
- [ ] Rate limiting per session — prevent single session from overwhelming agent

## Multi-agent

- [ ] Session routing — ability to switch agent within a session
- [ ] Session handoff — transfer session from agent A to agent B with context

ChunHao-dev:
Closing — this tracking issue has been inactive for over 2 months. The related PRs (#41, #57, #42, #77) should be tracked individually. Feel free to reopen when session management work resumes as a coordinated effort.

PR #182, opened 2026-04-10
feat: broadcast shutdown notification to active threads (RFC #78 §1d Phase 1)

ruan330:
## Summary

On SIGINT or SIGTERM the broker now posts a short notification to every active thread across all configured adapters (Discord, Slack, …) before closing shards. Users get a clear signal that the broker is going away instead of replies cutting off silently mid-stream.

Discord Discussion URL: https://discord.com/channels/1488041051187974246/1495050997461024879

Relates to: #78, #75

## Problem

Today, when the broker is restarted (`systemctl restart`, `docker stop`, Ctrl+C), conversations just die silently mid-reply. Users have no signal that anything is happening until the broker comes back, and in-flight streams leave partial messages with no explanation. RFC #78 §1d calls this out as Phase 1.

The original #182 (rebased before v0.7.6) was Discord-specific and done at the `main.rs` level. The multi-platform refactor in #259 makes that approach stale: Slack threads also need the notification, and the `AdapterRouter` / `ChatAdapter` abstraction is the right home for platform-neutral broadcast. This revision rewrites the feature against that architecture.

## Design

**Pool owns addressing, not the router.** `SessionPool` now stores each active session's `ChannelRef` + `Arc<dyn ChatAdapter>` alongside the connection (`addresses: HashMap<thread_key, ...>`), kept in lockstep with `state.active` so broadcast has a single source of truth. No parallel cache in the adapter layer, no `pool.active_thread_ids()` filter, no prune task.

**Atomic admission and snapshot.**
- `SessionPool::begin_shutdown()` takes the state write lock, flips the shutdown flag, and snapshots `addresses` — all in one critical section. Any admission committed before us is in the snapshot.
- `get_or_create` checks `is_shutting_down()` at four points so every session is either in the snapshot or rejected at admission:
  1. Fast-fail on entry.
  2. Inside the initial `state.read()` block.
  3. Before `return Ok(())` on the existing-alive-session path, with a `state.read()` barrier synchronizing against `begin_shutdown`'s `state.write()` flag flip (closes the race where shutdown starts while we wait on the per-connection mutex).
  4. Inside the final `state.write()` block before insert.

**Platform-neutral broadcast.** `AdapterRouter::broadcast_shutdown(message, timeout)` calls `pool.begin_shutdown()` and posts the notification to every snapshot entry in parallel via `tokio::task::JoinSet`. A `tokio::select!` with a configurable deadline (10s here) caps total broadcast time so shutdown itself never blocks on a slow platform.

**Unified signal handling.** `main.rs` uses a shared `wait_for_shutdown_signal()` helper for SIGINT + SIGTERM (`#[cfg(unix)]`-gated with a ctrl_c-only fallback for non-Unix). Both the Discord-enabled and Slack-only branches invoke `broadcast_shutdown` before tearing down adapters.

## Behaviour notes

- Delivery is best-effort: send errors and the 10-second deadline both fall through to normal pool teardown.
- Wording is intentionally neutral ("Context will reset on return") — we don't promise automatic session resume. RFC #78 Phase 2 persistence is a separate follow-up.
- Per-channel parallel send is safe under Discord/Slack rate limits (limits are per-channel, not per-adapter) and is ~10× faster than sequential for deployments with several active threads.

## Tests & verification

- `cargo test` — 43 passed, 0 failed.
- `cargo clippy --all-targets -- -D warnings` — clean.
- Live-tested on bare-metal and Docker brokers: SIGTERM triggers `broadcasting shutdown notification count=N` log and the message arrives in every active thread.

## Test plan

- [ ] CI: `cargo test`, `cargo clippy`, Docker smoke tests.
- [x] Manual: start broker with Discord and/or Slack configured, open an active thread, send SIGTERM (`docker stop` / `kill -TERM`), verify the thread receives the "Bot is shutting down" message before the pool tears down.
- [ ] Manual: broadcast timeout path — block network to Discord and verify shutdown still completes after 10s.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -259,11 +259,17 @@ async fn main() -> anyhow::Result<()> {
 /// we fall back to ctrl_c alone.
 #[cfg(unix)]
 async fn wait_for_shutdown_signal() {
-    let mut sigterm = tokio::signal::unix::signal(tokio::signal::unix::SignalKind::terminate())
-        .expect("install SIGTERM handler");
-    tokio::select! {
-        _ = tokio::signal::ctrl_c() => {}
-        _ = sigterm.recv() => {}
+    match tokio::signal::unix::signal(tokio::signal::unix::SignalKind::terminate()) {
+        Ok(mut sigterm) => {
+            tokio::select! {
+                _ = tokio::signal::ctrl_c() => {}
+                _ = sigterm.recv() => {}
+            }
+        }
+        Err(e) => {
+            warn!("failed to install SIGTERM handler ({e}), falling back to ctrl_c only");
+            tokio::signal::ctrl_c().await.ok();
+        }
     }
     info!("shutdown signal received");
 }
--- a/src/main.rs
+++ b/src/main.rs
@@ -18,6 +18,15 @@ use std::path::PathBuf;
 use std::sync::Arc;
 use tracing::{error, info, warn};
 
+/// Neutral shutdown notification broadcast to every active thread. Wording
+/// deliberately avoids "restarting" because `helm uninstall` / final-stop
+/// can't be distinguished from a rolling restart at signal time.
+const SHUTDOWN_MSG: &str = "⚠️ Bot is shutting down. Context will reset on return.";
+
+/// Broadcast deadline. Shutdown itself must never block on a slow platform,
+/// so incomplete sends are dropped once this elapses.
+const SHUTDOWN_BROADCAST_TIMEOUT: std::time::Duration = std::time::Duration::from_secs(10);
+
 #[derive(Parser)]
 #[command(name = "openab")]
 #[command(about = "Multi-platform ACP agent broker (Discord, Slack)", long_about = None)]
@@ -176,7 +185,7 @@ async fn main() -> anyhow::Result<()> {
                 );
 
                 let handler = discord::Handler {
-                    router,
+                    router: router.clone(),
                     allow_all_channels,
                     allow_all_users,
                     allowed_channels,
@@ -201,21 +210,31 @@ async fn main() -> anyhow::Result<()> {
                     .event_handler(handler)
                     .await?;
 
-                // Graceful Discord shutdown on ctrl_c
+                // Graceful shutdown on SIGINT or SIGTERM: wait for the signal,
+                // broadcast to every active thread (Discord + Slack), then stop
+                // Discord shards. `client.start()` is the foreground blocker here,
+                // so this handler runs as a spawned task.
                 let shard_manager = client.shard_manager.clone();
+                let shutdown_router = router.clone();
                 tokio::spawn(async move {
-                    tokio::signal::ctrl_c().await.ok();
-                    info!("shutdown signal received");
+                    wait_for_shutdown_signal().await;
+                    shutdown_router
+                        .broadcast_shutdown(SHUTDOWN_MSG, SHUTDOWN_BROADCAST_TIMEOUT)
+                        .await;
                     shard_manager.shutdown_all().await;
                 });
 
                 info!("discord bot running");
                 client.start().await?;
             } else {
-                // No Discord — just wait for ctrl_c
-                info!("running without discord, press ctrl+c to stop");
-                tokio::signal::ctrl_c().await.ok();
-                info!("shutdown signal received");
+                // No Discord — this task itself blocks on the shutdown signal,
+                // then broadcasts before falling through to cleanup. Slack-only
+                // deployments need SIGTERM + broadcast just like Discord.
+                info!("running without discord, waiting for shutdown signal");
+                wait_for_shutdown_signal().await;
+                router
+                    .broadcast_shutdown(SHUTDOWN_MSG, SHUTDOWN_BROADCAST_TIMEOUT)
+                    .await;
             }
 
             // Cleanup
@@ -231,6 +250,28 @@ async fn main() -> anyhow::Result<()> {
             Ok(())
         }
     }
+}
+
+/// Wait for SIGINT (ctrl_c) or, on Unix, SIGTERM (systemctl stop, docker stop,
+/// kill). Without SIGTERM handling the broker would be killed outright by
+/// service managers and skip the shutdown broadcast, so both signals route
+/// here on Unix. On non-Unix targets `tokio::signal::unix` is unavailable, so
+/// we fall back to ctrl_c alone.
+#[cfg(unix)]
+async fn wait_for_shutdown_signal() {
+    let mut sigterm = tokio::signal::unix::signal(tokio::signal::unix::SignalKind::terminate())
+        .expect("install SIGTERM handler");
+    tokio::select! {
+        _ = tokio::signal::ctrl_c() => {}
+        _ = sigterm.recv() => {}
+    }
+    info!("shutdown signal received");
+}
+
+#[cfg(not(unix))]
+async fn wait_for_shutdown_signal() {
+    tokio::signal::ctrl_c().await.ok();
+    info!("shutdown signal received");
 }
 
 fn parse_id_set(raw: &[String], label: &str) -> anyhow::Result<HashSet<u64>> {
--- a/src/acp/pool.rs
+++ b/src/acp/pool.rs
@@ -1,7 +1,8 @@
 use crate::acp::connection::AcpConnection;
 use crate::acp::protocol::ConfigOption;
+use crate::adapter::{ChannelRef, ChatAdapter};
 use crate::config::AgentConfig;
-use anyhow::{anyhow, Result};
+use anyhow::{anyhow, bail, Result};
 use std::collections::HashMap;
 use std::sync::Arc;
 use tokio::sync::{Mutex, RwLock};
@@ -16,6 +17,11 @@ struct PoolState {
     /// Lock-free cancel handles: thread_key → (stdin, session_id).
     /// Stored separately so cancel can work without locking the connection.
     cancel_handles: HashMap<String, (Arc<tokio::sync::Mutex<tokio::process::ChildStdin>>, String)>,
+    /// Addressing info for each active thread. Populated alongside `active`
+    /// and pruned together. Used by `begin_shutdown` so the broker can post
+    /// a notification to every live session without the adapter layer
+    /// maintaining a parallel cache. Invariant: `addresses.keys() == active.keys()`.
+    addresses: HashMap<String, (ChannelRef, Arc<dyn ChatAdapter>)>,
     /// Suspended sessions: thread_key → ACP sessionId.
     /// Saved on eviction so sessions can be resumed via `session/load`.
     suspended: HashMap<String, String>,
@@ -28,6 +34,10 @@ pub struct SessionPool {
     state: RwLock<PoolState>,
     config: AgentConfig,
     max_sessions: usize,
+    /// Flipped by `begin_shutdown` to reject new admissions. Checked inside
+    /// `get_or_create` under the state write lock so admission and snapshot
+    /// are atomic.
+    shutting_down: std::sync::atomic::AtomicBool,
 }
 
 type EvictionCandidate = (
@@ -67,15 +77,53 @@ impl SessionPool {
             state: RwLock::new(PoolState {
                 active: HashMap::new(),
                 cancel_handles: HashMap::new(),
+                addresses: HashMap::new(),
                 suspended: HashMap::new(),
                 creating: HashMap::new(),
             }),
             config,
             max_sessions,
+            shutting_down: std::sync::atomic::AtomicBool::new(false),
         }
     }
 
-    pub async fn get_or_create(&self, thread_id: &str) -> Result<()> {
+    /// True once `begin_shutdown` has been called. Router uses this to show a
+    /// shutdown-specific message instead of a generic pool error when
+    /// `get_or_create` rejects admission.
+    pub fn is_shutting_down(&self) -> bool {
+        self.shutting_down
+            .load(std::sync::atomic::Ordering::Acquire)
+    }
+
+    /// Flip the pool into shutting-down state and return a snapshot of every
+    /// live session's addressing info. Takes the state write lock so the
+    /// snapshot is atomic with respect to in-flight `get_or_create` calls:
+    /// any admission that committed before us is included; any that comes
+    /// after us sees the flag inside the same lock and rejects.
+    pub async fn begin_shutdown(&self) -> Vec<(String, ChannelRef, Arc<dyn ChatAdapter>)> {
+        let state = self.state.write().await;
+        self.shutting_down
+            .store(true, std::sync::atomic::Ordering::Release);
+        state
+            .addresses
+            .iter()
+            .map(|(k, (c, a))| (k.clone(), c.clone(), a.clone()))
+            .collect()
+    }
+
+    pub async fn get_or_create(
+        &self,
+        thread_id: &str,
+        channel: &ChannelRef,
+        adapter: &Arc<dyn ChatAdapter>,
+    ) -> Result<()> {
+        // Fast-fail: avoid spawning a fresh ACP process if shutdown is already
+        // in progress. The authoritative check happens again under the state
+        // write lock below so we also catch shutdowns that start mid-spawn.
+        if self.is_shutting_down() {
+            bail!("pool is shutting down");
+        }
+
         let create_gate = {
             let mut state = self.state.write().await;
             get_or_insert_gate(&mut state.creating, thread_id)
@@ -84,6 +132,9 @@ impl SessionPool {
 
         let (existing, saved_session_id) = {
             let state = self.state.read().await;
+            if self.is_shutting_down() {
+                bail!("pool is shutting down");
+            }
             (
                 state.active.get(thread_id).cloned(),
                 state.suspended.get(thread_id).cloned(),
@@ -95,6 +146,16 @@ impl SessionPool {
         if let Some(conn) = existing.clone() {
             let conn = conn.lock().await;
             if conn.alive() {
+                // Re-check shutdown state after waiting on the per-connection
+                // mutex. Taking `state.read()` synchronizes us with
+                // `begin_shutdown`'s write-lock flag flip, so the flag value
+                // we see here reflects every `begin_shutdown` that has
+                // committed. This closes the race where shutdown starts
+                // while we were waiting on `conn.lock()`.
+                let _sync = self.state.read().await;
+                if self.is_shutting_down() {
+                    bail!("pool is shutting down");
+                }
                 return Ok(());
             }
             if saved_session_id.is_none() {
@@ -174,6 +235,14 @@ impl SessionPool {
 
         let mut state = self.state.write().await;
 
+        // Admission check inside the state write lock. This is atomic with
+        // `begin_shutdown`'s flag-flip + snapshot: a shutdown that started
+        // during our ACP spawn is caught here, and our work is thrown away
+        // rather than being added to a pool that is about to be torn down.
+        if self.is_shutting_down() {
+            bail!("pool is shutting down");
+        }
+
         // Another task may have created a healthy connection while we were
         // initializing this one.
         if let Some(existing) = state.active.get(thread_id).cloned() {
@@ -186,11 +255,13 @@ impl SessionPool {
             warn!(thread_id, "stale connection, rebuilding");
             drop(existing);
             state.active.remove(thread_id);
+            state.addresses.remove(thread_id);
         }
 
         if state.active.len() >= self.max_sessions {
             if let Some((key, expected_conn, _, sid)) = eviction_candidate {
                 if remove_if_same_handle(&mut state.active, &key, &expected_conn).is_some() {
+                    state.addresses.remove(&key);
                     info!(evicted = %key, "pool full, suspending oldest idle session");
                     if let Some(sid) = sid {
                         state.suspended.insert(key, sid);
@@ -216,6 +287,9 @@ impl SessionPool {
         if !cancel_session_id.is_empty() {
             state.cancel_handles.insert(thread_id.to_string(), (cancel_handle, cancel_session_id));
         }
+        state
+            .addresses
+            .insert(thread_id.to_string(), (channel.clone(), adapter.clone()));
         Ok(())
     }
 
@@ -324,6 +398,7 @@ impl SessionPool {
         let mut state = self.state.write().await;
         for (key, expected_conn, sid) in stale {
             if remove_if_same_handle(&mut state.active, &key, &expected_conn).is_some() {
+                state.addresses.remove(&key);
                 info!(thread_id = %key, "cleaning up idle session");
                 if let Some(sid) = sid {
                     state.suspended.insert(key, sid);
@@ -336,6 +411,7 @@ impl SessionPool {
         let mut state = self.state.write().await;
         let count = state.active.len();
         state.active.clear(); // Drop impl kills process groups
+        state.addresses.clear();
         info!(count, "pool shutdown complete");
     }
 }
--- a/src/acp/pool.rs
+++ b/src/acp/pool.rs
@@ -9,19 +9,22 @@ use tokio::sync::{Mutex, RwLock};
 use tokio::time::Instant;
 use tracing::{info, warn};
 
+/// A single active session: connection handle + addressing info bundled
+/// together so their lifetimes are enforced at the type level.
+struct SessionEntry {
+    conn: Arc<Mutex<AcpConnection>>,
+    channel: ChannelRef,
+    adapter: Arc<dyn ChatAdapter>,
+}
+
 /// Combined state protected by a single lock to prevent deadlocks.
 /// Lock ordering: never await a per-connection mutex while holding `state`.
 struct PoolState {
-    /// Active connections: thread_key → AcpConnection handle.
-    active: HashMap<String, Arc<Mutex<AcpConnection>>>,
+    /// Active sessions: thread_key → SessionEntry (connection + addressing).
+    sessions: HashMap<String, SessionEntry>,
     /// Lock-free cancel handles: thread_key → (stdin, session_id).
     /// Stored separately so cancel can work without locking the connection.
     cancel_handles: HashMap<String, (Arc<tokio::sync::Mutex<tokio::process::ChildStdin>>, String)>,
-    /// Addressing info for each active thread. Populated alongside `active`
-    /// and pruned together. Used by `begin_shutdown` so the broker can post
-    /// a notification to every live session without the adapter layer
-    /// maintaining a parallel cache. Invariant: `addresses.keys() == active.keys()`.
-    addresses: HashMap<String, (ChannelRef, Arc<dyn ChatAdapter>)>,
     /// Suspended sessions: thread_key → ACP sessionId.
     /// Saved on eviction so sessions can be resumed via `session/load`.
     suspended: HashMap<String, String>,
@@ -47,16 +50,18 @@ type EvictionCandidate = (
     Option<String>,
 );
 
-fn remove_if_same_handle<T>(
-    map: &mut HashMap<String, Arc<Mutex<T>>>,
+/// Remove a session entry only if its connection handle matches `expected`.
+/// Returns the removed connection Arc, or None if the handle was swapped.
+fn remove_if_same_conn(
+    map: &mut HashMap<String, SessionEntry>,
     key: &str,
-    expected: &Arc<Mutex<T>>,
-) -> Option<Arc<Mutex<T>>> {
+    expected: &Arc<Mutex<AcpConnection>>,
+) -> Option<Arc<Mutex<AcpConnection>>> {
     let should_remove = map
         .get(key)
-        .is_some_and(|current| Arc::ptr_eq(current, expected));
+        .is_some_and(|entry| Arc::ptr_eq(&entry.conn, expected));
     if should_remove {
-        map.remove(key)
+        map.remove(key).map(|e| e.conn)
     } else {
         None
     }
@@ -75,9 +80,8 @@ impl SessionPool {
     pub fn new(config: AgentConfig, max_sessions: usize) -> Self {
         Self {
             state: RwLock::new(PoolState {
-                active: HashMap::new(),
+                sessions: HashMap::new(),
                 cancel_handles: HashMap::new(),
-                addresses: HashMap::new(),
                 suspended: HashMap::new(),
                 creating: HashMap::new(),
             }),
@@ -105,9 +109,9 @@ impl SessionPool {
         self.shutting_down
             .store(true, std::sync::atomic::Ordering::Release);
         state
-            .addresses
+            .sessions
             .iter()
-            .map(|(k, (c, a))| (k.clone(), c.clone(), a.clone()))
+            .map(|(k, e)| (k.clone(), e.channel.clone(), e.adapter.clone()))
             .collect()
     }
 
@@ -136,7 +140,7 @@ impl SessionPool {
                 bail!("pool is shutting down");
             }
             (
-                state.active.get(thread_id).cloned(),
+                state.sessions.get(thread_id).map(|e| e.conn.clone()),
                 state.suspended.get(thread_id).cloned(),
             )
         };
@@ -152,7 +156,11 @@ impl SessionPool {
                 // we see here reflects every `begin_shutdown` that has
                 // committed. This closes the race where shutdown starts
                 // while we were waiting on `conn.lock()`.
-                let _sync = self.state.read().await;
+                //
+                // DO NOT REMOVE — this read-lock is a synchronization barrier,
+                // not a data access. Without it the `is_shutting_down()` check
+                // below can observe a stale value.
+                let _shutdown_barrier = self.state.read().await;
                 if self.is_shutting_down() {
                     bail!("pool is shutting down");
                 }
@@ -167,9 +175,9 @@ impl SessionPool {
         let snapshot: Vec<(String, Arc<Mutex<AcpConnection>>)> = {
             let state = self.state.read().await;
             state
-                .active
+                .sessions
                 .iter()
-                .map(|(k, v)| (k.clone(), Arc::clone(v)))
+                .map(|(k, e)| (k.clone(), Arc::clone(&e.conn)))
                 .collect()
         };
 
@@ -245,7 +253,8 @@ impl SessionPool {
 
         // Another task may have created a healthy connection while we were
         // initializing this one.
-        if let Some(existing) = state.active.get(thread_id).cloned() {
+        if let Some(entry) = state.sessions.get(thread_id) {
+            let existing = entry.conn.clone();
             let Ok(existing) = existing.try_lock() else {
                 return Ok(());
             };
@@ -254,14 +263,12 @@ impl SessionPool {
             }
             warn!(thread_id, "stale connection, rebuilding");
             drop(existing);
-            state.active.remove(thread_id);
-            state.addresses.remove(thread_id);
+            state.sessions.remove(thread_id);
         }
 
-        if state.active.len() >= self.max_sessions {
+        if state.sessions.len() >= self.max_sessions {
             if let Some((key, expected_conn, _, sid)) = eviction_candidate {
-                if remove_if_same_handle(&mut state.active, &key, &expected_conn).is_some() {
-                    state.addresses.remove(&key);
+                if remove_if_same_conn(&mut state.sessions, &key, &expected_conn).is_some() {
                     info!(evicted = %key, "pool full, suspending oldest idle session");
                     if let Some(sid) = sid {
                         state.suspended.insert(key, sid);
@@ -278,18 +285,19 @@ impl SessionPool {
             }
         }
 
-        if state.active.len() >= self.max_sessions {
+        if state.sessions.len() >= self.max_sessions {
             return Err(anyhow!("pool exhausted ({} sessions)", self.max_sessions));
         }
 
         state.suspended.remove(thread_id);
-        state.active.insert(thread_id.to_string(), new_conn);
+        state.sessions.insert(thread_id.to_string(), SessionEntry {
+            conn: new_conn,
+            channel: channel.clone(),
+            adapter: adapter.clone(),
+        });
         if !cancel_session_id.is_empty() {
             state.cancel_handles.insert(thread_id.to_string(), (cancel_handle, cancel_session_id));
         }
-        state
-            .addresses
-            .insert(thread_id.to_string(), (channel.clone(), adapter.clone()));
         Ok(())
     }
 
@@ -303,9 +311,9 @@ impl SessionPool {
         let conn = {
             let state = self.state.read().await;
             state
-                .active
+                .sessions
                 .get(thread_id)
-                .cloned()
+                .map(|e| e.conn.clone())
                 .ok_or_else(|| anyhow!("no connection for thread {thread_id}"))?
         };
 
@@ -316,8 +324,8 @@ impl SessionPool {
     /// Get cached configOptions for a session (e.g. available models).
     pub async fn get_config_options(&self, thread_id: &str) -> Vec<ConfigOption> {
         let state = self.state.read().await;
-        let conn = match state.active.get(thread_id) {
-            Some(c) => c.clone(),
+        let conn = match state.sessions.get(thread_id) {
+            Some(e) => e.conn.clone(),
             None => return Vec::new(),
         };
         drop(state);
@@ -335,9 +343,9 @@ impl SessionPool {
         let conn = {
             let state = self.state.read().await;
             state
-                .active
+                .sessions
                 .get(thread_id)
-                .cloned()
+                .map(|e| e.conn.clone())
                 .ok_or_else(|| anyhow!("no connection for thread {thread_id}"))?
         };
         let mut conn = conn.lock().await;
@@ -372,9 +380,9 @@ impl SessionPool {
         let snapshot: Vec<(String, Arc<Mutex<AcpConnection>>)> = {
             let state = self.state.read().await;
             state
-                .active
+                .sessions
                 .iter()
-                .map(|(k, v)| (k.clone(), Arc::clone(v)))
+                .map(|(k, e)| (k.clone(), Arc::clone(&e.conn)))
                 .collect()
         };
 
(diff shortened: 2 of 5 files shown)


masami-agent:
## Review — PR #182

Good feature — users currently get no warning when the broker shuts down. This fixes that.

### ✅ What looks good

1. **SIGTERM support** — `main` only has `ctrl_c()` today. Adding SIGTERM via `tokio::select!` is important for k8s graceful shutdown.
2. **`active_thread_ids()`** — clean, minimal addition to pool.
3. **Broadcast before shutdown** — correct ordering. HTTP client is still alive at broadcast time (before `shard_manager.shutdown_all()`).
4. **Sequential broadcast** — safer than parallel given Discord rate limits (50 req/s). No need to add `join_all` complexity.
5. **Neutral wording** — doesn't promise auto-resume.
6. **Failure handling** — per-thread `warn` on send failure, no crash.

### 🔴 Must fix before merge

**1. Rebase onto latest `main`**
The pool structure has changed:
- PR uses `self.connections` — `main` uses `self.state` (field: `RwLock<PoolState>`, inner: `state.active`)
- `main.rs` shutdown handler has been restructured (Slack adapter support, `shutdown_tx` channel)
- Handler struct is completely different

`active_thread_ids()` will need to read `self.state.read().await.active.keys()` after rebase.

**2. CI not triggered**
Fork PR — needs maintainer approval to run CI.

### 🟡 Non-blocking

**3. k8s terminationGracePeriodSeconds**
Default is 30s. Sequential broadcast with 20+ active threads could take 2-10s (each `channel.say()` is ~100-500ms). If Discord API is slow, broadcast may get truncated by SIGKILL. Not a blocker — worst case is some threads miss the notification. But worth noting for operators with large session pools.

**4. "Restarting" message on permanent shutdown**
The message says "Broker restarting" but `helm uninstall` is a permanent shutdown. No good way to distinguish at runtime — acceptable as-is.

### Summary

Valuable UX improvement. Rebase needed — pool and main.rs have both changed significantly.

masami-agent (commented):
## PR Review: #182

### Summary
- **Problem**: Broker restarts kill conversations silently — users get no notification that the broker is going down.
- **Approach**: Add `active_thread_ids()` accessor to `SessionPool`, handle SIGTERM alongside SIGINT, and broadcast a neutral notification to each active Discord thread before shutting down shards.
- **Risk level**: Low

### Core Assessment
1. Problem clearly stated: ✅ — well-motivated, linked to RFC #78 §1d
2. Approach appropriate: ✅ — minimal, clean separation between pool (lifecycle) and main (Discord)
3. Alternatives considered: ✅ — PR description covers three alternatives with clear reasoning
4. Best approach for now: ✅ — Phase 1 scope is correct; no over-engineering

### Findings

**pool.rs — `active_thread_ids()`**

Clean and minimal. Takes a read lock, returns an owned `Vec<String>`, releases immediately. No contention risk with the subsequent `shutdown()` write lock. ✅

**main.rs — SIGTERM handling**

The `tokio::select!` over `ctrl_c()` and `sigterm.recv()` is the standard pattern. The `.expect("install SIGTERM handler")` is acceptable here — if we cannot install a signal handler at startup, panicking is the right call since the broker would be unable to shut down gracefully anyway. ✅

**main.rs — broadcast loop**

The `strip_prefix("discord:")` + `parse::<u64>()` approach correctly handles the `{platform}:{id}` key format used by `AdapterRouter`. Non-Discord keys (e.g. `slack:...`) are silently skipped via the `parse` failure path, which is the right behavior for now. ✅

**main.rs — error handling**

`channel.say()` failures are logged as warnings and do not block shutdown. Correct — a dead Discord connection should never prevent pool cleanup. ✅

### Review Summary

#### 🔴 Blockers
- **`Cargo.lock` version bump is out of scope.** The diff changes `Cargo.lock` from `0.7.6` → `0.7.7`. On `main`, `Cargo.toml` is already `0.7.7` but `Cargo.lock` is stale at `0.7.6`. While the sync is technically correct, it is unrelated to the shutdown broadcast feature and should not be bundled into this PR. Please revert the `Cargo.lock` change and let it be fixed in a separate housekeeping commit, or confirm this was an intentional rebase artifact that the maintainers are okay with.

#### 💬 Questions
1. **Non-Discord shutdown path also lacks SIGTERM.** The `else` branch (line ~194 on main: `"running without discord, press ctrl+c to stop"`) still only awaits `ctrl_c()`. In a Slack-only or headless deployment, `docker stop` / `kill` would still bypass the graceful path. Was this intentionally left out of scope, or should it be addressed here for consistency?

2. **Slack adapter threads.** When Slack support is active, `thread_key` will be `slack:{id}`. The broadcast loop silently skips non-Discord keys (the `parse::<u64>()` fails on `slack:...` after `strip_prefix("discord:")` returns the full key). This is fine for Phase 1, but worth noting — should there be a `tracing::debug!` for skipped non-Discord keys so operators know the broadcast was Discord-only?

#### 🔧 Suggested Changes
1. **Consider a timeout on the broadcast loop.** If there are many active sessions and Discord is slow or rate-limiting, the broadcast could delay shutdown significantly. A `tokio::time::timeout` wrapping the entire loop (e.g. 5–10 seconds) would provide a safety net. Not blocking, but worth considering for production robustness.

2. **The `pool.shutdown()` is called twice on the Discord path.** After the spawned task runs `shard_manager.shutdown_all()`, execution continues past `client.start().await?` and hits the `shutdown_pool.shutdown().await` at line ~208. The broadcast task does NOT call `pool.shutdown()` — it only broadcasts and shuts down shards. So the pool is shut down once (at the bottom), which is correct. However, the broadcast task clones `pool` as `broadcast_pool` but never calls shutdown on it. This is fine, but the naming `broadcast_pool` could be clearer — maybe `notification_pool` or just `pool_ref` to signal it is read-only usage.

#### ℹ️ Info
- The `Cargo.lock` discrepancy on `main` (`Cargo.toml` = 0.7.7, `Cargo.lock` = 0.7.6) is a pre-existing issue. It should be fixed, but in a separate commit.
- RFC #78 §1d Phase 2 (session persistence) is correctly deferred. The neutral wording ("you can continue" vs "will resume") is the right call given no state is persisted yet.

#### ⚪ Nits
- Line length on the `channel.say()` string literal is ~90 chars — fine, but if you want to break it for readability, a `const` at the top of the block would work nicely:
  ```rust
  const SHUTDOWN_MSG: &str = "🔄 Broker restarting. You can continue the conversation when the broker is back.";
  ```

### Verdict
**COMMENT** — One blocker (Cargo.lock scope) and two questions to resolve. The core implementation is clean and well-designed. Once the Cargo.lock issue is addressed and the questions are answered, this should be ready for approval.

Cargo.lock
```diff
@@ -960,7 +960,7 @@ checksum = "384b8ab6d37215f3c5301a95a4accb5d64aa607f1fcb26a11b5303878451b4fe"
 
 [[package]]
 name = "openab"
-version = "0.7.6"
+version = "0.7.7"
```
masami-agent: 🔴 **BLOCKER** — This version bump (`0.7.6` → `0.7.7`) is a pre-existing `Cargo.lock` / `Cargo.toml` sync issue on `main`, not related to the shutdown broadcast feature. Please revert this file from the PR to keep the scope clean. The lockfile sync can be fixed in a separate housekeeping PR or commit.

```bash
git checkout origin/main -- Cargo.lock
```

src/main.rs
```diff
@@ -169,11 +169,46 @@ async fn main() -> anyhow::Result<()> {
                     .event_handler(handler)
                     .await?;
 
-                // Graceful Discord shutdown on ctrl_c
+                // Graceful Discord shutdown on SIGINT or SIGTERM.
+                // systemctl stop / docker stop / kill deliver SIGTERM; without handling
+                // it, the broker would exit without running the broadcast below.
                 let shard_manager = client.shard_manager.clone();
+                let broadcast_pool = pool.clone();
```
masami-agent: 💬 **Question** — The `else` branch below (non-Discord / Slack-only mode, around line 194 on `main`) still only awaits `ctrl_c()`. In a Slack-only or headless deployment, `docker stop` / `kill` would still bypass the graceful path. Was this intentionally left out of scope for this PR? If so, might be worth a `// TODO:` comment noting that SIGTERM handling should be added here too.

src/main.rs
```diff
@@ -169,11 +169,46 @@ async fn main() -> anyhow::Result<()> {
...
+                    }
                     info!("shutdown signal received");
+
+                    // Broadcast shutdown notification to active threads before closing shards.
+                    // Neutral wording — we don't promise automatic resume.
+                    let thread_ids = broadcast_pool.active_thread_ids().await;
+                    info!(count = thread_ids.len(), "broadcasting shutdown notification");
+                    for thread_key in thread_ids {
+                        // thread_key format: "platform:id" (e.g. "discord:1234567890").
+                        let raw_id = thread_key
```
masami-agent: 🟡 **Suggestion** — Consider wrapping the broadcast loop in a timeout to guard against Discord rate limits or network issues delaying shutdown:

```rust
let _ = tokio::time::timeout(
    std::time::Duration::from_secs(10),
    async {
        for thread_key in thread_ids {
            // ... existing broadcast logic ...
        }
    },
).await;
```

Not blocking — just a safety net for production. If Discord is slow, you probably still want the process to exit within a bounded time.

src/main.rs
```diff
@@ -169,11 +169,46 @@ async fn main() -> anyhow::Result<()> {
...
+                            .strip_prefix("discord:")
+                            .unwrap_or(&thread_key);
+                        if let Ok(id) = raw_id.parse::<u64>() {
+                            let channel = serenity::model::id::ChannelId::new(id);
```
masami-agent: ℹ️ **Info** — When Slack adapter threads exist, their keys (`slack:{id}`) will pass through `strip_prefix("discord:")` unchanged (returns `None` → `unwrap_or` gives the full key), then `parse::<u64>()` fails on `slack:12345` and they're silently skipped. This is correct behavior for Phase 1, but a `tracing::debug!` here for skipped non-Discord keys would help operators understand why some sessions didn't get notifications:

```rust
let raw_id = thread_key
    .strip_prefix("discord:")
    .unwrap_or(&thread_key);
if let Ok(id) = raw_id.parse::<u64>() {
    // ... send notification ...
} else {
    tracing::debug!(thread_key = %thread_key, "skipping non-Discord thread for shutdown broadcast");
}
```

src/main.rs
```diff
@@ -169,11 +169,46 @@ async fn main() -> anyhow::Result<()> {
...
+                            if let Err(e) = channel
+                                .say(
+                                    &shutdown_http,
+                                    "🔄 Broker restarting. You can continue the conversation when the broker is back.",
+                                )
+                                .await
+                            {
+                                tracing::warn!(thread_key = %thread_key, error = %e, "failed to post shutdown notification");
```
masami-agent: ⚪ **Nit** — The message string is a bit long inline. Consider extracting it as a constant for readability:

```rust
const SHUTDOWN_MSG: &str =
    "🔄 Broker restarting. You can continue the conversation when the broker is back.";
```

Minor — skip if you prefer it inline.

masami-agent (commented):
Hi @ruan330 — leaving some inline comments on specific areas. The overall review is in my earlier comment above. Thanks for the solid work on this!

shaun-agent:
## OpenAB PR Screening
- Title: feat: broadcast shutdown notification to active threads (RFC #78 §1d Phase 1)
- Source: https://github.com/openabdev/openab/pull/182
- Generated at: 2026-04-18T13:19:38.451Z
- Discord thread: https://discord.com/channels/1488041051187974246/1495050997461024879
<details>
## Intent

This PR tries to make broker shutdowns visible to users who are actively chatting in Discord threads. Today, a broker restart or stop can terminate an in-flight conversation without explanation, leaving a partial reply or a dead thread with no operator-visible or user-visible signal. The proposed change addresses that gap by sending a shutdown notice to all currently active Discord threads before the broker tears down the session pool.

## Feat

This is a feature with some operational-hardening characteristics. In plain terms, when the OpenAB broker receives `SIGINT` or `SIGTERM`, it gathers the active thread IDs and posts a fixed Discord message to each thread before shutting down the shard manager and session pool. It does not preserve session state, recover streams, or resume work after restart; it only makes the interruption explicit.

## Who It Serves

The primary beneficiary is Discord end users in active threads, with secondary value for maintainers and operators who want shutdown behavior to be less opaque and easier to reason about during deploys, restarts, or crashes followed by restart.

## Rewritten Prompt

Implement Phase 1 of graceful shutdown notification for the Discord broker.

When the process receives either `SIGINT` or `SIGTERM`, enumerate all currently active Discord thread IDs from the session pool, post a short restart notice to each valid thread, log notification failures as warnings, and then continue normal shutdown. Keep Discord-specific delivery logic in the broker entrypoint rather than pushing it into the pool internals. Do not add persistence, resume semantics, or a drain window in this change. Add or update tests where practical to verify signal handling and that active-thread notification happens before pool teardown.

## Merge Pitch

This is worth advancing because it closes a user-facing failure mode with low implementation surface area: silent thread death during broker restarts. The change is operationally useful, narrow in scope, and aligned with RFC #78 Phase 1.

The risk profile is moderate-low. The likely reviewer concern is not whether the feature is useful, but whether shutdown-path messaging is too coupled to Discord delivery, too optimistic under rate limits, or under-tested for edge cases like invalid thread IDs and partial shutdown failures.

## Best-Practice Comparison

OpenClaw principles that are relevant here:
- Explicit delivery routing is relevant. This PR makes shutdown communication an explicit broker action aimed at the affected threads rather than leaving behavior implicit.
- Retry/backoff and run logs are somewhat relevant. The PR logs failures, but it does not retry notification delivery or produce durable shutdown event records.
- Gateway-owned scheduling is only weakly relevant. This is shutdown orchestration, not scheduled work.
- Durable job persistence is not addressed. The PR explicitly stops short of preserving or replaying active session state.
- Isolated executions are mostly not relevant to this specific change.

Hermes Agent principles that are relevant here:
- Fresh session per scheduled run is conceptually adjacent but not directly applicable. This PR does not create a new run model; it just acknowledges interruption.
- Self-contained prompts for scheduled tasks are not especially relevant.
- Atomic writes for persisted state and file locking to prevent overlap are not relevant because this phase intentionally avoids persistence.
- Gateway daemon tick model is also not directly relevant, since this is signal-driven shutdown behavior rather than periodic coordination.

Net assessment:
- The PR aligns best with the principle of explicit operational lifecycle handling.
- It does not yet align with the durability and recovery patterns seen in stronger production-grade systems.
- That is acceptable if the item is judged strictly as Phase 1 notification, not as full graceful shutdown.

## Implementation Options

### Option 1: Minimal signal-path broadcast

Keep the current approach: handle `SIGINT` and `SIGTERM` in `main`, fetch active thread IDs from the pool via a narrow accessor, post a fixed message sequentially, warn on failures, then continue shutdown.

This is the most conservative path. It preserves existing boundaries reasonably well and keeps the PR tightly scoped to user notification.

### Option 2: Lifecycle notifier abstraction

Introduce a small shutdown notifier component or trait owned by the broker layer. The pool exposes active session identifiers, while the broker invokes a notifier responsible for formatting and delivering shutdown notices. The initial implementation still targets Discord threads, but the shutdown path becomes easier to test and less ad hoc.

This is the balanced option. It adds a small abstraction without dragging Discord concerns into `SessionPool` or overengineering the feature.

### Option 3: Broader graceful-shutdown framework

Implement a more complete shutdown coordinator: optional drain window for in-flight streams, bounded concurrency for thread notifications, structured shutdown result logging, and groundwork for Phase 2 persistence or resumable session metadata.

This is the ambitious option. It better matches mature gateway/runtime patterns, but it expands scope and risks turning a clean Phase 1 merge into a broader design exercise.

## Comparison Table

|---|---|---|---|---|---|---|
| Minimal signal-path broadcast | Fast | Low | Medium | Medium | Immediate and clear | Strong |
| Lifecycle notifier abstraction | Medium | Medium | Medium-High | High | Immediate and clear | Strong if reviewers want cleaner boundaries |
| Broader graceful-shutdown framework | Slow | High | High | Medium-High | Highest long-term value | Weak for this PR, better as follow-up work |

## Recommendation

Recommend the balanced version of Option 1 leaning slightly toward Option 2: merge the shutdown notification behavior now, but keep the design disciplined enough that notification delivery is clearly broker-owned and easy to evolve.

That is the right step for follow-up discussion because it solves the real user problem immediately without pretending to solve persistence or resumability. If reviewers are comfortable with the current narrow accessor plus `main`-layer delivery, this can merge as-is with a request for tighter tests. If they want cleaner seams, the only acceptable expansion should be a very small notifier abstraction, not a full graceful-shutdown framework. Phase 2 persistence, drain timing, and rate-limit mitigation should stay split into later items.
</details>

ruan330:
Force-pushed `fa106cd` — apologies for the churn. The previous commit (`82c8a7e`) was a straight rebase of the original Discord-only design, which no longer matches the multi-adapter shape the codebase moved to after #259. This revision rewrites against it.

**Mapping to the screening concerns**:
- *Too coupled to Discord* — broadcast now routes through `ChatAdapter::send_message` via a new `AdapterRouter::broadcast_shutdown()`, so Slack threads get the notification alongside Discord. Discord specifics stay in the adapter.
- *Rate limits* — `JoinSet` parallel send bounded by a 10s `tokio::select!` deadline. Best-effort by design.
- *Edge cases* — addressing moved into `SessionPool` next to `state.active`, and `begin_shutdown()` flips the flag + snapshots under one state write lock. Admission invariant: every session is either in the snapshot or rejected inline at `get_or_create`.

**Design alternatives considered** (before picking pool-owned addressing):
- *Router-only, track in-flight `handle_message` calls* — rejected; narrows broadcast scope too much (recently-chatted threads that aren't mid-reply wouldn't be notified).
- *Router cache + pool filter* — rejected; coordinating admission/snapshot atomicity across two data sources produced multiple race windows in review.
- *Pool owns addressing* (chosen) — single source of truth, admission atomicity reuses pool's existing state lock, adapter layer stays stateless.

Tests: 43 passed, clippy clean. CI green.

ruan330:
Force-pushed `12aa620` to address the remaining feedback:

- **Extracted `SHUTDOWN_MSG` + `SHUTDOWN_BROADCAST_TIMEOUT` to module-level consts** (@masami-agent's ⚪ nit). The message is defined once and reused across the Discord and Slack-only shutdown paths.

- **Closed a Slack feedback-loop in the admission-reject path**. The earlier revision sent an inline "Bot is shutting down" message back to any sender during the shutdown window. When the Slack Socket Mode loop is still running, our own broadcast post is delivered back as a message event — under `allow_bot_messages = "all"` it routes into `handle_message`, hits the reject branch, and generates another bot-authored reply, looping until the bot-turn cap. Fix: skip the inline rejection when `sender.is_bot`; human senders still get the notice. Discord is unaffected because the handler filters `msg.author.id == bot_id` up front.

Tests: 43 passed, clippy clean. CI green.

shaun-agent:
small maintainer note: the earlier CI failure was not a design objection to the shutdown feature itself. the real issue was branch drift against current `main` — the adapter interface had moved to `sender_json`, so the PR branch no longer compiled cleanly on the merge result.\n\nwe took over that compatibility cleanup on the PR branch, pushed the follow-up fixes (`9e7deaa`, `13563f9`), and reran the workflows. current head `13563f9` is now green on CI and Docker smoke tests.\n\nso at this point the remaining discussion should be about the feature and review judgment, not merge-drift noise.

shaun-agent:
handoff note: the earlier CI failures were merge-drift noise, not a feature-design rejection. we rebased/fixed the branch against current `main`, pushed the compatibility cleanup, and current head `13563f9` is now green on Rust CI and Docker smoke tests.\n\nat this point the remaining question is review judgment on the shutdown-broadcast design and scope, not build breakage.

masami-agent (commented):
## PR Review: #182

### Summary
- **Problem**: When the broker shuts down (SIGINT/SIGTERM), active conversations die silently mid-reply with no user-facing signal. RFC #78 §1d Phase 1.
- **Approach**: Pool-owned addressing + atomic shutdown flag + parallel broadcast via `JoinSet` with timeout, platform-neutral through the `ChatAdapter` abstraction.
- **Risk level**: Medium — touches core pool admission logic and shutdown path.

### Core Assessment
1. Problem clearly stated: ✅ — Excellent PR description with design rationale, behavior notes, and test plan.
2. Approach appropriate: ✅ — Pool-owns-addressing is the right call. No parallel cache in the adapter layer, single source of truth.
3. Alternatives considered: ✅ — PR description explicitly contrasts with the original Discord-specific approach and explains why the multi-platform refactor makes this the right home.
4. Best approach for now: ✅ — Minimal, well-scoped. Deliberately avoids session persistence (Phase 2 follow-up).

### Findings

**Shutdown admission invariant (pool.rs)** — The 4-point check in `get_or_create` is thorough and correct:
1. Fast-fail on entry (cheap, avoids spawning ACP process).
2. Inside `state.read()` block (catches shutdowns before this task ran).
3. After `conn.lock()` with `state.read()` barrier (synchronizes with `begin_shutdown`'s write-lock flag flip — this is the subtle one, and the comment explains it well).
4. Inside final `state.write()` before insert (atomic with flag-flip + snapshot).

The `begin_shutdown()` method correctly holds the state write lock while flipping the `AtomicBool` and snapshotting `addresses`, so any admission that committed before the lock is in the snapshot, and any that comes after sees the flag. Sound.

**`addresses` invariant** — `addresses.keys() == active.keys()` is maintained across all mutation paths: insert, stale rebuild, eviction, cleanup_idle, and shutdown. I traced each one — all correct.

**Broadcast timeout (adapter.rs)** — The `biased` select with `set.shutdown().await` is correct. `set.shutdown()` cancels remaining tasks and waits for them to drop, which is the right behavior for HTTP calls that should be cancellable via tokio's cooperative cancellation.

**Bot-loop prevention (adapter.rs)** — Smart catch. During the shutdown window, the bot's own broadcast message could arrive as a new event on Slack, trigger `handle_message`, get rejected by the pool, and the rejection reply would loop. Deserializing `sender_json` to check `is_bot` breaks the cycle cleanly. The `Deserialize` derive on `SenderContext` is the minimal change needed.

**Signal handling (main.rs)** — `wait_for_shutdown_signal()` with `#[cfg(unix)]` SIGTERM + ctrl_c, and a ctrl_c-only fallback for non-Unix, is clean and correct. Both the Discord and Slack-only branches use the same flow.

**discord.rs / slack.rs** — Purely rustfmt line wrapping. No logic changes. ✅

### Review Summary

#### 🔴 Blockers
(none)

#### 💬 Questions
1. **`get_or_create` signature change** — The new `channel: &ChannelRef, adapter: &Arc<dyn ChatAdapter>` parameters are passed through from `handle_message`. Is there any other call site for `get_or_create` outside of `handle_message`? If so, those callers would need updating. (I only see the one call site in `adapter.rs`, so this should be fine — just confirming.)

#### 🔧 Suggested Changes
1. **Commit history cleanup** — The branch has merge commits from "Codex Temp" (`48034166`, `9e7deaa9`, `13563f9c`). Since the project uses squash merge, this will be cleaned up at merge time, but if you're planning any further pushes to this branch, consider rebasing to a single commit for cleaner `git log` during review.

#### ℹ️ Info
1. The `_sync = self.state.read().await` pattern in check #3 is unusual (acquiring a read lock purely for synchronization), but the comment explains the intent clearly. The RwLock acquire provides the happens-before relationship needed to make the `AtomicBool` store from `begin_shutdown` visible. This is a valid and well-documented pattern here.
2. `broadcast_shutdown` sends via `adapter.send_message()` (Discord/Slack HTTP API), not through ACP connections, so `pool.shutdown()` clearing `active` connections afterward does not interfere with in-flight broadcast sends. The sequencing in `main.rs` (broadcast → shutdown shards → pool.shutdown()) is correct.
3. The `AtomicBool` with Acquire/Release ordering is technically redundant with the RwLock-protected checks in points #2 and #4 (which already have happens-before from the lock), but it's needed for the fast-fail path (#1) and the read-barrier path (#3). Belt-and-suspenders for shutdown correctness is the right trade-off.

#### ⚪ Nits
(none — code is clean, comments are thorough)

### Verdict
**APPROVE** — This is a well-designed, well-documented feature. The shutdown admission invariant is sound, the race condition coverage is thorough, and the platform-neutral broadcast approach is the right architecture. No blockers.

obrutjack (approved):
Merge checklist verified:

- ✅ CI all green (cargo check + 7 Docker smoke tests)
- ✅ masami-agent reviewed — no blockers, thorough analysis of shutdown admission invariant
- ✅ Multi-platform broadcast via ChatAdapter (Discord + Slack)
- ✅ Parallel broadcast with 10s timeout (JoinSet + deadline)
- ✅ SIGTERM + SIGINT handling (unix + non-unix)
- ✅ Bot-loop prevention during shutdown window
- ✅ Pool admission 4-layer shutdown check — sound
- ✅ Neutral wording (no "restarting" assumption)
- ✅ No version regression

Well-designed feature. The pool-owned addressing approach is the right architecture. Pending @thepagent.

ruan330:
Rebased onto current `main` (beta.8) to clear the new conflicts introduced since the previous head. Three commits preserved, content-equivalent to the previous `13563f9` head — `shaun-agent`'s JSON-parse approach for the admission-reject `is_bot` check is kept intact, just replayed on the new base.

Summary of conflict resolution:
- **`src/acp/pool.rs`** — combined upstream's `cancel_handles` (from #478 `/cancel`) with F's `addresses` map side-by-side in `PoolState`; imports, init, and insert paths mirror the combine.
- **`src/main.rs`** — kept upstream's `allow_all_channels` / `allow_all_users` Handler fields (#476) and preserved `router: router.clone()` so F's shutdown handler can still clone it on the line below.
- **`src/adapter.rs`** — auto-merged cleanly; no manual changes.

Tests: `cargo test --release` → 51/51 passed. `cargo clippy --all-targets -- -D warnings` clean.

New head SHAs:
- `2cbfe26` (was `12aa620`) — original F
- `3fe3137` (was `9e7deaa`) — adapt to `sender_json` API
- `44a9cdd` (was `13563f9`) — derive `is_bot` from `sender_json`

Thanks for your patience with the rebase churn — happy to adjust further if reviewers want a different shape.

chaodu-agent:
## Triage Review — 3 Lights

🟢 **INFO**

- Shutdown ordering is correct: broadcast fires while the HTTP client is still alive (before `shard_manager.shutdown_all()`).
- Four-point shutdown check in `get_or_create` properly closes race conditions, including the gap while waiting on the per-connection mutex.
- `JoinSet` + 10s deadline is a pragmatic best-effort design — shutdown is never blocked by a slow platform.
- Bot feedback-loop fix (Slack `is_bot` parse) prevents a real production footgun where the broadcast message re-enters `handle_message`.
- `SHUTDOWN_MSG` / `SHUTDOWN_BROADCAST_TIMEOUT` extracted to module-level consts — clean and easy to find.

🟡 **NIT**

- The `addresses` invariant comment claims `addresses.keys() == active.keys()`, but `remove_if_same_handle` can return `None` (different handle) without triggering `addresses.remove`. Practically unlikely to diverge, but the comment overstates the guarantee. Consider softening to "superset" or adding a defensive remove on the `None` path.
- `wait_for_shutdown_signal()` uses `.expect("install SIGTERM handler")` — would panic on fd exhaustion. Extremely unlikely, but a graceful fallback to ctrl_c-only would be more defensive.

🔴 **SUGGESTED CHANGES**

- **Sync-only read lock readability**: The `_sync = self.state.read().await` in the existing-alive-session path is purely a synchronization barrier, not a data access. The inline comment explains it, but the pattern is unusual enough that a future contributor could easily remove it as "dead code." It might be worth considering a more explicit marker — e.g. renaming to `_shutdown_barrier` or adding a `// DO NOT REMOVE` annotation — so the intent survives refactors.
- **Parallel HashMaps → unified struct**: `addresses` and `active` share identical lifetimes and keys. Merging them into a single `SessionEntry { conn, channel, adapter }` would enforce the invariant at the type level and eliminate any divergence risk. This is a larger refactor — suggestion for maintainers to consider as a follow-up rather than blocking this PR.
- **Configurable shutdown message**: The hardcoded const works for Phase 1, but operators may want to customize it (e.g. "Back in 5 min" or i18n). A future `shutdown_message` config field would be low-effort and high-value — worth tracking as a follow-up item.

DC https://discord.com/channels/1491295327620169908/1492427166007300096/1495684384643223642

chaodu-agent:
## 📎 Related PR: #459 (persist thread→session mapping)

Cross-referencing for context: #459 adds `save_mapping()` to `pool.shutdown()`, persisting the thread→session HashMap to PVC so sessions survive pod restarts.

These two PRs are complementary:
- **#182** ensures `pool.shutdown()` is called on SIGTERM (not just ctrl_c)
- **#459** ensures `pool.shutdown()` actually persists the mapping before exit

Together they complete the graceful restart loop: signal → broadcast → persist mapping → restart → reload mapping → resume sessions.

chaodu-agent:
## How #182 and #459 connect — separation of concerns

```
┌─────────────────────────────────────────────────────────────┐
│  OS / Kubernetes                                            │
│                                                             │
│  kubectl rollout restart ──► SIGTERM to pod                 │
└─────────────────────────────┬───────────────────────────────┘
                              │
                              ▼
┌─────────────────────────────────────────────────────────────┐
│  main.rs  (#182 — signal handling)                          │
│                                                             │
│  wait_for_shutdown_signal()                                 │
│    ├─ tokio::signal::ctrl_c()    ← SIGINT                   │
│    └─ sigterm.recv()             ← SIGTERM                  │
│                                                             │
│  router.broadcast_shutdown()     ← notify active threads    │
│  pool.shutdown().await ──────────┐                          │
└──────────────────────────────────┼──────────────────────────┘
                                   │
                                   ▼
┌─────────────────────────────────────────────────────────────┐
│  pool.rs  (#459 — persist mapping)                          │
│                                                             │
│  pub async fn shutdown(&self) {                             │
│      // 1. collect active → suspended                       │
│      // 2. save_mapping() → thread_map.json on PVC          │
│      // 3. active.clear()                                   │
│  }                                                          │
└─────────────────────────────────────────────────────────────┘
                                   │
                                   ▼
┌─────────────────────────────────────────────────────────────┐
│  Pod restart                                                │
│                                                             │
│  SessionPool::new()                                         │
│    └─ load_mapping() ← read thread_map.json from PVC        │
│                                                             │
│  Thread resumes → lookup mapping → session/load → ✅         │
└─────────────────────────────────────────────────────────────┘
```

Key insight: **#182 owns "when" to shutdown** (signal handling + orchestration). **#459 owns "what to do" during shutdown** (persist the mapping). The contract between them is simply `pool.shutdown()` — a pre-existing public API that neither PR needed to change the signature of.

ruan330:
Thanks for the thorough triage + the architectural framing — both really helpful.

**On the triage review**:

Most of the larger suggestions already got addressed by @shaun-agent's follow-up commits while the branch was being made mergeable:

- **`a7e15d4`** adds a graceful fallback when the SIGTERM handler fails to install (NIT 2).
- **`7ac9958`** merges `active` + `addresses` into a unified `SessionEntry` struct (SUGGESTED 2). The `addresses.keys() == active.keys()` invariant from NIT 1 is now type-enforced — the two fields share the same lifetime because they are the same field.

For the remaining items:

- SUGGESTED 1 (`_sync` readability) — happy to rename to `_shutdown_barrier` + `// DO NOT REMOVE` if reviewers prefer. One-line change, just say the word.
- SUGGESTED 3 (configurable `shutdown_message`) — agreed, follow-up. Not expanding this PR's scope.

**On #182 ↔ #459**:

The separation-of-concerns framing matches my mental model too — #182 answers "when" (signal capture + orchestration), #459 answers "what to persist" (thread→session map). Since both sides meet at the pre-existing `pool.shutdown()` without touching its signature, they can land in either order. Happy to rebase if #459 goes first; expecting a trivial pool.rs merge.

Thanks again to @chaodu-agent for the review and @shaun-agent for the mergeability cleanup.

shaun-agent (approved):
approving as Human Shaun after manual review. conflict is resolved and CI is green.

shaun-agent:
@pahud Human Shaun has manually approved this PR. conflict is resolved and CI is green. please merge when ready.

chaodu-agent:
## 四法師 Collaborative Triage Review — PR #182 (2026-05-01)

**Verdict: CHANGES_REQUESTED** — Design is sound and well-documented. Blocking on rebase against current `main` (CONFLICTING). Contributor must rebase carefully to preserve features added to `main` since last push.

<details>
<summary>🟢 INFO — What looks good</summary>

**All four monks agree the core design is correct:**

- **4-Point Atomic Admission Check** (超渡 + 覺渡): The layered shutdown check in `get_or_create` — (1) lock-free fast-fail, (2) state read lock, (3) `_shutdown_barrier` read lock after per-connection mutex, (4) final state write lock — is thorough and correctly closes all race windows between `begin_shutdown` and in-flight admissions.
- **Pool-Owned Addressing** (超渡 + 覺渡 + 擺渡): `SessionEntry` bundling `conn + channel + adapter` is the right architecture. Single source of truth, no parallel cache in the adapter layer, type-level lifetime enforcement.
- **JoinSet Parallel Broadcast + 10s Timeout** (覺渡): `tokio::task::JoinSet` with `tokio::select!` hard deadline ensures shutdown is never blocked by a slow platform. `biased` select + `set.shutdown().await` correctly cancels remaining tasks.
- **Bot Feedback-Loop Prevention** (超渡 + 覺渡): `sender_is_bot` check in the admission-reject path prevents Slack broadcast → re-entry → rejection → loop.
- **SIGTERM + SIGINT Handling** (超渡): `wait_for_shutdown_signal()` with `#[cfg(unix)]` SIGTERM and graceful fallback on handler installation failure.
- **2 Existing Approvals**: obrutjack (merge checklist verified) + shaun-agent (manual human approval, CI green at time of review).

</details>

<details>
<summary>🟡 NIT — Non-blocking suggestions</summary>

- **`SessionEntry` visibility** (覺渡): Currently private. Future monitoring features (list active sessions) may benefit from exposing it or adding getters. Non-blocking — follow-up.
- **Broadcast log line** (覺渡): `wait_for_shutdown_signal` logs `shutdown signal received` but no log before broadcast starts. Adding `info!("broadcasting shutdown notifications...")` in `main.rs` before `broadcast_shutdown()` would improve observability.
- **`begin_shutdown` early-return guard** (普渡): Could add `if self.is_shutting_down() { return vec![]; }` at the top of `begin_shutdown` to make double-call safe. Only one call site today, so non-blocking.
- **Configurable shutdown message** (超渡 previous triage): Hardcoded `SHUTDOWN_MSG` const works for Phase 1. Operators may want customization (i18n, ETA). Track as follow-up.

</details>

<details>
<summary>🔴 SUGGESTED CHANGES — Blocking</summary>

### 1. Rebase Required (all four monks)

PR merge state is `CONFLICTING`. The branch must be rebased onto current `main` before merge.

### 2. Rebase Risk: Session Persistence Regression (擺渡)

**Critical finding**: Current `main` (post-#459) has `mapping_path`, `load_mapping()`, and `save_mapping()` in `SessionPool` for persisting thread→session mappings to `~/.openab/thread_map.json`. The PR head predates this and replaces the pool struct without these fields.

If the rebase naively takes the PR version of `PoolState`/`SessionPool::new()`, session persistence will silently regress. The contributor must:
1. Keep `main`'s `mapping_path`, `load_mapping`, `save_mapping` intact
2. Apply `SessionEntry` refactor and `shutting_down` flag **on top of** the persisted-pool implementation
3. Verify `cargo test` passes after rebase, specifically any session-persistence tests

### 3. Rebase Risk: main.rs Feature Regression (擺渡)

Current `main` has `cron`, `markdown`, `gateway` modules and startup paths that the PR head lacks (based on older entrypoint). Rebase must preserve all current `main` features — the shutdown broadcast should be additive only.

</details>

<details>
<summary>📋 Review Participants</summary>

| Monk | Backend | Key Contribution |
|------|---------|-----------------|
| 超渡法師 | Kiro | Orchestration, 4-point check validation, bot-loop analysis |
| 覺渡法師 | Gemini | Deep code audit, JoinSet/timeout analysis, rebase conflict verification |
| 擺渡法師 | Codex | Baseline check — identified session persistence + main.rs regression risks |
| 普渡法師 | Claude | Implementation detail verification, `begin_shutdown` guard suggestion |

</details>

---

**Next step for contributor @ruan330**: Please rebase onto current `main`, paying special attention to `src/acp/pool.rs` (preserve `mapping_path`/`load_mapping`/`save_mapping` from #459) and `src/main.rs` (preserve cron/markdown/gateway modules). Once rebased and CI green, this should be ready to merge given the existing approvals and the sound design.

chaodu-agent:
## Follow-up: Precise Rebase Guidance (四法師 supplemental)

普渡法師 verified current `main` baseline and found:

- **`main` already has `shutdown_signal()`** (line 29) — functionally equivalent to the PR's `wait_for_shutdown_signal()` (SIGINT OR SIGTERM, `#[cfg(unix)]` + fallback).
- **`main` does NOT have** `broadcast_shutdown`, `begin_shutdown`, `SessionEntry`, or the 4-point admission check — these are all net-new from this PR.

### Rebase checklist for @ruan330

| File | Action |
|------|--------|
| `src/main.rs` | Delete `wait_for_shutdown_signal()`. Use `main`'s existing `shutdown_signal()`. Insert `broadcast_shutdown()` call before shard teardown. Preserve `cron`/`markdown`/`gateway` modules. |
| `src/acp/pool.rs` | Apply `SessionEntry` + `shutting_down` flag + `begin_shutdown()` **on top of** current `main`'s `mapping_path`/`load_mapping`/`save_mapping`. Do not overwrite persistence. |
| `src/adapter.rs` | Apply `broadcast_shutdown()` + shutdown admission logic. Preserve `origin_event_id` and table-mode flow from current `main`. |
| `src/discord.rs` | Formatting only — should auto-merge cleanly. |
| `src/slack.rs` | Formatting only — should auto-merge cleanly. |

Once rebased and `cargo test` passes, the existing approvals + this triage should clear the path to merge.

chaodu-agent (commented):
## PR Review: #182 — Broadcast Shutdown Notification

**CHANGES REQUESTED** — Solid design with atomic admission + snapshot, but the PR has **merge conflicts** (`mergeStateStatus: DIRTY`) that must be resolved before merge. The main branch has evolved significantly since this PR was opened.

<details>

**Main branch already has:**
- `shutdown_signal()` in `main.rs` — handles SIGINT + SIGTERM with `#[cfg(unix)]` gating
- `pool.shutdown()` — clears active sessions on exit
- `router` is `Arc<AdapterRouter>`, already cloned across Discord/Slack/Gateway/Cron tasks
- `get_or_create(&self, thread_id: &str)` — single-arg signature
- `PoolState` uses `active: HashMap<String, Arc<Mutex<AcpConnection>>>` (flat map)

**Main does NOT have:**
- Any broadcast notification to active threads on shutdown
- `SessionEntry` struct or `ChannelRef`/`ChatAdapter` stored in pool
- `begin_shutdown()` / `is_shutting_down()` on `SessionPool`
- `broadcast_shutdown()` on `AdapterRouter`

**Verdict:** Feature is entirely net-new. No stale/duplicate content.

</details>

<details>
<summary>Four Questions</summary>

When the broker shuts down (SIGINT/SIGTERM, `docker stop`, `helm uninstall`), active conversations die silently. Users see replies cut off mid-stream with no explanation. RFC #78 Phase 1d calls this out.

### 2. How does it solve it?
- **Pool owns addressing**: `SessionPool` stores `ChannelRef` + `Arc<dyn ChatAdapter>` alongside each connection in a new `SessionEntry` struct, keeping them in lockstep with `sessions` (renamed from `active`).
- **Atomic admission + snapshot**: `begin_shutdown()` takes the state write lock, flips an `AtomicBool` flag, and snapshots all session addresses in one critical section.
- **Four-point admission guard**: `get_or_create` checks `is_shutting_down()` at entry, inside `state.read()`, after the per-connection mutex (with a read-lock barrier), and inside the final `state.write()` before insert.
- **Platform-neutral broadcast**: `AdapterRouter::broadcast_shutdown()` sends to all snapshot entries in parallel via `JoinSet`, capped by a 10s timeout.
- **Unified signal handling**: `wait_for_shutdown_signal()` handles both SIGINT and SIGTERM on Unix.

### 3. What was considered?
- This is a rewrite of the original Discord-specific approach against the multi-platform `AdapterRouter`/`ChatAdapter` abstraction from #259.
- Wording is intentionally neutral ("Context will reset on return") — no promise of session resume (that is RFC #78 Phase 2).
- Best-effort delivery: send errors and the 10s deadline both fall through to normal teardown.

Strengths:
- The atomic admission + snapshot design is well-reasoned. The four-point shutdown check in `get_or_create` is thorough and each check point is justified in comments.
- Pool-owns-addressing is cleaner than a parallel cache in the adapter layer.
- `JoinSet` + `tokio::select!` with deadline is the right pattern for bounded parallel broadcast.
- The bot-loop prevention (skipping shutdown rejection for bot-authored events) is a nice edge case catch.

Concerns listed in Traffic Light below.

</details>

<details>
<summary>Traffic Light Summary</summary>

- 🔴 **MERGE CONFLICTS** — PR is in DIRTY merge state. Main has evolved (e.g., `shutdown_signal()` already exists, `PoolState` structure may have changed). Contributor must rebase onto current main and resolve conflicts.
- 🔴 **Verify `is_bot` field** — The shutdown-rejection path deserializes `SenderContext` and reads `.is_bot`. Confirm this field exists in the current `SenderContext` struct on main. If not, it needs to be added or the deserialization will silently default.
- 🟡 **Consolidate signal handlers** — Main already has `shutdown_signal()`. The PR adds `wait_for_shutdown_signal()` with identical logic. After rebase, consolidate to one.
- 🟡 **`_shutdown_barrier` fragility** — Consider extracting the read-lock-as-barrier pattern into a named helper method to prevent accidental removal.
- 🟢 **Atomic admission design** — The four-point shutdown check with justified comments at each point is excellent defensive programming.
- 🟢 **Bot-loop prevention** — Smart edge case handling for Slack where the bot's own broadcast could trigger a re-entry.
- 🟢 **Timeout-bounded broadcast** — `JoinSet` + `tokio::select!` with 10s deadline ensures shutdown is never blocked by a slow platform.

</details>

CI is all green (cargo check + 7 Docker smoke tests pass), but the merge conflicts must be resolved first. Previous reviews: masami-agent commented (no blockers), shaun-agent approved, obrutjack's approval was dismissed (likely due to new commits).

**Action needed:** Contributor @ruan330 please rebase onto current `main` and resolve conflicts. The `shutdown_signal()` function already exists on main — consolidate with your `wait_for_shutdown_signal()` during rebase.

chaodu-agent:
Not planned for now.

PR #186, opened 2026-04-10
feat: /model slash command with native Discord autocomplete

sh14y:
**Design discussion:** https://discord.com/channels/1491295327620169908/1491365157010542652/1493920430283686049

## Summary

Adds runtime model switching via Discord's native `/model` slash command, with autocomplete-driven model selection.

## What changed

This PR contains three commits that build on each other:

1. **`feat: add !model command for runtime model switching via session/set_model`**
   - Wires up the ACP `session/set_model` JSON-RPC method (verified on Kiro CLI v1.29.3)
   - Parses `models.availableModels` from `session/new` response
   - Adds alias resolution: `auto`, `opus`, `sonnet`, `haiku`
   - Initial text-based `!model` interface (superseded in commit 3 below)

2. **`fix: self-bootstrap kiro creds and config.toml inside the binary`**
   - Required for the cloud deployment we're using to test this; restores credentials from `KIRO_CRED_B64` and generates `config.toml` from env vars when missing. Idempotent — no-op when files already exist.

3. **`feat: replace !model text command with /model slash command + autocomplete`** ← the main UX change
   - Removes the `!model` text intercept entirely
   - Registers `/model` as a guild command on `ready()` for instant propagation
   - Adds `interaction_create` handler for both `Command` and `Autocomplete` interactions
   - `SessionPool` gains a `cached_models` snapshot updated on every `session/new`, so autocomplete can answer within Discord's 3-second deadline without spawning a fresh agent (~10s cold start)
   - Background warmup task at startup populates the model cache before first use
   - Aliases (`auto`, `opus`, `sonnet`, `haiku`) appear in autocomplete as friendly shortcuts only when their target model is actually available
   - The set path defers the response since cold session creation can exceed 3s

## UX

| Action | Result |
|--------|--------|
| Type `/m` in Discord | Native command picker shows `/model` |
| Tab into the `model` field | Autocomplete returns aliases + all available model ids (current model marked) |
| Pick with ↑↓, hit enter | Bot defers, sets the model on the channel/thread session, replies `✅ Switched to <id>` |
| `/model` with no argument | Bot replies with the full list and marks the current selection |

## Important: bot invite scope

`/model` will not appear unless the bot is invited with **both** `bot` **and** `applications.commands` OAuth scopes. Existing installations with only `bot` scope must re-invite using a new URL — the bot account stays the same and no data is lost.

The README has been updated to call this out in the Quick Start section.

## Test plan

- [x] `cargo build --release` clean
[...]

PR #192, opened 2026-04-11
feat: /model slash command with native Discord autocomplete

sh14y:
## Summary

Replaces the text-based `!model` command (#186) with a Discord-native `/model` slash command that supports autocomplete and arrow-key model selection.

> **Depends on #186** — this branch is stacked on `feat/model-switch-command` (which adds the underlying `session/set_model` ACP plumbing). Please review and merge #186 first.

## What changed

- **`SessionPool` cached_models** — every `session/new` now updates a global snapshot of available models on the pool, so autocomplete can serve suggestions instantly without spawning a fresh agent (`~10s` cold start would blow Discord's 3s autocomplete deadline).
- **Startup warmup** — `main.rs` spawns a background `__warmup__` session at boot so the cache is populated before the first slash command fires.
- **`/model` registration** — registered as a *guild command* on `ready()` for every guild the bot is in. Guild commands appear instantly; global commands take up to 1 hour to propagate.
- **`interaction_create` handler** — dispatches `Command` (the actual `/model` invocation) and `Autocomplete` (live filtering as the user types) interactions. The set path defers the response since cold session creation may exceed 3s.
- **Aliases in autocomplete** — `auto`, `opus`, `sonnet`, `haiku` are surfaced as friendly shortcuts (`sonnet → claude-sonnet-4.6`) but only when their target is actually available.
- **`!model` removed** — `/model` is the replacement, not an addition. The text-command intercept in `discord.rs` is gone.
- **README** — documents the `applications.commands` OAuth scope requirement and the new `/model` UX.

## UX

| Action | Result |
|--------|--------|
| Type `/m` in Discord | Native command picker shows `/model` |
| Tab into the `model` field | Autocomplete returns aliases + all available model ids (current model marked) |
| Pick with ↑↓, hit enter | Bot defers, sets the model on the channel/thread session, replies `✅ Switched to <id>` |
| `/model` with no argument | Bot replies with the full list and marks the current selection |

## Important: bot invite scope

`/model` will not appear unless the bot is invited with **both** `bot` **and** `applications.commands` OAuth scopes. Existing installations that only have `bot` scope will need to re-invite the bot using a new URL — the bot account itself stays the same and no data is lost.

The README has been updated to call this out in the Quick Start section.

## Test plan

- [x] `cargo build --release` clean
- [ ] Verify `/model` (no arg) lists all available models with current marker
- [ ] Verify `/model auto` switches and replies `✅ Switched to auto`
- [ ] Verify `/model sonnet` resolves alias and switches
- [ ] Verify autocomplete shows live filtered suggestions as you type
- [ ] Verify the warmup task populates the model cache before first use
- [ ] Verify allowlist (channel + user) is enforced for the slash command path

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -764,6 +764,7 @@ name = "openab"
 version = "0.1.0"
 dependencies = [
  "anyhow",
+ "base64",
  "rand 0.8.5",
  "regex",
  "serde",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,3 +15,4 @@ uuid = { version = "1", features = ["v4"] }
 regex = "1"
 anyhow = "1"
 rand = "0.8"
+base64 = "0.22"
--- a/README.md
+++ b/README.md
@@ -36,8 +36,10 @@ In short:
 
 1. Go to https://discord.com/developers/applications and create an application
 2. Bot tab → enable **Message Content Intent**
-3. OAuth2 → URL Generator → scope: `bot` → permissions: Send Messages, Send Messages in Threads, Create Public Threads, Read Message History, Add Reactions, Manage Messages
+3. OAuth2 → URL Generator → scopes: **`bot` + `applications.commands`** → permissions: Send Messages, Send Messages in Threads, Create Public Threads, Read Message History, Add Reactions, Manage Messages
 4. Invite the bot to your server using the generated URL
+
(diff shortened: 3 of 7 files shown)


sh14y:
Superseded by #186 — fast-forwarded the slash command commit onto `feat/model-switch-command` because our deployment target is locked to that branch. PR #186 now contains the full feature stack (ACP plumbing + slash command UI). Closing this duplicate.

PR #196, opened 2026-04-11
fix: remove trailing comma in TOML inline table

### LINKED ISSUE #195: Helm chart generates invalid TOML: trailing comma in inline table
## Description

The Helm chart template produces a trailing comma in TOML inline tables, which is invalid per the TOML spec.

## Generated output (invalid)

```toml
env = { GH_TOKEN = "xxx",  }
```

## Expected output

```toml
env = { GH_TOKEN = "xxx" }
```

## Root Cause

The Go template `range` loop appends a `,` after every key-value pair, including the last one. TOML inline tables do not allow trailing commas, causing the TOML parser to fail and pods to crash on startup.

## Suggested Fix

Use a `$first` flag pattern to only insert commas between items:

```gotemplate
env = { {{ $first := true }}{{ range $k, $v := .Values.env }}{{ if not $first }}, {{ end }}{{ $k }} = "{{ $v }}"{{ $first = false }}{{ end }} }
```

JARVIS-coding-Agent:
Fixes #195

## Problem

The Go template `range` loop in `configmap.yaml` appends a trailing comma after every key-value pair in the TOML inline table:

```toml
env = { GH_TOKEN = "xxx",  }
```

TOML does not allow trailing commas in inline tables, causing parser errors and pod crash loops.

## Fix

Use a `$first` flag to insert commas only between items, not after the last one.

```diff
- env = { {{ range $k, $v := $cfg.env }}{{ $k }} = "{{ $v }}", {{ end }} }
+ env = { {{ $first := true }}{{ range $k, $v := $cfg.env }}{{ if not $first }}, {{ end }}{{ $k }} = "{{ $v }}"{{ $first = false }}{{ end }} }
```

### DIFF
--- a/charts/openab/templates/configmap.yaml
+++ b/charts/openab/templates/configmap.yaml
@@ -30,7 +30,7 @@ data:
     args = {{ if $cfg.args }}{{ $cfg.args | toJson }}{{ else }}[]{{ end }}
     working_dir = "{{ $cfg.workingDir | default "/home/agent" }}"
     {{- if $cfg.env }}
-    env = { {{ range $k, $v := $cfg.env }}{{ $k }} = "{{ $v }}", {{ end }} }
+    env = { {{ $first := true }}{{ range $k, $v := $cfg.env }}{{ if not $first }}, {{ end }}{{ $k }} = "{{ $v }}"{{ $first = false }}{{ end }} }
     {{- end }}
 
     [pool]


chaodu-agent (approved):
LGTM 👍 乾淨俐落的 one-liner fix，解決 TOML trailing comma 導致 pod crash loop 的問題。

thepagent (approved):

PR #206, opened 2026-04-11
fix: keep Cargo.toml at stable version on main

thepagent:
`Cargo.toml` on main should always reflect the stable version, not pre-release suffixes like `-beta.x`.

### Changes
- `Cargo.toml`: `0.6.4-beta.1` → `0.6.4` (stable)
- `release-pr.yml`:
  - Version source changed from `Cargo.toml` to `Chart.yaml`
  - Beta releases only update `Chart.yaml`; `Cargo.toml` only gets the stable version
  - Removed `rust-toolchain` and `cargo generate-lockfile` (no longer needed)

### DIFF
--- a/.github/workflows/release-pr.yml
+++ b/.github/workflows/release-pr.yml
@@ -42,7 +42,7 @@ jobs:
           if [ -n "${{ inputs.version }}" ]; then
             VERSION="${{ inputs.version }}"
           else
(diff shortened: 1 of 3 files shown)


pahud (approved):
LGTM

pahud (approved):
LGTM

PR #228, opened 2026-04-11
helm: add first-class STT config to chart

chaodu-agent:
## Summary

Add `stt` as a first-class config block in the Helm chart, eliminating the need to manually patch configmaps after `helm upgrade`.

Closes #227

## Before (manual patching required)

```bash
helm upgrade openab openab/openab --set agents.kiro.env.GROQ_API_KEY=gsk_xxx
# then manually patch configmap to add [stt] block
# then kubectl rollout restart
```

## After (single command)

```bash
helm upgrade openab openab/openab \
  --set agents.kiro.stt.enabled=true \
  --set agents.kiro.stt.apiKey=gsk_xxx
```

## Changes

| File | Change |
|---|---|
| `values.yaml` | Add `stt` defaults (enabled, apiKey, model, baseUrl) |
| `configmap.yaml` | Render `[stt]` section when enabled, using `${STT_API_KEY}` env var |
| `secret.yaml` | Store `apiKey` in K8s Secret (same pattern as `discord.botToken`) |
| `deployment.yaml` | Inject `STT_API_KEY` env var from Secret |

## Design

Follows the existing `DISCORD_BOT_TOKEN` pattern — API key is stored in a K8s Secret and injected as an env var, never in plaintext in the configmap:

```
values.yaml          Secret              Deployment env        ConfigMap
stt.apiKey ────► stt-api-key ────► STT_API_KEY ────► ${STT_API_KEY}
```

All STT fields are optional. When `stt.enabled` is false (default), no STT resources are rendered.

PR #239, opened 2026-04-12
fix: remove hardcoded image.tag to use Chart.AppVersion

### LINKED ISSUE #235: bug: image.tag hardcoded in values.yaml causes version mismatch after helm upgrade
## Summary

`charts/openab/values.yaml` has `image.tag` hardcoded to a commit hash (`94253a5`), which prevents Helm from using the `appVersion` defined in `Chart.yaml`. This causes the deployed pod to run an outdated binary even after upgrading the Helm chart.

## Impact

The running OpenAB instance reports version `0.1.0` (from the old commit) instead of the expected `0.7.0`, because the image tag never changes on `helm upgrade`.

## Reproduction Steps

1. Deploy OpenAB using the Helm chart:
   ```bash
   helm upgrade --install openab ./charts/openab
   ```

2. Verify the deployed image:
   ```bash
   kubectl get deployment openab-kiro -o jsonpath="{.spec.template.spec.containers[0].image}"
   # Output: ghcr.io/openabdev/openab:94253a5
   ```

3. Check the running version:
   ```bash
   kubectl exec deployment/openab-kiro -- openab --version
   # Output: 0.1.0 (expected: 0.7.0)
   ```

4. Note that `Chart.yaml` has `appVersion: "0.7.0-beta.1"` but the pod still uses the hardcoded commit hash image.

## Root Cause

In `charts/openab/values.yaml`:
```yaml
image:
  repository: ghcr.io/openabdev/openab
  # tag defaults to .Chart.AppVersion
  tag: "94253a5"          # ← hardcoded, overrides appVersion fallback
```

The comment says it defaults to `appVersion`, but since a value is explicitly set, the template never falls back.

## Fix

Set `image.tag` to an empty string so the template uses `.Chart.AppVersion`:
```yaml
image:
  tag: ""
```

JARVIS-coding-Agent:
## Summary

Set `image.tag` to an empty string in `charts/openab/values.yaml` so the Helm template falls back to `.Chart.AppVersion`.

## Root Cause

`image.tag` was hardcoded to a commit hash (`94253a5`), which prevented `helm upgrade` from picking up the `appVersion` defined in `Chart.yaml`. Deployed pods ran an outdated binary as a result.

## Fix

```yaml
image:
  tag: ""
```

The existing template logic in `_helpers.tpl`:
```
{{- $tag := default .ctx.Chart.AppVersion .ctx.Values.image.tag }}
```
will now correctly resolve to the chart's `appVersion` (`0.7.0-beta.1`).

Closes #235

### DIFF
--- a/charts/openab/values.yaml
+++ b/charts/openab/values.yaml
@@ -1,7 +1,7 @@
 image:
   repository: ghcr.io/openabdev/openab
   # tag defaults to .Chart.AppVersion
-  tag: "94253a5"
+  tag: ""
   pullPolicy: IfNotPresent
 
 podSecurityContext:


thepagent (changes requested):
The `image.tag: ""` fix is correct, but this PR also rewrites the entire `values.yaml` from the old flat structure (`discord.*`, `agent.*`, `pool.*`) to the new `agents.kiro.*` structure. That migration was already done on main — so this diff is re-applying a change that already exists.

If merged as-is, it could overwrite recent additions (e.g. the `stt` block from #228) or cause merge conflicts.

**Suggested fix:** Please rebase onto latest `main` and reduce the diff to just the one-line change:

```yaml
image:
  tag: ""    # was: "94253a5" (or "78f8d2c")
```

Everything else in `values.yaml` should already match main after rebase.

thepagent (approved):
LGTM — clean one-liner, falls back to Chart.AppVersion as intended.

pahud (dismissed):

PR #244, opened 2026-04-12
feat: add allowed_bots_from config for bot-to-bot messaging

### LINKED ISSUE #243: feat: allow bot-to-bot messaging via configurable allowlist
## Description

openab currently hard-rejects all messages from other bots (`msg.author.bot → return` in `discord.rs:41-43`). This is a sensible default to prevent loops, but it blocks a legitimate and increasingly common architecture: **a dispatcher bot that routes tasks to openab worker agents**.

### Proposed change

Add an `allowed_bots_from` config option that accepts a list of bot user IDs whose messages should be processed instead of ignored:

```toml
[discord]
allowed_bots_from = ["123456789012345678"]
```

The bot filter logic would become:

```rust
if msg.author.bot {
    // Self-loop guard: never process own messages
    if msg.author.id == bot_id {
        return;
    }
    // Allow explicitly trusted bots
    if !self.allowed_bots_from.contains(&msg.author.id.to_string()) {
        return;
    }
}
```

When `allowed_bots_from` is empty (default), behavior is identical to today — all bot messages are rejected.

## Use Case

In a multi-agent setup, a central dispatcher bot creates threads, assigns tasks, and @mentions specific openab agents. Without bot-to-bot messaging, this workflow is impossible — the dispatcher's mentions are silently ignored.

An allowlist approach (rather than a blanket "accept all bots" flag) keeps the default safe:
- Only explicitly trusted bot IDs are accepted
- Self-loop guard prevents an agent from processing its own messages, even if misconfigured
- Unrelated bots in the same server cannot accidentally trigger agents

This pattern is used in production multi-agent deployments today via a fork, and upstreaming it would benefit anyone building dispatcher/worker architectures with openab.

chengli:
## Summary

- Add `allowed_bots_from` to `[discord]` config: a list of bot user IDs whose messages should be accepted instead of ignored
- Self-loop guard ensures an agent never processes its own messages, even if misconfigured
- Default behavior is unchanged — when the list is empty, all bot messages are rejected (same as today)
- Helm chart updated with `discord.allowedBotsFrom` + ID validation (same pattern as existing `allowedChannels`/`allowedUsers`)

Closes #243

## Details

In multi-agent architectures, a dispatcher bot creates threads and @mentions specific openab worker agents to route tasks. Currently `discord.rs` hard-rejects all bot messages (`msg.author.bot → return`), making this pattern impossible.

The fix introduces a targeted allowlist rather than a blanket "accept all bots" flag:

```toml
[discord]
allowed_bots_from = ["123456789012345678"]
```

Logic in `discord.rs`:

```rust
if msg.author.bot {
    if msg.author.id == bot_id { return; }          // self-loop guard
    if !self.allowed_bots_from.contains(&id) { return; }  // allowlist check
}
```

### Files changed

| File | Change |
|------|--------|
| `src/config.rs` | Add `allowed_bots_from: Vec<String>` to `DiscordConfig` |
| `src/discord.rs` | Replace hard reject with self-loop guard + allowlist check |
| `src/main.rs` | Parse + wire `allowed_bots_from` to Handler |
| `config.toml.example` | Document new option |
| `charts/openab/values.yaml` | Add `discord.allowedBotsFrom` |
| `charts/openab/templates/configmap.yaml` | Template + ID validation |

## Test plan

- [ ] `cargo check` passes ✅ (verified locally)
- [ ] Default config (no `allowed_bots_from`): all bot messages rejected — same as before
- [ ] With `allowed_bots_from = ["<bot_id>"]`: messages from that bot are processed
- [ ] Self-loop: agent's own messages are always rejected regardless of allowlist
- [ ] Helm: `helm template` with `allowedBotsFrom: ["123"]` renders correct TOML

### DIFF
--- a/charts/openab/templates/configmap.yaml
+++ b/charts/openab/templates/configmap.yaml
@@ -24,6 +24,12 @@ data:
     {{- end }}
     {{- end }}
     allowed_users = {{ $cfg.discord.allowedUsers | default list | toJson }}
+    {{- range $cfg.discord.allowedBotsFrom }}
+    {{- if regexMatch "e\\+|E\\+" (toString .) }}
+    {{- fail (printf "discord.allowedBotsFrom contains a mangled ID: %s — use --set-string instead of --set for bot IDs" (toString .)) }}
+    {{- end }}
+    {{- end }}
+    allowed_bots_from = {{ $cfg.discord.allowedBotsFrom | default list | toJson }}
 
     [agent]
     command = "{{ $cfg.command }}"
--- a/charts/openab/values.yaml
+++ b/charts/openab/values.yaml
@@ -60,6 +60,8 @@ agents:
         - "YOUR_CHANNEL_ID"
       # ⚠️ Use --set-string for user IDs to avoid float64 precision loss
       allowedUsers: []  # empty = allow all users (default)
+      # ⚠️ Use --set-string for bot IDs to avoid float64 precision loss
+      allowedBotsFrom: []  # bot user IDs to accept messages from (default: reject all bots)
     workingDir: /home/agent
     env: {}
     envFrom: []
--- a/config.toml.example
+++ b/config.toml.example
@@ -2,6 +2,7 @@
 bot_token = "${DISCORD_BOT_TOKEN}"
 allowed_channels = ["1234567890"]
 # allowed_users = ["<YOUR_DISCORD_USER_ID>"]  # empty or omitted = allow all users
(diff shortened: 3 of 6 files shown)


thepagent:
Closing — superseded by #321 (merged in v0.7.4) which added `allow_bot_messages` (off/mentions/all) + `trustedBotIds`. The `allowed_bots_from` functionality is now covered. Thanks! 🙏

PR #259, opened 2026-04-12
feat: add Slack adapter with multi-platform ChatAdapter architecture

### LINKED ISSUE #93: feat: Slack adapter — bring agent-broker to Slack workspaces

dogzzdogzz:
**Discord discussion:** https://discord.com/channels/1491295327620169908/1491969620754567270/1491969624135438386

## Summary

- Introduces `ChatAdapter` trait + `AdapterRouter`, decoupling session management, streaming, and reactions from any specific chat platform
- Refactors existing Discord support to implement the new trait (no behavior change)
- Adds Slack adapter using Socket Mode (WebSocket, no public URL needed) with auto-reconnect
- Extracts shared image/audio processing into `media.rs` for cross-adapter reuse
- Both adapters can run simultaneously, sharing one `SessionPool` with namespaced keys
- Updates Helm chart to support Slack deployment (backward compatible)

Closes #93. Implements Phase 1–3 of RFC #94.

## Changes

| File | Change |
|------|--------|
| `src/adapter.rs` | **New** — `ChatAdapter` trait, `ChannelRef`/`MessageRef`/`SenderContext`, `AdapterRouter` with streaming tool collapse (`TOOL_COLLAPSE_THRESHOLD=3`) |
| `src/slack.rs` | **New** — `SlackAdapter` + Socket Mode event loop, `markdown_to_mrkdwn` conversion, bot-loop protection (`MAX_CONSECUTIVE_BOT_TURNS=10`), silent `already_reacted`/`no_reaction` handling |
| `src/media.rs` | **New** — shared image resize/compress + STT download (extracted from discord.rs) |
| `src/discord.rs` | Refactored — `DiscordAdapter` implements `ChatAdapter`, delegates to router |
| `src/reactions.rs` | Decoupled from serenity — uses `Arc<dyn ChatAdapter>` |
| `src/format.rs` | Added shared `shorten_thread_name()` |
| `src/config.rs` | `discord` now optional, added `SlackConfig` |
| `src/main.rs` | Multi-adapter startup, Slack as background task |
| `config.toml.example` | Added `[slack]` section with docs |
| `charts/openab/values.yaml` | Added `slack` config per agent (enabled, botToken, appToken, allowedChannels, allowedUsers) |
| `charts/openab/templates/configmap.yaml` | Conditionally generates `[discord]` and/or `[slack]` sections |
| `charts/openab/templates/secret.yaml` | Stores `slack-bot-token` + `slack-app-token` alongside discord token |
| `charts/openab/templates/deployment.yaml` | Injects `SLACK_BOT_TOKEN` + `SLACK_APP_TOKEN` env vars |
| `charts/openab/templates/NOTES.txt` | Added Slack setup instructions |
| `docs/slack-bot-howto.md` | **New** — step-by-step Slack bot setup guide |

## Design Decisions

- **Socket Mode** over Events API — no public URL needed, mirrors Discord's persistent WebSocket model
- **Session keys namespaced** by platform (`discord:xxx`, `slack:xxx`) to prevent collision
- **Sender context wrapping** centralized in `AdapterRouter` (not duplicated per adapter)
- **Image/audio processing** shared via `media.rs` with optional `auth_token` param (Discord CDN is public, Slack files need Bearer token)
- **Unicode → Slack emoji mapping** in adapter (Slack reactions API requires short names like `eyes` not `👀`)
- **Slack user name resolution** via `users.info` API with 5-minute TTL cache
- **Thread follow-ups** — Slack `message` events with `thread_ts` handled without requiring re-mention (parity with Discord)
- **Helm backward compatible** — existing Discord-only values work unchanged; `[discord]` section now conditional (omitted when no botToken)

## Review fixes (by @thepagent)

| # | Issue | Fix | Commit |
|---|-------|-----|--------|
| 1 | Tool display collapse regression — `compose_display` lost `streaming` mode | Restored `streaming: bool` param with `TOOL_COLLAPSE_THRESHOLD=3`, matching original discord.rs behavior | `5cf50b1` |
| 2 | Slack `handle_message` blocks WebSocket event loop | Both `app_mention` and `message` handlers now use `tokio::spawn` | `5cf50b1` |
| 3 | No bot-loop protection for Slack | Added `is_bot_loop()` using `conversations.replies` with `MAX_CONSECUTIVE_BOT_TURNS=10` (thread follow-ups only, fail-open) | `82dca86` |
| 4 | `already_reacted` / `no_reaction` noisy error logs | Silently ignore these benign Slack API errors in `add_reaction` / `remove_reaction` | `82dca86` |

## Helm Usage

```bash
# Discord only (unchanged)
helm install openab openab/openab \
  --set agents.kiro.discord.botToken="$DISCORD_TOKEN" \
  --set-string 'agents.kiro.discord.allowedChannels[0]=CHANNEL_ID'

# Slack only
helm install openab openab/openab \
  --set agents.kiro.slack.enabled=true \
  --set agents.kiro.slack.botToken="$SLACK_BOT_TOKEN" \
  --set agents.kiro.slack.appToken="$SLACK_APP_TOKEN" \
  --set-string 'agents.kiro.slack.allowedChannels[0]=C0123456789'

# Both simultaneously
helm install openab openab/openab \
  --set agents.kiro.discord.botToken="$DISCORD_TOKEN" \
  --set-string 'agents.kiro.discord.allowedChannels[0]=DISCORD_CHANNEL' \
  --set agents.kiro.slack.enabled=true \
  --set agents.kiro.slack.botToken="$SLACK_BOT_TOKEN" \
  --set agents.kiro.slack.appToken="$SLACK_APP_TOKEN"
```

## Test plan

- [x] `cargo build` — zero warnings
- [x] `cargo test` — 40 tests pass
- [x] `cargo clippy` — clean
- [x] `helm template` — Discord-only, Slack-only, and both modes validated
- [x] Discord-only config works identically (no regression)
- [x] Slack-only config starts Socket Mode and responds to @mentions
- [x] Discord + Slack simultaneous config works
- [x] Thread follow-up in Slack (reply without @mention)
- [x] Reactions display correctly in Slack
- [x] Image attachments processed in Slack
- [x] Auto-reconnect on Socket Mode disconnect

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "openab"
-version = "0.7.4"
+version = "0.7.5"
 edition = "2021"
 
 [dependencies]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,8 +17,11 @@ anyhow = "1"
 async-trait = "0.1"
 futures-util = "0.3"
 rand = "0.8"
-reqwest = { version = "0.12", default-features = false, features = ["rustls-tls", "multipart", "json"] }
+clap = { version = "4", features = ["derive"] }
+rpassword = "7"
+reqwest = { version = "0.12", default-features = false, features = ["rustls-tls", "multipart", "json", "blocking"] }
 base64 = "0.22"
 image = { version = "0.25", default-features = false, features = ["jpeg", "png", "gif", "webp"] }
+unicode-width = "0.2"
 libc = "0.2"
 tokio-tungstenite = { version = "0.21", features = ["rustls-tls-webpki-roots"] }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "openab"
-version = "0.7.3"
+version = "0.7.4"
 edition = "2021"
 
 [dependencies]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "openab"
-version = "0.7.5"
+version = "0.7.6"
 edition = "2021"
 
 [dependencies]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,8 +14,11 @@ serenity = { version = "0.12", default-features = false, features = ["client", "
 uuid = { version = "1", features = ["v4"] }
 regex = "1"
 anyhow = "1"
+async-trait = "0.1"
+futures-util = "0.3"
 rand = "0.8"
 reqwest = { version = "0.12", default-features = false, features = ["rustls-tls", "multipart", "json"] }
 base64 = "0.22"
 image = { version = "0.25", default-features = false, features = ["jpeg", "png", "gif", "webp"] }
 libc = "0.2"
+tokio-tungstenite = { version = "0.21", features = ["rustls-tls-webpki-roots"] }
--- a/src/adapter.rs
+++ b/src/adapter.rs
@@ -25,12 +25,6 @@ pub struct ChannelRef {
     pub parent_id: Option<String>,
 }
 
-impl ChannelRef {
-    pub fn is_thread(&self) -> bool {
-        self.thread_id.is_some() || self.parent_id.is_some()
-    }
-}
-
 /// Identifies a message across platforms.
 #[derive(Clone, Debug)]
 pub struct MessageRef {
@@ -96,10 +90,6 @@ impl AdapterRouter {
             pool,
             reactions_config,
         }
-    }
-
-    pub fn pool(&self) -> &SessionPool {
-        &self.pool
     }
 
     /// Handle an incoming user message. The adapter is responsible for
--- a/src/adapter.rs
+++ b/src/adapter.rs
@@ -104,16 +104,41 @@ impl AdapterRouter {
 
     /// Handle an incoming user message. The adapter is responsible for
     /// filtering, resolving the thread, and building the SenderContext.
-    /// This method handles session management and streaming.
+    /// This method handles sender context injection, session management, and streaming.
     pub async fn handle_message(
         &self,
         adapter: &Arc<dyn ChatAdapter>,
         thread_channel: &ChannelRef,
-        _sender: &SenderContext,
-        content_blocks: Vec<ContentBlock>,
+        sender: &SenderContext,
+        prompt: &str,
+        extra_blocks: Vec<ContentBlock>,
         trigger_msg: &MessageRef,
     ) -> Result<()> {
         tracing::debug!(platform = adapter.platform(), "processing message");
+
+        // Build content blocks: sender context + prompt text, then extra (images, transcripts)
+        let sender_json = serde_json::to_string(sender).unwrap();
+        let prompt_with_sender = format!(
+            "<sender_context>\n{}\n</sender_context>\n\n{}",
+            sender_json, prompt
+        );
+
+        let mut content_blocks = Vec::with_capacity(1 + extra_blocks.len());
+        // Prepend any transcript blocks (they go before the text block)
+        for block in &extra_blocks {
+            if matches!(block, ContentBlock::Text { .. }) {
+                content_blocks.push(block.clone());
+            }
+        }
+        content_blocks.push(ContentBlock::Text {
+            text: prompt_with_sender,
+        });
+        // Append non-text blocks (images)
+        for block in extra_blocks {
+            if !matches!(block, ContentBlock::Text { .. }) {
+                content_blocks.push(block);
+            }
+        }
 
         let thinking_msg = adapter.send_message(thread_channel, "...").await?;
 
--- a/src/adapter.rs
+++ b/src/adapter.rs
@@ -250,8 +250,12 @@ impl AdapterRouter {
                                     let content = buf_rx.borrow_and_update().clone();
                                     if content != last_content {
                                         let display = if content.chars().count() > streaming_limit {
-                                            let truncated = format::truncate_chars(&content, streaming_limit);
-                                            format!("{truncated}…")
+                                            // Tail-priority: keep the last N chars so user
+                                            // sees the most recent agent output
+                                            let total = content.chars().count();
+                                            let skip = total - streaming_limit;
+                                            let truncated: String = content.chars().skip(skip).collect();
+                                            format!("…(truncated)\n{truncated}")
                                         } else {
                                             content.clone()
                                         };
--- a/src/adapter.rs
+++ b/src/adapter.rs
@@ -289,7 +289,7 @@ impl AdapterRouter {
                                     }
                                     text_buf.push_str(&t);
                                     let _ =
-                                        buf_tx.send(compose_display(&tool_lines, &text_buf));
+                                        buf_tx.send(compose_display(&tool_lines, &text_buf, true));
                                 }
                                 AcpEvent::Thinking => {
                                     reactions.set_thinking().await;
@@ -308,7 +308,7 @@ impl AdapterRouter {
                                         });
                                     }
                                     let _ =
-                                        buf_tx.send(compose_display(&tool_lines, &text_buf));
+                                        buf_tx.send(compose_display(&tool_lines, &text_buf, true));
                                 }
                                 AcpEvent::ToolDone { id, title, status } => {
                                     reactions.set_thinking().await;
@@ -330,7 +330,7 @@ impl AdapterRouter {
                                         });
                                     }
                                     let _ =
-                                        buf_tx.send(compose_display(&tool_lines, &text_buf));
+                                        buf_tx.send(compose_display(&tool_lines, &text_buf, true));
                                 }
                                 _ => {}
                             }
@@ -342,7 +342,7 @@ impl AdapterRouter {
                     let _ = edit_handle.await;
 
                     // Final edit with complete content
-                    let final_content = compose_display(&tool_lines, &text_buf);
+                    let final_content = compose_display(&tool_lines, &text_buf, false);
                     let final_content = if final_content.is_empty() {
                         if let Some(err) = response_error {
                             format!("⚠️ {err}")
@@ -405,14 +405,51 @@ impl ToolEntry {
     }
 }
 
-fn compose_display(tool_lines: &[ToolEntry], text: &str) -> String {
+/// Maximum number of finished tool entries to show individually
+/// during streaming before collapsing into a summary line.
+const TOOL_COLLAPSE_THRESHOLD: usize = 3;
+
+fn compose_display(tool_lines: &[ToolEntry], text: &str, streaming: bool) -> String {
     let mut out = String::new();
     if !tool_lines.is_empty() {
-        for entry in tool_lines {
-            out.push_str(&entry.render());
-            out.push('\n');
+        if streaming {
+            let done = tool_lines.iter().filter(|e| e.state == ToolState::Completed).count();
+            let failed = tool_lines.iter().filter(|e| e.state == ToolState::Failed).count();
+            let running: Vec<_> = tool_lines.iter().filter(|e| e.state == ToolState::Running).collect();
+            let finished = done + failed;
+
+            if finished <= TOOL_COLLAPSE_THRESHOLD {
+                for entry in tool_lines.iter().filter(|e| e.state != ToolState::Running) {
+                    out.push_str(&entry.render());
+                    out.push('\n');
+                }
+            } else {
+                let mut parts = Vec::new();
+                if done > 0 { parts.push(format!("✅ {done}")); }
+                if failed > 0 { parts.push(format!("❌ {failed}")); }
+                out.push_str(&format!("{} tool(s) completed\n", parts.join(" · ")));
+            }
+
+            if running.len() <= TOOL_COLLAPSE_THRESHOLD {
+                for entry in &running {
+                    out.push_str(&entry.render());
+                    out.push('\n');
+                }
+            } else {
+                let hidden = running.len() - TOOL_COLLAPSE_THRESHOLD;
+                out.push_str(&format!("🔧 {hidden} more running\n"));
+                for entry in running.iter().skip(hidden) {
+                    out.push_str(&entry.render());
+                    out.push('\n');
+                }
+            }
+        } else {
+            for entry in tool_lines {
+                out.push_str(&entry.render());
+                out.push('\n');
+            }
         }
-        out.push('\n');
+        if !out.is_empty() { out.push('\n'); }
     }
     out.push_str(text.trim_end());
     out
--- a/src/adapter.rs
+++ b/src/adapter.rs
@@ -1,0 +1,400 @@
+use anyhow::Result;
+use async_trait::async_trait;
+use serde::Serialize;
+use std::sync::Arc;
+use tokio::sync::watch;
+use tracing::error;
+
+use crate::acp::{classify_notification, AcpEvent, ContentBlock, SessionPool};
+use crate::config::ReactionsConfig;
+use crate::error_display::{format_coded_error, format_user_error};
+use crate::format;
+use crate::reactions::StatusReactionController;
+
+// --- Platform-agnostic types ---
+
+/// Identifies a channel or thread across platforms.
+#[derive(Clone, Debug, Hash, Eq, PartialEq)]
+pub struct ChannelRef {
+    pub platform: String,
+    pub channel_id: String,
+    /// Thread within a channel (e.g. Slack thread_ts, Telegram topic_id).
+    /// For Discord, threads are separate channels so this is None.
+    pub thread_id: Option<String>,
+    /// Parent channel if this is a thread-as-channel (Discord).
+    pub parent_id: Option<String>,
+}
+
+impl ChannelRef {
+    pub fn is_thread(&self) -> bool {
+        self.thread_id.is_some() || self.parent_id.is_some()
+    }
+}
+
+/// Identifies a message across platforms.
+#[derive(Clone, Debug)]
+pub struct MessageRef {
+    pub channel: ChannelRef,
+    pub message_id: String,
+}
+
+/// Sender identity injected into prompts for downstream agent context.
+#[derive(Clone, Debug, Serialize)]
+pub struct SenderContext {
+    pub schema: String,
+    pub sender_id: String,
+    pub sender_name: String,
+    pub display_name: String,
+    pub channel: String,
+    pub channel_id: String,
+    pub is_bot: bool,
+}
+
+// --- ChatAdapter trait ---
+
+#[async_trait]
+pub trait ChatAdapter: Send + Sync + 'static {
+    /// Platform name for logging and session key namespacing.
+    fn platform(&self) -> &'static str;
+
+    /// Maximum message length for this platform (e.g. 2000 for Discord, 4000 for Slack).
+    fn message_limit(&self) -> usize;
+
+    /// Send a new message, returns a reference to the sent message.
+    async fn send_message(&self, channel: &ChannelRef, content: &str) -> Result<MessageRef>;
+
+    /// Edit an existing message in-place.
+    async fn edit_message(&self, msg: &MessageRef, content: &str) -> Result<()>;
+
+    /// Create a thread from a trigger message, returns the thread channel ref.
+    async fn create_thread(
+        &self,
+        channel: &ChannelRef,
+        trigger_msg: &MessageRef,
+        title: &str,
+    ) -> Result<ChannelRef>;
+
+    /// Add a reaction/emoji to a message.
+    async fn add_reaction(&self, msg: &MessageRef, emoji: &str) -> Result<()>;
+
+    /// Remove a reaction/emoji from a message.
+    async fn remove_reaction(&self, msg: &MessageRef, emoji: &str) -> Result<()>;
+}
+
+// --- AdapterRouter ---
+
+/// Shared logic for routing messages to ACP agents, managing sessions,
+/// streaming edits, and controlling reactions. Platform-independent.
+pub struct AdapterRouter {
+    pool: Arc<SessionPool>,
+    reactions_config: ReactionsConfig,
+}
+
+impl AdapterRouter {
+    pub fn new(pool: Arc<SessionPool>, reactions_config: ReactionsConfig) -> Self {
+        Self {
+            pool,
+            reactions_config,
+        }
+    }
+
+    pub fn pool(&self) -> &SessionPool {
+        &self.pool
+    }
+
+    /// Handle an incoming user message. The adapter is responsible for
+    /// filtering, resolving the thread, and building the SenderContext.
+    /// This method handles session management and streaming.
+    pub async fn handle_message(
+        &self,
+        adapter: &Arc<dyn ChatAdapter>,
+        thread_channel: &ChannelRef,
+        _sender: &SenderContext,
+        content_blocks: Vec<ContentBlock>,
+        trigger_msg: &MessageRef,
+    ) -> Result<()> {
+        tracing::debug!(platform = adapter.platform(), "processing message");
+
+        let thinking_msg = adapter.send_message(thread_channel, "...").await?;
+
+        let thread_key = format!(
+            "{}:{}",
+            adapter.platform(),
+            thread_channel
+                .thread_id
+                .as_deref()
+                .unwrap_or(&thread_channel.channel_id)
+        );
+
+        if let Err(e) = self.pool.get_or_create(&thread_key).await {
+            let msg = format_user_error(&e.to_string());
+            let _ = adapter
+                .edit_message(&thinking_msg, &format!("⚠️ {msg}"))
+                .await;
+            error!("pool error: {e}");
+            return Err(e);
+        }
+
+        let reactions = Arc::new(StatusReactionController::new(
+            self.reactions_config.enabled,
+            adapter.clone(),
+            trigger_msg.clone(),
+            self.reactions_config.emojis.clone(),
+            self.reactions_config.timing.clone(),
+        ));
+        reactions.set_queued().await;
+
+        let result = self
+            .stream_prompt(
+                adapter,
+                &thread_key,
+                content_blocks,
+                thread_channel,
+                &thinking_msg,
+                reactions.clone(),
+            )
+            .await;
+
+        match &result {
+            Ok(()) => reactions.set_done().await,
+            Err(_) => reactions.set_error().await,
+        }
+
+        let hold_ms = if result.is_ok() {
+            self.reactions_config.timing.done_hold_ms
+        } else {
+            self.reactions_config.timing.error_hold_ms
+        };
+        if self.reactions_config.remove_after_reply {
+            let reactions = reactions;
+            tokio::spawn(async move {
+                tokio::time::sleep(std::time::Duration::from_millis(hold_ms)).await;
+                reactions.clear().await;
+            });
+        }
+
+        if let Err(ref e) = result {
+            let _ = adapter
+                .edit_message(&thinking_msg, &format!("⚠️ {e}"))
+                .await;
+        }
+
+        result
+    }
+
+    async fn stream_prompt(
+        &self,
+        adapter: &Arc<dyn ChatAdapter>,
+        thread_key: &str,
+        content_blocks: Vec<ContentBlock>,
+        thread_channel: &ChannelRef,
+        thinking_msg: &MessageRef,
+        reactions: Arc<StatusReactionController>,
+    ) -> Result<()> {
+        let adapter = adapter.clone();
+        let thread_channel = thread_channel.clone();
+        let msg_ref = thinking_msg.clone();
+        let message_limit = adapter.message_limit();
+
+        self.pool
+            .with_connection(thread_key, |conn| {
+                let content_blocks = content_blocks.clone();
+                Box::pin(async move {
+                    let reset = conn.session_reset;
+                    conn.session_reset = false;
+
+                    let (mut rx, _) = conn.session_prompt(content_blocks).await?;
+                    reactions.set_thinking().await;
+
+                    let initial = if reset {
+                        "⚠️ _Session expired, starting fresh..._\n\n...".to_string()
+                    } else {
+                        "...".to_string()
+                    };
+                    let (buf_tx, buf_rx) = watch::channel(initial);
+
+                    let mut text_buf = String::new();
+                    let mut tool_lines: Vec<ToolEntry> = Vec::new();
+
+                    if reset {
+                        text_buf.push_str("⚠️ _Session expired, starting fresh..._\n\n");
+                    }
+
+                    // Spawn edit-streaming task — only edits the single message, never sends new ones.
+                    // Long content is truncated during streaming; final multi-message split happens after.
+                    let streaming_limit = message_limit.saturating_sub(100);
+                    let edit_handle = {
+                        let adapter = adapter.clone();
+                        let msg_ref = msg_ref.clone();
+                        let mut buf_rx = buf_rx.clone();
+                        tokio::spawn(async move {
+                            let mut last_content = String::new();
+                            loop {
+                                tokio::time::sleep(std::time::Duration::from_millis(1500)).await;
+                                if buf_rx.has_changed().unwrap_or(false) {
+                                    let content = buf_rx.borrow_and_update().clone();
+                                    if content != last_content {
+                                        let display = if content.chars().count() > streaming_limit {
+                                            let truncated = format::truncate_chars(&content, streaming_limit);
+                                            format!("{truncated}…")
+                                        } else {
+                                            content.clone()
+                                        };
+                                        let _ = adapter.edit_message(&msg_ref, &display).await;
+                                        last_content = content;
+                                    }
+                                }
+                                if buf_rx.has_changed().is_err() {
+                                    break;
+                                }
+                            }
+                        })
+                    };
+
+                    // Process ACP notifications
+                    let mut got_first_text = false;
+                    let mut response_error: Option<String> = None;
+                    while let Some(notification) = rx.recv().await {
+                        if notification.id.is_some() {
+                            if let Some(ref err) = notification.error {
+                                response_error = Some(format_coded_error(err.code, &err.message));
+                            }
+                            break;
+                        }
+
+                        if let Some(event) = classify_notification(&notification) {
+                            match event {
+                                AcpEvent::Text(t) => {
+                                    if !got_first_text {
+                                        got_first_text = true;
+                                    }
+                                    text_buf.push_str(&t);
+                                    let _ =
+                                        buf_tx.send(compose_display(&tool_lines, &text_buf));
+                                }
+                                AcpEvent::Thinking => {
+                                    reactions.set_thinking().await;
+                                }
+                                AcpEvent::ToolStart { id, title } if !title.is_empty() => {
+                                    reactions.set_tool(&title).await;
+                                    let title = sanitize_title(&title);
+                                    if let Some(slot) = tool_lines.iter_mut().find(|e| e.id == id) {
+                                        slot.title = title;
+                                        slot.state = ToolState::Running;
+                                    } else {
+                                        tool_lines.push(ToolEntry {
+                                            id,
+                                            title,
+                                            state: ToolState::Running,
+                                        });
+                                    }
+                                    let _ =
+                                        buf_tx.send(compose_display(&tool_lines, &text_buf));
+                                }
+                                AcpEvent::ToolDone { id, title, status } => {
+                                    reactions.set_thinking().await;
+                                    let new_state = if status == "completed" {
+                                        ToolState::Completed
+                                    } else {
+                                        ToolState::Failed
+                                    };
+                                    if let Some(slot) = tool_lines.iter_mut().find(|e| e.id == id) {
+                                        if !title.is_empty() {
+                                            slot.title = sanitize_title(&title);
+                                        }
+                                        slot.state = new_state;
+                                    } else if !title.is_empty() {
+                                        tool_lines.push(ToolEntry {
+                                            id,
+                                            title: sanitize_title(&title),
+                                            state: new_state,
+                                        });
+                                    }
+                                    let _ =
+                                        buf_tx.send(compose_display(&tool_lines, &text_buf));
+                                }
+                                _ => {}
+                            }
+                        }
+                    }
+
+                    conn.prompt_done().await;
+                    drop(buf_tx);
+                    let _ = edit_handle.await;
+
+                    // Final edit with complete content
+                    let final_content = compose_display(&tool_lines, &text_buf);
+                    let final_content = if final_content.is_empty() {
+                        if let Some(err) = response_error {
+                            format!("⚠️ {err}")
+                        } else {
+                            "_(no response)_".to_string()
+                        }
+                    } else if let Some(err) = response_error {
+                        format!("⚠️ {err}\n\n{final_content}")
+                    } else {
+                        final_content
+                    };
+
+                    let chunks = format::split_message(&final_content, message_limit);
+                    let mut current_msg = msg_ref;
+                    for (i, chunk) in chunks.iter().enumerate() {
+                        if i == 0 {
+                            let _ = adapter.edit_message(&current_msg, chunk).await;
+                        } else if let Ok(new_msg) =
+                            adapter.send_message(&thread_channel, chunk).await
+                        {
+                            current_msg = new_msg;
+                        }
+                    }
+
+                    Ok(())
+                })
+            })
+            .await
+    }
+}
+
+/// Flatten a tool-call title into a single line safe for inline-code spans.
+fn sanitize_title(title: &str) -> String {
+    title.replace('\r', "").replace('\n', " ; ").replace('`', "'")
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+enum ToolState {
+    Running,
+    Completed,
+    Failed,
+}
+
+#[derive(Debug, Clone)]
+struct ToolEntry {
+    id: String,
+    title: String,
+    state: ToolState,
+}
+
+impl ToolEntry {
+    fn render(&self) -> String {
+        let icon = match self.state {
+            ToolState::Running => "🔧",
+            ToolState::Completed => "✅",
+            ToolState::Failed => "❌",
+        };
+        let suffix = if self.state == ToolState::Running { "..." } else { "" };
+        format!("{icon} `{}`{}", self.title, suffix)
+    }
+}
+
+fn compose_display(tool_lines: &[ToolEntry], text: &str) -> String {
+    let mut out = String::new();
+    if !tool_lines.is_empty() {
+        for entry in tool_lines {
+            out.push_str(&entry.render());
+            out.push('\n');
+        }
+        out.push('\n');
+    }
+    out.push_str(text.trim_end());
+    out
+}
--- a/src/slack.rs
+++ b/src/slack.rs
@@ -205,7 +205,7 @@ impl ChatAdapter for SlackAdapter {
 
     async fn add_reaction(&self, msg: &MessageRef, emoji: &str) -> Result<()> {
         let name = unicode_to_slack_emoji(emoji);
-        self.api_post(
+        match self.api_post(
             "reactions.add",
             serde_json::json!({
                 "channel": msg.channel.channel_id,
@@ -213,13 +213,17 @@ impl ChatAdapter for SlackAdapter {
                 "name": name,
             }),
         )
-        .await?;
-        Ok(())
+        .await
+        {
+            Ok(_) => Ok(()),
+            Err(e) if e.to_string().contains("already_reacted") => Ok(()),
+            Err(e) => Err(e),
+        }
     }
 
     async fn remove_reaction(&self, msg: &MessageRef, emoji: &str) -> Result<()> {
         let name = unicode_to_slack_emoji(emoji);
-        self.api_post(
+        match self.api_post(
             "reactions.remove",
             serde_json::json!({
                 "channel": msg.channel.channel_id,
@@ -227,8 +231,12 @@ impl ChatAdapter for SlackAdapter {
                 "name": name,
             }),
         )
-        .await?;
-        Ok(())
+        .await
+        {
+            Ok(_) => Ok(()),
+            Err(e) if e.to_string().contains("no_reaction") => Ok(()),
+            Err(e) => Err(e),
+        }
     }
 }
 
@@ -478,6 +486,13 @@ async fn handle_message(
     let prompt = if is_mention {
         strip_slack_mention(&text)
     } else {
+        // Thread follow-up: check for bot loop before processing
+        if let Some(ref tts) = thread_ts {
+            if is_bot_loop(adapter, &channel_id, tts).await {
+                tracing::warn!(channel_id, thread_ts = tts, "bot loop detected, ignoring");
+                return;
+            }
+        }
         text.trim().to_string()
     };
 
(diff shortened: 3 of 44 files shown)


chaodu-agent:
## PR Review: feat: add Slack adapter with multi-platform ChatAdapter architecture

**Author:** Tony Lee (`dogzzdogzz`) | +1820 / -761 | 20 files changed

---

OpenAB previously only supported Discord. Many engineering teams use Slack as their primary communication tool, so not supporting Slack blocks a large segment of enterprise users. This PR does two things:

- Abstracts the Discord-hardcoded logic into a `ChatAdapter` trait, making the platform layer pluggable
- Implements a Slack adapter (Socket Mode) so Slack users can @mention the bot, open threads, and interact with ACP agents

Closes #93 (Slack adapter proposal) and implements Phase 1–3 of RFC #94 (Multi-Platform Adapter Architecture).

### 2. How does it solve it?

Clean architectural split:

- **`adapter.rs`** — Defines `ChatAdapter` trait (send/edit/thread/react) + `AdapterRouter` (shared session routing, streaming, reactions logic)
- **`discord.rs`** — `DiscordAdapter` implements `ChatAdapter`; the `Handler` becomes a thin serenity `EventHandler` that delegates to `AdapterRouter`
- **`slack.rs`** — `SlackAdapter` implements `ChatAdapter` using `tokio-tungstenite` for Socket Mode WebSocket directly — no third-party Slack SDK dependency
- **`media.rs`** — Image resize/compress and STT download extracted from `discord.rs`, with an `auth_token` parameter for Slack private file downloads
- **`reactions.rs`** — Decoupled from serenity, now uses `Arc<dyn ChatAdapter>`
- **`main.rs`** — Slack runs as a background tokio task, Discord runs foreground (blocking); both can run simultaneously
- Session keys namespaced by platform (`discord:xxx`, `slack:xxx`) to prevent collisions
- Slack emoji mapping via `unicode_to_slack_emoji()` (Unicode → short name)
- Helm chart is backward compatible; both `[discord]` and `[slack]` are now optional

### 3. Were alternatives considered?

From the PR description and RFC #94:

- **Events API + HTTP** vs **Socket Mode** — Socket Mode was chosen because it requires no public URL and mirrors Discord's Gateway WebSocket model, which is more K8S-friendly
- **`slack-morphism` crate** vs raw implementation — Issue #93 originally proposed `slack-morphism`, but the PR uses `tokio-tungstenite` + raw Slack API directly. This reduces dependencies at the cost of maintaining WebSocket reconnection and API calls manually
- **Trait design** — The RFC proposed a `ChatAdapter` with a `start()` method; the final implementation dropped it, letting each adapter manage its own lifecycle (Discord via serenity client, Slack via `run_slack_adapter()` function), which is more flexible

Overall architecture is solid, the refactor is clean, and tests were correctly relocated to `media.rs`.

---

🟢 **INFO**
- `ChatAdapter` trait design is concise; `ChannelRef` / `MessageRef` abstractions are generic enough that adding Telegram later should be straightforward
- Centralizing streaming/session/reaction logic in `AdapterRouter` is the right call — avoids duplicating a large block of code per adapter
- Helm chart backward compatibility is well handled; `discord.enabled` defaults to `true` so existing deployments won't break
- Tests moved from `discord.rs` to `media.rs` without loss; 26 tests pass
- `docs/slack-bot-howto.md` is thorough, and the troubleshooting section is practical

🟡 **NIT**
- `SlackAdapter` calls `users.info` API on every `handle_message` to resolve display names — under high traffic this could hit Slack rate limits. An LRU cache with TTL would be more robust
- `unicode_to_slack_emoji()` is a hardcoded mapping; if `ReactionEmojis` config is customized with emojis not in the map, it silently falls back to `grey_question`. It might be worth validating at config load time or documenting this limitation
- `get_socket_mode_url()` creates a new `reqwest::Client` on every reconnect — could reuse `media::HTTP_CLIENT` or the adapter's own client
- `DiscordAdapter` is instantiated via `Arc::new(DiscordAdapter::new(ctx.http.clone()))` on every `message()` event. It might be worth initializing once in `ready()` and storing in a `OnceCell<Arc<dyn ChatAdapter>>`

🔴 **SUGGESTED CHANGES**
- **Potential duplicate processing of Slack thread messages**: When an `app_mention` event occurs inside a thread, Slack also fires a `message` event (since `has_thread` is true). There doesn't appear to be a dedup mechanism, so the same message could be processed twice by `handle_message`. It might be worth considering a seen-message dedup (e.g. `HashSet<String>` keyed by `ts` with TTL), or checking in the `message` handler whether the text contains a bot mention and skipping it
- **Slack adapter lacks graceful shutdown**: Discord has `shard_manager.shutdown_all()`, but `run_slack_adapter` is an infinite loop that can only be killed via `handle.abort()`. The WebSocket connection won't be gracefully closed. Suggestion for maintainers: pass in a `CancellationToken` or `watch::Receiver` so the loop can exit cleanly
- **`discord.enabled: true` with empty `botToken` in `values.yaml`**: The configmap template guards with `if and ($cfg.discord).enabled ($cfg.discord).botToken`, so an empty token effectively disables it — which is correct. However, it might be worth explicitly documenting that `enabled: true` without a token is treated as disabled, to avoid user confusion

---

Overall this is a high-quality PR. The architecture aligns with the RFC, the refactor introduces no regressions, and Helm backward compatibility is maintained. The main areas to watch are Slack-side event dedup and graceful shutdown.

*Review by 超渡法師 🙏*

dogzzdogzz:
@chaodu-agent Thanks for the thorough review! Just pushed a commit addressing all feedback:

**🔴 Suggested Changes:**

- **Duplicate processing of Slack thread messages**: Fixed. The `message` handler now checks if the message @mentions the bot (using bot user ID from `auth.test` API) and skips it — `app_mention` handles those. This is precise: mentioning *other* users in thread replies won't be skipped.

- **Slack adapter graceful shutdown**: Fixed. `run_slack_adapter` now accepts a `watch::Receiver<bool>` shutdown signal. The WebSocket loop uses `tokio::select!` to listen for both messages and shutdown simultaneously. On shutdown, it sends a WebSocket `Close` frame before exiting. `main.rs` signals via `watch::channel` and waits up to 5 seconds for clean exit.

- **`discord.enabled: true` + empty `botToken`**: Documented in `values.yaml` — "enabled with empty botToken is treated as disabled".

**🟡 NITs:**

- **`users.info` LRU cache**: Added a `HashMap` cache with 5-minute TTL on `SlackAdapter`. Cache hit returns immediately, miss calls API then caches. Avoids hitting Slack rate limits under traffic.

- **`unicode_to_slack_emoji` fallback**: Documented — comment notes it only covers the default emoji set and custom emoji falls back to `grey_question`.

- **`DiscordAdapter` created per message**: Fixed. Now uses `OnceLock` on `Handler` — initialized on first message, reused for all subsequent events. Verified that serenity's `Arc<Http>` is stable across reconnects (created once at `Client` construction, never recreated).

- **`get_socket_mode_url` creates new client**: Skipped — only runs on reconnect (rare), negligible cost.

Commit: b04d9e2

antigenius0910:
Thanks for the detailed explanation @dogzzdogzz! The overall architecture is really clean. the ChatAdapter trait, centralising streaming/session/reaction logic in AdapterRouter, and the Socket Mode choice all make a lot of sense. We learned a lot reading through it.

One thing we wanted to double-check with you. regarding "prompt streaming doesn't hold the pool lock". We wrote a small timing test to verify:

```
// Reproduces the with_connection locking pattern
async fn with_lock(...) {
    let mut guard = map.write().await;    // write lock acquired
    let val = guard.get_mut(key).unwrap();
    f(val).await                          // lock held until f() completes
}
```
```
// Two tasks with different keys (discord: / slack:), each sleeping 1s, started concurrently
[BEFORE] Elapsed: 2.01s  ← serial
```

The conns WriteGuard inside with_connection() can't be dropped while conn borrows from it, so the write lock is held for the full duration of f(conn).await — which includes the entire streaming loop.

We tried a small fix: `changing RwLock<HashMap<String, AcpConnection>> to RwLock<HashMap<String, Arc<Mutex<AcpConnection>>>>` so the map lock is released right after the Arc::clone(), and only the per-session Mutex is held during streaming:

```
[AFTER]  Elapsed: 1.00s  ← concurrent
```
You can reproduce both results with `cargo test lock_contention -- --nocapture`.

This might be out of scope for this PR and if there's something we've missed or you see it differently, we'd love to hear it. Just wanted to make sure this was on the radar! 🙏

dogzzdogzz:
@antigenius0910 Great catch — you're absolutely right. `with_connection()` holds the pool write lock for the entire duration of the streaming closure, which serializes all concurrent sessions. Your timing test clearly demonstrates the issue.

This is a pre-existing upstream design in `pool.rs` — our PR doesn't modify `with_connection()`, we only call it. But multi-platform support makes it more visible since Discord and Slack now share the same `SessionPool`.

The `Arc<Mutex<AcpConnection>>` approach you tested is the correct fix, and aligns with what @ruan330 already validated in production (see #78 comment). Since upstream just refactored `pool.rs` in v0.7.3 (added `PoolState`, suspend/resume, LRU eviction), I'd suggest we keep this PR focused on the adapter layer and follow up on the per-connection lock change in #78 after this merges.

Your test case would be a great addition to that discussion — concrete proof of the lock contention and the fix. Would you be up for posting it there?

thepagent:
CI is failing.

Cargo.toml
```diff
@@ -14,8 +14,11 @@ serenity = { version = "0.12", default-features = false, features = ["client", "
...
 regex = "1"
 anyhow = "1"
+async-trait = "0.1"
+futures-util = "0.3"
 rand = "0.8"
 reqwest = { version = "0.12", default-features = false, features = ["rustls-tls", "multipart", "json"] }
 base64 = "0.22"
 image = { version = "0.25", default-features = false, features = ["jpeg", "png", "gif", "webp"] }
 libc = "0.2"
+tokio-tungstenite = { version = "0.21", features = ["rustls-tls-webpki-roots"] }
```
chaodu-agent, line 27: Add `lru` crate for bounded user cache:

```suggestion
libc = "0.2"
lru = "0.12"
tokio-tungstenite = { version = "0.21", features = ["rustls-tls-webpki-roots"] }
```

src/slack.rs
```diff
@@ -0,0 +1,566 @@
+use crate::acp::ContentBlock;
+use crate::adapter::{AdapterRouter, ChatAdapter, ChannelRef, MessageRef, SenderContext};
+use crate::config::SttConfig;
+use crate::media;
+use anyhow::{anyhow, Result};
+use async_trait::async_trait;
+use futures_util::{SinkExt, StreamExt};
+use std::collections::{HashMap, HashSet};
+use std::sync::{Arc, LazyLock};
```
chaodu-agent, line 9: Import `NonZeroUsize` for LRU capacity, drop `HashMap`:

```suggestion
use std::collections::HashSet;
use std::num::NonZeroUsize;
```

src/slack.rs
```diff
@@ -0,0 +1,566 @@
...
+        "💪" => "muscle",
+        "🦾" => "mechanical_arm",
+        "🥱" => "yawning_face",
+        "😨" => "fearful",
+        "✅" => "white_check_mark",
+        "❌" => "x",
+        "🔧" => "wrench",
+        _ => "grey_question",
+    }
+}
```
chaodu-agent, line 44: Return `Option` so unmapped emoji can be detected at startup. Add `validate_emoji_mapping()` to warn operators about missing mappings:

```suggestion
/// Map Unicode emoji to Slack short names for reactions API.
/// Returns `None` for unmapped emoji.
fn unicode_to_slack_emoji(unicode: &str) -> Option<&'static str> {
    match unicode {
        "👀" => Some("eyes"),
        "🤔" => Some("thinking_face"),
        "🔥" => Some("fire"),
        "👨\u200d💻" => Some("technologist"),
        "⚡" => Some("zap"),
        "🆗" => Some("ok"),
        "😱" => Some("scream"),
        "🚫" => Some("no_entry_sign"),
        "😊" => Some("blush"),
        "😎" => Some("sunglasses"),
        "🫡" => Some("saluting_face"),
        "🤓" => Some("nerd_face"),
        "😏" => Some("smirk"),
        "✌\ufe0f" => Some("v"),
        "💪" => Some("muscle"),
        "🦾" => Some("mechanical_arm"),
        "🥱" => Some("yawning_face"),
        "😨" => Some("fearful"),
        "✅" => Some("white_check_mark"),
        "❌" => Some("x"),
        "🔧" => Some("wrench"),
        _ => None,
    }
}

/// Validate that all configured reaction emojis have Slack mappings.
/// Logs a warning at startup for any unmapped emoji.
pub fn validate_emoji_mapping(emojis: &crate::config::ReactionEmojis) {
    for emoji in emojis.all_emojis() {
        if unicode_to_slack_emoji(emoji).is_none() {
            warn!(
                emoji,
                "reaction emoji has no Slack mapping — will show as :grey_question:"
            );
        }
    }
}
```

src/slack.rs
```diff
@@ -0,0 +1,566 @@
...
+
+/// TTL for cached user display names (5 minutes).
+const USER_CACHE_TTL: std::time::Duration = std::time::Duration::from_secs(300);
+
+pub struct SlackAdapter {
+    client: reqwest::Client,
+    bot_token: String,
+    bot_user_id: tokio::sync::OnceCell<String>,
+    user_cache: tokio::sync::Mutex<HashMap<String, (String, tokio::time::Instant)>>,
+}
```
chaodu-agent, line 56: Replace unbounded `HashMap` with `lru::LruCache` capped at 1000 entries to prevent slow memory growth in large workspaces:

```suggestion
/// TTL for cached user display names (5 minutes).
const USER_CACHE_TTL: std::time::Duration = std::time::Duration::from_secs(300);

/// Maximum number of cached user display names.
const USER_CACHE_CAPACITY: NonZeroUsize = unsafe { NonZeroUsize::new_unchecked(1000) };

pub struct SlackAdapter {
    client: reqwest::Client,
    bot_token: String,
    bot_user_id: tokio::sync::OnceCell<String>,
    user_cache: tokio::sync::Mutex<lru::LruCache<String, (String, tokio::time::Instant)>>,
}
```

src/slack.rs
```diff
@@ -0,0 +1,566 @@
...
+
+impl SlackAdapter {
+    pub fn new(bot_token: String) -> Self {
+        Self {
+            client: reqwest::Client::new(),
+            bot_token,
+            bot_user_id: tokio::sync::OnceCell::new(),
+            user_cache: tokio::sync::Mutex::new(HashMap::new()),
```
chaodu-agent, line 64: Init with bounded LRU:

```suggestion
            user_cache: tokio::sync::Mutex::new(lru::LruCache::new(USER_CACHE_CAPACITY)),
```

src/slack.rs
```diff
@@ -0,0 +1,566 @@
...
+        }
+        Ok(json)
+    }
+
+    /// Resolve a Slack user ID to display name via users.info API.
+    /// Results are cached for 5 minutes to avoid hitting Slack rate limits.
+    async fn resolve_user_name(&self, user_id: &str) -> Option<String> {
+        // Check cache first
+        {
+            let cache = self.user_cache.lock().await;
```
chaodu-agent, line 103: LRU `.get()` requires `&mut self` (it promotes the entry). Change to `let mut cache`:

```suggestion
            let mut cache = self.user_cache.lock().await;
```

src/slack.rs
```diff
@@ -0,0 +1,566 @@
...
+            .get("real_name")
+            .and_then(|v| v.as_str())
+            .filter(|s| !s.is_empty());
+        let name = user
+            .get("name")
+            .and_then(|v| v.as_str());
+        let resolved = display.or(real).or(name)?.to_string();
+
+        // Cache the result
+        self.user_cache.lock().await.insert(
```
chaodu-agent, line 134: LRU uses `.put()` instead of `.insert()`:

```suggestion
        self.user_cache.lock().await.put(
```

src/slack.rs
```diff
@@ -0,0 +1,566 @@
...
+        Ok(ChannelRef {
+            platform: "slack".into(),
+            channel_id: channel.channel_id.clone(),
+            thread_id: Some(trigger_msg.message_id.clone()),
+            parent_id: None,
+        })
+    }
+
+    async fn add_reaction(&self, msg: &MessageRef, emoji: &str) -> Result<()> {
+        let name = unicode_to_slack_emoji(emoji);
```
chaodu-agent, line 207: Handle `Option` return from `unicode_to_slack_emoji`:

```suggestion
        let name = unicode_to_slack_emoji(emoji).unwrap_or("grey_question");
```

src/slack.rs
```diff
@@ -0,0 +1,566 @@
...
+                "timestamp": msg.message_id,
+                "name": name,
+            }),
+        )
+        .await?;
+        Ok(())
+    }
+
+    async fn remove_reaction(&self, msg: &MessageRef, emoji: &str) -> Result<()> {
+        let name = unicode_to_slack_emoji(emoji);
```
chaodu-agent, line 225: Same here:

```suggestion
        let name = unicode_to_slack_emoji(emoji).unwrap_or("grey_question");
```

chaodu-agent (commented):
## NIT Fixes — Inline Suggested Changes

Here are the concrete fixes for the NITs identified in the review. Each suggestion can be applied individually via GitHub's "Apply suggestion" button.

**Summary:**
1. `Cargo.toml` — Add `lru` crate
2. `src/config.rs` — Add `ReactionEmojis::all_emojis()` for validation (posted as regular comment — outside diff hunk)
3. `src/slack.rs` — `unicode_to_slack_emoji()` returns `Option`, add `validate_emoji_mapping()` startup check
4. `src/slack.rs` — Replace unbounded `HashMap` user cache with `lru::LruCache` (cap 1000)
5. `src/slack.rs` — Reuse `media::HTTP_CLIENT` in `get_socket_mode_url()`
6. `src/adapter.rs` — `#[must_use]` on `ChatAdapter` trait methods, improved doc comments, `reactions_emojis()` accessor

---

### config.rs — outside diff hunk

Add after `impl Default for ReactionTiming` block (after line 182):

```rust
impl ReactionEmojis {
    /// Return all configured emoji values for validation purposes.
    pub fn all_emojis(&self) -> Vec<&str> {
        vec![
            &self.queued, &self.thinking, &self.tool, &self.coding,
            &self.web, &self.done, &self.error,
        ]
    }
}
```

*Suggestions by 超渡法師 🙏*

src/slack.rs
```diff
@@ -0,0 +1,566 @@
...
+pub async fn run_slack_adapter(
+    bot_token: String,
+    app_token: String,
+    allowed_channels: HashSet<String>,
+    allowed_users: HashSet<String>,
+    stt_config: SttConfig,
+    router: Arc<AdapterRouter>,
+    mut shutdown_rx: watch::Receiver<bool>,
+) -> Result<()> {
+    let adapter = Arc::new(SlackAdapter::new(bot_token.clone()));
```
chaodu-agent, line 256: Call emoji validation at startup:

```suggestion
    let adapter = Arc::new(SlackAdapter::new(bot_token.clone()));
    validate_emoji_mapping(&router.reactions_emojis());
```

src/slack.rs
```diff
@@ -0,0 +1,566 @@
...
+
+        warn!("reconnecting to Slack Socket Mode in 5s...");
+        tokio::time::sleep(std::time::Duration::from_secs(5)).await;
+    }
+}
+
+/// Call apps.connections.open to get a WebSocket URL for Socket Mode.
+async fn get_socket_mode_url(app_token: &str) -> Result<String> {
+    let client = reqwest::Client::new();
+    let resp = client
```
chaodu-agent, line 418: Reuse the shared static HTTP client instead of creating a new one per reconnect:

```suggestion
    let resp = media::HTTP_CLIENT
```

src/adapter.rs
```diff
@@ -0,0 +1,415 @@
+use anyhow::Result;
+use async_trait::async_trait;
+use serde::Serialize;
+use std::sync::Arc;
+use tokio::sync::watch;
+use tracing::error;
+
+use crate::acp::{classify_notification, AcpEvent, ContentBlock, SessionPool};
+use crate::config::ReactionsConfig;
```
chaodu-agent, line 9: Import `ReactionEmojis` for the new accessor:

```suggestion
use crate::config::ReactionsConfig;
use crate::config::ReactionEmojis;
```

src/adapter.rs
```diff
@@ -0,0 +1,415 @@
...
+use crate::error_display::{format_coded_error, format_user_error};
+use crate::format;
+use crate::reactions::StatusReactionController;
+
+// --- Platform-agnostic types ---
+
+/// Identifies a channel or thread across platforms.
+#[derive(Clone, Debug, Hash, Eq, PartialEq)]
```
chaodu-agent, line 17: Expanded doc comment for `ChannelRef`:

```suggestion
/// Identifies a channel or thread across chat platforms.
///
/// Each platform maps differently: Discord threads are separate channels,
/// Slack threads use `thread_ts` within a channel, etc. This struct
/// normalizes those differences into a single reference type.
#[derive(Clone, Debug, Hash, Eq, PartialEq)]
```

src/adapter.rs
```diff
@@ -0,0 +1,415 @@
...
+    pub channel_id: String,
+    /// Thread within a channel (e.g. Slack thread_ts, Telegram topic_id).
+    /// For Discord, threads are separate channels so this is None.
+    pub thread_id: Option<String>,
+    /// Parent channel if this is a thread-as-channel (Discord).
+    pub parent_id: Option<String>,
+}
+
+/// Identifies a message across platforms.
+#[derive(Clone, Debug)]
```
chaodu-agent, line 29: Expanded doc comment for `MessageRef`:

```suggestion
/// Identifies a specific message across chat platforms.
///
/// Used for editing messages, adding reactions, and referencing trigger messages.
#[derive(Clone, Debug)]
```

src/adapter.rs
```diff
@@ -0,0 +1,415 @@
...
+pub struct MessageRef {
+    pub channel: ChannelRef,
+    pub message_id: String,
+}
+
+/// Sender identity injected into prompts for downstream agent context.
+#[derive(Clone, Debug, Serialize)]
```
chaodu-agent, line 36: Expanded doc comment for `SenderContext`:

```suggestion
/// Sender identity injected into prompts so downstream ACP agents
/// know who sent the message and from which platform/channel.
#[derive(Clone, Debug, Serialize)]
```

src/adapter.rs
```diff
@@ -0,0 +1,415 @@
...
+#[async_trait]
+pub trait ChatAdapter: Send + Sync + 'static {
+    /// Platform name for logging and session key namespacing.
+    fn platform(&self) -> &'static str;
+
+    /// Maximum message length for this platform (e.g. 2000 for Discord, 4000 for Slack).
+    fn message_limit(&self) -> usize;
+
+    /// Send a new message, returns a reference to the sent message.
+    async fn send_message(&self, channel: &ChannelRef, content: &str) -> Result<MessageRef>;
```
chaodu-agent, line 58: `#[must_use]` catches accidentally dropped `Result`s at compile time:

```suggestion
    /// Send a new message, returns a reference to the sent message.
    #[must_use = "send_message returns a Result that should be checked"]
    async fn send_message(&self, channel: &ChannelRef, content: &str) -> Result<MessageRef>;
```

src/adapter.rs
```diff
@@ -0,0 +1,415 @@
...
+
+    /// Edit an existing message in-place.
+    async fn edit_message(&self, msg: &MessageRef, content: &str) -> Result<()>;
```
chaodu-agent, line 61:
```suggestion
    /// Edit an existing message in-place.
    #[must_use = "edit_message returns a Result that should be checked"]
    async fn edit_message(&self, msg: &MessageRef, content: &str) -> Result<()>;
```

src/adapter.rs
```diff
@@ -0,0 +1,415 @@
...
+
+    /// Create a thread from a trigger message, returns the thread channel ref.
```
chaodu-agent, line 63:
```suggestion
    /// Create a thread from a trigger message, returns the thread channel ref.
    #[must_use = "create_thread returns a Result that should be checked"]
```

src/adapter.rs
```diff
@@ -0,0 +1,415 @@
...
+    async fn create_thread(
+        &self,
+        channel: &ChannelRef,
+        trigger_msg: &MessageRef,
+        title: &str,
+    ) -> Result<ChannelRef>;
+
+    /// Add a reaction/emoji to a message.
+    async fn add_reaction(&self, msg: &MessageRef, emoji: &str) -> Result<()>;
```
chaodu-agent, line 72:
```suggestion
    /// Add a reaction/emoji to a message.
    #[must_use = "add_reaction returns a Result that should be checked"]
    async fn add_reaction(&self, msg: &MessageRef, emoji: &str) -> Result<()>;
```

src/adapter.rs
```diff
@@ -0,0 +1,415 @@
...
+
+    /// Remove a reaction/emoji from a message.
+    async fn remove_reaction(&self, msg: &MessageRef, emoji: &str) -> Result<()>;
```
chaodu-agent, line 75:
```suggestion
    /// Remove a reaction/emoji from a message.
    #[must_use = "remove_reaction returns a Result that should be checked"]
    async fn remove_reaction(&self, msg: &MessageRef, emoji: &str) -> Result<()>;
```

src/adapter.rs
```diff
@@ -0,0 +1,415 @@
...
+}
+
+// --- AdapterRouter ---
+
+/// Shared logic for routing messages to ACP agents, managing sessions,
+/// streaming edits, and controlling reactions. Platform-independent.
+pub struct AdapterRouter {
```
chaodu-agent, line 82: Expanded doc comment for `AdapterRouter`:

```suggestion
/// Shared logic for routing messages to ACP agents, managing sessions,
/// streaming edits, and controlling reactions.
///
/// Platform adapters call [`AdapterRouter::handle_message`] after resolving
/// platform-specific details (threads, sender names, file attachments).
/// The router owns the [`SessionPool`] and [`ReactionsConfig`], keeping
/// all session and streaming logic in one place rather than duplicated
/// per adapter.
pub struct AdapterRouter {
```

src/adapter.rs
```diff
@@ -0,0 +1,415 @@
...
+    reactions_config: ReactionsConfig,
+}
+
+impl AdapterRouter {
+    pub fn new(pool: Arc<SessionPool>, reactions_config: ReactionsConfig) -> Self {
+        Self {
+            pool,
+            reactions_config,
+        }
+    }
```
chaodu-agent, line 93: Add accessor for emoji validation at startup:

```suggestion
        }
    }

    /// Return the configured reaction emojis (for startup validation).
    pub fn reactions_emojis(&self) -> &ReactionEmojis {
        &self.reactions_config.emojis
    }
```

Cargo.toml
chaodu-agent, line 27: Add `lru` crate for bounded user cache:

```suggestion
libc = "0.2"
lru = "0.12"
tokio-tungstenite = { version = "0.21", features = ["rustls-tls-webpki-roots"] }
```

src/slack.rs
chaodu-agent, line 9: Import `NonZeroUsize` for LRU capacity, drop `HashMap`:

```suggestion
use std::collections::HashSet;
use std::num::NonZeroUsize;
```

src/slack.rs
chaodu-agent, line 44: Return `Option` so unmapped emoji can be detected at startup. Add `validate_emoji_mapping()` to warn operators about missing mappings:

```suggestion
/// Map Unicode emoji to Slack short names for reactions API.
/// Returns `None` for unmapped emoji.
fn unicode_to_slack_emoji(unicode: &str) -> Option<&'static str> {
    match unicode {
        "👀" => Some("eyes"),
        "🤔" => Some("thinking_face"),
        "🔥" => Some("fire"),
        "👨\u200d💻" => Some("technologist"),
        "⚡" => Some("zap"),
        "🆗" => Some("ok"),
        "😱" => Some("scream"),
        "🚫" => Some("no_entry_sign"),
        "😊" => Some("blush"),
        "😎" => Some("sunglasses"),
        "🫡" => Some("saluting_face"),
        "🤓" => Some("nerd_face"),
        "😏" => Some("smirk"),
        "✌\ufe0f" => Some("v"),
        "💪" => Some("muscle"),
        "🦾" => Some("mechanical_arm"),
        "🥱" => Some("yawning_face"),
        "😨" => Some("fearful"),
        "✅" => Some("white_check_mark"),
        "❌" => Some("x"),
        "🔧" => Some("wrench"),
        _ => None,
    }
}

/// Validate that all configured reaction emojis have Slack mappings.
/// Logs a warning at startup for any unmapped emoji.
pub fn validate_emoji_mapping(emojis: &crate::config::ReactionEmojis) {
    for emoji in emojis.all_emojis() {
        if unicode_to_slack_emoji(emoji).is_none() {
            warn!(
                emoji,
                "reaction emoji has no Slack mapping — will show as :grey_question:"
            );
        }
    }
}
```

src/slack.rs
chaodu-agent, line 56: Replace unbounded `HashMap` with `lru::LruCache` capped at 1000 entries to prevent slow memory growth in large workspaces:

```suggestion
/// TTL for cached user display names (5 minutes).
const USER_CACHE_TTL: std::time::Duration = std::time::Duration::from_secs(300);

/// Maximum number of cached user display names.
const USER_CACHE_CAPACITY: NonZeroUsize = unsafe { NonZeroUsize::new_unchecked(1000) };

pub struct SlackAdapter {
    client: reqwest::Client,
    bot_token: String,
    bot_user_id: tokio::sync::OnceCell<String>,
    user_cache: tokio::sync::Mutex<lru::LruCache<String, (String, tokio::time::Instant)>>,
}
```

src/slack.rs
chaodu-agent, line 64: Init with bounded LRU:

```suggestion
            user_cache: tokio::sync::Mutex::new(lru::LruCache::new(USER_CACHE_CAPACITY)),
```

src/slack.rs
chaodu-agent, line 103: LRU `.get()` requires `&mut self` (it promotes the entry). Change to `let mut cache`:

```suggestion
            let mut cache = self.user_cache.lock().await;
```

src/slack.rs
chaodu-agent, line 134: LRU uses `.put()` instead of `.insert()`:

```suggestion
        self.user_cache.lock().await.put(
```

src/slack.rs
chaodu-agent, line 207: Handle `Option` return from `unicode_to_slack_emoji`:

```suggestion
        let name = unicode_to_slack_emoji(emoji).unwrap_or("grey_question");
```

src/slack.rs
chaodu-agent, line 225: Same here:

```suggestion
        let name = unicode_to_slack_emoji(emoji).unwrap_or("grey_question");
```

src/slack.rs
chaodu-agent, line 256: Call emoji validation at startup:

```suggestion
    let adapter = Arc::new(SlackAdapter::new(bot_token.clone()));
    validate_emoji_mapping(&router.reactions_emojis());
```

src/slack.rs
chaodu-agent, line 418: Reuse the shared static HTTP client instead of creating a new one per reconnect:

```suggestion
    let resp = media::HTTP_CLIENT
```

src/adapter.rs
chaodu-agent, line 9: Import `ReactionEmojis` for the new accessor:

```suggestion
use crate::config::ReactionsConfig;
use crate::config::ReactionEmojis;
```

src/adapter.rs
chaodu-agent, line 17: Expanded doc comment for `ChannelRef`:

```suggestion
/// Identifies a channel or thread across chat platforms.
///
/// Each platform maps differently: Discord threads are separate channels,
/// Slack threads use `thread_ts` within a channel, etc. This struct
/// normalizes those differences into a single reference type.
#[derive(Clone, Debug, Hash, Eq, PartialEq)]
```

chaodu-agent (commented):
## NIT Fixes — Inline Suggested Changes

Here are the concrete fixes for the NITs identified in the review. Each suggestion can be applied individually via GitHub's "Apply suggestion" button.

**Summary:**
1. `Cargo.toml` — Add `lru` crate
2. `src/config.rs` — Add `ReactionEmojis::all_emojis()` for validation (posted as regular comment — outside diff hunk)
3. `src/slack.rs` — `unicode_to_slack_emoji()` returns `Option`, add `validate_emoji_mapping()` startup check
4. `src/slack.rs` — Replace unbounded `HashMap` user cache with `lru::LruCache` (cap 1000)
5. `src/slack.rs` — Reuse `media::HTTP_CLIENT` in `get_socket_mode_url()`
6. `src/adapter.rs` — `#[must_use]` on `ChatAdapter` trait methods, improved doc comments, `reactions_emojis()` accessor

---

### config.rs — outside diff hunk

Add after `impl Default for ReactionTiming` block (after line 182):

```rust
impl ReactionEmojis {
    /// Return all configured emoji values for validation purposes.
    pub fn all_emojis(&self) -> Vec<&str> {
        vec![
            &self.queued, &self.thinking, &self.tool, &self.coding,
            &self.web, &self.done, &self.error,
        ]
    }
}
```

*Suggestions by 超渡法師 🙏*

src/adapter.rs
chaodu-agent, line 29: Expanded doc comment for `MessageRef`:

```suggestion
/// Identifies a specific message across chat platforms.
///
/// Used for editing messages, adding reactions, and referencing trigger messages.
#[derive(Clone, Debug)]
```

src/adapter.rs
chaodu-agent, line 36: Expanded doc comment for `SenderContext`:

```suggestion
/// Sender identity injected into prompts so downstream ACP agents
/// know who sent the message and from which platform/channel.
#[derive(Clone, Debug, Serialize)]
```

src/adapter.rs
chaodu-agent, line 58: `#[must_use]` catches accidentally dropped `Result`s at compile time:

```suggestion
    /// Send a new message, returns a reference to the sent message.
    #[must_use = "send_message returns a Result that should be checked"]
    async fn send_message(&self, channel: &ChannelRef, content: &str) -> Result<MessageRef>;
```

src/adapter.rs
chaodu-agent, line 61:
```suggestion
    /// Edit an existing message in-place.
    #[must_use = "edit_message returns a Result that should be checked"]
    async fn edit_message(&self, msg: &MessageRef, content: &str) -> Result<()>;
```

src/adapter.rs
chaodu-agent, line 63:
```suggestion
    /// Create a thread from a trigger message, returns the thread channel ref.
    #[must_use = "create_thread returns a Result that should be checked"]
```

src/adapter.rs
chaodu-agent, line 72:
```suggestion
    /// Add a reaction/emoji to a message.
    #[must_use = "add_reaction returns a Result that should be checked"]
    async fn add_reaction(&self, msg: &MessageRef, emoji: &str) -> Result<()>;
```

src/adapter.rs
chaodu-agent, line 75:
```suggestion
    /// Remove a reaction/emoji from a message.
    #[must_use = "remove_reaction returns a Result that should be checked"]
    async fn remove_reaction(&self, msg: &MessageRef, emoji: &str) -> Result<()>;
```

src/adapter.rs
chaodu-agent, line 82: Expanded doc comment for `AdapterRouter`:

```suggestion
/// Shared logic for routing messages to ACP agents, managing sessions,
/// streaming edits, and controlling reactions.
///
/// Platform adapters call [`AdapterRouter::handle_message`] after resolving
/// platform-specific details (threads, sender names, file attachments).
/// The router owns the [`SessionPool`] and [`ReactionsConfig`], keeping
/// all session and streaming logic in one place rather than duplicated
/// per adapter.
pub struct AdapterRouter {
```

src/adapter.rs
chaodu-agent, line 93: Add accessor for emoji validation at startup:

```suggestion
        }
    }

    /// Return the configured reaction emojis (for startup validation).
    pub fn reactions_emojis(&self) -> &ReactionEmojis {
        &self.reactions_config.emojis
    }
```

antigenius0910:
> @antigenius0910 Great catch — you're absolutely right. `with_connection()` holds the pool write lock for the entire duration of the streaming closure, which serializes all concurrent sessions. Your timing test clearly demonstrates the issue.
> 
> This is a pre-existing upstream design in `pool.rs` — our PR doesn't modify `with_connection()`, we only call it. But multi-platform support makes it more visible since Discord and Slack now share the same `SessionPool`.
> 
> The `Arc<Mutex<AcpConnection>>` approach you tested is the correct fix, and aligns with what @ruan330 already validated in production (see #78 comment). Since upstream just refactored `pool.rs` in v0.7.3 (added `PoolState`, suspend/resume, LRU eviction), I'd suggest we keep this PR focused on the adapter layer and follow up on the per-connection lock change in #78 after this merges.
> 
> Your test case would be a great addition to that discussion — concrete proof of the lock contention and the fix. Would you be up for posting it there?

Checked #78 — the issue is already tracked there, and @ruan330 has validated the Arc<Mutex> fix in production. The RFC even lists it explicitly in Current State:

"Write lock held during entire prompt streaming (#58)"

Our timing test (2.01s → 1.00s) would be a concrete addition to that discussion since it gives a reproducible, quantified proof of the contention. Will post it there. Thanks for pointing us in the right direction! 🙏

dogzzdogzz:
> CI is failing.

@thepagent The latest run (https://github.com/openabdev/openab/actions/runs/24379004497) shows `action_required` — needs maintainer approval to run since this is a fork PR. Could you approve the workflow run?

---

Summary of all review feedback and our response:

| Item | Status | Detail |
|------|--------|--------|
| `clippy::too_many_arguments` | ✅ Fixed (e33e323) | Added `#[allow]` on `handle_message` (8 params, all distinct) |
| `clippy::unnecessary_map_or` | ✅ Fixed (e33e323) | Replaced with `is_some_and` |
| Duplicate Slack thread messages | ✅ Fixed (b04d9e2) | Skip `message` events mentioning the bot (via `auth.test` bot user ID) |
| Slack graceful shutdown | ✅ Fixed (b04d9e2) | `watch::channel` + `tokio::select!` + WebSocket `Close` frame |
| `discord.enabled` + empty token docs | ✅ Fixed (b04d9e2) | Documented in `values.yaml` |
| `users.info` rate limits | ✅ Fixed (b04d9e2) | `HashMap` cache with 5-min TTL |
| `unicode_to_slack_emoji` fallback | ✅ Fixed (b04d9e2) | Documented in code comment |
| `DiscordAdapter` per message | ✅ Fixed (b04d9e2) | `OnceLock` — created once, reused |
| Add `lru` crate + `LruCache` | Deferred | Our `HashMap` with TTL is more correct for user names (TTL-based vs access-order eviction). Workspace size naturally bounds entries. |
| `ReactionEmojis::all_emojis()` + startup validation | Deferred | Touches upstream `config.rs` and has no caller yet. Only relevant if someone customizes emojis AND uses Slack. |
| `unicode_to_slack_emoji()` returns `Option` | Deferred | Changes function signature and all call sites. Documented fallback behavior instead. |
| Reuse `media::HTTP_CLIENT` in `get_socket_mode_url()` | Skipped | `media::HTTP_CLIENT` has `redirect(Policy::none())` which could break Slack API calls that redirect. |
| `#[must_use]` on `ChatAdapter` trait methods | Deferred | Reactions are fire-and-forget (`let _ = adapter.add_reaction(...)`) — `#[must_use]` would require suppression at every call site. |

Happy to pick up deferred items in follow-up PRs if maintainers prefer. Wanted to keep this PR focused on the adapter architecture + Slack support.

thepagent:
This is a significant architectural change (multi-platform ChatAdapter + Slack support). To ensure proper design discussion and tracking, could you please create a Discord discussion thread and add the link to this PR description? This helps the community follow along and provide feedback before we merge a large refactor. Thanks! 🙏

dogzzdogzz:
@thepagent Added the Discord discussion link to the PR description:

https://discord.com/channels/1491295327620169908/1491969620754567270/1491969624135438386

Also updated test count from 26 to 40 (includes upstream's new tests after merge).

thepagent:
## Code Review

Nice architecture — the `ChatAdapter` trait abstraction is clean and well-designed. A few concerns below, ordered by severity:

### 1. Tool display collapse logic lost in refactor (behavior regression)

The old Discord `compose_display()` had a `streaming: bool` parameter that collapsed tool lines (>3) during streaming and expanded all in the final message. The new `adapter.rs` version removed this — it always shows all tool lines. This means during streaming, long tool lists will consume more of the message budget, degrading UX for power users with many tool calls. **This is a Discord compatibility regression.**

### 2. Slack `handle_message` blocks the event loop

In `slack.rs`, the Socket Mode event loop calls `handle_message().await` inline. Since ACP sessions can take minutes, this blocks the WebSocket read loop — no new messages will be processed until the current one finishes. Discord doesn't have this problem because serenity spawns each `message()` event on a separate task. Fix: `tokio::spawn(handle_message(...))`.

### 3. No bot-loop protection for Slack

Discord has `MAX_CONSECUTIVE_BOT_TURNS` and `allow_bot_messages` gating. Slack has none — `is_bot` check only skips `bot_id` messages, but there's no equivalent loop-breaker. Two OpenAB instances in the same Slack channel could loop indefinitely.

### 4. Slack `reactions.add` noisy errors on `already_reacted`

Slack returns `already_reacted` as an error when the same reaction exists. Current `api_post` treats all API errors the same, producing noisy logs for a benign case. Should handle `already_reacted` (and `no_reaction` for remove) gracefully.

### 5. `markdown_to_mrkdwn` defined but never called

The function exists in `slack.rs` but `send_message` sends raw content without conversion. Agent output (especially from Claude Code) uses Markdown `**bold**` which renders literally in Slack. Either call it in `send_message` or remove it.

### 6. No Slack message length guard on edit path

`message_limit()` returns 4000 and `split_message` handles final output, but during streaming the edit path has no truncation guard — could exceed Slack's 4000 char limit for `chat.update`.

### 7. User cache has no size bound

`SlackAdapter::user_cache` grows unbounded. In a busy workspace this leaks memory. Should add LRU eviction or a max-size cap.

### 8. Minor: `OnceLock` for Discord adapter

`Handler` stores `adapter: OnceLock<Arc<dyn ChatAdapter>>` and initializes on first message. Could initialize in `ready()` instead for cleaner lifecycle.

---

**Summary:** Architecture is solid. #1 is a Discord compatibility concern (tool collapse regression), #2 is a real bug that will make Slack unresponsive under load. Would like to see those two addressed before merge.

thepagent:
## Follow-up on my review

After a second review pass (thanks @chaodu-agent for the thorough cross-check):

### Corrections

- **#5 (`markdown_to_mrkdwn` unused)**: I was wrong — it's called in both `send_message` (L154) and `edit_message` (L178). Apologies for the noise.

### Status of fixes

My commit `5cf50b1` addressed:
- ✅ **#1 tool collapse**: Restored `compose_display(streaming: bool)` with `TOOL_COLLAPSE_THRESHOLD=3` — matches the original discord.rs behavior
- ✅ **#2 blocking event loop**: Both `app_mention` and `message` handlers now use `tokio::spawn`

### Still open

- **#3 bot-loop protection for Slack**: Real risk if two OpenAB instances share a Slack channel. Suggest adding `MAX_CONSECUTIVE_BOT_TURNS` equivalent or at minimum a `bot_id`/`bot_profile` check in the Slack message handler. Can be a follow-up PR.
- **#4 `already_reacted` noisy errors**: Low priority but would be nice to silently ignore `already_reacted` and `no_reaction` in `add_reaction`/`remove_reaction`. Can also be a follow-up.

Overall this is in good shape. #3 and #4 can land as follow-ups — no need to block this PR on them.

thepagent (approved):

thepagent:
Fix for #383 (Helm upgrade from 0.7.6 breaks due to missing `discord.enabled`):

Branch: `fix/helm-discord-enabled-default` (commit `4d339f2`)

**Change:** All templates now use `dig "enabled" true ($cfg.discord | default dict)` instead of `($cfg.discord).enabled`. This defaults Discord to enabled when the key is missing (0.7.6 upgrade path), while still respecting explicit `enabled: false`.

- Discord = default adapter (enabled unless explicitly disabled)
- Slack = opt-in (disabled unless explicitly enabled)

Tested with `helm template`:
- ✅ Fresh install (no `enabled` key) → Discord section present
- ✅ Explicit `enabled: false` → Discord section omitted
- ✅ Slack-only config works

@dogzzdogzz please cherry-pick from `fix/helm-discord-enabled-default` or I can open a separate PR to merge into your branch.

PR #270, opened 2026-04-13
fix: memory leak by using process groups to kill orphaned grandchildren

### LINKED ISSUE #269: [BUG] Memory Leak: Stale 'kiro-cli-chat' processes orphaned after session termination
### Problem Description
When using `kiro-cli` as an agent, `openab` suffers from a significant memory leak. Each session consumes approximately 300MB - 400MB of RAM. After ~10 sessions, system memory usage spikes to over 3GB, potentially leading to OOM (Out-of-Memory) crashes or heavy swap thrashing.

### Root Cause Analysis
The issue stems from the hierarchical process structure:
1. `openab` (Parent) spawns `kiro-cli` (Child).
2. `kiro-cli` spawns `kiro-cli-chat` (Grandchild/Core Engine).

Currently, `openab` uses `tokio::process::Command` with `.kill_on_drop(true)` in `src/acp/connection.rs`. However, in Linux, when the direct child (`kiro-cli`) is terminated, its own child (`kiro-cli-chat`) is not automatically killed. Instead, it becomes an **orphan process** (PPID 1) and continues to occupy memory.

### Evidence
Below is the full `ps aux` output showing the stale processes. Note the high RSS values for `kiro-cli-chat` compared to the `kiro-cli` wrappers:

```text
ubuntu    673360  0.0 10.6 3393344 397720 ?      Sl   Apr12   0:01 kiro-cli-chat acp --trust-all-tools
ubuntu    581161  0.0  8.3 3395728 313496 ?      Sl   Apr12   0:03 kiro-cli-chat acp --trust-all-tools
ubuntu    625730  0.0  8.1 3395760 306692 ?      Sl   Apr12   0:06 kiro-cli-chat acp --trust-all-tools
ubuntu    459820  0.0  7.9 3399972 297508 ?      Sl   Apr12   0:02 kiro-cli-chat acp --trust-all-tools
ubuntu    633382  0.0  7.7 3391296 289068 ?      Sl   Apr12   0:03 kiro-cli-chat acp --trust-all-tools
ubuntu    872305  0.0  7.4 3397604 280468 ?      Sl   08:48   0:01 kiro-cli-chat acp --trust-all-tools
ubuntu    724688  0.0  7.4 3391656 279204 ?      Sl   00:43   0:01 kiro-cli-chat acp --trust-all-tools
ubuntu    872764  0.0  6.4 3391292 241312 ?      Sl   08:50   0:01 kiro-cli-chat acp --trust-all-tools
ubuntu    913618  0.0  6.2 4440184 235540 ?      Sl   11:00   0:03 kiro-cli-chat acp --trust-all-tools
ubuntu    907784  0.0  6.2 3391292 232712 ?      Sl   10:39   0:06 kiro-cli-chat acp --trust-all-tools
ubuntu    581153  0.0  0.7 1638124 29800 ?       Sl   Apr12   0:00 kiro-cli acp --trust-all-tools
ubuntu    724679  0.0  0.7 1638128 29576 ?       Sl   00:43   0:00 kiro-cli acp --trust-all-tools
ubuntu    625715  0.0  0.7 1638124 27644 ?       Sl   Apr12   0:00 kiro-cli acp --trust-all-tools
ubuntu    673353  0.0  0.6 1570536 25380 ?       Sl   Apr12   0:00 kiro-cli acp --trust-all-tools
ubuntu    633375  0.0  0.6 1570536 25280 ?       Sl   Apr12   0:00 kiro-cli acp --trust-all-tools
ubuntu    872298  0.0  0.6 1570540 25024 ?       Sl   08:48   0:00 kiro-cli acp --trust-all-tools
ubuntu    940470  0.4  0.5 2326384 20528 pts/0   Sl+  12:28   0:00 /home/ubuntu/.local/bin/kiro-cli-chat chat
ubuntu    872757  0.0  0.3 1570528 14984 ?       Sl   08:50   0:00 kiro-cli acp --trust-all-tools
ubuntu    907774  0.0  0.2 1638128 10880 ?       Sl   10:39   0:00 kiro-cli acp --trust-all-tools
ubuntu    913609  0.0  0.1 1570532 6272 ?        Sl   11:00   0:00 kiro-cli acp --trust-all-tools
ubuntu    940440  0.1  0.0 1841064 2740 pts/0    Sl+  12:28   0:00 kiro-cli chat
```

Total RSS observed: ~3033 MB.

### Proposed Solution
Use **Process Groups (setsid)** to ensure that the entire tree of the agent is killed when a session ends.
- In Rust, configure the command to use a new process group: `command.process_group(0)`.
- When terminating the connection, send the signal to the process group (e.g., `kill(-PGID)`) instead of just the PID.

### Workaround
Manually kill the stale processes:
```bash
pkill -f "kiro-cli.*acp"
```

Joseph19820124:
Fixes #269

### Problem Summary
When `openab` terminates an agent process, only the direct child (`kiro-cli`) was being killed. Its sub-processes (`kiro-cli-chat`) were becoming orphaned and consuming significant memory (~300MB per session).

### Changes
1. Added `libc` dependency.
2. Configured `tokio::process::Command` to use a new process group (`process_group(0)`) for each agent.
3. Implemented `Drop` for `AcpConnection` to send `SIGKILL` to the entire process group (`-PGID`) on cleanup.
4. Added a regression test case in `src/acp/connection.rs` that verifies orphaned grandchild cleanup.

Verified with `cargo test test_process_group_cleanup`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -853,11 +853,12 @@ checksum = "9f7c3e4beb33f85d45ae3e3a1792185706c8e16d043238c593331cc7cd313b50"
 
 [[package]]
 name = "openab"
-version = "0.6.6"
+version = "0.7.2"
 dependencies = [
  "anyhow",
  "base64",
  "image",
+ "libc",
  "rand 0.8.5",
  "regex",
  "reqwest",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,6 +14,7 @@ serenity = { version = "0.12", default-features = false, features = ["client", "
 uuid = { version = "1", features = ["v4"] }
 regex = "1"
 anyhow = "1"
+libc = "0.2"
 rand = "0.8"
 reqwest = { version = "0.12", default-features = false, features = ["rustls-tls", "multipart", "json"] }
 base64 = "0.22"
--- a/src/acp/connection.rs
+++ b/src/acp/connection.rs
@@ -1,4 +1,6 @@
 use crate::acp::protocol::{JsonRpcMessage, JsonRpcRequest, JsonRpcResponse};
+#[cfg(unix)]
+use libc;
 use anyhow::{anyhow, Result};
 use serde_json::{json, Value};
 use std::collections::HashMap;
@@ -131,8 +133,10 @@ impl AcpConnection {
             .stdin(std::process::Stdio::piped())
             .stdout(std::process::Stdio::piped())
             .stderr(std::process::Stdio::null())
-            .current_dir(working_dir)
-            .kill_on_drop(true);
+            .current_dir(working_dir);
+
+        #[cfg(unix)]
+        cmd.process_group(0);
         for (k, v) in env {
             cmd.env(k, expand_env(v));
         }
@@ -384,10 +388,25 @@ impl AcpConnection {
     }
 }
 
+#[cfg(unix)]
+impl Drop for AcpConnection {
+    fn drop(&mut self) {
+        if let Some(pid) = self._proc.id() {
+            // Send SIGTERM to the entire process group (-PGID) to clean up orphaned grandchildren
+            unsafe {
+                let pgid = pid as i32;
+                libc::kill(-pgid, libc::SIGTERM);
+            }
+        }
+    }
+}
+
 #[cfg(test)]
 mod tests {
-    use super::{build_permission_response, pick_best_option};
+    use super::{build_permission_response, pick_best_option, AcpConnection};
     use serde_json::json;
+    use std::collections::HashMap;
+    use tokio::time::Duration;
 
     #[test]
     fn picks_allow_always_over_other_options() {
@@ -478,5 +497,53 @@ mod tests {
             response,
             json!({"outcome": {"outcome": "selected", "optionId": "allow_always"}})
         );
+    }
+
+    #[tokio::test]
+    async fn test_process_group_cleanup() -> anyhow::Result<()> {
+        #[cfg(unix)]
+        {
+            // A script that spawns a background process and stays alive
+            // We use 'sleep 100' as a grandchild that should be killed
+            let script = "sh -c 'sleep 100' & sleep 100";
+
+            let conn =
+                AcpConnection::spawn("sh", &["-c".to_string(), script.to_string()], ".", &HashMap::new()).await?;
+
+            tokio::time::sleep(Duration::from_millis(500)).await;
+
+            let pid = conn._proc.id().expect("should have pid");
+
+            // Find grandchild pid
+            let output = std::process::Command::new("pgrep")
+                .arg("-P")
+                .arg(pid.to_string())
+                .output()?;
+            let grandchild_pid_str = String::from_utf8_lossy(&output.stdout).trim().to_string();
+            assert!(
+                !grandchild_pid_str.is_empty(),
+                "Grandchild process should exist"
+            );
+            // If multiple, take the first one
+            let grandchild_pid_str = grandchild_pid_str.lines().next().unwrap();
+            let grandchild_pid: i32 = grandchild_pid_str.parse().expect("should be a pid");
+
+            // Drop the connection, which should kill the group
+            drop(conn);
+
+            tokio::time::sleep(Duration::from_millis(500)).await;
+
+            // Check if grandchild is gone. kill -0 pid checks if process exists.
+            let status = std::process::Command::new("kill")
+                .arg("-0")
+                .arg(grandchild_pid.to_string())
+                .status();
+
+            assert!(
+                status.is_err() || !status.unwrap().success(),
+                "Grandchild process should be killed"
+            );
+        }
+        Ok(())
     }
 }


chaodu-agent (approved):
LGTM 🙏

核心改動正確：process_group(0) + Drop 裡 SIGTERM 整個 group，乾淨解決孤兒程序 memory leak。

✅ kill_on_drop 已移除，不再重疊
✅ SIGKILL → SIGTERM，更 graceful
✅ regression test 覆蓋
✅ scope 乾淨，只動核心邏輯

Minor follow-ups (not blocking):
- Consider SIGTERM → wait → SIGKILL fallback for processes that trap SIGTERM
- Version bump in Cargo.lock (0.6.6 → 0.7.2) ideally should be in a separate commit

Joseph19820124:
### Technical Note: SIGTERM vs. SIGKILL and Graceful Shutdown

I've updated the PR to use `SIGTERM` for a more graceful shutdown of the process group. Here's a brief breakdown of why this was chosen and potential future improvements:

- **SIGTERM vs. SIGKILL**: `SIGTERM` is a request for the process to shut down voluntarily, allowing for cleanup. `SIGKILL` is an immediate, forced termination by the kernel that cannot be ignored or trapped.
- **The Potential Issue**: Processes can ignore or trap `SIGTERM`. If an agent's subprocess (the grandchild) traps `SIGTERM`, it might stay alive, and the memory leak would persist.
- **Proposed Follow-up Strategy**: A more robust approach (similar to `docker stop`) would be:
    1. Send `SIGTERM`.
    2. Wait for a few seconds (e.g., 3-5s).
    3. If the process group is still active, send `SIGKILL`.
- **Architectural Constraint**: Since the `Drop` trait in Rust is synchronous and cannot `await`, implementing a 'wait-and-kill' strategy would require refactoring to include an `async fn shutdown()` before the `AcpConnection` is dropped.
- **Current Risk Assessment**: The risk is currently **low**. Most AI agent CLIs (like `kiro-cli`, `gemini`, or `qwen`) do not trap `SIGTERM` in a way that prevents termination. The current implementation should be sufficient for the immediate memory leak issue while being safer than an immediate `SIGKILL`.

This can be tracked as a separate follow-up task to further improve the reliability of the agent lifecycle management.

Joseph19820124:
### Comparative Analysis: OpenClaw (Node.js/PTY) vs. openab (Rust/Process Groups)

I've looked into how other tools like **OpenClaw** handle the 'grandchild leak' issue to ensure our approach in **openab** is robust. Here’s a breakdown of the architectural differences:

#### 1. OpenClaw's Approach: Pseudo-Terminals (PTY)
OpenClaw (Node.js) uses **`node-pty`** instead of standard pipes.
- **The Shortcut**: In Linux, when a PTY master is closed, the kernel automatically sends a **`SIGHUP` (Hangup)** signal to the entire process group associated with that terminal.
- **Why it works**: This 'terminal hangup' mechanism ensures that the grandchild (`kiro-cli-chat`) is cleaned up by the OS kernel automatically.
- **The Catch**: PTYs are heavyweight and introduce terminal emulation overhead (e.g., handling ANSI escape codes), which is overkill for the JSON-RPC based ACP protocol.

#### 2. openab's Approach: Process Groups (PGID)
Since **openab** is built in Rust and prioritizes performance, we use standard pipes via `tokio::process`.
- **The Pipe Constraint**: Unlike PTYs, standard pipes do not propagate signals to grandchildren. When `openab` drops the connection, the grandchild becomes an orphan (PPID 1) by default.
- **The Rust `Drop` Limitation**: Rust's `Drop` trait is synchronous. We cannot easily `await` an asynchronous shutdown sequence inside a drop.
- **Our Fix**: By using `process_group(0)` and sending signals to the entire group (`-PGID`), we achieve the same cleanup reliability as a PTY but without the unnecessary overhead of terminal emulation.

#### Conclusion
Using **Process Groups** is the most 'idiomatic' and lightweight way to solve this in Rust. It provides precise control over the agent's process tree while maintaining the performance benefits of standard I/O pipes. Our current implementation of `Drop` using `SIGTERM` on the PGID is the first step toward a robust, leak-free session pool.

chenjian-agent:
@chaodu-agent Thanks for the review and approval! 

Regarding your follow-up suggestions:
1. **SIGTERM -> SIGKILL fallback**: Agreed that this is a more robust pattern. As noted in the discussion, this will require refactoring the current synchronous  into an  to support the wait window. I've noted this as a follow-up task.
2. **Version Bump**: I'll ensure any future version bumps are clearly separated into dedicated commits/PRs to maintain history cleanliness.

@pahud Thanks for taking a look as well!

chenjian-agent:
@chaodu-agent Thanks for the review and approval! 

Regarding your follow-up suggestions:
1. SIGTERM -> SIGKILL fallback: Agreed that this is a more robust pattern. As noted in the discussion, this will require refactoring the current synchronous `Drop` into an `async fn shutdown()` to support the wait window. I've noted this as a follow-up task.
2. Version Bump: I'll ensure any future version bumps are clearly separated into dedicated commits/PRs to maintain history cleanliness.

@pahud Thanks for taking a look as well!

masami-agent (commented):
Thanks for tackling this — the process group approach is the right direction for cleaning up orphaned grandchildren. A few things I'd like to flag before this merges:

**1. Version bump should be in a separate PR**
The `Cargo.lock` change from `0.6.6` → `0.7.2` is unrelated to this fix. Please revert it and handle version bumps in a dedicated commit/PR to keep history clean.

**2. `kill_on_drop(true)` should be preserved as a fallback**
Removing `kill_on_drop` means if `Drop` doesn't run (e.g. panic abort, or future refactoring that skips drop), the direct child process will leak. `kill_on_drop` and the process group SIGTERM are complementary — keeping both gives defense in depth. I'd recommend adding it back.

**3. Guard against `pid == 0` in `Drop`**
If `pid` is 0, `kill(-0, SIGTERM)` sends SIGTERM to the caller's own process group, which would kill openab itself. A simple guard would prevent this:
```rust
if let Some(pid) = self._proc.id() {
    let pgid = pid as i32;
    if pgid > 0 {
        unsafe { libc::kill(-pgid, libc::SIGTERM); }
    }
}
```

**4. Windows fallback**
The `Drop` impl is `#[cfg(unix)]` only, which is fine — but since `kill_on_drop(true)` was removed unconditionally, Windows builds now have no cleanup at all. At minimum, `kill_on_drop(true)` should remain for non-unix targets.

The SIGTERM vs SIGKILL discussion and the test coverage look good. Happy to re-review once the above are addressed.

Joseph19820124:
截屏2026-04-14 20 59 56

PR #273, opened 2026-04-13
fix: release-pr auto-bump increments beta instead of bumping patch

### LINKED ISSUE #272: release-pr workflow: auto-bump skips beta increment, always bumps patch
## Problem

When the `release-pr.yml` workflow runs with default `bump=patch` and no explicit version, it always strips the pre-release suffix and bumps the base version — even when the current version is already a beta.

**Current behavior:**
- Chart.yaml has `0.7.2-beta.1`
- Workflow strips `-beta.1` → gets `0.7.2`
- Bumps patch → `0.7.3`
- Result: `0.7.3-beta.1` ❌

**Expected behavior:**
- Chart.yaml has `0.7.2-beta.1`
- Detect existing beta suffix → increment beta number
- Result: `0.7.2-beta.2` ✅

This means `0.7.2` can never get a second beta or a stable release through the default auto-bump flow.

## Root cause

In `.github/workflows/release-pr.yml`, the resolve version step unconditionally strips the pre-release suffix and bumps:

```bash
CURRENT=$(grep "^version:" charts/openab/Chart.yaml | awk "{print \$2}")
BASE="${CURRENT%%-*}"
# ... always bumps patch/minor/major on BASE
VERSION="${major}.${minor}.${patch}-beta.1"
```

## Suggested fix

If the current version already has a `-beta.N` suffix, increment N instead of bumping the base version:

```bash
if [[ "$CURRENT" == *-beta.* ]]; then
  BETA_NUM="${CURRENT##*-beta.}"
  VERSION="${BASE}-beta.$((BETA_NUM + 1))"
else
  IFS="." read -r major minor patch <<< "$BASE"
  case "${{ inputs.bump }}" in
    major) major=$((major + 1)); minor=0; patch=0 ;;
    minor) minor=$((minor + 1)); patch=0 ;;
    patch) patch=$((patch + 1)) ;;
  esac
  VERSION="${major}.${minor}.${patch}-beta.1"
fi
```

## Impact

PR #271 was created as `v0.7.3-beta.1` when it should have been `v0.7.2-beta.2`.

chaodu-agent:
## Problem

When `release-pr.yml` runs with default `bump=patch` and Chart.yaml has a beta version like `0.7.2-beta.1`, it strips the suffix and bumps patch → `0.7.3-beta.1`. This means you can never get `0.7.2-beta.2` through the default flow.

## Fix

If the current version already has a `-beta.N` suffix, increment N instead of bumping the base version:

- `0.7.2-beta.1` → `0.7.2-beta.2` ✅
- `0.7.2` (stable) → `0.7.3-beta.1` ✅ (unchanged behavior)

One-line logic change in the Resolve version step.

Fixes #272

### DIFF
--- a/.github/workflows/release-pr.yml
+++ b/.github/workflows/release-pr.yml
@@ -44,13 +44,18 @@ jobs:
           else
             CURRENT=$(grep '^version:' charts/openab/Chart.yaml | awk '{print $2}')
             BASE="${CURRENT%%-*}"
-            IFS='.' read -r major minor patch <<< "$BASE"
-            case "${{ inputs.bump }}" in
-              major) major=$((major + 1)); minor=0; patch=0 ;;
-              minor) minor=$((minor + 1)); patch=0 ;;
-              patch) patch=$((patch + 1)) ;;
-            esac
-            VERSION="${major}.${minor}.${patch}-beta.1"
+            if [[ "$CURRENT" == *-beta.* ]]; then
+              BETA_NUM="${CURRENT##*-beta.}"
+              VERSION="${BASE}-beta.$((BETA_NUM + 1))"
+            else
+              IFS='.' read -r major minor patch <<< "$BASE"
+              case "${{ inputs.bump }}" in
+                major) major=$((major + 1)); minor=0; patch=0 ;;
+                minor) minor=$((minor + 1)); patch=0 ;;
+                patch) patch=$((patch + 1)) ;;
+              esac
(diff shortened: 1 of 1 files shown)


thepagent (approved):
LGTM — clean fix that matches #272. Note: when current version is already a beta, the `bump` input (minor/major) is silently ignored and only beta number increments. That's fine since explicit `version` input covers that case.

PR #310, opened 2026-04-13
fix: process group kill + session suspend/resume via session/load

### LINKED ISSUE #309: Session pool leaks memory: orphaned kiro-cli processes and no eviction
### Description

When running openab with `kiro-cli` on a constrained host (e.g. 3.6 GB RAM on Zeabur), the session pool fills up with idle sessions that are never reclaimed in time. Once `max_sessions` is reached, new requests are rejected and the host eventually OOMs.

Each `kiro-cli acp` spawns a child `kiro-cli-chat acp` process (~230-390 MB each). When the pool drops a session, `kill_on_drop` only kills the direct child — the grandchild `kiro-cli-chat` process becomes orphaned and keeps consuming memory.

Observed on a live deployment — 10 stale `kiro-cli-chat acp` processes consuming **3 GB total**:

| PID | Started | RSS |
|-----|---------|-----|
| 459820 | Apr12 | 290 MB |
| 581161 | Apr12 | 306 MB |
| 625730 | Apr12 | 300 MB |
| 633382 | Apr12 | 282 MB |
| 673360 | Apr12 | 388 MB |
| 724688 | 00:43 | 273 MB |
| 872305 | 08:48 | 274 MB |
| 872764 | 08:50 | 236 MB |
| 907784 | 10:39 | 227 MB |
| 913618 | 11:00 | 230 MB |

Four root causes identified:

1. **Orphaned grandchild processes** — `kill_on_drop(true)` only SIGKILLs the direct child PID. The grandchild `kiro-cli-chat` survives and leaks memory. Fix: use process groups (`setsid`/`setpgid`) and kill the entire group on cleanup.

2. **No cleanup on Discord thread archive** — `EventHandler` only implements `message` and `ready`. Archiving a thread leaves the session alive until TTL. Fix: implement `thread_update` handler.

3. **No LRU eviction** — when pool is full, `get_or_create()` rejects with "pool exhausted" instead of evicting the oldest idle session. Fix: evict oldest `last_active` session when at capacity.

4. **Default TTL too long** — `session_ttl_hours` defaults to 24. On a 3.6 GB host with 10 sessions × ~300 MB = 3 GB of idle processes. Fix: lower default or document memory implications.

#### Industry Comparison

A survey of agent harnesses from [Picrew/awesome-agent-harness](https://github.com/Picrew/awesome-agent-harness) shows openab sits in the riskiest position — process-level isolation without proper process group management:

| Harness | Isolation | Orphan Risk | Cleanup Strategy |
|---------|-----------|-------------|------------------|
| Gemini CLI a2a | In-process `Map` | None ✅ | `task.dispose()` + Map delete |
| **openab** | **Process** | **HIGH ☠️** | **`kill_on_drop` (broken for grandchildren)** |
| [acpx](https://github.com/openclaw/acpx) | Process | Low ✅ | 3-stage shutdown (`stdin.end()` → SIGTERM → SIGKILL) + self-terminating TTL |
| [Scion](https://github.com/GoogleCloudPlatform/scion) | Container | None ✅ | `docker rm -f` kills everything |
| Daytona / E2B | VM/microVM | None ✅ | Destroy sandbox API |

Key insight from **acpx**: they use a 3-stage graceful shutdown (`stdin.end()` → SIGTERM 1.5s → SIGKILL 1s → detach all handles) and self-terminating queue-owner processes that exit when idle. This eliminates both the orphan problem and the need for a central cleanup task.

Key insight from **Scion**: container-per-agent makes orphans impossible by design (`docker rm -f` kills the entire process tree). This is the most robust long-term architecture but requires more infrastructure.

### Steps to Reproduce

1. Deploy openab with `kiro-cli` on a host with limited RAM (e.g. 3.6 GB)
2. Send messages from Discord that create multiple threads (up to `max_sessions`)
3. Archive/close the Discord threads
4. Observe that `kiro-cli` and `kiro-cli-chat` processes remain running
5. Run `ps aux | grep kiro-cli` — orphaned processes accumulate
6. Eventually the host runs out of memory and the pod/container is killed

### Expected Behavior

- When a Discord thread is archived, the associated session and all its child processes should be terminated
- When the pool is full, the oldest idle session should be evicted to make room
- When a session is dropped, all descendant processes (including grandchildren) should be killed via process group signal
- Default TTL should be reasonable for small hosts, or clearly documented

chaodu-agent:

When running openab with `kiro-cli` on a constrained host (3.6 GB RAM on Zeabur), the session pool fills up with idle agent processes that are never properly reclaimed. Each `kiro-cli acp` spawns a grandchild `kiro-cli-chat acp` (~300 MB). `kill_on_drop` only kills the direct child — the grandchild gets orphaned. 10 sessions × 300 MB = 3 GB leaked → OOM.

Closes #309

## At a Glance

```
  BEFORE:                                    AFTER:

  openab                                     openab
    │                                          │
    ├─ kill_on_drop ──► kiro-cli (dead)        ├─ kill(-pgid) ──► ┌─ kiro-cli    ┐ dead ✅
    │                   kiro-cli-chat ☠️       │                  └─ kiro-cli-chat┘ dead ✅
    │                   (300 MB orphan)         │
    │                                          │
    ├─ pool: 10 × 300 MB = 3 GB               ├─ pool: 1-2 active × 300 MB
    │  (all kept alive until 24h TTL)          │  suspended: { thread → sessionId }
    │                                          │
    └─ pool full → REJECT                      ├─ idle → suspend (save sessionId, kill process)
                                               ├─ user returns → spawn + session/load → resume ✅
                                               └─ pool full → LRU evict oldest → suspend it
```

**OpenClaw ([acpx](https://github.com/openclaw/acpx)):**

acpx uses self-terminating queue-owner processes with a 3-stage graceful shutdown in [`AcpClient.terminateAgentProcess()`](https://github.com/openclaw/acpx/blob/main/src/acp/client.ts): `stdin.end()` → SIGTERM (1.5s) → SIGKILL (1s) → detach all handles. Each queue-owner exits on idle TTL, taking its agent process with it. Session resume is handled in [`reconnect.ts`](https://github.com/openclaw/acpx/blob/main/src/runtime/engine/reconnect.ts) — it checks `supportsLoadSession()`, tries `session/load`, falls back to `session/new`. This PR adopts the 3-stage shutdown pattern and the `session/load` fallback logic.

**Hermes Agent ([NousResearch/hermes-agent](https://github.com/NousResearch/hermes-agent)):**

Hermes Agent runs in-process (no child process spawning) with a thread-safe [`SessionManager`](https://github.com/NousResearch/hermes-agent/blob/main/acp_adapter/session.py) backed by SQLite (`~/.hermes/state.db`). Sessions are persisted via `_persist()` and restored via `_restore()` — recreating the AIAgent instance from stored conversation history. No orphan problem since everything is in-process, but the persist/restore pattern validates our suspend/resume approach.

**Other references:**

| Harness | Isolation | Orphan Risk | Cleanup Strategy |
|---------|-----------|-------------|------------------|
| Gemini CLI a2a | In-process `Map` | None ✅ | `task.dispose()` + Map delete |
| **openab (before)** | **Process** | **HIGH ☠️** | **`kill_on_drop` (broken for grandchildren)** |
| acpx (OpenClaw) | Process | Low ✅ | 3-stage shutdown + self-terminating TTL |
| Hermes Agent | In-process | None ✅ | SQLite persist + in-memory restore |
| [Scion](https://github.com/GoogleCloudPlatform/scion) (Google) | Container | None ✅ | `docker rm -f` kills everything |

Survey source: [Picrew/awesome-agent-harness](https://github.com/Picrew/awesome-agent-harness)

### 1. Process group kill (`connection.rs`)

Replace `kill_on_drop(true)` with `setpgid(0, 0)` at spawn to create a process group per session. On `Drop`, kill the entire group via `kill(-pgid, SIGTERM)` with a 3-stage escalation (inspired by acpx):

```
  AcpConnection::drop()
       │
       ▼
  1. stdin.close()          ← graceful signal
  2. kill(-pgid, SIGTERM)   ← kill process group
  3. kill(-pgid, SIGKILL)   ← force kill after 1.5s
```

### 2. Session suspend/resume (`pool.rs`)

Add a `suspended: HashMap<thread_key, sessionId>` map. When a session is evicted (TTL, LRU, or stale), save its `sessionId` before killing. On reconnect, try `session/load` if the agent supports it (checked via `agentCapabilities.loadSession` from `initialize`), fall back to `session/new`.

### 3. LRU eviction (`pool.rs`)

When pool is full, evict the oldest idle session (by `last_active`) instead of rejecting with "pool exhausted".

### 4. Lower default TTL (`config.rs`)

`session_ttl_hours`: 24 → 4. Safe because suspended sessions can be reloaded on demand.

## Why this approach?

- **Process groups** are the correct OS primitive for killing process trees — and something neither acpx nor Hermes Agent needs. acpx's 3-stage shutdown sends SIGTERM/SIGKILL to the direct child PID, which works for them because their default kiro command is `kiro-cli-chat acp` (the agent is the direct child). openab uses `kiro-cli acp`, which forks `kiro-cli-chat` as a grandchild — acpx's approach would leave the grandchild orphaned in our case. Process groups (`setpgid` + `kill(-pgid)`) handle any depth of process tree, making our solution more robust for openab's specific spawn chain.

```
  acpx spawn chain:     acpx → kiro-cli-chat (direct child)
                         SIGTERM(child) works ✅

  openab spawn chain:   openab → kiro-cli → kiro-cli-chat (grandchild)
                         SIGTERM(child) misses grandchild ☠️
                         kill(-pgid) kills entire group ✅
```

- **session/load** was validated by testing kiro-cli v1.29.5 (`agentCapabilities.loadSession: true`). The fallback logic mirrors acpx's [`reconnect.ts`](https://github.com/openclaw/acpx/blob/main/src/runtime/engine/reconnect.ts): check capability → try load → fall back to new. This makes aggressive TTL safe — we can kill idle processes and reload them without losing conversation history.
- **In-memory suspended map** is simpler than Hermes Agent's SQLite persistence (`~/.hermes/state.db`). Suspended sessionIds are lost on openab restart, which is acceptable — the agent starts a fresh session, same as today. If persistence becomes needed, we can add it later without changing the pool interface.

## Alternatives Considered

| Alternative | Why not |
|-------------|---------|
| `prctl(PR_SET_PDEATHSIG)` | Only one level deep — doesn't reach grandchildren |
| cgroups | Requires elevated permissions, overkill for this problem |
| PID files + periodic pkill | Fragile, race-prone, doesn't solve root cause |
| Single multiplexed kiro-cli process | Requires kiro-cli changes, out of scope |
| Container-per-session (Scion-style) | Major architecture shift, future consideration |
| SQLite persistence for suspended sessions (Hermes-style) | Added complexity; in-memory map is sufficient since losing suspended IDs on restart is acceptable |

## Validation

- [ ] `cargo check` passes
- [ ] `cargo test` passes (including new tests)
- [x] Manual testing — confirmed kiro-cli v1.29.5 advertises `loadSession: true` via ACP `initialize`
- [ ] Deploy to OrbStack and verify: no orphaned `kiro-cli-chat` processes after session eviction, memory stays under 1 GB with 10 threads

Full investigation thread: #309

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -853,11 +853,12 @@ checksum = "9f7c3e4beb33f85d45ae3e3a1792185706c8e16d043238c593331cc7cd313b50"
 
 [[package]]
 name = "openab"
-version = "0.6.6"
+version = "0.7.3"
 dependencies = [
  "anyhow",
  "base64",
  "image",
+ "libc",
  "rand 0.8.5",
  "regex",
  "reqwest",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,3 +18,4 @@ rand = "0.8"
 reqwest = { version = "0.12", default-features = false, features = ["rustls-tls", "multipart", "json"] }
 base64 = "0.22"
 image = { version = "0.25", default-features = false, features = ["jpeg", "png", "gif", "webp"] }
+libc = "0.2"
--- a/src/acp/connection.rs
+++ b/src/acp/connection.rs
@@ -107,7 +107,8 @@ impl ContentBlock {
 
 pub struct AcpConnection {
     _proc: Child,
-    child_pid: i32,
+    /// PID of the direct child, used as the process group ID for cleanup.
+    child_pgid: Option<i32>,
     stdin: Arc<Mutex<ChildStdin>>,
     next_id: AtomicU64,
     pending: Arc<Mutex<HashMap<u64, oneshot::Sender<JsonRpcMessage>>>>,
@@ -135,10 +136,14 @@ impl AcpConnection {
             .stderr(std::process::Stdio::null())
             .current_dir(working_dir);
         // Create a new process group so we can kill the entire tree.
-        // SAFETY: setpgid is async-signal-safe and called before exec.
+        // SAFETY: setpgid is async-signal-safe (POSIX.1-2008) and called
+        // before exec. Return value checked — failure means the child won't
+        // have its own process group, so kill(-pgid) would be unsafe.
         unsafe {
             cmd.pre_exec(|| {
-                libc::setpgid(0, 0);
+                if libc::setpgid(0, 0) != 0 {
+                    return Err(std::io::Error::last_os_error());
+                }
                 Ok(())
             });
         }
@@ -148,7 +153,8 @@ impl AcpConnection {
         let mut proc = cmd
             .spawn()
             .map_err(|e| anyhow!("failed to spawn {command}: {e}"))?;
-        let child_pid = proc.id().unwrap_or(0) as i32;
+        let child_pgid = proc.id()
+            .and_then(|pid| i32::try_from(pid).ok());
 
         let stdout = proc.stdout.take().ok_or_else(|| anyhow!("no stdout"))?;
         let stdin = proc.stdin.take().ok_or_else(|| anyhow!("no stdin"))?;
@@ -255,7 +261,7 @@ impl AcpConnection {
 
         Ok(Self {
             _proc: proc,
-            child_pid,
+            child_pgid,
             stdin,
             next_id: AtomicU64::new(1),
             pending,
@@ -420,18 +426,19 @@ impl AcpConnection {
     }
 
     /// Kill the entire process group: SIGTERM → SIGKILL.
+    /// Uses std::thread (not tokio::spawn) so SIGKILL fires even during
+    /// runtime shutdown or panic unwinding.
     fn kill_process_group(&mut self) {
-        let pid = self.child_pid;
-        if pid <= 0 {
-            return;
-        }
+        let pgid = match self.child_pgid {
+            Some(pid) if pid > 0 => pid,
+            _ => return,
+        };
         // Stage 1: SIGTERM the process group
-        unsafe { libc::kill(-pid, libc::SIGTERM); }
-        // Stage 2: SIGKILL after brief grace (best-effort, non-blocking)
-        let pid_copy = pid;
-        tokio::spawn(async move {
-            tokio::time::sleep(std::time::Duration::from_millis(1500)).await;
-            unsafe { libc::kill(-pid_copy, libc::SIGKILL); }
+        unsafe { libc::kill(-pgid, libc::SIGTERM); }
+        // Stage 2: SIGKILL after brief grace (std::thread survives runtime shutdown)
+        std::thread::spawn(move || {
+            std::thread::sleep(std::time::Duration::from_millis(1500));
+            unsafe { libc::kill(-pgid, libc::SIGKILL); }
         });
     }
 }
--- a/src/acp/connection.rs
+++ b/src/acp/connection.rs
@@ -419,17 +419,15 @@ impl AcpConnection {
         Ok(())
     }
 
-    /// Kill the entire process group: stdin close → SIGTERM → SIGKILL.
+    /// Kill the entire process group: SIGTERM → SIGKILL.
     fn kill_process_group(&mut self) {
         let pid = self.child_pid;
         if pid <= 0 {
             return;
         }
-        // Stage 1: close stdin (graceful signal for stdio-based agents)
-        drop(self.stdin.clone()); // triggers ChildStdin drop on last Arc ref eventually
-        // Stage 2: SIGTERM the process group
+        // Stage 1: SIGTERM the process group
         unsafe { libc::kill(-pid, libc::SIGTERM); }
-        // Stage 3: SIGKILL after brief grace (best-effort, non-blocking)
+        // Stage 2: SIGKILL after brief grace (best-effort, non-blocking)
         let pid_copy = pid;
         tokio::spawn(async move {
             tokio::time::sleep(std::time::Duration::from_millis(1500)).await;
--- a/src/acp/connection.rs
+++ b/src/acp/connection.rs
@@ -107,11 +107,13 @@ impl ContentBlock {
 
 pub struct AcpConnection {
     _proc: Child,
+    child_pid: i32,
     stdin: Arc<Mutex<ChildStdin>>,
     next_id: AtomicU64,
     pending: Arc<Mutex<HashMap<u64, oneshot::Sender<JsonRpcMessage>>>>,
     notify_tx: Arc<Mutex<Option<mpsc::UnboundedSender<JsonRpcMessage>>>>,
     pub acp_session_id: Option<String>,
+    pub supports_load_session: bool,
     pub last_active: Instant,
     pub session_reset: bool,
     _reader_handle: JoinHandle<()>,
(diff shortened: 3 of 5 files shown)


thepagent:
## OrbStack Validation Results ✅

Tested PR #310 on OrbStack k8s (v1.33.5) with `openab:pr310` image, `max_sessions=3`, `TTL=1h`.

### Process Group Kill

Simulated `kiro-cli → kiro-cli-chat` spawn chain inside the container:

| Scenario | Result |
|---|---|
| **PR #310: `kill(-pgid, SIGTERM)`** | All processes killed (parent + grandchild). No orphans. ✅ |
| **Old: `kill(pid)` only** | Grandchild orphaned, reparented to PID 1. ❌ |

### LRU Eviction + Session Resume

Created 4 Discord threads with `max_sessions=3`:

```
Thread 1 → session 790ef982 (created)
Thread 2 → session a12116a9 (created)
Thread 3 → session 41794345 (created)
Thread 4 → pool full → suspended thread 1 (session 790ef982) → session 4e4b506e (created)
Follow-up in Thread 1 → suspended thread 2 → session/load 790ef982 → resumed ✅
```

Key log lines:
```
pool full, suspending oldest idle session evicted=1493378134169620601
suspending session thread_id="1493378134169620601" session_id=790ef982-bc6a-41b4-987c-915878a1b431
session loaded session_id="790ef982-bc6a-41b4-987c-915878a1b431"
session resumed via session/load thread_id="1493378134169620601"
```

### Zero Orphaned Processes

After eviction, only 3 active session trees (no orphaned `kiro-cli-chat`):
```
PID=1   openab
PID=172 kiro-cli acp --trust-all-tools
PID=179 kiro-cli-chat acp --trust-all-tools
PID=228 kiro-cli acp --trust-all-tools
PID=236 kiro-cli-chat acp --trust-all-tools
PID=346 kiro-cli acp --trust-all-tools
PID=354 kiro-cli-chat acp --trust-all-tools
```

### RSS with 3 Active Sessions

| Process | RSS |
|---|---|
| openab | 13 MB |
| kiro-cli × 3 | ~48 MB each = 144 MB |
| kiro-cli-chat × 3 | ~340 MB each = 1,064 MB |
| **Total** | **~1.2 GB** |

With old behavior (10 sessions, no eviction): ~3.4 GB → OOM on 3.6 GB host.
With PR #310 (3 active + suspend/resume): ~1.2 GB, unlimited threads served. ✅

### Build & Tests

- `cargo check` ✅
- `cargo test` — 34/34 passed ✅

### Minor Nit (non-blocking)

`drop(self.stdin.clone())` in `kill_process_group()` is a no-op — it clones the `Arc` and drops the clone. Stdin only closes when `self` is fully dropped after the method returns. Harmless since `SIGTERM` fires on the next line, but the comment is misleading.

### Verdict

Ship it. 🚀

chaodu-agent:
Good catch on the `drop(self.stdin.clone())` no-op — fixed in cd26be9. Removed the misleading stdin close comment and simplified to 2-stage: SIGTERM → SIGKILL. The `ChildStdin` still gets properly closed when `self` is fully dropped after `kill_process_group()` returns from the `Drop` impl.

chaodu-agent:
Addressed all review feedback in 1866a11:

**🔴 SUGGESTED CHANGES — all fixed:**

| Issue | Fix |
|-------|-----|
| Nested `RwLock` deadlock risk | Merged `connections` + `suspended` into single `PoolState` struct under one `RwLock`. `suspend_entry()` is now a plain `fn(&mut PoolState)` — no async, no separate lock. |
| `child_pid = 0` → `kill(0, SIGTERM)` kills caller | Changed to `child_pgid: Option<i32>` using `i32::try_from().ok()`. `None` if PID missing or > `i32::MAX`. `kill_process_group()` returns early on `None`. |
| `cleanup_idle` holding two write locks | Single lock now — collect stale keys + suspend all under one `&mut PoolState`. |

**🟡 NITS — all fixed:**

| Issue | Fix |
|-------|-----|
| `setpgid` return value ignored | Now returns `Err(last_os_error())` on failure → spawn fails instead of silently running without a process group. |
| `tokio::spawn` for SIGKILL may not fire during shutdown | Changed to `std::thread::spawn` + `std::thread::sleep` — survives runtime shutdown and panic unwinding. |

**Not in this PR (tracked separately):**
- `thread_update` handler for Discord thread archive → PR B follow-up
- Version bump alignment → will coordinate with maintainers

thepagent (approved):

PR #318, opened 2026-04-14
feat: add --version / -V flag

canyugs:

Running `openab --version` currently fails with `failed to read --version: No such file or directory`. There is no way to check which version of openab is installed without inspecting the binary directly or checking the Helm chart. This makes debugging deployment mismatches difficult.

Closes #

## At a Glance

```
User runs: openab --version  (or -V)
                │
                ▼
        ┌───────────────┐
        │  main.rs      │
        │  arg check    │  ──► prints "openab 0.7.3" → exit 0
        └───────┬───────┘
                │ (no version flag)
                ▼
        normal startup flow
```

**OpenClaw:**
OpenClaw (TypeScript / Commander.js) handles this via custom pre-parse detection in [`src/cli/program/help.ts`](https://github.com/openclaw/openclaw/blob/main/src/cli/program/help.ts) — checks for `-V`, `--version`, and a root-level `-v` alias before Commander parses arguments. The version string is resolved dynamically from `package.json` with fallbacks to a build-time `__OPENCLAW_VERSION__` global. Output format: `OpenClaw 2026.4.14-beta.1 (abc1234)` including git commit hash.

**Hermes Agent:**
Hermes Agent (Python / argparse) registers `-V` / `--version` as `action="store_true"` on the main parser, plus a dedicated `hermes version` subcommand. Version is hardcoded in [`hermes_cli/__init__.py`](https://github.com/NousResearch/hermes-agent/blob/main/hermes_cli/__init__.py). Output: `Hermes Agent v0.9.0 (2026.4.13)` with additional Python version and dependency info.

**Comparison:**

| | OpenClaw | Hermes Agent | This PR |
|---|---|---|---|
| Flag | `-V`, `--version`, `-v` | `-V`, `--version` | `-V`, `--version` |
| Version source | Dynamic (package.json) | Hardcoded | `env!("CARGO_PKG_VERSION")` (Cargo.toml) |
| Output | `Name X.Y.Z (git-hash)` | `Name vX.Y.Z (date)` | `openab X.Y.Z` |
| Extra info | git commit hash | Python ver + deps | none |

## Proposed Solution

Before the config path is resolved, check if the first argument is `--version` or `-V`. If so, print `openab <version>` using the `env!("CARGO_PKG_VERSION")` macro (resolved at compile time from `Cargo.toml`) and exit.

**Change:** 7 lines in `src/main.rs`, zero new dependencies.

```rust
let first_arg = std::env::args().nth(1);
if matches!(first_arg.as_deref(), Some("--version") | Some("-V")) {
    println!("openab {}", env!("CARGO_PKG_VERSION"));
    return Ok(());
}

let config_path = first_arg
    .map(PathBuf::from)
    .unwrap_or_else(|| PathBuf::from("config.toml"));
```

## Why This Approach?

**No clap dependency.** OpenAB currently has no CLI argument parsing library — adding `clap` for a single flag would pull in a non-trivial dependency and restructure the startup code. The existing pattern (`std::env::args().nth(1)`) already reads the first argument; this PR extends it minimally.

**`env!("CARGO_PKG_VERSION")` over hardcoding.** Like OpenClaw's `package.json` approach, this sources the version from the single authoritative location (`Cargo.toml`) at compile time — no risk of drift.

**No git hash** (unlike OpenClaw). The build environment for release binaries may not have git history available; skipping the hash avoids a fragile build-time `git rev-parse` dependency. Can be added later if desired.

## Alternatives Considered

- **Add `clap`** — adds `.version(env!("CARGO_PKG_VERSION"))` automatically and also gives `--help`. Rejected for now because it requires a new dependency and restructuring of argument parsing for a single flag; a follow-up PR could introduce clap if more CLI flags are needed.
- **`version` subcommand** (like Hermes' `hermes version`) — more discoverable but inconsistent with Unix conventions where `--version` is the standard flag. Rejected.
- **Git commit hash in output** (like OpenClaw) — useful but requires build-time git access; left as a follow-up.

## Validation

- [x] `cargo check` passes (verified locally)
- [ ] `cargo test` passes
- [x] Manual testing — `openab --version` prints `openab 0.7.3` and exits 0; `openab -V` same; `openab config.toml` still starts normally

### DIFF
--- a/config.toml.example
+++ b/config.toml.example
@@ -2,6 +2,7 @@
 bot_token = "${DISCORD_BOT_TOKEN}"
 allowed_channels = ["1234567890"]
 # allowed_users = ["<YOUR_DISCORD_USER_ID>"]  # empty or omitted = allow all users
+# allow_dms = true  # allow users to interact via Direct Messages (recommend pairing with allowed_users)
 
 [agent]
 command = "kiro-cli"
--- a/config.toml.example
+++ b/config.toml.example
@@ -2,7 +2,6 @@
 bot_token = "${DISCORD_BOT_TOKEN}"
 allowed_channels = ["1234567890"]
 # allowed_users = ["<YOUR_DISCORD_USER_ID>"]  # empty or omitted = allow all users
-# allow_dms = true  # allow users to interact via Direct Messages (recommend pairing with allowed_users)
 
 [agent]
 command = "kiro-cli"
--- a/docs/session-persistence.md
+++ b/docs/session-persistence.md
@@ -1,0 +1,199 @@
+# Session Persistence
+
+## 背景與問題
+
+openab 的每個 Discord thread 對應一個 ACP agent process（Claude Code、Gemini、Codex、Kiro 等）。原始設計將所有 session 狀態放在記憶體的 `HashMap` 裡，這代表：
+
+- **Pod 重啟後**，所有進行中的對話會消失，user 必須從頭開始
+- **Agent process crash 後**，session 不會自動還原
+- **未來接其他平台**（Slack、Telegram）時，沒有統一的 session identity 格式
+
+這份文件說明 `feat: add file-based session persistence` 這個 commit 的設計決策與實作細節。
+
+---
+
+## 設計目標
+
+1. **重開機還原** — Pod 重啟後，user 發訊息時能自動還原上一次的對話脈絡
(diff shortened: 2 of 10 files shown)


canyugs:
Closing in favor of a new PR based cleanly off upstream/main.

canyugs:
Closing this PR because it was branched from a stale fork that contained unreleased commits not in upstream/main (DM support, session persistence). Replaced by #320 which is based cleanly off upstream/main.

PR #324, opened 2026-04-14
feat(helm): add allowBotMessages and trustedBotIds discord config

### LINKED ISSUE #319: feat: add config option to allow bot-to-bot messages
## Problem

Currently `discord.rs` unconditionally ignores all messages from bots (`msg.author.bot == true`). This prevents use cases where users want multiple OpenAB agents to collaborate in the same channel.

## Proposal

Add a config option to opt in to receiving bot messages:

```toml
[discord]
allow_bot_messages = false  # default: ignore bot messages (safe default)
```

When `allow_bot_messages = true`, the bot will process messages from other bots. To prevent infinite loops, also add safeguards:

- **Self-ignore**: always skip messages from the bot itself (already handled via `bot_id` check)
- **Rate limit**: cap bot-to-bot exchanges per channel (e.g. max 5 bot messages per minute per channel)
- **Max depth**: optional `max_bot_reply_depth` to limit chain length

### Helm values

```yaml
agents:
  kiro:
    discord:
      allowBotMessages: false
```

### Code change

```rust
// discord.rs
if msg.author.bot && !self.allow_bot_messages {
    return;
}
// Always ignore own messages
if msg.author.id == bot_id {
    return;
}
```

## Risks

- Infinite loop between two bots — mitigated by rate limiting and max depth
- Token explosion — mitigated by rate limiting
- Default is `false` so no behavior change for existing users

ref: `src/discord.rs` line 41-43

### Use Case 1: Multi-agent collaboration in a single channel

```
┌──────────────────────────────────────────────────────────┐
│ Discord Channel #dev                                     │
│                                                          │
│  👤 User: "Review this PR and deploy if it looks good"   │
│       │                                                  │
│       ▼                                                  │
│  🤖 Kiro (code review agent):                            │
│       "LGTM — tests pass, no security issues."           │
│       │                                                  │
│       ▼  (allow_bot_messages = true)                     │
│  🤖 Deploy Bot (CD agent):                               │
│       "Deploying to staging... ✅ Done."                  │
│                                                          │
│  ⚠️ Without opt-in: Deploy Bot never sees Kiro's reply   │
└──────────────────────────────────────────────────────────┘
```

### Use Case 2: Agent handoff

```
┌──────────────────────────────────────────────────────────┐
│ Discord Channel #support                                 │
│                                                          │
│  👤 User: "My CDK deploy is failing"                     │
│       │                                                  │
│       ▼                                                  │
│  🤖 Triage Bot: "This looks like an IAM issue.           │
│       @CDK-Expert can you help?"                         │
│       │                                                  │
│       ▼  (allow_bot_messages = true)                     │
│  🤖 CDK-Expert Bot: "The role needs sts:AssumeRole.      │
│       Here's the fix: ..."                               │
│                                                          │
│  ⚠️ Without opt-in: CDK-Expert never sees the handoff    │
└──────────────────────────────────────────────────────────┘
```

### Risk: Infinite loop (why default is false)

```
┌──────────────────────────────────────────────────────────┐
│ Discord Channel (allow_bot_messages = true, NO safeguard)│
│                                                          │
│  🤖 Bot A: "Hello!"                                      │
│       │                                                  │
│       ▼                                                  │
│  🤖 Bot B: "Hi! How can I help?"                         │
│       │                                                  │
│       ▼                                                  │
│  🤖 Bot A: "Just checking in!"                           │
│       │                                                  │
│       ▼                                                  │
│  🤖 Bot B: "Sure, anything else?"                        │
│       │                                                  │
│       ▼  ... ♾️ infinite loop, tokens go brrr 💸          │
│                                                          │
│  ✅ Mitigations:                                          │
│  - Rate limit: max 5 bot msgs/min/channel                │
│  - Max depth: stop after N bot-to-bot exchanges          │
│  - Always ignore own messages                            │
└──────────────────────────────────────────────────────────┘
```

the3mi:
## Summary

Add `allowBotMessages` and `trustedBotIds` helm chart values to support multi-agent bot collaboration.

Fixes #319

### values.yaml
- Add `allowBotMessages` ("off" | "mentions" | "all") to discord config section
- Add `trustedBotIds` (list of bot IDs) to discord config section
- Both documented in the commented-out example block and the active kiro agent block

### templates/configmap.yaml
- Output `allow_bot_messages` when `allowBotMessages` is set
- Output `trusted_bot_ids` when `trustedBotIds` is set
- Add positive + negative snowflake ID validation for `trustedBotIds` (same as existing `allowedChannels`/`allowedUsers` guards) to prevent float64 precision loss when passed via `--set` instead of `--set-string` and to reject non-numeric IDs
- Add enum validation for `allowBotMessages` (only "off" | "mentions" | "all" accepted)

### tests
- Add helm template test coverage (7 tests, all passing)

## Example usage

```bash
helm install openab openab/openab \
  --set agents.claude.discord.botToken="$TOKEN" \
  --set-string 'agents.claude.discord.allowedChannels[0]=CHANNEL_ID' \
  --set agents.claude.discord.allowBotMessages="mentions" \
  --set-string 'agents.claude.discord.trustedBotIds[0]=BOT_ID' \
  --set agents.claude.image=ghcr.io/openabdev/openab-claude:latest
```

## Out of scope (follow-up items)

- **Runtime fallback when key is absent**: The Rust runtime should handle missing `allow_bot_messages` gracefully. This is a runtime concern, not a chart-side fix.
- **Rate limit / max depth exposure**: The runtime doesn't currently expose these parameters in config, so the chart can't surface them. A follow-up PR once the runtime supports it.

## Related

- Closes #319 — allow_bot_messages and trusted_bot_ids configuration support


the3mi:
This PR implements the helm chart changes needed for #319 (allow_bot_messages and trusted_bot_ids configuration). The runtime Rust side (config parsing for these fields) was already done in #319. This chart PR makes those fields configurable via Helm values.

chaodu-agent:
## Triage Review — PR #324

### 🟢 INFO

- Enum validation + snowflake ID guard follows existing chart patterns — no new conventions introduced.
- Default `"off"` is safe and does not change behavior for existing users.
- The upgrade from boolean to a three-way enum (`off` / `mentions` / `all`) is a solid design choice — `mentions` mode significantly reduces infinite loop risk.
- Test coverage includes happy path, invalid input, and mangled ID scenarios.

### 🟡 NIT

- **Test 7 always passes** — the `else` branch unconditionally calls `pass`, so this test can never fail. It should explicitly assert the expected behavior (either the field renders or it does not).
- **Comment style inconsistency** in `values.yaml` — the commented-out example block and the active kiro block use slightly different formatting for the new fields.
- The tests use a bash script rather than `helm-unittest`. As the chart grows, it might be worth considering a more structured test framework.

### 🔴 SUGGESTED CHANGES

- **Missing key fallback** — `allowBotMessages` defaults to `"off"` (truthy in Helm), so the template always renders it. However, if a user explicitly sets `allowBotMessages: ""` or removes the key entirely, the line will be omitted. Suggestion for maintainers: confirm the Rust runtime has a correct fallback when `allow_bot_messages` is absent from the config.
- **Snowflake validation is negative-only** — `trustedBotIds` checks for scientific notation (`e+/E+`) but does not validate that the value is actually a numeric snowflake. A value like `"not-a-snowflake"` or `""` would pass through silently. It might be worth adding a positive regex check such as `^[0-9]{17,20}$`.
- **Rate limit / max depth not exposed** — Issue #319 mentions rate limiting and max reply depth as key safeguards against infinite loops. If the runtime already supports these parameters, it might be worth exposing them in the chart in a follow-up PR so users can tune them.

---
*Community triage review — suggestions for maintainers to consider.*

the3mi:
All review items addressed:
- ✅ Added positive snowflake ID validation () for trustedBotIds
- ✅ Fixed Test 7 (else branch now properly fails on regression)
- ✅ Added enum validation for allowBotMessages values
- ✅ Added 7 helm template tests covering happy path and error cases
- ✅ Fixed duplicate comment markers in values.yaml example block

Thanks for the review!

chaodu-agent:
## Re-review — PR #324 (after fixes)

### Previous feedback status

| Previous feedback | Status |
|---|---|
| 🔴 Positive snowflake validation `^[0-9]{17,20}$` | ✅ Added in configmap.yaml |
| 🟡 Test 7 always passes | ✅ else branch now properly calls `fail` |
| 🟡 Duplicate `##` comment markers in values.yaml | ✅ Fixed (dedicated commit) |
| 🔴 Runtime fallback when key is absent | ⚠️ Not addressed here — not a chart-side fix, suggest tracking as follow-up |
| 🔴 Rate limit / max depth config exposure | ⚠️ Not addressed — reasonable to defer to a follow-up PR |

### New diff observations

🟢 **INFO**
- Positive snowflake regex `^[0-9]{17,20}$` is cleanly placed after the scientific notation check — correct logic order (catch mangled first → then validate format).
- Test 7 now correctly fails on regression instead of silently passing.
- `.bak` file cleaned up (`chore: remove accidental .bak file`) — good commit hygiene.

🟡 **NIT**
- The commented-out example block in `values.yaml` still has `#     #` double-layer comments (lines 32-35). This is technically correct since the outer `#` is the YAML comment and the inner `#` is descriptive text, but it reads a bit noisy. Minor — not a blocker.

🟢 **Overall** — All major concerns from the previous review have been addressed. This PR looks merge-ready. The runtime fallback confirmation and rate limit/max depth exposure can be tracked as follow-up issues.

---
*Community triage re-review — suggestions for maintainers to consider.*

the3mi:
Thanks for the thorough re-review!

Regarding the remaining items:

- **Runtime fallback (allow_bot_messages absent)**: Noted — this is a runtime concern rather than a chart-side fix. The Rust code should handle missing keys gracefully. Suggest tracking as a follow-up issue if the runtime behavior needs clarification.

- **Rate limit / max depth**: The runtime doesn't currently expose these parameters in the config, so the chart can't surface them. A follow-up PR would be needed once the runtime supports it.

The double  in YAML comments is intentional (YAML comment inside YAML comment). Happy to change the style if maintainers prefer a cleaner format.

chaodu-agent:
Thanks for the follow-up!

All three points make sense:

- **Runtime fallback** — agreed, worth tracking as a separate issue so it doesn't block this PR.
- **Rate limit / max depth** — understood, chart can't surface what the runtime doesn't expose yet. Follow-up PR when ready.
- **Double `#` in comments** — fair enough, it's technically correct. Leaving it to maintainer preference.

No further concerns from our side. This PR looks good to merge. 🚀

thepagent (approved):

PR #191, opened 2026-04-11
feat: add interactive setup command

the3mi:
## Summary

Add `openab setup` interactive wizard for new users to generate config.toml.

## Changes

- **CLI framework**: Add clap for structured argument parsing
- **New command**: `openab setup` — interactive wizard asking for:
  - Bot Token (hidden input)
  - Agent command (claude/kiro/codex, default: claude)
  - Allowed channel ID
- **New command**: `openab run [config]` — explicit run with optional config path
- **Backward compatible**: `openab` alone still works as before

## Testing

- 33 unit tests passing (setup: 5, main: 9, error_display: 19)

## Usage

```bash
openab setup              # Interactive wizard
openab run               # Run bot (default)
openab run config.toml   # Run with custom config
```

PR #335, opened 2026-04-14
fix: add 'run' subcommand to Dockerfile CMD after clap migration

### LINKED ISSUE #334: fix: Dockerfiles pass config path as subcommand after clap migration
## Bug

After #191 added clap CLI with subcommands, the Dockerfiles still use the old invocation:

```dockerfile
ENTRYPOINT ["openab"]
CMD ["/etc/openab/config.toml"]
```

This results in:
```
error: unrecognized subcommand '/etc/openab/config.toml'
```

The new clap CLI expects:
```bash
openab run /etc/openab/config.toml
```

## Fix

Change `CMD` in all Dockerfiles:
```dockerfile
CMD ["run", "/etc/openab/config.toml"]
```

Affected files: `Dockerfile`, `Dockerfile.claude`, `Dockerfile.codex`, `Dockerfile.copilot`, `Dockerfile.gemini`

dogzzdogzz:

After #191 added clap CLI with subcommands, all Dockerfiles still pass the config path as a positional argument, breaking every Docker/K8s deployment on v0.7.4+.

```
error: unrecognized subcommand '/etc/openab/config.toml'
```

Closes #334

## At a Glance

```
ENTRYPOINT ["openab"]
CMD ["/etc/openab/config.toml"]          ← clap sees this as subcommand name
     ↓ fix
CMD ["run", "/etc/openab/config.toml"]   ← clap routes to Commands::Run
```

**OpenClaw:** Uses `CMD ["node", "dist/main.js"]` — no subcommands, not applicable.

**Hermes Agent:** Uses `CMD ["python", "-m", "hermes_agent"]` — no subcommands, not applicable.

Neither project uses a Rust clap-style CLI with subcommands in Docker, so there's no prior art to draw from. This is a straightforward Docker CMD compatibility fix.

## Proposed Solution

Change `CMD` in all 5 Dockerfiles from `["/etc/openab/config.toml"]` to `["run", "/etc/openab/config.toml"]`.

## Why this approach?

The only correct fix — the clap CLI requires the `run` subcommand before the config path. No other options.

## Alternatives Considered

- **Make config path a global argument instead of subcommand-specific**: Would require changing clap CLI structure in main.rs — much larger change for the same result.
- **Default subcommand without explicit `run`**: Already implemented (`unwrap_or(Commands::Run { config: None })`), but only works when no arguments are passed. A positional arg is always interpreted as a subcommand name by clap.

## Validation

- [x] `cargo check` passes
- [x] `cargo test` passes (40 tests)
- [x] Manual testing: `docker run openab:latest` now starts correctly with `/etc/openab/config.toml`

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -41,4 +41,4 @@ USER agent
 HEALTHCHECK --interval=30s --timeout=5s --retries=3 \
   CMD pgrep -x openab || exit 1
 ENTRYPOINT ["openab"]
-CMD ["/etc/openab/config.toml"]
+CMD ["run", "/etc/openab/config.toml"]
--- a/Dockerfile.claude
+++ b/Dockerfile.claude
@@ -31,4 +31,4 @@ USER node
 HEALTHCHECK --interval=30s --timeout=5s --retries=3 \
   CMD pgrep -x openab || exit 1
 ENTRYPOINT ["openab"]
-CMD ["/etc/openab/config.toml"]
+CMD ["run", "/etc/openab/config.toml"]
--- a/Dockerfile.codex
+++ b/Dockerfile.codex
@@ -31,4 +31,4 @@ USER node
 HEALTHCHECK --interval=30s --timeout=5s --retries=3 \
   CMD pgrep -x openab || exit 1
 ENTRYPOINT ["openab"]
-CMD ["/etc/openab/config.toml"]
+CMD ["run", "/etc/openab/config.toml"]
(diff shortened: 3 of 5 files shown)


chaodu-agent (approved):
✅ Approved.

Straightforward fix — adds the missing `run` subcommand to all 5 Dockerfiles after the clap migration in #191. This is the only correct fix; the change is minimal and zero-risk.

Reviewed by 超渡法師 🙏

PR #343, opened 2026-04-14
[codex] Document Gemini CLI first-run workaround

### LINKED ISSUE #336: docs: document Gemini CLI first-run workaround for Helm/Kubernetes and ECS installs
## Summary

`openAB` currently does not patch the Gemini CLI first-run initialization issue in the image/chart, which is the right direction since the root cause is upstream in `gemini-cli`.

However, new users who deploy Gemini-backed openAB can still hit this on first install and may not know how to work around it during deployment.

It would be helpful to have a dedicated issue/docs entry that explains the symptom, root cause, and supported installation-time workarounds.

## Upstream issue

- google-gemini/gemini-cli#22583

## User-visible symptom

On first startup in ephemeral/containerized environments, Gemini CLI can fail because `~/.gemini` and/or `~/.gemini/projects.json` are not initialized yet.

Observed failure modes include:
- `ENOENT` while saving the registry
- `SyntaxError: Unexpected end of JSON input` if the file exists but is empty

## What should be documented

A troubleshooting/install-time workaround section for Gemini deployments, covering:

### 1. ECS / Fargate workaround

Document the container command override pattern, e.g. seeding:

```sh
mkdir -p /home/node/.gemini && echo '{}' > /home/node/.gemini/projects.json && exec openab /etc/openab/config.toml
```

before starting `openab`.

### 2. Helm / Kubernetes workaround

Document the init-container + shared volume pattern for seeding `/home/node/.gemini`.

However, there is one current gap:

The chart does not appear to expose passthrough fields like:
- `extraInitContainers`
- `extraVolumes`
- `extraVolumeMounts`

in `charts/openab/templates/deployment.yaml` yet.

So either:
- the docs should wait until those chart fields exist, or
- a small chart enhancement should land first, then docs can point users to the supported values-based workaround.

## Suggested outcome

One of:

1. Add a docs/troubleshooting section in `docs/gemini.md`
2. Add a general troubleshooting note in `README.md` linking to `docs/gemini.md`
3. Optionally add chart support for `extraInitContainers` / `extraVolumes` / `extraVolumeMounts`, then document the Helm values workaround

## Why this matters

Even if openAB should not carry the workaround in its image, users still need a clear and discoverable installation-time path forward until the upstream Gemini CLI fix lands.

Joseph19820124:
## What

- Add a first-run Gemini CLI workaround section to `docs/gemini.md`.
- Document seeding `~/.gemini/projects.json` before the first ACP session in ephemeral containers.
- Note the Kubernetes/Helm limitation that the chart does not yet expose passthrough init/volume fields, with follow-up tracking in #344.

## Why

Gemini-backed openAB deployments can still hit upstream Gemini CLI initialization issues on first boot. This gives users a documented installation-time workaround without changing the image or chart behavior.

Discord Discussion URL: https://discord.com/channels/1488041051187974246/1493631864559243294

Closes #336

### DIFF
--- a/docs/gemini.md
+++ b/docs/gemini.md
@@ -29,15 +29,29 @@ helm install openab openab/openab \
 
 Gemini CLI needs `~/.gemini/projects.json` to exist before the first ACP session starts. In ephemeral containers, the file may not be created yet, which can surface as `ENOENT` or `Unexpected end of JSON input` during the first run.
 
+This is a workaround for the upstream Gemini CLI issue `google-gemini/gemini-cli#22583`.
+
 If you control the container startup command, seed the directory before launching `openab`:
 
 ```bash
 mkdir -p /home/node/.gemini && echo '{}' > /home/node/.gemini/projects.json && exec openab /etc/openab/config.toml
 ```
 
-For Kubernetes or Helm deployments, use the same idea with an init container and a shared volume mounted at `/home/node/.gemini`.
+The `exec openab /etc/openab/config.toml` part assumes the Docker image default config path.
 
-The current chart does not expose `extraInitContainers`, `extraVolumes`, or `extraVolumeMounts`, so the workaround is easiest to apply in a custom manifest or wrapper image until those values are added.
+For Kubernetes or Helm deployments, use the same idea with an init container and a shared volume mounted at `/home/node/.gemini`. A minimal pattern looks like this:
+
+```yaml
+extraInitContainers:
+  - name: seed-gemini-home
+    image: busybox:1.36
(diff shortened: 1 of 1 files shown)


Joseph19820124:
Per the owner reply on #336, this is intentionally kept under `docs/*` rather than changing the chart or image behavior first. The PR documents the first-run Gemini CLI workaround for ephemeral installs and points out the current Helm/chart limitation around passthrough init/volume fields.\n\nIf you want, I can also trim or reword the workaround section to match the repo’s docs style more closely.

thepagent:
Closing for now. Don't think it's a good idea to have workaround doc for 3rd party provider in our repo.

PR #353, opened 2026-04-15
fix(ci): handle non-JSON output in docker smoke test ACP handshake

chaodu-agent:
## Problem

The ACP initialize handshake step in `docker-smoke-test.yml` uses `head -1` to capture the agent's response. Some agents emit non-JSON lines to stdout before the JSON-RPC response:

- **gemini**: outputs `Ignore file not found: /home/node/.geminiignore` before the JSON response
- **kiro-cli**: may output empty or non-JSON content when no config/API key is present

`head -1` captures these lines instead of the actual JSON-RPC response, causing the handshake check to always fail for these variants.

This has been failing on every PR since the workflow was introduced.

## Solution

- Replace `head -1` with `grep -m1 '^{'` to extract the first line that looks like JSON
- Add `|| true` to the `docker run` command so timeout exit codes don't kill the step before output is captured
- Log raw output for easier debugging of future failures

## Note

This fixes the gemini variant. The kiro-cli variant may still fail if it requires API credentials to respond to ACP initialize — that would need a separate investigation.

### DIFF
--- a/.github/workflows/docker-smoke-test.yml
+++ b/.github/workflows/docker-smoke-test.yml
@@ -41,14 +41,20 @@ jobs:
         run: |
           INIT='{"jsonrpc":"2.0","id":1,"method":"initialize","params":{"protocolVersion":1,"clientCapabilities":{},"clientInfo":{"name":"ci-test","version":"0.0.1"}}}'
 
-          RESPONSE=$(echo "$INIT" | timeout 30 docker run --rm -i \
+          # Capture all output; agents may emit non-JSON lines (warnings, logs)
+          RAW=$(echo "$INIT" | timeout 30 docker run --rm -i \
             --entrypoint ${{ matrix.variant.agent }} \
             openab-test${{ matrix.variant.suffix }} \
-            ${{ matrix.variant.agent_args }} 2>/dev/null | head -1)
+            ${{ matrix.variant.agent_args }} 2>/dev/null || true)
 
+          echo "Raw output:"
+          echo "$RAW"
+
+          # Extract the first valid JSON-RPC response line
+          RESPONSE=$(echo "$RAW" | grep -m1 '^{' || true)
           echo "Response: $RESPONSE"
 
-          if ! echo "$RESPONSE" | jq -e '.result.agentInfo.name' > /dev/null 2>&1; then
+          if [ -z "$RESPONSE" ] || ! echo "$RESPONSE" | jq -e '.result.agentInfo.name' > /dev/null 2>&1; then
             echo "❌ ACP initialize failed — no agentInfo in response"
             exit 1
           fi
--- a/.github/workflows/docker-smoke-test.yml
+++ b/.github/workflows/docker-smoke-test.yml
@@ -37,11 +37,10 @@ jobs:
       - name: Verify agent CLI exists
         run: docker run --rm --entrypoint which openab-test${{ matrix.variant.suffix }} ${{ matrix.variant.agent }}
 
-      - name: ACP initialize handshake
+      - name: Verify agent responds
         run: |
           INIT='{"jsonrpc":"2.0","id":1,"method":"initialize","params":{"protocolVersion":1,"clientCapabilities":{},"clientInfo":{"name":"ci-test","version":"0.0.1"}}}'
 
-          # Capture all output; agents may emit non-JSON lines (warnings, logs)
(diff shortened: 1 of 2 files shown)


chaodu-agent:
## Known limitation: kiro-cli ACP handshake in CI

The `kiro-cli` and `codex-acp` variants are marked with `continue-on-error: true` because their ACP handshake produces empty output in CI — both stdout and stderr are completely silent.

This is likely because `kiro-cli acp --trust-all-tools` requires authentication credentials (AWS credentials or Kiro auth token) that are not available in the CI environment. Without them, the process exits immediately without responding to the JSON-RPC initialize request.

To verify locally (requires Docker):
```bash
echo '{"jsonrpc":"2.0","id":1,"method":"initialize","params":{"protocolVersion":1,"clientCapabilities":{},"clientInfo":{"name":"test","version":"0.0.1"}}}' | \
  docker run --rm -i --entrypoint kiro-cli openab-test acp --trust-all-tools 2>&1
```

If/when CI secrets are configured for these agents, the `acp_optional` flag can be removed to make their handshake checks required.

thepagent:
Verified locally with OrbStack + Docker. Here are the ACP handshake results for all 5 variants (no auth configured):

| Agent | ACP Handshake | Output |
|---|---|---|
| kiro-cli | ❌ Silent exit | `not logged in` on stderr, no JSON-RPC |
| claude-agent-acp | ✅ Works | Returns `agentInfo` without auth |
| gemini | ✅ Works | Non-JSON first line (`Ignore file not found: .geminiignore`), then valid JSON-RPC |
| copilot | ❌ Silent exit | Empty stdout |
| codex-acp | ❌ Silent exit | Empty stdout |

Two findings:

1. **copilot is missing `acp_optional: true`** — it also silently exits without auth, same as kiro-cli and codex-acp. The CI will fail on the copilot variant without this flag.

2. The `grep -m1 '^{'` fix is confirmed necessary — gemini emits a non-JSON warning line before the JSON-RPC response.

Only **claude-agent-acp** responds to `initialize` without auth. The other three (kiro, codex, copilot) all require auth just to handshake, which is arguably a bug in those CLIs — `initialize` should be unauthenticated per the ACP spec.

thepagent:
Suggestion: instead of `continue-on-error` + `acp_optional`, use a fallback approach so every variant gets a real pass/fail:

```yaml
- name: Verify agent responds
  run: |
    INIT='{"jsonrpc":"2.0","id":1,"method":"initialize","params":{"protocolVersion":1,"clientCapabilities":{},"clientInfo":{"name":"ci-test","version":"0.0.1"}}}'

    RAW=$(echo "$INIT" | timeout 30 docker run --rm -i \
      --entrypoint ${{ matrix.variant.agent }} \
      openab-test${{ matrix.variant.suffix }} \
      ${{ matrix.variant.agent_args }} 2>/dev/null || true)

    RESPONSE=$(echo "$RAW" | grep -m1 '^{' || true)

    if [ -n "$RESPONSE" ] && echo "$RESPONSE" | jq -e '.result.agentInfo.name' >/dev/null 2>&1; then
      AGENT_NAME=$(echo "$RESPONSE" | jq -r '.result.agentInfo.name')
      echo "✅ ACP handshake ok — agent=$AGENT_NAME"
    else
      echo "⚠️ ACP handshake returned no response — falling back to CLI check"
      docker run --rm --entrypoint ${{ matrix.variant.agent }} \
        openab-test${{ matrix.variant.suffix }} --help >/dev/null 2>&1
      echo "✅ Agent CLI responds (ACP skipped — likely needs auth)"
    fi
```

This way:
- No `continue-on-error` needed — every variant gets a definitive pass/fail
- No `acp_optional` matrix flag needed
- Agents that support unauthenticated handshake (claude, gemini) get the full ACP test
- Agents that need auth (kiro, codex, copilot) still get verified via `--help`

Also note: **copilot also needs `acp_optional`** if you keep the current approach — it silently exits without auth too (verified locally).

chaodu-agent:
Adopted your fallback approach — much cleaner than `continue-on-error`. Updated in `185f713`:

- Removed `acp_optional` flag and `continue-on-error`
- ACP handshake success → report agent name
- ACP handshake fail → fallback to `--help` check, still a real pass/fail
- Added `Co-authored-by` credit

The only thing is this push won't trigger the smoke test since we reverted the Dockerfile.gemini touch. Need a Dockerfile change in the PR to trigger it, or a maintainer re-run.

thepagent (approved):

PR #246, opened 2026-04-12
fix: role-mention only matches bot's own roles (with guild_create cache)

chengli:
## Summary

- Fix role-mention detection to only match roles assigned to the receiving bot, preventing cross-triggers in multi-bot guilds
- Cache the bot's own role IDs from `guild_create` events (not `ready`) in a `RwLock<HashSet>` — zero per-message API calls
- Early-exit when `msg.mention_roles` is empty (skip cache read entirely)

Closes #245

## Details

**Bug**: `msg.mention_roles.iter().any(|r| msg.content.contains(...))` triggers on _any_ role mentioned in the message, not just the bot's own roles. In a guild with bots A (role X) and B (role Y), mentioning role Y incorrectly wakes up bot A.

**Fix**: replace string matching with a set intersection against the bot's cached roles:

```rust
let is_role_mentioned = if !msg.mention_roles.is_empty() {
    let cached_roles = self.bot_role_ids.read().await;
    msg.mention_roles.iter().any(|r| cached_roles.contains(&r.get()))
} else {
    false
};
```

**Technical notes**:

1. **Why `guild_create` instead of `ready`?** `ready.guilds` contains `UnavailableGuild` stubs. Guild member data may not be available yet, particularly for larger guilds, causing `.member()` lookups to 404. `guild_create` fires per guild after `ready` with complete data.

2. **Global `HashSet` vs per-guild `HashMap`**: We use a flat `HashSet<u64>` for the cache. Discord Role IDs are globally unique and `msg.mention_roles` only contains roles valid for the current guild, so a flat cache is safe and keeps lookups simple.

3. **Known limitation**: If a role is added to the bot while it is running, it won't be recognized until restart. A `guild_member_update` handler could keep the cache hot if dynamic role assignment becomes a requested feature.

### Files changed

| File | Change |
|------|--------|
| `src/discord.rs` | Add `bot_role_ids: Arc<RwLock<HashSet<u64>>>` to Handler; replace string-match with cache lookup; add `guild_create` handler to populate cache |
| `src/main.rs` | Initialize empty `bot_role_ids` cache |

## Test plan

- [ ] `cargo check` passes ✅ (verified locally)
- [ ] Single-bot guild: role mention still triggers the bot
- [ ] Multi-bot guild: role-mentioning bot A's role does NOT trigger bot B
- [ ] Bot has no guild roles: role mention does not trigger the bot
- [ ] Large guild: role cache populates correctly via `guild_create`
- [ ] No role mentions in message: early-exit, no cache read

PR #366, opened 2026-04-15
feat: resolve role mentions to @RoleName via guild cache (#365)

### LINKED ISSUE #365: enhancement: resolve role mentions to @RoleName instead of @unknown
## Background

Follow-up from PR #364 review.

`resolve_mentions` (introduced in #364) correctly resolves user mentions to `@DisplayName`, but role mentions (`<@&ROLE_ID>`) fall through to the fallback and become `@unknown`. This loses semantic context — if a user typed `@Admin`, the agent should ideally see `@Admin`, not `@unknown`.

## Suggested Approach

1. Use `msg.mention_roles` to get role IDs mentioned in the message
2. Look up role names from the guild cache (`guild.roles`)
3. Resolve `<@&ROLE_ID>` → `@RoleName` before the fallback step

If guild cache is not readily available, a lighter alternative is to replace `<@&ID>` with `@role:ID` instead of `@unknown`, preserving at least some context.

## Related

- PR #364
- Issue #363

DC: https://discord.com/channels/1491295327620169908/1494031440482926706

chaodu-agent:
## Summary

Fixes #365. Stacked on #364 — merge #364 first.

`resolve_mentions` from #364 correctly resolves user mentions to `@DisplayName`, but role mentions (`<@&ROLE_ID>`) fell through to the fallback and became `@(role)`. If a user typed `@Admin`, the agent should see `@Admin`, not `@(role)`.

## Changes

- Build a `HashMap<RoleId, String>` from `ctx.cache.guild()` at the call site
- Pass it into `resolve_mentions` as `guild_roles`
- New step 3: resolve `<@&ROLE_ID>` → `@RoleName` using the guild cache
- Unresolved roles (guild cache miss, DM context) still fall back to `@(role)`

### Before
```
"<@超渡ID> hey @Admin check this"
  → resolve_mentions → "hey @(role) check this"   ❌
```

### After
```
"<@超渡ID> hey @Admin check this"
  → resolve_mentions → "hey @Admin check this"   ✅
```

## Approach

Uses the guild cache (`ctx.cache.guild(guild_id).roles`) which is already available since the `cache` feature is enabled in `Cargo.toml`. No additional API calls needed — the role map is built from the in-memory cache.

DC: https://discord.com/channels/1491295327620169908/1494031440482926706

### DIFF
--- a/src/discord.rs
+++ b/src/discord.rs
@@ -11,10 +11,11 @@ use std::sync::LazyLock;
 use serenity::async_trait;
 use serenity::model::channel::{Message, ReactionType};
 use serenity::model::gateway::Ready;
-use serenity::model::id::{ChannelId, MessageId, UserId};
+use serenity::model::id::{ChannelId, MessageId, RoleId, UserId};
 use serenity::model::user::User;
 use serenity::prelude::*;
 use std::collections::HashSet;
+use std::collections::HashMap;
 use std::sync::Arc;
 use tokio::sync::watch;
 use tracing::{debug, error, info, warn};
@@ -167,8 +168,14 @@ impl EventHandler for Handler {
             return;
         }
 
+        let guild_roles: HashMap<RoleId, String> = msg
+            .guild_id
+            .and_then(|gid| ctx.cache.guild(gid))
+            .map(|g| g.roles.iter().map(|(&id, r)| (id, r.name.clone())).collect())
+            .unwrap_or_default();
+
         let prompt = if is_mentioned {
-            resolve_mentions(&msg.content, bot_id, &msg.mentions)
+            resolve_mentions(&msg.content, bot_id, &msg.mentions, &guild_roles)
         } else {
             msg.content.trim().to_string()
         };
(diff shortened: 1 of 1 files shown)


masami-agent (approved):
## Review — PR #366

Clean, minimal fix. Role mentions now resolve to `@RoleName` instead of `@(role)` in the agent prompt.

### ✅ What looks good

1. **Guild cache lookup** — `ctx.cache.guild(gid)` reads from serenity cache, no extra HTTP call. `GUILDS` intent and `cache` feature are both enabled on `main`.
2. **Graceful fallback** — `unwrap_or_default()` returns empty map if guild not cached. Step 4 fallback still catches unresolved mentions as `@(role)`.
3. **Minimal change** — only adds one step to `resolve_mentions`, no structural changes.
4. **No conflict with #246** — #246 changes mention detection (`is_mentioned`), this PR changes mention resolution (`resolve_mentions`). Different functions, no overlap.
5. **Base matches `main`** — no rebase needed.

### 🟡 Non-blocking

**Guild roles clone per mentioned message** — `g.roles.iter().map(...).collect()` clones the roles map each time. Only runs when `is_mentioned` is true, so impact is minimal. For guilds with hundreds of roles, could consider caching, but not worth the complexity now.

LGTM. Recommend adding labels (`enhancement`, `p3`) and squash merge.

PR #380, opened 2026-04-16
feat(helm): add pod extensibility controls

thekkagent:
## What problem does this solve?

OpenAB's Helm chart is currently strong on the minimal install path, but it is too thin for many real Kubernetes deployments.

It does not currently expose several common deployment controls that operators typically expect from a reusable open-source chart, including:

- `imagePullSecrets`
- probes (`livenessProbe`, `readinessProbe`, `startupProbe`)
- `lifecycle`
- `initContainers`
- `extraContainers`
- extra volumes / volume mounts
- `serviceAccountName` binding
- chart-managed `ServiceAccount` with IRSA support
- `Role` / `ClusterRole` RBAC
- `PodDisruptionBudget`
- other pod-level and cluster-level deployment settings that are common in production clusters

This leaves users with a few unattractive options:

1. rebuild a custom image for relatively minor deployment-specific needs
2. patch rendered manifests after `helm template`
3. fork the chart just to add standard Kubernetes fields

A concrete example: operators who need agents to access AWS resources (via IRSA) or the Kubernetes API currently have to fork the chart to add ServiceAccount annotations and RBAC rules.

Closes: #397

Discord Discussion URL: https://discordapp.com/channels/1491295327620169908/1493841502529523732

## At a Glance

```text
┌─────────────────────────────────────┐
│ Current OpenAB Helm Chart           │
│ minimal install path                │
└──────────────┬──────────────────────┘
               │
               ▼
┌─────────────────────────────────────┐
│ Missing common chart hooks          │
│ - imagePullSecrets                  │
│ - probes / lifecycle                │
│ - initContainers / extraContainers  │
│ - extra volumes / mounts            │
│ - ServiceAccount + RBAC             │
│ - PodDisruptionBudget               │
│ - polymorphic env (valueFrom)       │
└──────────────┬──────────────────────┘
               │
               ▼
┌─────────────────────────────────────┐
│ This PR                             │
│ pod extensibility + SA/RBAC/PDB     │
│ + env + template fixes              │
└─────────────────────────────────────┘
```

## Prior Art & Industry Research

**OpenClaw:**

I reviewed the local OpenClaw repository and its Kubernetes deployment manifests. While OpenClaw does not currently ship a Helm chart in this repo, it does treat startup bootstrap as a first-class deployment concern. In particular, it uses an `initContainer` to prepare configuration and workspace state before the main container starts.

This suggests:

- `initContainers` are a reasonable place for startup preparation
[...]

PR #393, opened 2026-04-16
fix(helm): decouple [discord] section from botToken presence

### LINKED ISSUE #392: [Bug] [Helm] Helm chart 0.7.7-beta.1: [discord] section in config.toml depends on botToken being set at render time
### Summary
In the `0.7.7-beta.1` Helm chart, the `[discord]` block in the generated `config.toml` is conditionally rendered only if `($cfg.discord).botToken` is non-empty. This creates a "Chicken and Egg" problem for users who manage secrets securely (e.g., via AWS Secrets Manager injected into Kubernetes Secrets).

### Problem Description
To keep secrets secure, we inject the actual Discord Bot Token into an environment variable (e.g., `DISCORD_BOT_TOKEN`) from a Secret at runtime. We then use a placeholder `${DISCORD_BOT_TOKEN}` in the configuration.

However, in the current `configmap.yaml` template:
```yaml
{{- if and ($cfg.discord).enabled ($cfg.discord).botToken }}
[discord]
bot_token = "${DISCORD_BOT_TOKEN}"
...
```
The logic requires `botToken` to be provided to Helm during `helm install/upgrade` to even generate the `[discord]` section. If we leave it empty, the agent crashes because no adapter is configured. If we provide it, the sensitive token is stored in the Helm release history (metadata), which is a security concern.

### Suggested Fix
The generation of the `[discord]` (and `[slack]`) sections should be decoupled from the presence of the token at render time. It should be possible to enable the adapter and rely on runtime environment variable expansion for the token.

**Proposal:**
Change the condition in `templates/configmap.yaml` from:
`{{- if and ($cfg.discord).enabled ($cfg.discord).botToken }}`
to:
`{{- if ($cfg.discord).enabled }}`

This way, users can provide a dummy value or leave it empty in the Chart, while still having the `[discord]` block generated for runtime use.

### Environment
- **Chart Version**: 0.7.7-beta.1
- **Deployment**: EKS (Tokyo)
- **Secret Management**: AWS Secrets Manager -> K8s Secret -> Env Var

### Context
Discord discussion: https://discord.com/channels/1491295327620169908/1491365162869985283

Joseph19820124:
### Summary
This PR decouples the `[discord]` section generation in `config.toml` from the presence of `botToken` at render time. 

### Problem
Previously, the `[discord]` block would only be generated if a `botToken` was provided to Helm. This forced users who manage secrets at runtime (e.g., via AWS Secrets Manager) to either:
1. Provide a dummy token to Helm (confusing).
2. Provide the real token to Helm (security risk, as it's stored in Helm history).

### Solution
The template now only checks for `($cfg.discord).enabled`. This allows the block to be generated while relying on runtime environment variable expansion (`${DISCORD_BOT_TOKEN}`) for the actual authentication.

### Context
- Fixes #392
- Discord discussion: https://discord.com/channels/1491295327620169908/1491365162869985283
- Reference: #380

### DIFF
--- a/charts/openab/templates/configmap.yaml
+++ b/charts/openab/templates/configmap.yaml
@@ -10,6 +10,7 @@ metadata:
     {{- include "openab.labels" $d | nindent 4 }}
 data:
   config.toml: |
+    {{- if ($cfg.discord).enabled }}
     [discord]
     bot_token = "${DISCORD_BOT_TOKEN}"
     {{- range $cfg.discord.allowedChannels }}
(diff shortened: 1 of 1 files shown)


thepagent:
Thank you for the contribution and for identifying this issue! I've opened #403 which rebases this fix onto current `main` (the Slack section already exists upstream, and the configmap condition has evolved since this PR was authored). Closing in favor of #403.

Issue #382, opened 2026-04-16
[Tracking] 0.7.7-beta — Slack adapter stabilization


thepagent:
## Overview

Tracking issue for stabilizing the Slack adapter ([#259](https://github.com/openabdev/openab/pull/259)) before 0.7.7 GA release.

## Sub-issues

- [ ] #383 — Helm: `discord.enabled` migration issue on upgrade from 0.7.6
- [ ] #384 — Slack: `user_cache` unbounded memory growth
- [ ] #385 — Slack: `markdown_to_mrkdwn` control char assumption
- [ ] #386 — Slack: `get_socket_mode_url` creates new reqwest::Client per reconnect
- [ ] #387 — Discord: `OnceLock` adapter initialization in `ready()` instead of first message

## Done (in PR #259)

- [x] Tool display collapse regression — `compose_display` restored `streaming: bool` (`f641bec`)
- [x] Slack `handle_message` blocks event loop — `tokio::spawn` (`f641bec`)
- [x] Slack bot-loop protection — `is_bot_loop()` + `MAX_CONSECUTIVE_BOT_TURNS=10` (`4684c6a`)
- [x] Slack `already_reacted` / `no_reaction` noisy errors — silently ignored (`4684c6a`)

## Release criteria

All sub-issues resolved or triaged as won't-fix before tagging 0.7.7 GA.

PR #399, opened 2026-04-16
feat(slack): AllowUsers/AllowBots gating, DM support, participation cache

### LINKED ISSUE #396: Slack: bot does not respond to direct messages (message.im channel type)
## Summary

The Slack adapter does not respond to direct messages (DMs). When a user sends a message directly to the bot, the bot is silent.

## Root Cause

In `src/slack.rs`, the `"message"` event handler only processes messages that are already in a thread (`has_thread && ...`):

```rust
if has_thread && !is_bot && !skip_subtype && !mentions_bot {
    handle_message(...)
}
```

In DMs (`message.im`), the initial message has **no `thread_ts`**, so `has_thread` is `false` and the message is dropped silently. The `app_mention` branch also never fires in DMs because users cannot @mention the bot in a DM.

## Affected Event Types

- `message` with `channel_type: "im"` (DM channels, IDs start with `D`)

## Proposed Fix

Detect DM channels via `channel_type == "im"` (or channel ID prefix `D`) and allow those messages through regardless of `has_thread`:

```rust
let channel_type = event["channel_type"].as_str().unwrap_or("");
let channel_id = event["channel"].as_str().unwrap_or("");
let is_dm = channel_type == "im" || channel_id.starts_with('D');

if (has_thread || is_dm) && !is_bot && !skip_subtype && !mentions_bot {
    handle_message(...)
}
```

## Tracking

Part of #382

dogzzdogzz:

The Slack adapter has three gaps compared to Discord:

1. **Thread follow-up bug** — bot replies to *any* thread message, even threads it was never mentioned in ([reported in Discord](https://discord.com/channels/1491295327620169908/1491969620754567270/1494205732562800771))
2. **No bot-to-bot gating** — Discord has `allow_bot_messages` + `trusted_bot_ids`; Slack ignores all bot messages with no way to enable collaboration
3. **No DM support** — Slack DMs (`message.im`) are silently dropped because the event handler requires `thread_ts` (fixes #396)

Part of #382 (Slack adapter stabilization for 0.7.7 GA).

## Prior art

**Discord adapter** (`src/discord.rs`) already implements:
- `AllowBots` enum (Off/Mentions/All) with consecutive bot turn cap (fail-closed)
- `trusted_bot_ids` allowlist
- Thread-aware message gating (only respond in threads created by mention)

**OpenClaw** uses a similar "involved" concept — bots only respond in conversations they've been invited into.

**Hermes Agent** inspired the original `AllowBots` 3-value design (see `config.rs` doc comment).

## What's in this PR

| File | Purpose |
|---|---|
| `src/config.rs` | New `AllowUsers` enum (Involved/Mentions), `AllowBots` + `trusted_bot_ids` added to `SlackConfig` |
| `src/slack.rs` | Event handler rewrite: DM detection, AllowUsers gating, AllowBots gating, participation cache, `api_get()` for read methods, STT reaction |
| `src/main.rs` | Wire new config params to `run_slack_adapter()` |
| `charts/openab/templates/configmap.yaml` | Helm template for new Slack config options with validation |
| `charts/openab/values.yaml` | Default values for new options |
| `config.toml.example` | Document new Slack options |

### `AllowUsers` enum (Involved / Mentions)

| Mode | Channel/MPDM (non-thread) | Thread (bot NOT involved) | Thread (bot involved) |
|------|---------------------------|---------------------------|----------------------|
| `involved` (default) | @mention required | ignored | reply without @mention |
| `mentions` | @mention required | @mention required | @mention required |

- **DMs** (`D` prefix channel ID): always processed (implicit mention) — `AllowUsers` does not apply
- **"Involved" definition**: bot posted at least one message in the thread, OR thread parent @mentions the bot

### Participation cache

- **Positive-only**: only caches `true` (involved is irreversible — bot can't un-post)
- **`false` never cached**: next message re-queries API (bot may have replied since)
- **TTL = `session_ttl_hours`**: cache entries expire with ACP sessions
- **Bounded at 1000 entries**: evicts expired first, then oldest half

### Fail-closed on API errors

Consistent with Discord adapter (`discord.rs:182`: *"rejecting (fail-closed)"*). If `conversations.replies` fails, the message is dropped. User can always @mention to retry.

### `api_get()` for Slack read methods

`conversations.replies` is a GET method that does not accept JSON body POST. The previous `api_post()` call silently failed with `invalid_arguments`. Added `api_get()` with query parameters.

### `tokio::spawn` preserved

The old `feat/generalize-discord-features` branch removed `tokio::spawn` for `handle_message`, which would block the WebSocket read loop and cause Slack to disconnect on slow agent responses. This PR keeps `tokio::spawn`.

## Alternatives considered

1. **`AllowUsers::All`** (respond to all thread messages) — rejected as impractical; bot would reply in unrelated threads
2. **Fail-open for participation check** — rejected for consistency with Discord; silent message drops are preferable to replying in unrelated threads (user can @mention to retry)
3. **Caching negative results** — rejected because thread state changes with every message; stale `false` would permanently prevent bot from responding in a thread it later joins

## Test plan

- [ ] **Channel (involved mode)**: post without mention → no reply; @mention → reply in thread; thread follow-up without mention → reply
- [ ] **Channel (mentions mode)**: thread follow-up without mention → no reply even in bot-involved threads
- [ ] **DM**: send message without mention → reply; thread follow-up → reply
- [ ] **Bot gating**: set `allow_bot_messages = "all"` → bot responds to other bot messages; loop protection caps at 10 consecutive bot turns
- [ ] `cargo check` + `cargo clippy` — clean
- [ ] Helm: `helm template` renders new config options correctly

### DIFF
--- a/charts/openab/templates/configmap.yaml
+++ b/charts/openab/templates/configmap.yaml
@@ -60,6 +60,21 @@ data:
     {{- end }}
     {{- end }}
     allowed_users = {{ ($cfg.slack).allowedUsers | default list | toJson }}
+    {{- if ($cfg.slack).allowBotMessages }}
+    {{- if not (has ($cfg.slack).allowBotMessages (list "off" "mentions" "all")) }}
+    {{- fail (printf "agents.%s.slack.allowBotMessages must be one of: off, mentions, all — got: %s" $name ($cfg.slack).allowBotMessages) }}
+    {{- end }}
+    allow_bot_messages = "{{ ($cfg.slack).allowBotMessages }}"
+    {{- end }}
+    {{- if ($cfg.slack).trustedBotIds }}
+    trusted_bot_ids = {{ ($cfg.slack).trustedBotIds | toJson }}
+    {{- end }}
+    {{- if ($cfg.slack).allowUserMessages }}
+    {{- if not (has ($cfg.slack).allowUserMessages (list "involved" "mentions")) }}
+    {{- fail (printf "agents.%s.slack.allowUserMessages must be one of: involved, mentions — got: %s" $name ($cfg.slack).allowUserMessages) }}
+    {{- end }}
+    allow_user_messages = "{{ ($cfg.slack).allowUserMessages }}"
+    {{- end }}
     {{- end }}
 
     [agent]
--- a/charts/openab/values.yaml
+++ b/charts/openab/values.yaml
@@ -133,6 +133,14 @@ agents:
       appToken: ""       # App-Level Token (xapp-...) for Socket Mode
       allowedChannels: []  # empty = allow all channels
       allowedUsers: []     # empty = allow all users
+      # allowBotMessages: "off" (default) | "mentions" | "all"
+      allowBotMessages: "off"
+      # trustedBotIds: Bot User IDs (U...) — find via Slack UI: click bot profile → Copy member ID
+      trustedBotIds: []
+      # allowUserMessages: "involved" (default) | "mentions"
+      # "involved" = respond to thread follow-ups without @mention if bot has participated
+      # "mentions" = always require @mention
+      allowUserMessages: "involved"
     workingDir: /home/agent
     env: {}
     envFrom: []
--- a/config.toml.example
+++ b/config.toml.example
@@ -13,6 +13,11 @@ allowed_channels = ["1234567890"]       # empty or omitted = allow all channels
 # app_token = "${SLACK_APP_TOKEN}"     # App-Level Token (xapp-...) for Socket Mode
 # allowed_channels = ["C0123456789"]   # empty or omitted = allow all channels
 # allowed_users = ["U0123456789"]      # empty or omitted = allow all users
+# allow_bot_messages = "off"           # "off" (default) | "mentions" | "all"
+# trusted_bot_ids = []                 # empty = any bot (mode permitting); set to restrict
+# allow_user_messages = "involved"     # "involved" (default) | "mentions"
+                                        # "involved" = reply in threads bot has participated in
+                                        # "mentions" = always require @mention
 
 [agent]
 command = "kiro-cli"
--- a/src/config.rs
+++ b/src/config.rs
@@ -88,6 +88,30 @@ pub struct DiscordConfig {
     pub trusted_bot_ids: Vec<String>,
 }
 
+/// Controls whether the bot responds to user messages in threads without @mention.
+///
+/// - `Involved` (default): respond to thread messages only if the bot has participated
+///   in the thread (posted at least one message, or the thread parent @mentions the bot).
+///   Channel/MPDM messages always require @mention. DMs always process (implicit mention).
+/// - `Mentions`: always require @mention, even in threads the bot is participating in.
+#[derive(Debug, Clone, Copy, Default, PartialEq, Eq)]
+pub enum AllowUsers {
+    #[default]
+    Involved,
+    Mentions,
+}
+
+impl<'de> Deserialize<'de> for AllowUsers {
+    fn deserialize<D: serde::Deserializer<'de>>(deserializer: D) -> Result<Self, D::Error> {
+        let s = String::deserialize(deserializer)?;
+        match s.to_lowercase().as_str() {
+            "involved" => Ok(Self::Involved),
+            "mentions" => Ok(Self::Mentions),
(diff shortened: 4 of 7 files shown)


chaodu-agent:
## Community Triage Review

Nice work bringing Slack to feature parity with the Discord adapter. The `AllowUsers`/`AllowBots` gating, participation cache, and DM support all look well-structured. A few observations:

🟢 **INFO**

- `AllowBots` enum reused from Discord — consistent, no duplication.
- Fail-closed on participation check aligns with Discord (`discord.rs:182`).
- Positive-only participation cache is the right call — "involved" is irreversible.
- Keeping `tokio::spawn` for `handle_message` prevents blocking the WebSocket read loop.
- `api_get()` fixes the silent `conversations.replies` failure (POST → GET).
- Helm template validation with `fail` on invalid enum values is a nice DX touch.

🟡 **NIT**

- `bot_participated_in_thread()` uses `limit=200` — for very long threads the bot's participation could be missed. Might be worth a doc comment noting this boundary.
- `bot_id_cache` has no eviction. Practically fine (few bot IDs), but worth a comment explaining why unbounded is acceptable.
- `handle_message` and `run_slack_adapter` both suppress `clippy::too_many_arguments` now. A context struct could help long-term, but not a blocker for this PR.
- The 🎤 reaction on audio attachments when STT is disabled is a nice UX addition — worth mentioning in the PR description.

🔴 **SUGGESTED CHANGES**

- **Duplicate API calls per thread**: In `AllowBots::All` mode, each bot message calls `conversations.replies` for loop protection. If the same thread also triggers `bot_participated_in_thread()`, that's two API calls for the same thread. It might be worth considering merging these queries or having the loop protection call update the participation cache as a side effect.

- **Inconsistent fail behavior on API errors**: The participation check is fail-closed (message dropped on error), but the loop protection in `AllowBots::All` is fail-open (`if let Ok(resp)` — API failure skips the check and processes the message). During an API outage this could allow a runaway bot loop. Might be worth aligning both to fail-closed for consistency.

- **Missing `inclusive: true` in loop protection query**: `bot_participated_in_thread()` passes `("inclusive", "true")` to `conversations.replies`, but the loop protection call in `AllowBots::All` does not. This could cause the parent message to be excluded from the consecutive-bot-turn count. Suggestion: add `("inclusive", "true")` there as well.

- **DM + `app_mention` edge case**: The code correctly routes DM mentions through the `message` event (`if mentions_bot && !is_dm { continue; }`). It might be worth confirming that Slack never fires `app_mention` in DMs, or adding a DM guard in the `app_mention` handler to be safe.

dogzzdogzz:
Thanks for the thorough review! Addressed the suggested changes in `9fae5e2`:

### 🔴 SUGGESTED CHANGES — resolution

| # | Feedback | Action | Detail |
|---|----------|--------|--------|
| 1 | Duplicate API calls per thread | **No change needed** | `AllowBots` gating and `AllowUsers` gating are mutually exclusive paths (`is_bot` vs `!is_bot`), so the same event never triggers both `conversations.replies` calls. Added no code change, but happy to add a comment if helpful. |
| 2 | Loop protection is fail-open | **Fixed** ✅ | Changed `if let Ok(resp)` to `match` with explicit `Err` arm that logs and rejects (fail-closed). Consistent with Discord adapter (`discord.rs:182`) and `bot_participated_in_thread()`. |
| 3 | Missing `inclusive: true` in loop protection | **Fixed** ✅ | Added `("inclusive", "true")` to the loop protection `conversations.replies` query, matching `bot_participated_in_thread()`. |
| 4 | DM + `app_mention` edge case | **No change** | Slack does not fire `app_mention` in DMs per their documentation. The current `message` handler correctly routes DM mentions. Could add a defensive guard later if needed, but keeping it simple for now. |

chaodu-agent (approved):
Re-reviewed after `9fae5e2`. All 🔴 items resolved:

- **#2 fail-closed** ✅ — loop protection now rejects on API error, consistent with Discord adapter and `bot_participated_in_thread()`.
- **#3 inclusive param** ✅ — added to loop protection query.
- **#1 duplicate API calls** — confirmed `is_bot` / `!is_bot` are mutually exclusive paths; no duplicate calls. Original concern was incorrect.
- **#4 DM + app_mention** — Slack docs confirm `app_mention` does not fire in DMs. No guard needed.

No remaining blockers from triage. LGTM 🚀

thepagent (approved):

PR #461, opened 2026-04-19
feat(discord): preserve user mention UIDs for LLM mention-back

### LINKED ISSUE #460: feat(discord): preserve UID in resolved mentions as <@UID|Name>
### Description

`resolve_mentions()` currently replaces `<@UID>` with `@DisplayName`, discarding the UID. Change it to emit `<@UID|DisplayName>` instead, so the LLM sees a human-readable name while the UID is preserved for mention-back in replies.

**Inbound** (Discord → ACP): resolve other-user mentions to `<@UID|DisplayName>`:

```rust
// current (line 491-495)
let display = format!("@{}", label);

// proposed
let display = format!("<@{}|{}>", user.id, label);
```

**Outbound** (ACP → Discord): strip the `|Name` suffix before sending to Discord, since Discord does not support Slack-style `<@UID|Name>` syntax — it only renders `<@UID>`:

```rust
// before sending reply to Discord
let out = regex::Regex::new(r"<@(\d+)\|[^>]+>")
    .unwrap()
    .replace_all(&reply, "<@$1>");
```

https://github.com/openabdev/openab/blob/6887db629f492651fdbf528c6deb5c4761c25974/src/discord.rs#L481-L501

### Investigation: How Other Projects Handle This

We investigated how OpenClaw and Hermes handle Discord mentions:

| | Bot own mention | Other user mentions |
|---|---|---|
| **OpenAB** (current) | Stripped | `<@UID>` → `@DisplayName` (UID lost) |
| **Hermes** ([NousResearch/hermes-agent](https://github.com/NousResearch/hermes-agent)) | Stripped | Passed through as raw `<@UID>` — LLM sees Discord syntax but no human-readable name |
| **OpenClaw** ([openclaw/openclaw](https://github.com/openclaw/openclaw)) | Stripped via `stripPatterns: ["<@!?\\d+>"]` | All mentions stripped — LLM sees neither UID nor name |

- **Hermes** (`gateway/platforms/discord.py` line ~2992): only strips the bot own mention, leaves other `<@UID>` raw in `message.content` passed to the agent.
- **OpenClaw** (`extensions/discord/src/channel.ts`): declares `stripPatterns: () => ["<@!?\\d+>"]` which removes all user mentions before the message reaches the LLM. Outbound mentions are reconstructed via a directory lookup (`mentions.ts` → `formatMention()`).

Neither project preserves both UID and display name for the agent. The `<@UID|DisplayName>` format (borrowed from Slack native syntax) gives the LLM human-readable context while keeping the UID available for mention-back.

### Use Case

When a user mentions another user (e.g. `@AgentDealer`) in a message to the bot, the agent receives `@AgentDealer` but has no way to mention that user back in its reply because the UID is lost. Preserving both UID and display name enables the agent to reference users in responses that Discord can render as proper mentions.

The outbound regex cleanup ensures Discord renders `<@UID>` as a clickable mention rather than showing the raw `<@UID|Name>` as literal text.

chaodu-agent:
## Summary

Closes #460

Simplify `resolve_mentions()` to keep other-user `<@UID>` mentions as-is instead of resolving them to `@DisplayName` (which discards the UID). This lets the LLM copy `<@UID>` into its reply for correct Discord mention rendering — no outbound transformation needed.

## Changes (`src/discord.rs`, +6/-24)

- **Remove step 2** — no longer resolves `<@UID>` → `@DisplayName`. User mentions pass through as raw `<@UID>`.
- **Remove `USER_MENTION_RE`** — no longer needed since user mentions are preserved.
- **Remove `mentions` parameter** — `resolve_mentions` no longer needs the serenity `User` list.
- **Call unconditionally** — not just when `is_mentioned`, so thread follow-ups also get bot-mention stripping.

### What stays the same

- Bot own mention: stripped
- Role mentions: `@(role)` fallback
- `send_message` / outbound: untouched

## Why preserve raw `<@UID>` instead of `<@UID|DisplayName>`

We explored `<@UID|DisplayName>` (Slack-style) but LLMs don't reliably preserve token formats — the model might respond with `@Name`, `<@Name>`, or `<@UID|Name>`. Only raw `<@UID>` is safe:

- LLM copies `<@UID>` → Discord renders clickable mention ✅
- LLM writes `@Name` → plain text, no worse than before ⚠️
- No outbound regex needed — zero risk of mangling

## Recommended usage: ice-breaking in multi-bot chatrooms

Since `<@UID>` is now preserved, bots need a UID→name mapping to know who is who. The recommended approach is to send an ice-breaking greeting at the start of a multi-bot chatroom (via system prompt or context entry):

```
We have 3 people in this room:

ME @MY_NAME
BOT_A_NAME @BOT_A
BOT_B_NAME @BOT_B

Make sure to use <@UID> to mention someone in the messages.
```

This lets each bot build the UID→name mapping in its own context from the very beginning and learn how to mention others correctly.

## Design notes

**Slack isolation:** Slack has its own independent mention handling in `src/slack.rs`. Not touched.

**Thread title:** `shorten_thread_name()` will show raw `<@UID>` in thread titles. Pre-existing gap (old code stripped mentions entirely), can be addressed separately.

DC: https://discord.com/channels/1491295327620169908/1494739741600387204/1495368292146217071

### DIFF
--- a/docs/discord.md
+++ b/docs/discord.md
@@ -127,6 +127,16 @@ Empty (default) = any bot can pass through (subject to the mode check).
 
 Role mentions are ignored because they are shared across bots and cause false positives in multi-bot setups. This is intentional since v0.7.8-beta.3 (#420, #440).
 
+### User mention UIDs
+
+When a user mentions another user (e.g. `@SomeUser`) in a message to the bot, the raw Discord mention `<@UID>` is preserved in the prompt sent to the LLM. This means:
+
+- The LLM can copy `<@UID>` into its reply to produce a clickable Discord mention
+- The bot's own mention is stripped (so the bot doesn't see itself being triggered)
+- Role mentions are replaced with `@(role)` placeholder
+
+To help the LLM know who each UID refers to, provide a UID→name mapping via system prompt or context entry (see [Multi-Bot Setup](#multi-bot-setup) below).
+
 ---
 
 ## Thread Behavior
@@ -171,6 +181,22 @@ To enable bots to collaborate (e.g. code review → deploy handoff):
 [discord]
 allow_bot_messages = "mentions"
 ```
+
+### Ice-breaking: teaching bots who's in the room
+
+Since user mentions are preserved as raw `<@UID>`, bots need a UID→name mapping to know who is who. Add an ice-breaking greeting to each bot's system prompt or context entry:
+
+```
+We have 3 participants in this room:
+
+MY_NICIKNAME    <@MY_NAME>
+BOT1_NICKNAME   <@BOT1>
+BOT2_NICKNAME   <@BOT2>
+
+Always use <@UID> format to mention someone in your messages.
+```
+
+This lets each bot build the mapping in its own context from the start and correctly mention others using `<@UID>`.
 
 See [multi-agent.md](multi-agent.md) for detailed examples.
 
--- a/src/discord.rs
+++ b/src/discord.rs
@@ -10,7 +10,6 @@ use serenity::http::Http;
 use serenity::model::channel::{AutoArchiveDuration, Message, ReactionType};
 use serenity::model::gateway::Ready;
 use serenity::model::id::{ChannelId, MessageId, UserId};
-use serenity::model::user::User;
 use serenity::prelude::*;
 use std::collections::{HashMap, HashSet};
 use std::sync::{Arc, OnceLock};
@@ -321,11 +320,7 @@ impl EventHandler for Handler {
             return;
         }
 
-        let prompt = if is_mentioned {
-            resolve_mentions(&msg.content, bot_id, &msg.mentions)
-        } else {
-            msg.content.trim().to_string()
-        };
+        let prompt = resolve_mentions(&msg.content, bot_id);
 
         // No text and no attachments → skip
         if prompt.is_empty() && msg.attachments.is_empty() {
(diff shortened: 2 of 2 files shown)


thepagent (approved):
LGTM — clean implementation.

- Inbound `resolve_mentions()` now runs unconditionally (good: thread follow-ups also get enrichment)
- Outbound `strip_mention_names()` correctly applied in both streaming and final send paths
- Regex safety verified: `USER_MENTION_RE` won't clobber resolved `<@UID|Name>` mentions
- No Slack impact (separate code paths)
- Empirically confirmed Discord does not render `<@UID|Name>` (see #460 comment)

Minor note: for thread follow-ups without explicit @mention, `msg.mentions` may be sparse — unresolved UIDs fall through to `@(user)` which is acceptable.

thepagent (changes requested):
The `docs/discord.md` ice-breaking example uses real participant names and UIDs. Please replace with mock/generic names, e.g.:

```
We have 3 participants in this room:

ME      MyBot        <@1111111111111111111>
BOT     OtherBot     <@2222222222222222222>
HUMAN   SomeUser     <@3333333333333333333>

Always use <@UID> format to mention someone in your messages.
```

Same for the earlier section that references `@Pahud` — use a generic name like `@SomeUser` instead.

thepagent (approved):

PR #470, opened 2026-04-19
fix(acp): close notify channel on EOF to prevent stream hang

### LINKED ISSUE #295: bug: with_connection global write lock + unclosed notify channel freezes bot when ACP process goes stale
### Description

When an ACP child process becomes unresponsive (stale session, auth expired, CLI hung), the bot freezes completely — it continues posting `...` placeholders for every incoming mention but never processes any prompt. This affects **all threads**, not just the one with the stale session. Confirmed on v0.7.2 (latest stable) — the affected code paths are unchanged.

**Two bugs combine:**

**Bug 1: `with_connection` holds global write lock during streaming (`pool.rs:72-80`)**

```rust
pub async fn with_connection<F, R>(&self, thread_id: &str, f: F) -> Result<R> {
    let mut conns = self.connections.write().await;  // global write lock
    let conn = conns.get_mut(thread_id)...;
    f(conn).await  // held for ENTIRE duration of stream_prompt (minutes/hours)
}
```

`stream_prompt` runs inside `f`, so the `RwLock` write guard on `self.connections` is held for the entire streaming duration. While held, **all** other `get_or_create` (read lock), `with_connection` (write lock), and `cleanup_idle` (write lock) calls block. Tokio's `RwLock` blocks new readers when a writer is waiting, so even the read-lock fast path in `get_or_create` is blocked.

**Bug 2: `rx.recv()` hangs forever when ACP process dies (`connection.rs:152-153`)**

When the reader task hits EOF (child process died), it does not close the notify channel:

```rust
let sub = notify_tx.lock().await;
drop(sub);  // drops MutexGuard, NOT the Option<Sender> inside
```

The `mpsc::UnboundedSender` remains alive in `Arc<Mutex<Option<Sender>>>`, so `rx.recv()` in `stream_prompt` (`discord.rs:269`) **never returns `None`** — it waits forever.

**Combined effect:**
1. ACP process becomes stale (CLI session expired, API auth timeout, OOM, etc.)
2. `stream_prompt` sends prompt via `session_prompt` → `rx.recv()` hangs forever (sender not closed)
3. Write lock on `pool.connections` held forever
4. All subsequent `get_or_create` calls block (read lock blocked by held write lock)
5. `cleanup_idle` also blocks (needs write lock) — can't clean up the stale session
6. Bot posts `...` for every incoming mention (before pool access) but never updates them
7. **Entire bot is frozen** — not just one thread, ALL threads

**Suggested fixes:**
- **Fix 1 (critical):** Close notify channel on EOF: `*sub = None;` instead of `drop(sub);`
- **Fix 2 (critical):** Don't hold global lock during streaming — use per-connection locks or extract connection from pool during streaming
- **Fix 3 (recommended):** Add timeout to `rx.recv()` in streaming loop
- **Fix 4 (defense-in-depth):** Improve `alive()` to also check child process status

Analysis validated independently by Kiro (kiro-cli) and Codex (codex-acp).

### Steps to Reproduce

1. Deploy openab with any ACP agent (claude-agent-acp, gemini --acp, kiro-cli acp, codex-acp)
2. Send a mention that triggers an ACP session — bot processes it normally
3. Wait for the ACP child process to become stale (e.g., CLI session expires overnight, API auth times out, or the process hangs)
4. Send another mention to the same bot
5. Bot posts `...` placeholder but never updates it — `stream_prompt` hangs on `rx.recv()` holding the pool write lock
6. Send mentions targeting different threads — ALL are also frozen because the global write lock is held

### Expected Behavior

- When an ACP process dies or becomes unresponsive, the notify channel should close (`rx.recv()` returns `None`), the error is surfaced to the user (e.g., `⚠️ Failed`), and the stale session is cleaned up
- Other threads should not be affected by one stale session — the pool lock should not be held during streaming

### Environment

- openab version: 0.6.0 on our VPS, but **confirmed unchanged in 0.7.2** (`pool.rs` never modified since initial commit, `connection.rs` EOF cleanup unchanged)
- Helm chart: 0.6.0 (latest stable: 0.7.3-beta.56)
- ACP agents: claude-agent-acp, gemini --acp, kiro-cli acp, codex-acp
- K3s on Zeabur VPS (2 vCPU, 4GB RAM)

### Screenshots / Logs

**Logs** — after freeze, only `accepted bot message` appears. No `spawning agent`, no `pool error`, no errors at all:
```
[dispatcher] Dispatched #192 (code-review) → itachi (prompt delivered)
# itachi log:
INFO openab::discord: accepted bot message (in allowed_bots_from) bot_id=1490975142803669113
# ... nothing else. No spawning, no errors. Repeated every 5 minutes for 10+ hours.
```

**Discord thread** — 25 consecutive `...` messages from bots, none ever updated:
```
[11:24:55] 千手扉間 (Dispatcher): <@itachi> (phase: code-review) — please pick up #192
[11:24:56] 宇智波鼬: ...
[11:29:39] 千手扉間 (Dispatcher): <@itachi> (phase: code-review) — please pick up #192
[11:29:40] 宇智波鼬: ...
# ... repeated 25 times, none updated beyond "..."
```

**Scale**: 4 agents frozen, 432+ mentions over ~10 hours, zero processed.

masami-agent:
## Summary

Closes #295

When the ACP child process dies (EOF on stdout), the reader task in `connection.rs` fails to close the notify channel. This causes `rx.recv()` in the streaming loop to hang forever, holding the per-connection mutex and permanently leaking the pool slot.

## Changes

| File | Change |
|------|--------|
| `src/acp/connection.rs` | Fix EOF handler: `*sub = None` instead of `drop(sub)` — drops the `UnboundedSender`, closing the channel |
| `src/adapter.rs` | Add 10-minute timeout to `rx.recv()` as defense-in-depth |

## Root Cause

In the reader task EOF handler:

```rust
// Before (bug)
let sub = notify_tx.lock().await;
drop(sub);  // drops MutexGuard, NOT the Option<Sender> inside
```

`drop(sub)` only drops the `MutexGuard<Option<UnboundedSender>>`. The `Option` remains `Some(sender)`, so the `UnboundedSender` is never dropped and `rx.recv()` never returns `None`.

```rust
// After (fix)
let mut sub = notify_tx.lock().await;
*sub = None;  // drops the Sender → channel closes → rx.recv() returns None
```

## Defense-in-Depth

Added a 10-minute timeout around `rx.recv()` in the streaming loop (`adapter.rs`). If no notification arrives within 10 minutes, the loop breaks with an "Agent stopped responding" error. This prevents indefinite hangs even if other channel-closing bugs exist.

## Scope Note

Issue #295 also describes a global write lock problem in `with_connection`. That was already fixed in a prior refactor — `with_connection` now uses a read lock + per-connection mutex. See [validation comment](https://github.com/openabdev/openab/issues/295#issuecomment-4276043728) for details. This PR only addresses the remaining confirmed bug (notify channel EOF).

## Validation

- Code-level verification: `*sub = None` drops the `UnboundedSender`, which is the standard tokio `mpsc` channel close mechanism
- `tokio::time::timeout` wrapping `rx.recv()` is a well-established pattern for stall detection

### DIFF
--- a/src/acp/connection.rs
+++ b/src/acp/connection.rs
@@ -253,9 +253,9 @@ impl AcpConnection {
                         params: None,
                     });
                 }
-                // Signal subscriber
-                let sub = notify_tx.lock().await;
-                drop(sub);
+                // Close the notify channel so rx.recv() returns None
+                let mut sub = notify_tx.lock().await;
+                *sub = None;
             })
         };
 
--- a/src/adapter.rs
+++ b/src/adapter.rs
@@ -273,7 +273,16 @@ impl AdapterRouter {
 
                     // Process ACP notifications
                     let mut response_error: Option<String> = None;
-                    while let Some(notification) = rx.recv().await {
+                    let recv_timeout = std::time::Duration::from_secs(600);
+                    loop {
+                        let notification = match tokio::time::timeout(recv_timeout, rx.recv()).await {
+                            Ok(Some(n)) => n,
+                            Ok(None) => break, // channel closed
+                            Err(_) => {
+                                response_error = Some("Agent stopped responding".into());
+                                break;
+                            }
+                        };
                         if notification.id.is_some() {
                             if let Some(ref err) = notification.error {
                                 response_error = Some(format_coded_error(err.code, &err.message));


obrutjack (approved):
Merge checklist verified:

- ✅ CI all green (cargo check + 7 Docker smoke tests)
- ✅ MERGEABLE, upstream branch
- ✅ Fixes real bug: EOF didn't close notify channel → stream_prompt hangs forever
- ✅ 10min timeout as safety net for unresponsive agents
- ✅ Minimal change (+13/-4), correct logic
- ✅ No version regression

Pending @thepagent code owner review.

PR #480, opened 2026-04-20
feat(slack): align with Discord — shared BotTurnTracker + MultibotMentions

dogzzdogzz:
## Summary

- **Shared \`BotTurnTracker\`**: Move \`BotTurnTracker\`, \`TurnResult\`, and \`HARD_BOT_TURN_LIMIT\` to \`adapter.rs\` so both Discord and Slack use the same implementation
- **Slack bot turn limits**: Replace the old API-based consecutive cap with the soft + hard limit system (matching Discord). Adds \`max_bot_turns\` to \`SlackConfig\` (default 20)
- **Slack \`MultibotMentions\`**: Previously a no-op (identical to \`Involved\`). Now fully implemented:
  - \`multibot_threads\` cache on \`SlackAdapter\` for O(1) multibot detection
  - Early detection: any bot message in a thread immediately updates the cache
  - \`bot_participated_in_thread()\` returns \`(involved, other_bot_present)\`
  - When other bots detected + no @mention → reject (same as Discord)

## Discord link
https://discord.com/channels/1491295327620169908/1494739741600387204/1495409492089769984

## Test plan
- [ ] \`cargo clippy -- -D warnings\` passes
- [ ] \`cargo test\` — 51 tests pass
- [ ] Manual: Slack thread with another bot, \`MultibotMentions\` mode requires @mention
- [ ] Manual: \`Involved\` mode does NOT require @mention in same scenario
- [ ] Manual: Bot turn soft limit sends warning message; hard limit is silent
- [ ] Manual: Human reply resets both counters

### DIFF
--- a/src/adapter.rs
+++ b/src/adapter.rs
@@ -18,9 +18,16 @@ pub const HARD_BOT_TURN_LIMIT: u32 = 100;
 
 #[derive(Debug, PartialEq, Eq)]
 pub enum TurnResult {
+    /// Counter below limits — continue normally.
     Ok,
+    /// Counter == soft_limit — warn once, then stop.
     SoftLimit(u32),
+    /// Counter > soft_limit — silently stop (already warned).
+    Throttled,
+    /// Counter == HARD_BOT_TURN_LIMIT — warn once, then stop.
     HardLimit,
+    /// Counter > HARD_BOT_TURN_LIMIT — silently stop (already warned).
+    Stopped,
 }
 
 pub struct BotTurnTracker {
@@ -37,9 +44,13 @@ impl BotTurnTracker {
         let (soft, hard) = self.counts.entry(thread_id.to_string()).or_insert((0, 0));
         *soft += 1;
         *hard += 1;
-        if *hard >= HARD_BOT_TURN_LIMIT {
+        if *hard > HARD_BOT_TURN_LIMIT {
+            TurnResult::Stopped
+        } else if *hard == HARD_BOT_TURN_LIMIT {
             TurnResult::HardLimit
-        } else if *soft >= self.soft_limit {
+        } else if *soft > self.soft_limit {
+            TurnResult::Throttled
+        } else if *soft == self.soft_limit {
             TurnResult::SoftLimit(*soft)
         } else {
             TurnResult::Ok
--- a/src/adapter.rs
+++ b/src/adapter.rs
@@ -1,6 +1,7 @@
 use anyhow::Result;
 use async_trait::async_trait;
 use serde::Serialize;
+use std::collections::HashMap;
 use std::sync::Arc;
 use tracing::error;
 
@@ -9,6 +10,49 @@ use crate::config::ReactionsConfig;
 use crate::error_display::{format_coded_error, format_user_error};
 use crate::format;
 use crate::reactions::StatusReactionController;
+
+// --- Bot turn tracking ---
+
+/// Absolute per-thread cap on bot turns. Cannot be overridden by config or human intervention.
+pub const HARD_BOT_TURN_LIMIT: u32 = 100;
+
+#[derive(Debug, PartialEq, Eq)]
+pub enum TurnResult {
+    Ok,
+    SoftLimit(u32),
+    HardLimit,
+}
+
+pub struct BotTurnTracker {
+    soft_limit: u32,
+    counts: HashMap<String, (u32, u32)>,
+}
+
+impl BotTurnTracker {
+    pub fn new(soft_limit: u32) -> Self {
+        Self { soft_limit, counts: HashMap::new() }
+    }
+
+    pub fn on_bot_message(&mut self, thread_id: &str) -> TurnResult {
+        let (soft, hard) = self.counts.entry(thread_id.to_string()).or_insert((0, 0));
+        *soft += 1;
+        *hard += 1;
+        if *hard >= HARD_BOT_TURN_LIMIT {
+            TurnResult::HardLimit
+        } else if *soft >= self.soft_limit {
+            TurnResult::SoftLimit(*soft)
+        } else {
+            TurnResult::Ok
+        }
+    }
+
+    pub fn on_human_message(&mut self, thread_id: &str) {
+        if let Some((soft, hard)) = self.counts.get_mut(thread_id) {
+            *soft = 0;
+            *hard = 0;
+        }
+    }
+}
 
 // --- Platform-agnostic types ---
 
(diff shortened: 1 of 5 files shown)


shaun-agent:
## OpenAB PR Screening
- Title: feat(slack): align with Discord — shared BotTurnTracker + MultibotMentions
- Source: https://github.com/openabdev/openab/pull/480
- Generated at: 2026-04-20T05:01:43.291Z
- Discord thread: https://discord.com/channels/1488041051187974246/1495650449750233180
<details>
## Intent

PR #480 is trying to remove behavior drift between the Discord and Slack adapters for two concrete operator-visible problems:

1. Slack currently handles bot turn limits differently from Discord, which means threaded conversations can terminate or continue under different rules depending on platform.
2. Slack’s `MultibotMentions` mode is effectively not enforcing multibot mention discipline, so in threads where multiple bots participate, Slack can respond without an explicit mention even though Discord would reject that turn.

The practical goal is cross-platform parity: shared bot-turn tracking logic, the same soft/hard limit semantics, and Slack multibot mention handling that matches Discord’s current behavior.

## Feat

This is primarily a feature alignment PR with some cleanup/refactor characteristics.

In plain language, it does three things:

- Moves `BotTurnTracker`, `TurnResult`, and `HARD_BOT_TURN_LIMIT` into shared adapter-level code so both Discord and Slack use the same turn-limiting implementation.
- Replaces Slack’s older consecutive-turn cap with the same soft-warning plus hard-stop model used by Discord, adding `max_bot_turns` to `SlackConfig` with a default of `20`.
- Implements real `MultibotMentions` handling for Slack by detecting when another bot has participated in a thread and requiring an explicit `@mention` before OpenAB responds in that mode.

Type-wise, this is best categorized as a feature parity change plus a small refactor.

## Who It Serves

The primary beneficiaries are maintainers and runtime operators who need Discord and Slack behavior to be predictable and supportable as one product, not two divergent adapters.

Secondarily, Slack end users benefit because thread reply behavior becomes more intentional in multibot channels, reducing accidental bot pile-ons and making mention rules easier to understand.

## Rewritten Prompt

Unify Slack and Discord bot-turn behavior by moving `BotTurnTracker`, `TurnResult`, and the hard-turn limit into shared adapter code, then switch Slack to the same soft-warning and hard-stop turn policy already used by Discord. Implement Slack `MultibotMentions` so it detects other bot participation per thread, rejects unmentioned replies only when another bot is present, and preserves `Involved` behavior without mention enforcement. Add `max_bot_turns` to `SlackConfig` with a sane default, keep the logic testable at the adapter boundary, and cover config defaults, human-reset behavior, soft-limit warnings, hard-limit silence, and Slack multibot mention gating.

## Merge Pitch

This item is worth advancing because it reduces cross-platform behavioral drift in two areas that are easy to misoperate: runaway bot exchanges and ambiguous multibot thread routing. The upside is immediate: fewer adapter-specific surprises, a clearer mental model for operators, and less duplicated logic to maintain.

The risk profile is moderate. The likely reviewer concern is false positives or stale state in Slack multibot detection, especially if thread participation caching is updated too aggressively or not reset correctly. The other concern is whether “Discord parity” is being used as a proxy for correctness without enough Slack-specific tests.

## Best-Practice Comparison

Against OpenClaw, the relevant principle is not scheduling but routing and execution control. This PR aligns with OpenClaw-style explicit delivery routing by making Slack refuse turns in multibot threads unless explicitly addressed. The shared `BotTurnTracker` also moves toward a more consistent execution policy across adapters. Durable job persistence, isolated executions, and retry/run-log concerns are mostly not relevant here because this PR is about inbound thread eligibility and turn caps, not background job lifecycle.

Against Hermes Agent, several principles are relevant at a smaller scale. The strongest fit is “self-contained prompts for scheduled tasks” only in spirit: each adapter should make a deterministic respond/do-not-respond decision from local thread state rather than hidden platform quirks. Atomic writes and file locking are not relevant unless the Slack multibot cache becomes persisted or shared across processes, which does not appear to be the case here. The more relevant Hermes lesson is fresh, explicit state boundaries: if Slack caches bot participation, that cache needs clear update/reset semantics so decision-making stays deterministic and safe under concurrent event handling.

Overall, the best-practice signal is:

- Relevant: explicit routing rules, consistent execution policy, deterministic turn gating
- Partially relevant: state isolation and freshness
- Not especially relevant: scheduling ticks, persistent job stores, file locks, retry backoff infrastructure

## Implementation Options

### Option 1: Conservative parity patch

Keep the current PR shape: centralize `BotTurnTracker` in shared adapter code, add `max_bot_turns` to Slack config, and implement Slack multibot detection with an in-memory per-thread cache.

This is the fastest path and gets Discord/Slack semantics aligned without changing broader architecture. The downside is that Slack-specific thread state remains adapter-local and may be harder to reason about if event ordering or cache invalidation gets more complex later.

### Option 2: Balanced adapter-policy extraction

Extract a shared “reply eligibility” policy layer that owns turn-limit evaluation and multibot mention decisions, while keeping platform-specific data collection in each adapter. Discord and Slack would both feed normalized thread facts into the same policy functions.

This is a better long-term shape because it separates “what facts did Slack observe?” from “given these facts, should the bot reply?”. It increases up-front work a bit, but it makes parity easier to test and reduces future drift.

### Option 3: Ambitious conversation-state subsystem

Introduce a more formal conversation/thread state model shared across adapters, with explicit participant tracking, turn counters, reset semantics, and observability hooks for why a turn was accepted or rejected.

This would produce the cleanest long-term behavior and make future platform support easier, but it is probably too large for the immediate need. It raises scope, review burden, and the chance of merge delay.

## Comparison Table

|---|---|---|---|---|---|---|
| Conservative parity patch | High | Low | Medium | Medium | High | High |
| Balanced adapter-policy extraction | Medium | Medium | High | High | High | Very high |
| Ambitious conversation-state subsystem | Low | High | High if completed well | Very high | Medium near-term, high long-term | Medium |

## Recommendation

Recommend the balanced path: merge toward shared policy, not just shared structs.

Concretely, the next agent should preserve the current goal of shipping parity now, but frame the implementation around a small shared decision layer for turn limits and multibot gating. That gives OpenAB a mergeable step that is still compact, while avoiding a future where Slack and Discord once again share names but not decision logic.

If scope needs to stay tight, split sequencing like this:

1. Land shared bot-turn policy and Slack config parity.
2. Land Slack `MultibotMentions` with focused tests around thread facts and reply eligibility.
3. Follow up later only if needed with better thread-state observability or cache hardening.

That sequence keeps the PR reviewable, improves correctness immediately, and sets Masami or Pahud up for a cleaner merge discussion.
</details>

Issue #21, opened 2026-04-04
ci: deduplicate Rust build stage across Dockerfiles


thepagent:
## Problem

The Rust builder stage is duplicated across `Dockerfile`, `Dockerfile.codex`, and `Dockerfile.claude`. CI builds the same binary 6 times (3 variants × 2 platforms).

## Proposed solution

Build the `agent-broker` binary once per platform, then share it across variant images. Options:

1. **Separate build job** — build binary as artifact, `COPY` into each variant Dockerfile
2. **Shared base image** — `Dockerfile.builder` pushed to GHCR, variants use `COPY --from=ghcr.io/.../agent-broker-builder`
3. **Multi-stage with `--target`** — single Dockerfile with build args for the runtime stage

GHA cache mitigates this for now, but deduplication would reduce CI time and improve maintainability.

Follow-up from #20.

chaodu-agent:
Makes sense to deduplicate the build. A few questions to help scope this:

1. **How long does CI take currently?** What's the total build time with the duplicated Rust stage across 3 variants × 2 platforms?
2. **How much time would deduplication save?** Rough estimate is fine.
3. **Which approach do you prefer?** The issue lists 3 options — do you have a preference based on your CI experience?

If the current build time is tolerable with GHA cache, this might be lower priority than other open items.

CHC-Agent:
Hi @thepagent — this issue has been waiting for additional information for over 30 days. Is this still something you'd like to pursue?

If there's no update within 7 days, this issue will be closed. Feel free to reopen when ready.

PR #484, opened 2026-04-20
fix(discord): move bot turn tracker before gating (#483)

### LINKED ISSUE #483: fix(discord): bot turn limit never triggers in multi-bot ping-pong
### Description

Bot turn limits (soft=20, hard=100) never trigger when two bots ping-pong in a thread because the tracker runs **after** bot message gating and self-message filtering.

```
Each bot's event handler:

  Broker's own reply → line 232: self-check → return (tracker never called)
  Dealer's message   → line 248: bot gating → return (tracker never called)
  Human's message    → passes gating → tracker.on_human_message() → resets to 0

  Result: Broker's tracker counter is ALWAYS 0
          Dealer's tracker counter is ALWAYS 0
          Two bots can ping-pong infinitely
```

### Steps to Reproduce

1. Set `allow_bot_messages = "mentions"` on both bots
2. Human starts a thread, both bots join
3. Bots mention each other and start ping-ponging
4. Observe: no soft limit warning at 20 turns, no hard limit at 100

### Expected Behavior

Bot turn limit should trigger after 20 consecutive bot turns (soft) or 100 total (hard), regardless of which bot sent the message.

### Fix

Move the bot turn tracker next to the early multibot detection — before bot gating and self-check — so ALL bot messages (including filtered ones) count toward the limit. Same pattern as #481.

thepagent:
Closes #483

## The Bug

Bot turn limits (soft=20, hard=100) never trigger in multi-bot ping-pong because the tracker runs after both the self-check and bot message gating.

```
AgentBroker receives message
   │
   ▼
┌─────────────────────────────┐
│ Self-check                  │
│ own reply → return          │  ← not counted
└──────────────┬──────────────┘
               ▼
┌─────────────────────────────┐
│ Bot gating                  │
│ other bot msg → return      │  ← not counted
└──────────────┬──────────────┘
               ▼
┌─────────────────────────────┐
│ Tracker                     │
│ only human msgs reach here  │
│ → counter always 0          │
└─────────────────────────────┘
```

## The Fix

Move tracker before self-check and gating. Count ALL bot messages including own.

```
AgentBroker receives message
   │
   ▼
┌─────────────────────────────┐
│ Multibot detection          │
│ Bot turn tracker    ← NEW   │
│                             │
│ Own reply?    → +1 counted  │
│ Other bot?    → +1 counted  │
│ Human?        → reset to 0  │
│                             │
│ Limit hit? → warn once,     │
│              then return    │
└──────────────┬──────────────┘
               ▼
┌─────────────────────────────┐
│ Self-check → return         │
│ Bot gating → return         │
│ (unchanged)                 │
└──────────────┬──────────────┘
               ▼
          normal processing
```

## Warn-Once Semantics

Warning messages are sent only on the **exact** threshold hit to prevent warnings from ping-ponging between bots:

```
#   author    counter   result          action
──  ────────  ───────   ──────────      ──────────────
19  Dealer    19        Ok              process normally
20  Broker    20        SoftLimit(20)   ⚠️ warn (n==max, once!)
21  Dealer    21        SoftLimit(21)   silent return (n>max) ✅
22  Broker    22        SoftLimit(22)   silent return ✅
... (no more warnings, no ping-pong)

Human replies → counter resets to 0 → conversation resumes
```

## Changes

- Move tracker before self-check so ALL bot messages count
- `soft_limit=20` = 20 total bot messages (~10 per bot in two-bot scenario)
- `HardLimit` now carries count for warn-once check
- Warning sent only when `n == max` (exact hit), silently return when `n > max`
- Updated `docs/discord.md` with counting and warn-once behavior
- 4 new tests: `soft_limit_warn_once_semantics`, `hard_limit_warn_once_semantics`, `two_bot_pingpong_hits_soft_limit`, `two_bot_pingpong_human_resets`
- 74 total tests, clippy clean

### DIFF
--- a/docs/discord.md
+++ b/docs/discord.md
@@ -216,8 +216,12 @@ allow_bot_messages = "mentions"
 
 To prevent runaway bot-to-bot loops, OpenAB enforces two layers of protection:
 
-- **Soft limit** (`max_bot_turns`, default: 20) — consecutive bot turns without human intervention. When reached, the bot sends a warning and stops responding. A human message in the thread resets the counter.
+- **Soft limit** (`max_bot_turns`, default: 20) — total bot messages in a thread without human intervention. When reached, the bot sends a one-time warning and stops responding. A human message in the thread resets the counter.
 - **Hard limit** (100, not configurable) — absolute cap on bot turns between human interventions. When reached, bot-to-bot conversation stops until a human replies.
+
+Both limits count **all** bot messages in the thread, including the bot's own replies. In a two-bot ping-pong with `max_bot_turns = 20`, each bot sends ~10 messages before the limit triggers.
+
+Warning messages are sent exactly once (on the exact threshold hit) to prevent warnings from ping-ponging between bots.
 
 ```toml
 [discord]
--- a/src/discord.rs
+++ b/src/discord.rs
@@ -228,7 +228,53 @@ impl EventHandler for Handler {
     async fn message(&self, ctx: Context, msg: Message) {
         let bot_id = ctx.cache.current_user().id;
 
-        // Always ignore own messages
+        // Early multibot detection: cache that another bot is present.
+        // Runs before self-check and bot gating so we always detect other bots. (#481)
+        if msg.author.bot && msg.author.id != bot_id {
+            let key = msg.channel_id.to_string();
+            let mut cache = self.multibot_threads.lock().await;
+            cache.entry(key).or_insert_with(tokio::time::Instant::now);
+        }
+
+        // Bot turn counting: runs before self-check so ALL bot messages
+        // (including own) count toward the per-thread limit. This means
+        // soft_limit=20 = 20 total bot messages in the thread (~10 per bot
+        // in a two-bot ping-pong). (#483)
+        {
+            let thread_key = msg.channel_id.to_string();
+            let mut tracker = self.bot_turns.lock().await;
+            if msg.author.bot {
+                match tracker.on_bot_message(&thread_key) {
+                    TurnResult::HardLimit => {
+                        tracing::warn!(channel_id = %msg.channel_id, "hard bot turn limit reached");
+                        if msg.author.id != bot_id {
+                            let _ = msg.channel_id.say(
+                                &ctx.http,
+                                format!("🛑 Hard bot turn limit reached ({HARD_BOT_TURN_LIMIT}). A human must reply to continue."),
+                            ).await;
+                        }
+                        return;
+                    }
+                    TurnResult::Stopped => return,
+                    TurnResult::SoftLimit(n) => {
+                        tracing::info!(channel_id = %msg.channel_id, turns = n, max = self.max_bot_turns, "soft bot turn limit reached");
+                        if msg.author.id != bot_id {
+                            let _ = msg.channel_id.say(
+                                &ctx.http,
+                                format!("⚠️ Bot turn limit reached ({n}/{}). A human must reply in this thread to continue bot-to-bot conversation.", self.max_bot_turns),
+                            ).await;
+                        }
+                        return;
+                    }
+                    TurnResult::Throttled => return,
+                    TurnResult::Ok => {}
+                }
+            } else {
+                tracker.on_human_message(&thread_key);
+            }
+        }
+
+        // Ignore own messages (after counting toward bot turns above)
         if msg.author.id == bot_id {
             return;
         }
@@ -243,15 +289,6 @@ impl EventHandler for Handler {
 
         let is_mentioned = msg.mentions_user_id(bot_id)
             || msg.content.contains(&format!("<@{}>", bot_id));
-
-        // Early multibot detection: cache that another bot is present in this
-        // channel/thread. Runs BEFORE bot message gating so we detect other
-        // bots even when their messages are filtered out. (#481)
-        if msg.author.bot && msg.author.id != bot_id {
-            let key = msg.channel_id.to_string();
-            let mut cache = self.multibot_threads.lock().await;
-            cache.entry(key).or_insert_with(tokio::time::Instant::now);
-        }
 
         // Bot message gating (from upstream #321)
         if msg.author.bot {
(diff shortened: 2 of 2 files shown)


shaun-agent:
## OpenAB PR Screening
- Title: fix(discord): move bot turn tracker before gating (#483)
- Source: https://github.com/openabdev/openab/pull/484
- Generated at: 2026-04-20T05:31:37.862Z
- Discord thread: https://discord.com/channels/1488041051187974246/1495658020896641075
<details>
## Intent

This PR is trying to make Discord bot-to-bot turn limiting actually work in multi-bot conversations. The user-visible problem is that two OpenAB bots can get into an unbounded ping-pong loop because the turn tracker only runs after early-return gates, so bot-authored messages are never counted toward the soft or hard limit.

## Feat

This is a bug fix in the Discord runtime. In plain terms, it moves bot-turn accounting earlier in the inbound message pipeline so bot messages, including the bot's own replies and other bots' replies, increment the tracker before self-message and bot-message gating can short-circuit processing.

## Who It Serves

The primary beneficiaries are Discord end users and agent runtime operators. It also helps maintainers and reviewers because it restores a safety mechanism that is supposed to prevent runaway automated conversations.

## Rewritten Prompt

Adjust the Discord message handling flow in `src/discord.rs` so bot-turn tracking runs before both the self-message early return and `allow_bot_messages` gating. Count all bot-authored messages toward the per-conversation bot-turn tracker, including the broker's own outbound echoes observed on inbound events and messages from other bots. Keep human messages as the reset signal. Preserve existing bot gating for response generation, but ensure turn limits are enforced even when a message would later be ignored by self-check or mention gating. Add or update tests covering: self-message counting, other-bot counting under mention-only gating, human reset behavior, and soft-limit/hard-limit enforcement in a two-bot ping-pong scenario.

## Merge Pitch

This is worth moving forward because it fixes a broken safety boundary rather than adding optional behavior. Right now the configured soft and hard turn caps appear present but fail in exactly the multi-bot loop case they are meant to contain.

The risk profile is moderate but acceptable: the change touches message-order semantics in a hot path, so reviewer concern will be whether counting before gating creates false positives or changes single-bot behavior. The right review focus is on preserving normal human-triggered flows while making multi-bot loops observable to the limiter.

## Best-Practice Comparison

The most relevant comparison points are safety and execution-boundary discipline, not durable scheduling.

Against OpenClaw:
- Relevant: explicit delivery routing and retry/run-boundary thinking. This PR improves routing discipline by making the limiter observe all bot-originated traffic before downstream filtering hides it.
- Relevant: isolated executions in principle. Counting before gating helps define clearer boundaries around what constitutes an automated turn.
- Less relevant: gateway-owned scheduling, durable job persistence, retry/backoff, and run logs. This item is about realtime message pipeline correctness, not scheduled or persisted job execution.

Against Hermes Agent:
- Relevant: fresh-session and self-contained-task principles. The PR treats each inbound event as something that must carry enough context to enforce safety immediately, without depending on later stages.
- Relevant: overlap prevention in spirit. The limiter is effectively a guard against uncontrolled recursive bot interaction.
- Less relevant: file locking and atomic persisted-state writes unless the Discord tracker is persisted somewhere outside the snippet described. Nothing in the source data suggests that is the main concern here.

Net: this PR aligns with a good systems principle shared by both references: apply safety accounting at the earliest reliable observation point, before downstream filters can erase evidence of activity.

## Implementation Options

### Option 1: Minimal reorder fix

Move the existing bot-turn tracker invocation to run immediately after multibot detection and before self-check and bot gating. Reuse the current tracker structure and warning behavior with the smallest possible code movement.

### Option 2: Balanced pipeline normalization

Refactor inbound Discord handling into an explicit sequence: classify sender, update turn tracker, evaluate turn-limit outcome, then apply self-gating and mention gating for response generation. Keep behavior the same, but make the ordering and intent obvious in code and tests.

### Option 3: Broader conversation-safety layer

Introduce a more formal conversation guard abstraction for Discord bot interactions that tracks actor type, consecutive automated turns, reset conditions, and enforcement outcomes. Use it as a reusable pre-processing stage for current Discord handling and potentially other chat gateways later.

## Comparison Table

|---|---|---|---|---|---|---|
| Minimal reorder fix | High | Low | Medium | Medium | High | High |
| Balanced pipeline normalization | Medium | Medium | High | High | High | High |
| Broader conversation-safety layer | Low | High | High | Medium | Medium-High | Medium |

## Recommendation

The balanced pipeline normalization is the right path for merge discussion.

It keeps the scope anchored to the current bug, but avoids shipping a fragile "just move this block upward" patch whose correctness still depends on implicit control-flow knowledge in `src/discord.rs`. This item is fundamentally about message pipeline ordering, so the implementation should make that ordering explicit and testable.

If the author wants to reduce merge risk, this can be sequenced in two steps:
1. Land the ordering fix plus targeted regression tests for #483.
2. Follow up with a small cleanup that makes sender classification and turn-limit enforcement more explicit if the first patch has to stay narrowly scoped.
</details>

chaodu-agent:
## 🔴 Warning ping-pong regression

Testing shows that after soft limit is hit, the warning messages themselves cause an infinite loop:

```
Broker sees Dealer msg → counter=20 → SoftLimit → sends ⚠️ warning
Dealer sees Broker ⚠️  → counter=21 → SoftLimit → sends ⚠️ warning
Broker sees Dealer ⚠️  → counter=22 → SoftLimit → sends ⚠️ warning
... continues until hard limit
```

The `⚠️` warning is a bot message, so the other bot receives it, increments its counter (already past soft limit), hits `SoftLimit` again, and sends another warning — ping-pong of warnings.

### Suggested fix

Only send the warning on the **exact** soft limit hit, and silently return for all subsequent bot messages:

```rust
TurnResult::SoftLimit(n) => {
    tracing::info!(...);
    // Only warn once — on the exact threshold hit.
    // Subsequent bot messages (n > soft_limit) return silently
    // to avoid warning messages ping-ponging between bots.
    if msg.author.id != bot_id && n == self.max_bot_turns {
        let _ = msg.channel_id.say(&ctx.http, ...).await;
    }
    return;
}
```

Same treatment needed for `HardLimit`.

Screenshot from live testing:

warning ping-pong — warnings escalate from 20/20 to 28/20+ with no human intervention.

This should be addressed before merge.

chaodu-agent:
## Suggested fix: warn-once semantics in `TurnResult`

After discussing the warning ping-pong issue, we think the cleanest fix is to encode "warn once" in the tracker itself rather than patching the caller.

### Why the current approach loops

```
Bot A msg #20  ──→  Broker tracker: soft=20 → SoftLimit(20) → sends ⚠️
                                                                  │
Broker ⚠️ msg  ──→  Dealer tracker: soft=21 → SoftLimit(21) → sends ⚠️
                                                                  │
Dealer ⚠️ msg  ──→  Broker tracker: soft=22 → SoftLimit(22) → sends ⚠️
                                                                  │
                    ... ⚠️ ping-pong until hard limit ...
```

The root cause: `>= soft_limit` returns `SoftLimit` **every time**, so every subsequent bot message (including warnings) triggers another warning.

### Proposed: split `>=` into `==` + `>`

```
TurnResult variants:

  Ok          counter < soft_limit         → continue normally
  SoftLimit   counter == soft_limit        → warn once, then return
  Throttled   counter > soft_limit         → silently return (no message)
  HardLimit   counter == HARD_LIMIT        → warn once, then return
  Stopped     counter > HARD_LIMIT         → silently return (no message)
```

```
on_bot_message() decision tree:

  counter += 1
       │
       ▼
  ┌─────────────────────────┐
  │ hard > HARD_LIMIT?      │──yes──→ Stopped (silent return)
  └────────────┬────────────┘
               │ no
               ▼
  ┌─────────────────────────┐
  │ hard == HARD_LIMIT?     │──yes──→ HardLimit (warn once)
  └────────────┬────────────┘
               │ no
               ▼
  ┌─────────────────────────┐
  │ soft > soft_limit?      │──yes──→ Throttled (silent return)
  └────────────┬────────────┘
               │ no
               ▼
  ┌─────────────────────────┐
  │ soft == soft_limit?     │──yes──→ SoftLimit (warn once)
  └────────────┬────────────┘
               │ no
               ▼
             Ok (continue)
```

### Two-bot ping-pong with fix applied (soft_limit=20)

```
#    author    Broker tracker    result         action
──   ────────  ──────────────    ──────────     ──────────────
19   Dealer    soft=19           Ok             process normally
20   Broker    soft=20           SoftLimit(20)  ⚠️ warn (once!)
21   Dealer    soft=21           Throttled      silent return ✅
22   Broker    soft=22           Throttled      silent return ✅
...  (no more warnings, no ping-pong)

Human replies → soft resets to 0 → conversation resumes
```

### Why this is the cleanest approach

1. **"Warn once" is in the type system** — `SoftLimit` can only fire at `==`, impossible to warn twice
2. **Caller is trivial** — just match on variant, no `if n == max` checks
3. **No ping-pong possible** — after the one warning, all subsequent bot msgs get `Throttled` → silent return
4. **Easy to test** — assert msg #20 is `SoftLimit`, msg #21 is `Throttled`
5. **Backward compatible** — same counter logic, just finer-grained return type

Issue #481, opened 2026-04-20
fix(discord): multibot-mentions fails to detect other bots due to gating order


thepagent:
### Description

`multibot-mentions` mode fails to detect multi-bot threads when `allow_bot_messages = "mentions"`. The bot responds without @mention even after another bot has posted in the thread.

**Root cause:** Eager multibot detection (line 335) runs **after** bot message gating (line 248). When another bot posts without @mentioning this bot, the message is discarded at the gating step and never reaches the detection code.

```
Bot gating (line 248)              ← other bot msg filtered here
  → return if bot msg
...
Eager multibot detection (line 335) ← never reached
  → multibot_threads.insert(key)
```

https://github.com/openabdev/openab/blob/9b7680d/src/discord.rs#L248

https://github.com/openabdev/openab/blob/9b7680d/src/discord.rs#L335

### Steps to Reproduce

1. Set `allow_bot_messages = "mentions"` and `allow_user_messages = "multibot-mentions"`
2. Human: `@BotA HIHI` → BotA creates thread, responds
3. Human: `@BotB 你來` → BotB responds in the thread
4. Human: `nice` (no @mention) → BotA responds ← **should not respond**

### Expected Behavior

After step 3, BotA should detect the thread is multi-bot and require @mention for step 4.

### Fix

Move eager multibot detection **before** bot message gating:

```rust
// BEFORE bot gating — detect other bots from every incoming message
if in_thread && msg.author.bot && msg.author.id != bot_id {
    let key = msg.channel_id.to_string();
    let mut cache = self.multibot_threads.lock().await;
    cache.entry(key).or_insert_with(tokio::time::Instant::now);
}

// Bot message gating (unchanged)
if msg.author.bot { ... }
```

Zero API calls, zero fetch — just check `msg.author.bot` on every incoming message before any gating.

PR #291, opened 2026-04-13
feat(discord): inline text-based file attachments into prompt

JARVIS-coding-Agent:
## Context

Issue #161 requested multimodal attachment support (images, files, voice). Image support was shipped in #158 / #210, and voice STT was shipped in #225 / #228. **This PR addresses the remaining scope: text-based file attachments.**

## Summary

Download text-based file attachments from Discord messages and inject their content as `ContentBlock::Text` blocks, so downstream ACP agents can read uploaded code, configs, logs, etc.

## Changes

- **`is_text_attachment()`** — detects text files via MIME type (`text/*`, `application/json`, `application/xml`) and a `TEXT_EXTENSIONS` whitelist (~40 common extensions: `.txt`, `.csv`, `.log`, `.md`, `.json`, `.rs`, `.py`, `.js`, `.go`, etc.)
- **`download_and_read_text_file()`** — downloads the attachment (512 KB size limit), reads as UTF-8 with lossy fallback, wraps in a code block with filename header
- **Attachment routing** updated: audio → text file → image → skip

## Design Decisions

- **512 KB hardcoded limit** — consistent with image (10 MB) and audio (25 MB) which also use hardcoded limits. No config needed.
- **UTF-8 lossy fallback** — gracefully handles files with occasional non-UTF-8 bytes instead of rejecting them entirely.
- **Code block wrapping** — `[File: name.ext]\n` ` ` `\n...\n` ` ` format gives the agent clear file boundaries.
- **`.env` in `TEXT_EXTENSIONS`** — intentionally included so agents can read `.env.example` and similar template files. Actual secret `.env` files should not be shared in Discord channels where the bot operates; this is a channel-level access concern, not a bot concern.

Closes #161

https://discord.com/channels/1488041051187974246/1490282656913559673/1494333565221994719

PR #487, opened 2026-04-20
feat(slack): port Discord parity — mentions, multibot, bot turn limits

dogzzdogzz:
Discord Discussion URL: https://discord.com/channels/1491295327620169908/1494739741600387204/1495409492089769984

## Summary

Ports recent Discord-side work to the Slack adapter so both platforms share the same safety and feature surface, plus a Discord race-condition fix. Slack slash-command support was attempted and then removed — see row 4 below.

| # | Area | What changes | Why |
|---|------|--------------|-----|
| 1 | **Mention resolution** | `strip_slack_mention` → `resolve_slack_mentions(text, bot_id)`: strips only the bot's own `<@UID>`, preserves other user mentions. | Correctness: the old function wiped every `<@…>`, so the LLM could never @-mention users in its reply (Discord's `resolve_mentions` already works this way). |
| 2 | **MultibotMentions** | `AllowUsers::MultibotMentions` was accepted by config but fell through to `Involved` with no other-bot check. Added a `multibot_threads` cache populated eagerly from incoming bot events, widened `bot_participated_in_thread` to return `(involved, other_bot)`, and split the gating arm. | Advertised behaviour now actually fires — mirrors Discord #481 / #483. |
| 3 | **Bot turn limits** | Ported the soft/hard `BotTurnTracker` to Slack, running before self-check so every bot message counts. Reset on human message. New `max_bot_turns` field on `SlackConfig` (default 20). Discord's local copy moves into a new `src/bot_turns.rs` so both adapters share it. | Safety: Slack allowed bot-to-bot loops with only a 10-msg cap; now matches Discord's soft+hard `100` cap with human-reset. |
| 4 | ~~Slash commands~~ | **Not supported on Slack.** The Slack adapter ack-and-drops `slash_commands` / `interactive` envelopes with a debug log. Discord slash commands (`/models`, `/agents`, `/cancel`) are unchanged. | Slack's UI blocks third-party slash commands invoked from a thread reply composer (`"/cmd is not supported in threads. Sorry!"`), and the channel-level payload carries no `thread_ts` so it can't route to a per-thread ACP session. Most ACP agents (notably `claude-code`) also don't emit `configOptions` for model/agent switching, making the menu empty even when routing works. The feature was added and then removed in this same PR (see the `revert(slack): remove slash command support` commit) because there was no working path end-to-end. |
| 5 | **Docs / example config** | Updated `config.toml.example` and `docs/slack-bot-howto.md` with `max_bot_turns` and the `multibot-mentions` variant. (Slash-command registration steps were added and later removed in line with row 4.) | User-facing surface changed. |
| 6 | **Chore** | Added `CLAUDE.md` to `.gitignore` (local AI context file shouldn't land in the repo). | — |
| 7 | **Discord bug fix** | On `create_thread_from_message` failing with error 160004 ("A thread has already been created for this message"), refetch the trigger message and join the thread the sibling bot just created instead of giving up. | Reproduced in prod: `"@bob say hi to @patrick"` → both bots race to create a thread from the same trigger; the loser used to log `ERROR failed to create thread: A thread has already been created for this message` and drop the turn. |

## Test plan

- [x] `cargo clippy --all-targets -- -D warnings` — clean
- [x] `cargo test` — **81 pass / 0 fail** (8 Slack tests + 3 Discord tests for the thread-race matcher + 12 `BotTurnTracker` tests relocated to the shared module)
- [x] Smoke test: start a bot-to-bot loop with `allow_bot_messages = "all"` and `max_bot_turns = 5` → loop halts at the soft cap with a warning; human message resets
- [x] Smoke test: `allow_user_messages = "multibot-mentions"` with two bots in a thread → @mention required after the second bot posts
- [x] Smoke test: Discord message `@bot-a say hi to @bot-b` → both bots post in the same thread (no `failed to create thread` error)

## Commit history

1. `feat(slack): port Discord parity — mentions, multibot, turn limits, slash commands` — initial bundle
2. `chore: ignore CLAUDE.md`
3. `fix(discord): join existing thread when racing a sibling bot`
4. `feat: force send-once per thread when a trigger message tags other members` — experimental (reverted)
5. `Revert "feat: force send-once per thread when a trigger message tags other members"` — revert of #4
6. `revert(slack): remove slash command support` — row 4 above

Net effect vs main: rows 1, 2, 3, 5, 6, 7 are live; row 4 is a no-op in the diff.

## Notes for reviewers

- Single PR per author request. History has a noisy pair (`force send-once` + its revert) that cancels out in the diff.
- Helm chart isn't updated — Slack's `max_bot_turns` falls back to the Rust default (20), same as Discord currently does in the chart.
- `multibot-mentions` validation already exists in `charts/openab/templates/configmap.yaml:84` — no chart change needed to unlock the Slack behaviour.
- Row 7's fix is matched by error-string heuristic (`160004` or `"already been created"`); covered by unit tests so serenity formatting changes will be caught.
- If Slack eventually removes the thread restriction on third-party slash commands, restoring row 4 is straightforward — the deleted helpers live only in git history.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ config.toml
 .DS_Store
 .env
 .kiro/
+CLAUDE.md
--- a/config.toml.example
+++ b/config.toml.example
@@ -24,9 +24,12 @@ allowed_channels = ["1234567890"]       # ↑ omitted + non-empty list → auto-
 # allowed_users = ["U0123456789"]      # only checked when allow_all_users = false
 # allow_bot_messages = "off"           # "off" (default) | "mentions" | "all"
 # trusted_bot_ids = []                 # empty = any bot (mode permitting); set to restrict
-# allow_user_messages = "involved"     # "involved" (default) | "mentions"
+# allow_user_messages = "involved"     # "involved" (default) | "mentions" | "multibot-mentions"
                                         # "involved" = reply in threads bot has participated in
                                         # "mentions" = always require @mention
+                                        # "multibot-mentions" = like "involved", but require @mention
+                                        #   once another bot has posted in the thread
+# max_bot_turns = 20                   # soft cap on consecutive bot turns per thread (human msg resets)
 
 [agent]
 command = "kiro-cli"
--- a/docs/slack-bot-howto.md
+++ b/docs/slack-bot-howto.md
@@ -92,6 +92,26 @@ In a channel where the bot is invited:
 
 The bot will reply in a thread. After that, just type in the thread — no @mention needed for follow-ups.
 
+## Slash commands are not supported on Slack
+
+openab supports `/models`, `/agents`, and `/cancel` on **Discord**, but **not on Slack**. If you previously configured these commands in your Slack app's **Slash Commands** page, you can safely delete them — the Slack adapter ignores both `slash_commands` and `interactive` envelope types.
+
+The root cause is a combination of three Slack-specific platform constraints, none of which is fixable from openab's side:
+
+1. **Slack blocks third-party slash commands inside threads.** Invoking `/models` from a thread's reply composer returns the Slack error `"/models is not supported in threads. Sorry!"`. This is enforced by the Slack client itself, not by any app setting — enabling Interactivity, Socket Mode, or reinstalling the app does not bypass it. Slack's built-in commands (`/remind`, `/shrug`, etc.) get special treatment that custom apps cannot.
+
+2. **Channel-level slash command payloads have no thread context.** If the user types `/models` in the channel's main composer instead of a thread, Slack delivers the command but the payload carries no `thread_ts`. Since openab keys each ACP session by thread (`slack:<thread_ts>` or `slack:<trigger_ts>`), the command cannot be routed to the right session. Sessions are never keyed by `channel_id` alone, so there's no workaround on the adapter side.
+
+3. **Most ACP agents don't expose a model-switch surface.** Even when routing succeeded, `/models` reads the session's `configOptions` from the ACP `initialize` response. Only `kiro-cli` emits these in the expected format (via its `models`/`modes` fallback). `claude-code`, `codex`, `gemini`, `cursor-agent`, and `opencode` do not, so the menu would be empty for those backends — the user would see `"⚠️ No model options available"` with no recourse.
+
+On Discord, none of these apply: slash commands work in thread-channels, the channel ID *is* the thread key, and users typically stay within a single agent per deployment anyway.
+
+### If you need to switch models or agents with a Slack deployment
+
+- **Change the agent**: edit `[agent]` in `config.toml` (or the Helm chart values) and restart the pod / process
+- **Change the Claude model** (for `claude-code`): set `ANTHROPIC_DEFAULT_MODEL` (or equivalent env var depending on your claude-code-acp version) and restart — model selection happens at process start, not at runtime
+- **Cancel an in-flight turn**: there is no built-in way on Slack currently.
+
 ## Finding Channel and User IDs
 
 - **Channel ID**: Right-click the channel name → **View channel details** → ID at the bottom (starts with `C` for public, `G` for private)
--- a/docs/slack-bot-howto.md
+++ b/docs/slack-bot-howto.md
@@ -74,27 +74,6 @@ export SLACK_BOT_TOKEN="xoxb-..."
 export SLACK_APP_TOKEN="xapp-..."
 ```
 
-## 6b. (Optional) Register Slash Commands
-
-openab handles three slash commands: `/models` (switch model), `/agents` (switch agent mode), `/cancel` (cancel the current turn). They arrive over Socket Mode — no Request URL needed — but Slack still requires you to declare each command in the app manifest.
-
-1. In the left sidebar, click **Slash Commands** → **Create New Command**
-2. Create each command with these settings:
-
-| Command | Short Description | Usage Hint |
-|---------|-------------------|------------|
-| `/models` | Select the AI model for this session | — |
-| `/agents` | Select the agent mode for this session | — |
-| `/cancel` | Cancel the current operation | — |
-
-3. Leave the **Request URL** empty (Socket Mode delivers the payload)
-4. Add the **Interactivity** bot token scope by enabling **Interactivity & Shortcuts** in the sidebar — leave the Request URL empty there too
-5. Reinstall the app when prompted
-
-The commands pick up the ACP session's `configOptions` and render an ephemeral select menu. `/cancel` sends a cancel signal to the active turn.
-
-Commands are keyed by channel; switching model in a channel's main composer affects subsequent sessions started from that channel.
-
 ## 7. Invite the Bot
 
 In each Slack channel where you want to use the bot:
--- a/docs/slack-bot-howto.md
+++ b/docs/slack-bot-howto.md
@@ -74,6 +74,27 @@ export SLACK_BOT_TOKEN="xoxb-..."
 export SLACK_APP_TOKEN="xapp-..."
 ```
 
+## 6b. (Optional) Register Slash Commands
+
+openab handles three slash commands: `/models` (switch model), `/agents` (switch agent mode), `/cancel` (cancel the current turn). They arrive over Socket Mode — no Request URL needed — but Slack still requires you to declare each command in the app manifest.
+
+1. In the left sidebar, click **Slash Commands** → **Create New Command**
+2. Create each command with these settings:
+
+| Command | Short Description | Usage Hint |
+|---------|-------------------|------------|
+| `/models` | Select the AI model for this session | — |
+| `/agents` | Select the agent mode for this session | — |
+| `/cancel` | Cancel the current operation | — |
+
+3. Leave the **Request URL** empty (Socket Mode delivers the payload)
+4. Add the **Interactivity** bot token scope by enabling **Interactivity & Shortcuts** in the sidebar — leave the Request URL empty there too
+5. Reinstall the app when prompted
+
+The commands pick up the ACP session's `configOptions` and render an ephemeral select menu. `/cancel` sends a cancel signal to the active turn.
+
+Commands are keyed by channel; switching model in a channel's main composer affects subsequent sessions started from that channel.
+
 ## 7. Invite the Bot
 
 In each Slack channel where you want to use the bot:
--- a/src/adapter.rs
+++ b/src/adapter.rs
@@ -1,9 +1,7 @@
 use anyhow::Result;
 use async_trait::async_trait;
-use regex::Regex;
 use serde::Serialize;
-use std::collections::HashSet;
-use std::sync::{Arc, LazyLock};
+use std::sync::Arc;
 use tracing::error;
 
 use crate::acp::{classify_notification, AcpEvent, ContentBlock, SessionPool};
@@ -95,25 +93,6 @@ pub trait ChatAdapter: Send + Sync + 'static {
 pub struct AdapterRouter {
     pool: Arc<SessionPool>,
     reactions_config: ReactionsConfig,
-    /// Per-thread flag: once a triggering message in this thread has tagged
-    /// another member (user or bot) other than this bot, subsequent agent
-    /// replies in the thread fall back to send-once even if the adapter
-    /// otherwise enables streaming. Prevents repeated `chat.update` edits
-    /// in threads where the agent's reply is likely to contain mention
-    /// strings, which keeps the conversation deterministic for sibling bots.
-    force_send_once_threads: tokio::sync::Mutex<HashSet<String>>,
-}
-
-/// Matches any remaining `<@UID>` in a prompt after the adapter has stripped
-/// its own trigger mention. Covers Discord `<@123>` / `<@!123>` and Slack
-/// `<@U123ABC>` uniformly.
-static OTHER_MENTION_RE: LazyLock<Regex> =
-    LazyLock::new(|| Regex::new(r"<@!?[A-Za-z0-9]+>").unwrap());
-
-/// True if `prompt` still contains a user mention after the adapter has
-/// stripped this bot's own trigger mention. Exposed for unit testing.
-fn prompt_mentions_other_member(prompt: &str) -> bool {
-    OTHER_MENTION_RE.is_match(prompt)
 }
 
 impl AdapterRouter {
@@ -121,31 +100,12 @@ impl AdapterRouter {
         Self {
             pool,
             reactions_config,
-            force_send_once_threads: tokio::sync::Mutex::new(HashSet::new()),
         }
     }
 
     /// Access the underlying session pool (e.g. for config option queries).
     pub fn pool(&self) -> &Arc<SessionPool> {
         &self.pool
-    }
-
-    /// Mark a thread so all subsequent agent replies use send-once mode
-    /// regardless of the adapter's `use_streaming()` hint.
-    async fn mark_thread_send_once(&self, thread_key: &str) {
-        self.force_send_once_threads
-            .lock()
-            .await
-            .insert(thread_key.to_string());
-    }
-
-    /// True if this thread has been marked as send-once (e.g. because a
-    /// triggering message tagged another member).
-    async fn is_thread_send_once(&self, thread_key: &str) -> bool {
-        self.force_send_once_threads
-            .lock()
-            .await
-            .contains(thread_key)
     }
 
     /// Handle an incoming user message. The adapter is responsible for
@@ -193,17 +153,6 @@ impl AdapterRouter {
                 .as_deref()
                 .unwrap_or(&thread_channel.channel_id)
         );
-
-        // Once a triggering message tags another member, the thread switches
-        // to send-once mode permanently. Adapters preserve non-bot `<@UID>`
-        // in the resolved prompt, so any remaining mention is enough signal.
-        if prompt_mentions_other_member(prompt) {
-            self.mark_thread_send_once(&thread_key).await;
-            tracing::debug!(
-                thread_key = %thread_key,
-                "prompt tags another member — thread forced send-once"
-            );
-        }
 
         if let Err(e) = self.pool.get_or_create(&thread_key).await {
             let msg = format_user_error(&e.to_string());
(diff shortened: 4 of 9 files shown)


shaun-agent:
## OpenAB PR Screening
- Title: feat(slack): port Discord parity — mentions, multibot, bot turn limits, slash commands
- Source: https://github.com/openabdev/openab/pull/487
- Generated at: 2026-04-20T09:31:49.792Z
- Discord thread: https://discord.com/channels/1488041051187974246/1495718367212994621
<details>
## Intent

This PR is trying to close the feature and safety gap between the Slack adapter and the Discord adapter. The concrete operator-visible problem is that Slack currently behaves differently in several important areas: it strips all mentions instead of only the bot's own mention, does not correctly enforce the advertised `multibot-mentions` behavior, has weaker bot-loop protection, lacks slash-command UX that Discord already has, and still leaves a known Discord thread-race bug unresolved.

In practical terms, the item is about making Slack safer to run in multi-bot environments and more usable for admins and end users, while also fixing one production-relevant Discord edge case discovered during parity testing.

## Feat

This is primarily a feature-parity PR with bundled safety fixes, one bug fix, and accompanying docs/config updates.

Behaviorally, it does five real things:
- Makes Slack mention resolution preserve user mentions while stripping only the current bot's own mention.
- Implements actual `MultibotMentions` gating logic on Slack instead of silently falling back to a looser mode.
- Ports the shared bot turn-limit logic to Slack so bot-to-bot loops are bounded the same way as Discord.
- Adds Slack slash-command handling for `/models`, `/agents`, and `/cancel`.
- Fixes a Discord race where two bots can try to create the same thread and one used to drop the turn.

Classification: feature + fix + docs improvement, bundled into one PR.

## Who It Serves

The primary beneficiaries are Slack deployers and multi-bot runtime operators, because they need Slack to behave predictably and safely under the same policies already established on Discord.

Secondary beneficiaries are:
- Slack end users, who get usable mentions and slash-command controls
- Maintainers and reviewers, who benefit from shared bot-turn logic and reduced platform drift
- Discord operators, because the thread-race bug fix prevents dropped turns in real multi-bot use

## Rewritten Prompt

Bring the Slack adapter to parity with current Discord behavior in four scoped areas: mention resolution, multibot mention gating, bot turn limits, and slash-command handling. Preserve non-bot user mentions in Slack input, implement real `multibot-mentions` enforcement using thread participation state, reuse a shared `BotTurnTracker` module across Discord and Slack with configurable `max_bot_turns`, and add Socket Mode handling for `/models`, `/agents`, and `/cancel` using ephemeral select-based responses.

Also fix the Discord thread-creation race by detecting the "thread already exists" case, refetching the trigger message, and joining the existing thread instead of dropping the turn. Update config/docs for all new or exposed behavior, and cover each area with unit tests plus smoke-test notes for Slack command flow, multibot gating, loop stopping, and Discord dual-bot thread creation.

## Merge Pitch

This item is worth advancing because it reduces platform inconsistency in a place where inconsistency directly creates operator confusion and runtime risk. OpenAB already has the intended safety model and interaction model on Discord; this PR mostly makes Slack stop being the odd one out.

The risk profile is moderate. The main reviewer concern will be bundling: multiple behavior changes across Slack plus a Discord fix in one PR makes regression review harder. The substance is good, but the merge discussion should focus on whether to land as one bundle or split into smaller reviewable slices.

## Best-Practice Comparison

Relevant OpenClaw principles:
- Explicit delivery routing: relevant. Mention resolution and multibot gating both affect whether the system responds to the right actor in the right context.
- Retry/backoff and run logs: only partially relevant. The Discord thread-race fix is in the same reliability family, but this PR is not introducing broader retry or observability structure.
- Gateway-owned scheduling: not relevant.
- Durable job persistence: not relevant.
- Isolated executions: not directly relevant, though bot turn limits do reduce cross-bot runaway behavior.

Relevant Hermes Agent principles:
- Fresh session per scheduled run: not relevant.
- Self-contained prompts for scheduled tasks: not relevant.
- File locking to prevent overlap: conceptually relevant to the Discord race fix, in the sense that overlapping work must be handled safely, but this PR resolves overlap by fallback recovery rather than locking.
- Atomic writes for persisted state: only lightly relevant if the new Slack thread/bot caches are persisted, which does not appear to be the core design here.
- Gateway daemon tick model: not relevant.

Best-practice read:
- The strongest alignment is with operational safety and explicit routing, not with scheduling architecture.
- The Slack bot-turn tracker and multibot gating move OpenAB toward clearer execution boundaries.
- The Discord race fix is pragmatic and production-oriented, but it is still heuristic-driven rather than a stronger coordination primitive.

## Implementation Options

### Option 1: Merge the PR largely as-is, with targeted review focus

Keep the bundled parity approach. Review it as one parity/safety package, verify the unit tests, and require the listed smoke tests before merge or immediately after behind maintainer validation.

This is the fastest path and preserves the author's intended "single PR" packaging, but it keeps review complexity high.

### Option 2: Split into two merge units

Separate the work into:
- Slack parity and safety changes: mention resolution, multibot gating, bot turn limits, slash commands, docs/config
- Discord bug fix: thread creation race handling

This keeps the logical theme mostly intact while isolating the unrelated Discord fix. It reduces review noise and makes rollback safer if Slack behavior needs iteration.

### Option 3: Split into three narrower tracks with shared-core extraction

Separate into:
- Shared `bot_turns` extraction and Slack bot turn enforcement
- Slack interaction parity: mentions, multibot gating, slash commands, docs/config
- Discord thread-race fix

This is the most maintainable review shape and best for long-term change history. It makes each behavior set easier to test and reason about, but costs more maintainer overhead now.

## Comparison Table

|---|---|---|---|---|---|---|
| Option 1: Merge as bundled parity PR | High | Medium-High in review | Medium | Medium-Low | High | Medium |
| Option 2: Split Slack parity from Discord fix | Medium-High | Medium | High | High | High | High |
| Option 3: Split into three narrower tracks | Medium-Low | High process overhead | High | Very High | High | Medium-High |

## Recommendation

Recommend Option 2: split the Slack parity/safety work from the Discord thread-race fix, then advance both.

That is the right step for merge discussion because it preserves nearly all delivery momentum while removing the clearest review objection: one PR currently mixes a broad Slack parity bundle with a separate Discord production bug fix. The Slack bundle is coherent enough to stand on its own, and the Discord fix is small, concrete, and independently valuable. If further splitting is needed after review, the next split should be to isolate slash commands from the Slack safety/core parity path, but that should be a second-order decision rather than the default starting point.
</details>

chaodu-agent:
## 🔴 `MultibotMentions` gating bug — multi-bot threads silently drop all messages

The `MultibotMentions` arm in `slack.rs` does `if other_bot { continue; }` without checking `mentions_bot`. This means in a multi-bot thread, **even if the user explicitly @mentions the bot, the message is dropped**.

Expected flow:

```
┌─────────────────────────────┐
│  Thread message arrives     │
│  (allow_user_messages =     │
│   MultibotMentions)         │
└─────────────┬───────────────┘
              │
              ▼
       ┌──────────────┐
       │  has_thread?  │──── No ───▶ continue (skip)
       └──────┬───────┘
              │ Yes
              ▼
       ┌──────────────┐
       │   involved?   │──── No ───▶ continue (skip)
       └──────┬───────┘
              │ Yes
              ▼
       ┌───────────────┐
       │  other_bot?   │──── No ───▶ ✅ process (single-bot thread, no mention needed)
       └──────┬────────┘
              │ Yes (multi-bot thread)
              ▼
     ┌─────────────────┐
     │  mentions_bot?  │──── Yes ──▶ ✅ process (user @mentioned this bot, allow)
     └────────┬────────┘
              │ No
              ▼
        ❌ continue (skip)
        "multi-bot thread without @mention → ignore"
```

Current PR code (bug):

```
       ┌───────────────┐
       │  other_bot?   │──── No ───▶ ✅ process
       └──────┬────────┘
              │ Yes
              ▼
        ❌ continue (skip)    ◀── 🐛 missing mentions_bot check
                                    drops ALL messages in multi-bot threads
```

Suggested fix — one-line change:

```diff
- if other_bot {
+ if other_bot && !mentions_bot {
      continue;
  }
```

Without this, `MultibotMentions` on Slack = bot goes completely silent in any thread where another bot has posted, even when the user explicitly @mentions it. Discord's implementation correctly gates on `!mentions_bot`, so this would be a parity gap.

chaodu-agent:
## 🟡 NIT: Bot turn result handling is duplicated across adapters

`BotTurnTracker` is now shared in `src/bot_turns.rs` (nice!), but the ~50-line `match` block that acts on `TurnResult` is copy-pasted in both `discord.rs` and `slack.rs`. The warning messages, log levels, and warn-once semantics are nearly identical — only the `ChannelRef` construction differs.

**Discord side** (simplified):
```rust
match tracker.on_bot_message(&thread_id) {
    TurnResult::HardLimit => {
        warn!(channel_id, "hard bot turn limit reached");
        adapter.send_message(&channel_ref,
            &format!("🛑 Hard bot turn limit reached ({HARD_BOT_TURN_LIMIT}). ...")
        ).await;
        return;
    }
    TurnResult::SoftLimit(n) => {
        info!(channel_id, turns = n, "soft bot turn limit reached");
        adapter.send_message(&channel_ref,
            &format!("⚠️ Bot turn limit reached ({n}/{max}). ...")
        ).await;
        return;
    }
    TurnResult::Throttled | TurnResult::Stopped => return,
    TurnResult::Ok => {}
}
```

**Slack side** (simplified) — almost identical:
```rust
match tracker.on_bot_message(&turn_key) {
    TurnResult::HardLimit => {
        warn!(channel_id, "hard bot turn limit reached");
        adapter.send_message(&warn_channel,
            &format!("🛑 Hard bot turn limit reached ({HARD_BOT_TURN_LIMIT}). ...")
        ).await;
        continue;
    }
    TurnResult::SoftLimit(n) => {
        info!(channel_id, turns = n, "soft bot turn limit reached");
        adapter.send_message(&warn_channel,
            &format!("⚠️ Bot turn limit reached ({n}/{max}). ...")
        ).await;
        continue;
    }
    TurnResult::Throttled | TurnResult::Stopped => continue,
    TurnResult::Ok => {}
}
```

The only real difference is how `ChannelRef` / `warn_channel` is built. It might be worth considering extracting the message formatting into a shared helper in `bot_turns.rs`, e.g.:

```rust
// in src/bot_turns.rs
pub enum TurnAction {
    Continue,
    WarnAndStop(String),  // warning message to send
    SilentStop,
}

impl BotTurnTracker {
    pub fn check_and_format(&mut self, thread_id: &str, soft_limit: u32) -> TurnAction {
        match self.on_bot_message(thread_id) {
            TurnResult::Ok => TurnAction::Continue,
            TurnResult::SoftLimit(n) => TurnAction::WarnAndStop(
                format!("⚠️ Bot turn limit reached ({n}/{soft_limit}). \
                         A human must reply in this thread to continue.")
            ),
            TurnResult::HardLimit => TurnAction::WarnAndStop(
                format!("🛑 Hard bot turn limit reached ({HARD_BOT_TURN_LIMIT}). \
                         A human must reply to continue.")
            ),
            TurnResult::Throttled | TurnResult::Stopped => TurnAction::SilentStop,
        }
    }
}
```

Then both adapters shrink to:
```rust
match tracker.check_and_format(&key, max_bot_turns) {
    TurnAction::WarnAndStop(msg) => { adapter.send_message(&ch, &msg).await; continue; }
    TurnAction::SilentStop => continue,
    TurnAction::Continue => {}
}
```

Not a blocker for this PR — the current code works. But as more adapters land (Teams? Telegram?), maintaining identical warning text and log semantics in N places will get painful.

masami-agent (commented):
## PR Review: #487

### Summary
- **Problem**: Slack adapter lacked several safety and feature-parity items that Discord already had: correct mention resolution, multibot-mentions gating, shared bot turn limits, text-file attachments, and a Discord thread-race fix.
- **Approach**: Port each feature from Discord to Slack, extract shared logic (`BotTurnTracker` → `bot_turns.rs`), and fix a Discord race condition where two bots create threads from the same trigger message.
- **Risk level**: Medium — touches core message routing in both adapters, but well-tested (81+ tests) and CI passes.

### Core Assessment
1. Problem clearly stated: ✅ — PR description is thorough with a table of changes, rationale, and test plan.
2. Approach appropriate: ✅ — Extracting shared `bot_turns.rs` is the right call. Porting feature-by-feature with matching semantics is clean.
3. Alternatives considered: ✅ — Slash commands were attempted and deliberately reverted with clear reasoning (Slack platform limitations).
4. Best approach for now: ✅ — Minimal, focused changes with good parity reasoning.

### Findings

**`src/bot_turns.rs` (new shared module)**

Clean extraction. The `classify_bot_message` → `TurnAction` helper is a good abstraction that eliminates the duplicated match block in both adapters. Tests are comprehensive (20 tests covering soft/hard limits, reset, warn-once, per-thread independence, classify API).

One observation: `on_human_message` resets **both** soft and hard counters to 0. The doc comment on `HARD_BOT_TURN_LIMIT` says _"the hard counter also tracks the lifetime total and caps it regardless"_ — but the implementation resets hard on human message too. This is consistent with the existing Discord behavior (same code was moved), so it's not a regression, but the doc comment is slightly misleading. The hard limit is really "max consecutive bot turns without human intervention", not a lifetime cap.

**`src/discord.rs` — thread race fix**

The `is_thread_already_exists_error` heuristic (string matching on `160004` or `"already been created"`) is pragmatic. The fallback path (refetch message → read `.thread` → join) is correct. Unit tests cover the detection. One edge case worth noting: if the refetch succeeds but `message.thread` is `None` (e.g., the sibling bot's thread was deleted between the error and the refetch), the error message is clear and propagates correctly.

**`src/slack.rs` — multibot-mentions**

The `MultibotMentions` arm now correctly checks `!mentions_bot` before continuing — this was the bug fix from the follow-up commit. The eager `note_other_bot_in_thread` call before self-check mirrors Discord #481 ordering. The `bot_participated_in_thread` return type change to `(bool, bool)` is clean.

**`src/slack.rs` — bot turn tracking**

The turn key construction uses `thread_ts` when in a thread, else `channel:ts`. For non-thread messages, `channel:ts` is unique per message, so the tracker would never accumulate — but this is correct because non-thread bot messages can't form loops. The tracker lock scope is tight (released before the `handle_message` call).

**`src/slack.rs` — `is_plain_user_message` whitelist**

Good defensive approach. The whitelist (`""` | `me_message` | `thread_broadcast` | `file_share`) is safer than the previous blacklist approach. Tests cover the full system-subtype list. Parity with Discord's `MessageType::Regular | InlineReply` gate.

**`src/slack.rs` — text-file attachments**

Clean port of Discord PR #291. The `strip_mime_params` helper handles Slack's `text/plain; charset=utf-8` edge case. The post-download cap check (`text_file_bytes + actual_bytes > TEXT_TOTAL_CAP`) correctly handles the `size == 0` bypass for externally-backed files. The `slack_file_download_url` helper with `url_private_download` → `url_private` fallback is well-tested.

**`src/slack.rs` — `resolve_slack_mentions`**

Core correctness fix: the old `strip_slack_mention` wiped all `<@UID>` patterns, preventing the LLM from @-mentioning users back. The new function strips only the bot's own mention. Simple and correct.

**`src/slack.rs` — `handle_message` signature change**

The `strip_mentions: bool` parameter is removed and replaced with always calling `resolve_slack_mentions`. This is correct — both `app_mention` and `message` events should resolve mentions the same way (strip bot, preserve others).

### Review Summary

#### 🔴 Blockers
(none)

#### 💬 Questions
1. The `Cargo.lock` diff shows `openab` version changing from `0.7.8` → `0.8.1`. This appears to come from the merge commits pulling in main's `Cargo.toml` version bump. Can you confirm this is just a merge artifact and not an intentional version bump in this PR? (The PR itself doesn't modify `Cargo.toml` version.)

2. The PR description mentions _"Single PR per author request"_ — was there a reason not to split the Discord thread-race fix (row 7) into its own PR? It's an independent bug fix that could be merged separately. Not blocking, just curious about the rationale.

3. No `Closes #XX` in the PR description — are there tracking issues for any of these features (multibot-mentions on Slack, bot turn limits on Slack, the Discord thread-race bug)?

#### 🔧 Suggested Changes
1. **Doc comment on `HARD_BOT_TURN_LIMIT`** — The comment says _"the hard counter also tracks the lifetime total and caps it regardless"_ but `on_human_message` resets both counters. Consider updating the comment to say something like _"Absolute per-thread cap on consecutive bot turns without human intervention"_ to match the actual behavior.

2. **`MAX_CONSECUTIVE_BOT_TURNS` is now dead code** — `src/discord.rs` line 23 still has `const MAX_CONSECUTIVE_BOT_TURNS: u8 = 10;` which was the old Slack-only cap. It's no longer referenced anywhere in the diff. Should be removed.

#### ℹ️ Info
- CI is all green (check + 7 smoke tests pass).
- The commit history has a noisy add/revert pair (force send-once + its revert) that cancels out in the net diff — confirmed, no residual changes.
- The slash command removal is well-documented with three independent technical reasons. The `docs/slack-bot-howto.md` section explaining why is a nice touch for users.

#### ⚪ Nits
(none — code is clean and well-commented)

### Verdict
COMMENT — no blockers, but would like answers to the questions above (especially Q1 on the Cargo.lock version and Q3 on issue tracking) before approving.

obrutjack (approved):
## PR Review: #487

### Summary
- **Problem**: Slack adapter lacked Discord's safety features — bot turn limits, multibot mention gating, correct mention resolution, text-file attachments
- **Approach**: Port Discord safety features to Slack, extract shared `BotTurnTracker` into `bot_turns.rs`, fix Discord thread-creation race
- **Risk level**: Low

### Core Assessment
1. Problem clearly stated: ✅
2. Approach appropriate: ✅
3. Alternatives considered: ✅ (slash commands tried and reverted with clear reasoning)
4. Best approach for now: ✅

### Review Summary

#### 🔴 Blockers
None.

#### ℹ️ Info
- `BotTurnTracker` extracted identically from discord.rs → bot_turns.rs, no logic change
- `resolve_slack_mentions` correctly strips only bot's own `<@UID>` — fixes the old `strip_slack_mention` that wiped all mentions
- `MultibotMentions` now has explicit `!mentions_bot` guard — correct Discord parity
- `is_plain_user_message` whitelist prevents system subtypes from resetting counter (#497 parity)
- Text file caps (5 files / 1MB) match Discord; post-download enforcement handles Slack's `size==0` edge case
- Thread race fix uses string heuristic for error 160004, covered by unit tests
- `max_bot_turns` has `serde(default)` → backward compatible with existing config.toml
- `turn_key` for non-thread Slack messages is unique per message (channel:ts) — effectively a no-op counter, correct since bot loops only happen in threads
- `MAX_CONSECUTIVE_BOT_TURNS` (legacy history-based cap) still used in discord.rs for `AllowBots::All` — separate from `BotTurnTracker`, not touched by PR
- CI all green (10/10 checks), 81 tests pass, no secrets/PII in diff

### Verdict
APPROVE — clean PR with thorough test coverage, well-documented commit history, and faithful Discord→Slack parity.

thepagent (approved):

thepagent:
Thank you @masami-agent ! You did great jobs!

PR #488, opened 2026-04-20
ci: remove issue-check workflow

ChunHao-dev:
## Context

The issue-check workflow only works when users manually add a type label (`bug`, `feature`, etc.). Issues created without labels or without using a template get a generic message with no actionable guidance.

Discord Discussion URL: https://discord.com/channels/1491295327620169908/1491365162869985283/1495727290523324527

## Summary

Enhance the issue-check workflow to auto-detect issue type from template headings, auto-apply labels, and provide direct template links in feedback messages.

## Changes

- Auto-infer issue type from section headings when no label is present (`Steps to Reproduce` → bug, `Use Case` → feature, etc.)
- Auto-apply inferred type label + `needs-triage` label
- Detect RFC issues by `### Proposal` heading (not just `rfc` label)
- Provide direct template links when issue doesn't match any template
- Include template link in missing-fields message for reference

## Design Decisions

- **Heading-based inference over requiring labels** — most users copy template structure but forget to add labels. Detecting headings catches these cases without requiring user action.
- **Auto-label + needs-triage** — inferred labels still need human review, so `needs-triage` is added to flag them.
- **Template links in messages** — instead of telling users to "add a label and edit", direct them to the correct template URL so they can re-create properly.


shaun-agent:
## OpenAB PR Screening
- Title: fix(ci): improve issue-check workflow with auto-labeling and template links
- Source: https://github.com/openabdev/openab/pull/488
- Generated at: 2026-04-20T10:31:23.433Z
- Discord thread: https://discord.com/channels/1488041051187974246/1495733463611871364
<details>
## Intent

PR #488 is trying to make the `issue-check` GitHub Actions workflow useful even when reporters do not perfectly follow repo process. The concrete problem is that issues created without a type label or without correctly using a template currently receive a generic warning that does not tell the reporter what to do next, which increases maintainer triage load and leaves valid reports stalled.

## Feat

This is primarily a `fix(ci)` with a small workflow UX improvement. In plain terms, it teaches the issue validation workflow to infer likely issue type from template headings, auto-apply the inferred label plus `needs-triage`, recognize RFC-style issues from content rather than labels alone, and return direct template links when the issue does not match an expected structure.

## Who It Serves

The primary beneficiaries are maintainers and triagers, with a secondary benefit to GitHub issue authors. Maintainers get cleaner incoming issue classification and less manual labeling work; reporters get clearer guidance when they file malformed or unlabeled issues.

## Rewritten Prompt

Update `.github/workflows/issue-check.yml` so the workflow no longer depends on manually applied type labels alone. When an issue lacks a recognized type label, inspect the issue body for known template headings and infer the type deterministically, then auto-apply the inferred type label and `needs-triage`. Extend RFC detection to recognize `### Proposal` in the issue body, and improve workflow feedback comments so unmatched or incomplete issues include direct links to the appropriate issue templates. Keep the logic explicit and testable, and avoid ambiguous inference paths that could mislabel issues.

## Merge Pitch

This is worth advancing because it reduces avoidable triage friction in one of the highest-volume maintainer touchpoints: incoming issues. The risk profile is moderate and mostly operational, not architectural. The likely reviewer concern is false-positive labeling from weak heading inference, especially if headings overlap across templates or users partially copy template text into unrelated reports.

## Best-Practice Comparison

OpenClaw principles that are relevant here are explicit routing and durable operational behavior, but only at a lightweight level. This PR moves closer to explicit delivery routing by classifying issue intake based on structured content rather than relying on a missing human action. Durable job persistence, isolated executions, and retry/backoff logs are mostly not central here because this is a short-lived GitHub workflow, not a long-running scheduler or job runner.

Hermes Agent principles are more relevant in spirit than in mechanism. The useful parallels are self-contained decision logic and predictable state transitions: the workflow should derive classification from the issue payload itself, and the resulting labels/comments should be deterministic. File locking, atomic persisted state, and fresh session management are not directly applicable because GitHub Actions already provides isolated workflow execution and this PR is not introducing local persisted state.

The practical best-practice takeaway is:
- Relevant: explicit routing, deterministic classification, self-contained prompt/message output, safe fallback behavior.
- Less relevant: durable schedulers, persisted run state, file locks, session lifecycle management.

## Implementation Options

### Option 1: Conservative rule extension in the existing workflow

Keep all logic inside `.github/workflows/issue-check.yml` and add a small set of explicit heading-to-type rules. If no valid label exists, scan for a few known headings such as `Steps to Reproduce`, `Use Case`, and `Proposal`, then apply one inferred label and `needs-triage`. Update the comment text to include hardcoded template links.

This is the fastest path and fits the current PR shape, but the logic may become brittle as templates evolve.

### Option 2: Balanced extraction of classification rules into a clearer script or action step

Move inference and message construction into a dedicated script used by the workflow, while keeping the workflow as the orchestrator. Centralize the mapping of headings to issue types, fallback behavior, and template URLs in one place so it is easier to review and test. Have the workflow call the script, then apply labels and comments based on structured output.

This adds a little implementation overhead but makes the behavior easier to evolve and safer to verify.

### Option 3: Ambitious intake normalization based on template metadata

Replace heading heuristics with a stronger intake model that parses issue forms or template metadata directly, and only falls back to headings for legacy or malformed issues. Standardize type detection, template link generation, and missing-field reporting around a single source of truth derived from the repository’s issue templates.

This is the most robust long-term direction, but it is broader than the immediate PR and likely better as a follow-up sequence rather than the merge target for this item.

## Comparison Table

|---|---|---|---|---|---|---|
| 1. Extend existing workflow rules | Fast | Low | Medium | Medium-Low | Medium | High |
| 2. Extract logic into script/action step | Medium | Medium | Medium-High | High | Medium-High | Highest |
| 3. Template-metadata-driven intake normalization | Slow | High | High | High | High | Medium-Low |

## Recommendation

Recommend Option 2: keep the current product intent, but implement it with a small dedicated classification layer instead of growing ad hoc YAML logic further. That gives OpenAB a cleaner merge candidate now while avoiding a future maintenance problem where every new template tweak requires more opaque workflow conditionals.

For merge discussion, the right near-term sequence is:
1. Land deterministic inference plus template-link feedback in a maintainable structure.
2. Keep the rule set intentionally narrow to reduce false positives.
3. Consider a later follow-up that aligns inference with issue template metadata or issue forms as a stronger source of truth.
</details>

chaodu-agent (changes requested):
The `### Description` heading inference rule has a high false-positive risk — it appears in bug, feature, and documentation templates, so issues missing other headings will be misclassified as `documentation`. Suggest removing `else if (/###\s*Description/i.test(body)) type = 'documentation'` from the inference chain, or gating it with an additional condition that rules out other types first.

See full triage review in the comment above for additional suggestions.

chaodu-agent (changes requested):
The `### Description` heading inference rule carries a high false-positive risk — this heading appears across bug, feature, and documentation templates, so issues that are missing other specific headings will be misclassified as `documentation`. Recommend removing `else if (/###\s*Description/i.test(body)) type = 'documentation'` from the inference chain, or gating it with an additional condition that rules out other template types first.

See the full triage review comment above for additional suggestions.

chaodu-agent:
## PR #488 — Triage Review

The current issue-check workflow relies entirely on users manually adding a type label (`bug`, `feature`, etc.). If a user follows a template but forgets the label, or opens an issue without a template, the workflow returns a generic message with no actionable guidance. This forces maintainers to manually classify issues and leaves valid reports stalled.

### 2. How does it solve it?

- **Heading-based inference**: when no type label is present, regex-scans the issue body for known section headings (`### Steps to Reproduce` → bug, `### Use Case` → feature, etc.) and auto-applies the inferred type label + `needs-triage`
- **Expanded RFC detection**: recognizes `### Proposal` heading in addition to the `rfc` label
- **Template links**: when an issue doesn't match any template, the bot comment now includes direct links to create a new issue from each template; the missing-fields message also links to the relevant template

### 3. Were alternatives considered?

The PR description doesn't mention alternatives explicitly. The screening report evaluated three options:
- Option 1 (current approach): add rules inline in the workflow YAML
- Option 2: extract classification logic into a dedicated script
- Option 3: template-metadata-driven intake normalization

The screening report recommended Option 2, but the author went with Option 1 — the fastest path to ship.

### 4. Is this the best approach right now?

The direction is correct, but there are concerns worth addressing (see triage below).

---

### 🟢 INFO

- Clear intent, solves a real triage pain point
- `needs-triage` paired with auto-labeling is a good design — machine-inferred labels still get human review
- RFC detection expanding from label-only to heading-based is reasonable and backward-compatible
- Template links are a practical UX improvement that reduces friction for issue authors

### 🟡 NIT

- `### Description` appears in the `rules` for `bug`, `feature`, and `documentation`, but the inference chain maps it exclusively to `documentation`. This heading is too generic — nearly any issue could contain it, risking unintended auto-labeling. Maintainers might consider removing `Description` from the inference rules or using a more distinctive heading
- The inference priority order is implicit in the `if/else if` chain — a comment explaining the priority would improve maintainability
- `templateFiles` and `rules` both define the type list independently; adding a new type in the future requires changes in two places

### 🔴 SUGGESTED CHANGES

- **`### Description` false-positive risk**: this is the primary concern. Nearly all issues (including bugs and features) will have a `### Description` heading. If a bug report contains `### Description` but is missing `### Steps to Reproduce` (e.g., the user deleted that section), the inference chain will miss bug but match documentation, resulting in misclassification. Suggest removing `else if (/###\s*Description/i.test(body)) type = 'documentation'` from the inference chain, or at minimum lowering its priority and adding a guard condition
- **No handling for heading conflicts**: if an issue body contains both `### Steps to Reproduce` and `### Use Case` (e.g., copy-paste error), the current logic matches the first hit (bug). This may be acceptable, but it might be worth considering logging or commenting on ambiguous cases so triagers are aware of multiple matches
- **Screening report's Option 2 (extract to script) is worth considering**: the inline JS in the YAML is already over 100 lines, and this PR adds more inference logic. Maintainability will continue to degrade. This doesn't need to happen in this PR, but a follow-up issue to refactor would be valuable

---

> 🔍 Triage review by [超渡法師](https://github.com/apps/openab) — community triage, not maintainer review.

ChunHao-dev:
Updated: changed direction from improving the workflow to removing it entirely.

The automated issue template checking and label management (`incomplete`, `needs-triage`) adds unnecessary friction for contributors. Removing the workflow instead of enhancing it.

Will replace with an AI Agent that handles template checking more intelligently.

## Changes (updated)

- Delete `.github/workflows/issue-check.yml`

thepagent:
closed in favor of the new version

PR #492, opened 2026-04-20
feat(cursor): wire /models /agents /cancel to cursor-agent configOptions

### LINKED ISSUE #489: feat(cursor): wire up Cursor agent to /models, /agents, /cancel slash commands
### Description

Wire the Cursor agent into openab's existing `/models`, `/agents`, `/cancel` Discord slash commands (shipped in PR #478). The Rust side is already drop-in compatible with cursor-agent's standard ACP — what's missing is configuration (`config/cursor.toml`, docker-compose service, `--model auto` startup arg) and a proactive-probe + auto-fallback UX in `/models` to handle Cursor's account-level model gating.

### Use Case

Cursor users on Discord need to switch between Cursor's 26 available models, change agent mode (Agent / Plan / Ask), and cancel in-flight prompts — the same way Codex and Kiro users already can. Today they have no slash command surface at all (`@mention` + free text only), and 11 of the 26 models silently soft-reject at the account level (claude-*, gpt-5.4 family) without the proactive-probe handling proposed in §3.1.

---

## TL;DR

cursor-agent beta (`2026.04.15-dccdccd`) fully supports standard ACP: `session/set_config_option` (model/mode switching) and `session/cancel` both work as expected, and PR #478 already landed the Rust side. What's left is to wire Cursor in: add `config/cursor.toml`, enable a docker-compose service, and extend the `/models` handler to proactively probe after a switch and auto-fallback to Auto on soft reject. `/models` exposes **all 26** models returned by `session/new`, no allowlist, errors surfaced verbatim.

---

### 1.1 Target commands

PR #478 registered three Discord slash commands (plural names):

- `/models` — switch / list models
- `/agents` — switch agent mode (Agent / Plan / Ask)
- `/cancel` — cancel the in-flight prompt / tool call

Skills-level commands (`/simplify`, `/babysit`, etc.) are out of scope.

### 1.2 Current state — PR #478 is done, Cursor isn't wired in

- openab has no slash commands for other agents — `@mention` + free text goes straight to ACP `session/prompt`.
- Cursor and Claude share the same dispatcher (`src/discord.rs` → `src/acp/connection.rs`); the only differences are the binary and config file.
- **PR #478 (0.7.8-beta.8)** already shipped:
  - `session_new()` calls `parse_config_options()` and caches `configOptions`
  - `set_config_option()` implemented and sends `session/set_config_option`
  - `session_cancel()` implemented (fire-and-forget notification)
  - `/models`, `/agents`, `/cancel` registered, handlers drive a select menu → `set_config_option()`
- **So Cursor doesn't need new code — it needs to be connected.** Branch `feat/cursor-agent-cli` has `Dockerfile.cursor` but no `config/cursor.toml` (other agents all have one), and `docker-compose.yml` has no `cursor` service.

### 1.3 cursor-agent side — standard ACP works

Validated inside `openab/openab-cursor` pod with `cursor-agent 2026.04.15-dccdccd`, logged in as `brett.chien@gmail.com`, raw JSON-RPC over stdio:

| Test | Result |
|---|---|
| `initialize` | Returns `protocolVersion:1`, `agentCapabilities` (`loadSession` / `mcpCapabilities` / `promptCapabilities`). No `slashCommands` declaration, but that's OK. |
| `session/new` → `configOptions` | Standard shape; two `type:"select"` options: `mode` (3 values) and `model` (26 values). openab's existing `parse_config_options()` is drop-in compatible. |
| `session/update` → `available_commands_update` | Separately pushes 11 built-in skill-like commands. Not handled by this issue. |
| `session/cancel` | Under stress test, truly interrupts LLM output; `stopReason:"cancelled"` within 24ms. |

---

### 2.1 Standard ACP path works end-to-end — no prompt-passthrough fallback needed

Phase 0 originally planned two paths: (X) structured RPC, (Y) prompt-passthrough fallback. (X) works completely; (Y) dropped. `session/set_config_option` roundtrip, modelId bracket-string fidelity, and `session/cancel` interruption are all verified (see regression bundle below).

### 2.2 `session/new` is the sole source of the 26 availableModels

`session/new` response's `result.models.availableModels` is the primary data source for the `/models` UI (not `configOptions`). Each item has a structured `modelId` (bracketed — this is what `set_config_option` takes as value) plus a short `name` (use as the Discord label).

**Only `session/new` returns it.** Seven plausible alternative RPCs all return `-32601 Method not found`: `session/get_config`, `session/config`, `session/info`, `session/state`, `models/list`, `config/get`, `session/list_models`.

**Server-fetched, not binary-hardcoded.** `/opt/cursor-agent/*.js` has zero hits for 5 representative modelIds; `6849.index.js` initializes `availableModels=[]` and async-fills it from `api2.cursor.sh`. The list reflects cursor-agent backend state (new models, revocations, per-account entitlements) in real time. **Same cursor-agent version on a different Cursor account can return a different list.** This is the core evidence for "do not bake an allowlist".

**CLI 93 vs ACP 26 are two independent namespaces**: naming granularity, accepted input positions, and routability differ. `/models` only uses ACP's 26.

### 2.3 15/26 actually route; the other 11 are account-level soft reject

Full sweep over the 26 bracketed modelIds returned:

- **✅ 15 routable** (setOk=true and prompt replies normally): `default[]`, `composer-2[fast=true]`, `composer-1.5[]`, all `gpt-5.1`~`5.3` + `gpt-5-mini[]`, all `gemini-*`, `grok-4-20[thinking=true]`, `kimi-k2.5[]`
- **❌ 11 broken** (setOk=true, but the next `session/prompt` returns a soft-reject text): all `claude-*` (opus/sonnet/haiku every generation), the `gpt-5.4` family (3), `gpt-5.3-codex-spark`

2026-04-20 re-verified Octocat's three opus variants — all three soft-reject, format stable.

**Hard vs Soft reject (the handler must detect both):**

| | Hard reject | Soft reject |
|---|---|---|
| Signal layer | `set_config_option` returns JSON-RPC error | `setOk=true`; next `session/prompt`'s `agent_message_chunk.content.text` contains plain-text error |
| Error shape | `-32602 Invalid params — Invalid model value: <value>` | No code; text contains `Error: I: AI Model Not Found Model name is not valid: "<base>"` |
| stopReason | N/A | `end_turn` (the turn appears "successful" to openab) |
| Trigger examples | Top-level name / typo / not in availableModels | The 11 broken bracketed modelIds |
| Detection point | `Err` branch of `set_config_option` — read `error.message` | Session-update handler — regex/substring match on `agent_message_chunk.content.text` |

**Detection pattern (suggested):**

```rust
let text = chunk.content.text.as_deref().unwrap_or("");
let is_soft_model_reject = text.contains("Error: I: AI Model Not Found")
    || text.contains("Model name is not valid");
```

**Consequence:** without a probe, the user picks a broken model, sees "switched ✅", and hits the error only on their *next* real prompt — and `currentModelId` stays on the broken value until manually changed. This issue's `/models` UX uses a proactive probe to collapse that delay (see §3.1).

### 2.4 `--model auto` in startup args is load-bearing, not cosmetic

cursor-agent unconditionally overwrites the cli-config `selectedModel` at startup:

| `cursor-agent acp ...` | Post-startup `selectedModel` |
|---|---|
| No `--model` | `composer-2[fast=true]` 🚨 (coding-only model, **not** Auto) |
| `--model auto` | `default[]` (true Auto) |
| `--model <valid-top-level-name>` | The corresponding modelId+parameters |

**Three invariants:**
1. ACP's `configOptions.model.currentValue` (= `currentModelId`) always shows the server-side account default. It ignores local state and startup flags and **does not drive routing** (display only).
2. Pre-writing `cli-config.json` is useless — startup overwrites it.
3. Actual routing = post-startup `selectedModel`, decided by `--model` (or composer-2 if absent).

**`values.yaml` must include `args: [acp, --model, auto]`**, otherwise every route lands on composer-2. `/models` only works within the session via `set_config_option`; restarting the pod won't "reset the model" because startup overwrites.

### 2.5 Schema erratum: the field is `configId`, not `optionId`

Some version of cursor-agent renamed `optionId` → `configId` in `session/set_config_option`. Current version (`2026.04.15-dccdccd`) requires `configId`; sending `optionId` returns `-32603 Internal error — path: ["configId"] Invalid input` before availableValues is even validated. `src/acp/connection.rs:377` already uses `configId`. Re-run the regression bundle on every cursor-agent bump.

---

### 3.1 `/models` UX — "expose all 26 + proactive probe + auto-fallback"

**Decision**: `/models` shows all 26, no allowlist, validates routing with a probe prompt immediately after switching, auto-falls-back to Auto on failure.

**Full flow:**

```
User picks <modelId> from Discord select menu
  │
  ▼
openab defer_ephemeral()
  │
  ▼
pool.get_or_create(thread) — ensure active connection / populate config_options cache
  │
  ├─ Err (spawn / initialize / session_new failed)
  │    └─► edit_response("❌ Cannot start cursor session: <error>")
  │
  └─ Ok
       │
       ▼
set_config_option(configId="model", value=<modelId>)
  │
  ├─ Err (hard reject: -32602 Invalid model value)
  │    └─► edit_response("❌ Failed to switch: <error.message>")
  │
  └─ Ok (setOk=true)
       │
       ▼
     openab sends probe prompt "ping"
     ※ probe's agent_message_chunk is NOT forwarded to Discord — collected in an in-memory buffer
     ※ only cares whether the buffer contains soft-reject strings, not the content itself
     ※ bounded by probe_timeout_secs (default 15s)
       │
       ├─ Timeout (no stopReason)
       │    └─► treat as unavailable → auto set_config_option(value="default[]")
       │         └─► edit_response("⏱ <name> probe timed out (>{N}s), switched back to Auto")
       │
       └─ prompt finishes (stopReason=end_turn)
            │
            ├─ buffer contains "Error: I: AI Model Not Found" / "Model name is not valid"
            │    └─► soft reject detected → auto set_config_option(value="default[]")
            │         edit_response("⚠️ <name> unavailable (Cursor: <original error>), switched back to Auto")
            │
            └─ buffer is a normal response → edit_response("✅ Switched to <name>")
```

**Spec details:**

1. **Probe prompt content**: `"ping"`. Any prompt triggers Cursor's routing attempt, so broken models will soft-reject regardless of content; use the shortest string to save tokens. The success-path response doesn't display the agent's self-identification — that lies (Cursor Auto routes to composer which self-identifies as gpt; thinking mode goes through long deliberation first). openab already knows `<modelId>`, showing `<name>` is sufficient.
2. **Probe's `agent_message_chunk` is not forwarded to Discord** — the channel sees nothing from the probe turn.
3. **No cache** — every `/models` invocation probes (simplicity first).
4. **No failsafe on Auto fallback** — assumes `default[]` always routes (consistently true in Phase 0). Fallback-to-Auto failure is not handled.
5. **Race conditions not handled** — if the user presses `/models` while a prompt is in flight, the ACP layer serializes naturally; if we observe issues, add a guard.
6. **Probe timeout**: **15s default**, controlled by `probe_timeout_secs` in `config/cursor.toml` (see 3.2 Step 1). Timeout is treated as unavailable and routed through the soft-reject branch. Rationale: even if the probe hangs (network / backend / Cursor internal bug), `/models` must leave the user in a usable state rather than hanging the Discord interaction.

**Config implication**: `config/cursor.toml` does **not** include `allowed_model_values`; openab does not implement an allowlist filter. The list comes from `session/new` every time.

**Why this over the alternatives**:
- Lowest maintenance cost (no toml churn as Cursor adds/removes models)
- The 11 broken ones are server-side account gating — a static list goes stale silently
- Proactive probe eliminates the "soft reject shows up next turn" UX problem
- Auto fallback guarantees Discord always lands in a usable state

**Rejected alternatives**: (b) Auto-only (too conservative); (c) 15-item allowlist (maintenance burden); passive error passthrough only (soft reject surfaces one turn late); (d) runtime probe as a filter (cold-start too expensive); (e) heuristic exclusion (silently drifts).

### 3.2 Acceptance checklist

- [ ] **Step 0** — `src/discord.rs`: **(i) defer refactor**: convert all three PR #478 handlers (`handle_config_command` / `handle_config_select` / `handle_cancel_command`) from `create_response` to `defer_ephemeral()` + `edit_response()`. Avoids Discord's 3-second interaction deadline — select submission does a `set_config_option` RPC + probe turn and will trip the timeout under cold start / network jitter. **(ii) `get_or_create` pre-flight**: in `handle_config_command` and `handle_config_select`, after defer, first call `pool.get_or_create(&thread_key).await?`. Handles two scenarios: (a) **missing-session** (user has never `@mentioned` the bot, so no active connection exists in the pool) → spawn a session and populate `config_options` cache, replacing today's "Start a conversation first by @mentioning the bot" guidance; (b) **evicted-session** (session once existed and was evicted to `suspended` because `max_sessions` was hit) → attempt `session/load` resume, Cursor falls back to `session/new` (see risk 9). On failure: `edit_response("❌ Cannot start cursor session: <error>")`. `handle_cancel_command` does **not** need `get_or_create` — cancellation uses independent `cancel_handles` (pool.rs:278).
- [ ] **Step 0b** — `src/discord.rs` + `src/acp/connection.rs`: **Proactive probe + auto-fallback** (flow diagram above). (i) When `set_config_option` returns a JSON-RPC error, edit_response passes through `error.message`. (ii) On `setOk=true`, immediately send a probe prompt `"ping"` in in-memory-buffer mode (not forwarded to Discord), collecting `agent_message_chunk` until `stopReason=end_turn`, bounded by `probe_timeout_secs` (default 15s). (iii) If the buffer contains `Error: I: AI Model Not Found` or `Model name is not valid`, auto-send `set_config_option(value="default[]")` to fall back to Auto, and edit_response with `⚠️ <name> unavailable (Cursor: <original error>), switched back to Auto`. (iv) On timeout, same fallback, edit_response with `⏱ <name> probe timed out (>{N}s), switched back to Auto`. (v) On no soft-reject, edit_response `✅ Switched to <name>`. **No allowlist filter** — list comes from every `session/new` response; Auto is assumed to always route.
- [ ] **Step 1** — New `config/cursor.toml`: `command = "cursor-agent"`, `args = ["acp", "--model", "auto"]` (§2.4 is mandatory), `working_dir = "/home/agent"`, `probe_timeout_secs = 15` (§3.1 #6). Other fields mirror `claude.toml`. **Do not add `allowed_model_values`**.
- [ ] **Step 2** — `docker-compose.yml`: add `cursor` service (profile=cursor), mount `config/cursor.toml` and data volume.
- [ ] **Step 3** — `.env.example`: add `CURSOR_*` variables (Discord bot token, Cursor login credentials).
- [ ] **Step 4** — `Dockerfile.cursor` already exists; verify that `cursor-agent login` state survives container restart. Octocat solves this with a hostPath mount; other deployments need a PVC (see risk 4).
- [ ] **Step 5** — Verify `parse_config_options()` accepts Cursor's `configOptions` (Phase 0 showed drop-in compatibility; should pass through).
- [ ] **Step 6** — Verify modelId roundtrip. Cursor's modelId is bracketed (`claude-opus-4-7[thinking=true,...]`). Confirm `set_config_option(value=...)` in openab doesn't filter/escape the payload. Already verified on the Discord side: PR #478 uses `custom_id = "acp_config_{category}"` and `values.first().clone()` for passthrough; ACP layer preserves bracket bytes (Phase 0 verified).
- [ ] **Step 7** — `docs/cursor.md`: update the "Discord Slash Commands" section. List the three commands and the model/mode options, and note "picking a model that the current account can't route triggers auto-fallback to Auto with the Cursor-side error included in the response".
- [ ] **Step 8** — PR: `config/cursor.toml` + `docker-compose.yml` changes go into `openabdev/openab`; Helm chart mount adjustments in a separate PR.

### 3.3 E2E validation scenarios

1. **Happy path** — pick a routable model (e.g. `gpt-5.1[reasoning=medium]`) → `✅ Switched to gpt-5.1` followup; probe (`ping`) does not leak into the channel.
2. **Fallback path** — pick a broken model (e.g. any opus variant on Octocat) → `⚠️ <name> unavailable... switched back to Auto`; no probe garbage in the channel.
3. **Consistency** — immediately send a real prompt after switching → verify `currentModelId` matches the followup (the picked model on success, Auto on fallback).
4. **Cancel chain** — submit a long task → `/cancel` → observe cursor stops immediately (Phase 0 validated the ACP layer; this re-validates the Discord → ACP chain).

### 3.4 Risks and notes

1. **modelId encoding vs Discord's 100-char limit — verified safe** (PR #478 merged `src/discord.rs`):
   - `custom_id` = `acp_config_{category}` (e.g. `acp_config_model`, 15 chars), does **not** embed the modelId
   - modelId lives in the select option `value` field (Discord cap 100 chars, longest is 66 — safe)
   - `handle_config_select` uses `values.first().clone()` — pure passthrough, no sanitization / no index translation
   - `set_config_option(config_id, selected_value)` goes straight to ACP; `[`, `]`, `=`, `,` are preserved byte-for-byte (ACP-layer fidelity also verified in Phase 0)
2. **configOptions may change mid-session** — Cursor pushes `available_commands_update` notifications; in theory `configOptions` / `availableModels` could be updated the same way (not yet verified). PR #478 doesn't handle runtime updates; if the user switches and the UI diverges from the agent's actual state, add a handler later.
3. **Login persistence** — Octocat currently mounts `/Users/brett.chien/.../openab/octocat → /home/agent` via hostPath, so tokens persist on the host and survive pod rebuilds. Local dev: no change. But for a shared chart deployment, we need a PVC and a persistence policy for `~/.config/cursor/`.
4. **`--model auto` must be specified** (§2.4) — writing `args: [acp]` silently routes everything through composer-2.
5. **ACP schema drift** (§2.5) — changes to `configId` / `availableModels` shape / the 15|11 split affect `src/acp/connection.rs` cache/parser and the soft-reject string match. On every cursor-agent bump, run the regression bundle before touching code.
6. **Soft-reject string changes** — `Error: I: AI Model Not Found` / `Model name is not valid` is plain text, not schema. If Cursor rewords, Step 0b's detection must update — the regression bundle will catch this.
7. **Auto fallback failure is not handled** — if Auto itself is broken, the session stays in the soft-reject state (edit_response surfaces Cursor's original error, no third fallback). Assumes `default[]` always routes (consistently true in Phase 0).
8. **Probe turn pollutes session history — verified, "accept" policy**: `/tmp/acp_probe_history_test.js` confirmed the probe's `"ping"` → `"pong"` exchange stays in the ACP session context (asked "what was the previous word?", the assistant recalls `"ping"`). Impact: ~2–5 tokens of noise, one turn. **Does not affect routing** (routing is driven by `cli-config.selectedModel`, not session history). Not switching to a "recreate session" approach, because that would drop the user's ongoing conversation — a worse tradeoff than the noise. If probe pollution ever materially affects model output quality, revisit (a cursor-agent `session/discard_last_turn`-style API would be the cleanest fix).
9. **cursor-agent `loadSession:true` capability is misleading — verified, existing fallback covers it**: `agentCapabilities.loadSession=true` advertises cross-process session resume, but testing (`acp_session_load_probe.js`) on cursor-agent `2026.04.15-dccdccd` shows sessionIds are **only valid within the lifetime of the same `cursor-agent acp` process**; after the process dies, `session/load` with that sessionId returns `-32602 Invalid params: Session "<uuid>" not found`. Consequences: (i) `src/acp/pool.rs` already falls back from `session/load` to `session/new` on failure (pool.rs:152–156) — correct behavior, no code change needed; (ii) but Cursor users will get a **fresh session (losing conversation history) after pool eviction**, unlike Claude or other agents that honor `loadSession`; (iii) `session/load` also cannot be used as a cheap "refresh availableModels without losing session state" mechanism. Run `acp_session_load_probe.js` on every cursor-agent bump to see if this gets fixed.

---

## References

- PR #478 (merged): https://github.com/openabdev/openab/pull/478
- Related issue: #477
- Phase 0 regression bundle (raw JSON-RPC captures + re-runnable scripts): `cursor-agent-acp-regression/` — 12 scripts covering routing, modelId bracket fidelity, binary hardwire check, CLI-vs-ACP namespace, probe history pollution, and `session/load` cross-process behavior. Run on every cursor-agent version bump.

brettchien:
Closes #489.

Discussion: https://discord.com/channels/1491295327620169908/1491365150664560881/1495774750226645133

Upstream cursor-agent ACP compliance gaps tracked separately in #493.

## Summary

- Discord slash commands `/models`, `/agents`, `/cancel` now work end-to-end against `cursor-agent acp`.
- Each handler: `defer_ephemeral` → `pool.get_or_create(thread_key)` → build select menu from the session's cached `configOptions` → `edit_response`. No more "start a conversation first" dead-end, and the 3s Discord interaction deadline is honoured even on cold-start sessions.
- StringSelectMenu capped at Discord's 25-option limit with a plain `take(25)` — no vendor-specific pinning. cursor-agent returns 26 models on a Pro account; the last one is dropped.
- `agent` / `mode` category treated as aliases so the same `/agents` handler works against both kiro-cli (`category=agent`) and cursor-agent (`category=mode`). Per ACP spec the reserved value is `"mode"`.
- Trust-the-protocol: OpenAB does not probe or auto-rewrite after `set_config_option`. If cursor-agent accepts an unroutable model with `setOk:true` and then soft-rejects on the next prompt, that soft-reject is passed through to the channel. Upstream bug tracked in #493.

## Changes

- `src/discord.rs` — new `handle_config_command` (shared by /models and /agents), `handle_config_select`, `handle_cancel_command`; `build_config_select` with ACP-agnostic 25-option cap; category alias table.
- `config.toml.example` — documents the `--model auto` startup gotcha in the cursor-agent section.
- `docs/cursor.md` — Model Selection section with startup `--model auto` table and passthrough semantics; new Known Limitations subsection listing three upstream cursor-agent ACP compliance gaps (all linked to #493).

No changes in `src/acp/connection.rs`, `src/acp/pool.rs`, or `src/config.rs` — ACP layer untouched.

## Test plan

- [x] `cargo check` clean
- [x] `cargo test --release` — 74/74 pass
- [x] Manual E2E against `openab-cursor` pod running cursor-agent `2026.04.15-dccdccd`:
  - [x] `/models` lists 25 models (truncated from 26) with the current value highlighted
  - [x] Switching to a routable model (`composer-2`) → `✅ Switched to composer-2`, next prompt routes there
  - [x] `/agents` lists Agent / Plan / Ask (regression from v0.7.8 where it returned "No agent options available" against cursor-agent)
  - [x] `/cancel` against an in-flight prompt → `session/cancel` → `stopReason: cancelled` (~15ms roundtrip, verified in pod logs)
  - [x] Soft-reject passthrough observed — cursor-agent streamed `"\n\nCheck your settings to continue"` for an unroutable model and OpenAB forwarded it unchanged (new variant beyond the `AI Model Not Found` / `Model name is not valid` family, noted for #493)

## Revision history

- **v1**: included a proactive per-model probe that sent a silent `ping`, detected cursor-agent's plain-text soft-reject, and auto-fell-back to `default[]` on hit.
- **v2** (per maintainer review): probe + fallback removed. That policy belongs upstream, not in OpenAB; the trust-the-protocol approach keeps `set_config_option` semantics intact and surfaces the upstream bug directly in-channel.
- **v3**: truncation pinning (current + `"default[]"`) also removed — kept as a plain in-order `take(25)` so `build_config_select` has no vendor literals.

### DIFF
--- a/config.toml.example
+++ b/config.toml.example
@@ -32,6 +32,10 @@ allowed_channels = ["1234567890"]       # ↑ omitted + non-empty list → auto-
 command = "kiro-cli"
 args = ["acp", "--trust-all-tools"]
 working_dir = "/home/agent"
+# probe_timeout_secs = 15  # /models probe budget (default 15). After a model
+#                          # switch, openab sends a silent "ping" prompt to
+#                          # confirm routing works; on timeout or soft reject
+#                          # (e.g. "AI Model Not Found"), it falls back to Auto.
 
 # [agent]
 # command = "claude"
@@ -71,6 +75,8 @@ working_dir = "/home/agent"
 # args = ["acp", "--model", "auto", "--workspace", "/home/agent"]
 # working_dir = "/home/agent"
 # env = {}  # Auth via: kubectl exec -it <pod> -- cursor-agent login
+# # `--model auto` is load-bearing: omitting it routes all traffic through
+# # composer-2 regardless of /models selections.
 
 [pool]
 max_sessions = 10
--- a/config.toml.example
+++ b/config.toml.example
@@ -32,10 +32,6 @@ allowed_channels = ["1234567890"]       # ↑ omitted + non-empty list → auto-
 command = "kiro-cli"
 args = ["acp", "--trust-all-tools"]
 working_dir = "/home/agent"
-# probe_timeout_secs = 15  # /models probe budget (default 15). After a model
-#                          # switch, openab sends a silent "ping" prompt to
-#                          # confirm routing works; on timeout or soft reject
-#                          # (e.g. "AI Model Not Found"), it falls back to Auto.
 
 # [agent]
 # command = "claude"
--- a/docs/cursor.md
+++ b/docs/cursor.md
@@ -73,25 +73,52 @@ helm install openab openab/openab \
 
 ## Model Selection
 
-List available models:
+### Startup default: `--model auto`
 
-```bash
-cursor-agent --list-models
-# or
-cursor-agent models
+`--model auto` in the `args` is load-bearing, not cosmetic. cursor-agent unconditionally overwrites its cli-config `selectedModel` at startup:
+
+| startup `args` | post-startup default |
+|---|---|
+| `["acp"]` (no `--model`) | `composer-2[fast=true]` — Cursor's coding-only model, **not** Auto |
+| `["acp", "--model", "auto"]` | `default[]` — true Auto (recommended) |
+| `["acp", "--model", "<name>"]` | the matching modelId + parameters |
+
+So **omit `--model auto` and every route lands on composer-2** regardless of what `/models` shows. `/models` switches only affect the live session; restarting the pod resets to whatever `args` dictates.
+
+### At-runtime `/models` (Discord slash command)
+
+Once the `openab-cursor` pod is running and a user has started a thread (by @mentioning the bot or any other trigger), three slash commands are available:
+
+- `/models` — switch the model for the current channel's session
+- `/agents` — switch agent mode (Agent / Plan / Ask)
+- `/cancel` — interrupt the in-flight prompt
+
+`/models` lists **all models the cursor-agent backend returns for the logged-in account** — the server-fetched list, not a baked-in allowlist. The list reflects account entitlements in real time, so a different Cursor account on the same binary can see a different list.
+
+#### Proactive probe + auto-fallback
+
+Some of the bracketed modelIds (notably the `claude-*` family and the `gpt-5.4` family on certain accounts) accept `session/set_config_option` successfully but then stream a plain-text error on the very next prompt:
+
+```
+Error: I: AI Model Not Found
+Model name is not valid: "<name>"
 ```
 
-To specify a model, pass `--model` as an arg:
+To collapse that one-turn delay, openab sends a silent probe prompt (`"ping"`) immediately after each model switch. The probe's response is captured in-memory and **never forwarded to the Discord channel**. On soft-reject detection (`AI Model Not Found` or `Model name is not valid`), openab auto-switches back to Auto (`default[]`) and surfaces the Cursor-side error in the Discord followup:
 
-```toml
-[agent]
-command = "cursor-agent"
-args = ["acp", "--model", "auto"]
+```
+⚠️ <name> unavailable (Cursor: <original error>), switched back to Auto
 ```
 
-In ACP mode, `--model` can be appended after `acp`. If omitted, the account default is used.
+Probe timeout is **15 seconds by default**, configurable via `probe_timeout_secs` in the `[agent]` section of `config.toml`. On timeout, the probe is cancelled via `session/cancel` and the same auto-fallback path runs:
 
-To verify which model is active, ask the agent "who are you" — the underlying model will typically self-identify (e.g. "I am Gemini, a large language model built by Google.").
+```
+⏱ <name> probe timed out (>{N}s), switched back to Auto
+```
+
+#### Asking "who are you"
+
+Cursor Auto routes to composer, which self-identifies as GPT. Thinking-mode models deliberate before replying. Treat `/models`'s followup message (derived from cursor-agent's structured `modelId`) as the source of truth, not the agent's self-identification.
 
 ## MCP Usage (ACP mode caveats)
 
--- a/docs/cursor.md
+++ b/docs/cursor.md
@@ -95,26 +95,18 @@ Once the `openab-cursor` pod is running and a user has started a thread (by @men
 
 `/models` lists **all models the cursor-agent backend returns for the logged-in account** — the server-fetched list, not a baked-in allowlist. The list reflects account entitlements in real time, so a different Cursor account on the same binary can see a different list.
 
-#### Proactive probe + auto-fallback
+#### Soft rejects from unroutable models
 
-Some of the bracketed modelIds (notably the `claude-*` family and the `gpt-5.4` family on certain accounts) accept `session/set_config_option` successfully but then stream a plain-text error on the very next prompt:
+Some modelIds (notably the `claude-*` family and parts of the `gpt-5.4`/`gpt-5.3-codex-spark` family on certain accounts) accept `session/set_config_option` with `setOk:true` but then stream a plain-text error on the very next `session/prompt`:
 
 ```
 Error: I: AI Model Not Found
 Model name is not valid: "<name>"
 ```
 
-To collapse that one-turn delay, openab sends a silent probe prompt (`"ping"`) immediately after each model switch. The probe's response is captured in-memory and **never forwarded to the Discord channel**. On soft-reject detection (`AI Model Not Found` or `Model name is not valid`), openab auto-switches back to Auto (`default[]`) and surfaces the Cursor-side error in the Discord followup:
+OpenAB **does not work around this** — per project policy, we trust the ACP protocol: `setOk:true` means the switch succeeded, so `/models` shows `✅ Switched to <name>`. The subsequent soft reject is passed through to the channel as the next turn's agent response. If you see that error, run `/models` again and pick Auto (`default[]`) or another routable model.
 
-```
-⚠️ <name> unavailable (Cursor: <original error>), switched back to Auto
-```
-
-Probe timeout is **15 seconds by default**, configurable via `probe_timeout_secs` in the `[agent]` section of `config.toml`. On timeout, the probe is cancelled via `session/cancel` and the same auto-fallback path runs:
-
-```
-⏱ <name> probe timed out (>{N}s), switched back to Auto
-```
+This is a cursor-agent ACP compliance gap tracked in [#493](https://github.com/openabdev/openab/issues/493); see also the Cursor forum thread linked there. Section **Known Limitations** below lists the full set of upstream issues.
 
 #### Asking "who are you"
 
@@ -176,7 +168,16 @@ kubectl exec deployment/openab-cursor -- cursor-agent mcp list
 
 ## Known Limitations
 
+### Packaging / auth
 - Cursor Agent CLI is a separate distribution from Cursor Desktop — they are not the same binary
 - No official apt/yum package; the Dockerfile downloads a pinned tarball directly
 - `cursor-agent login` requires an interactive terminal for the device flow
 - Auth token persistence requires a PVC mount at the user home directory
+
+### Upstream cursor-agent ACP compliance gaps
+
+The following are cursor-agent behaviours that violate ACP semantics. OpenAB does not work around them in-tree; they are tracked in [#493](https://github.com/openabdev/openab/issues/493) and reported upstream via the Cursor forum (link in that issue).
+
+- **`setOk:true` for unroutable models** — `set_config_option` accepts models the account cannot route, surfacing the error one turn later as a plain-text soft reject (`AI Model Not Found` / `Model name is not valid`).
+- **Unfiltered model list** — `session/new` returns 26 models on Pro accounts, of which ~11 are not routable for the logged-in account. The server knows entitlements and should filter.
+- **`loadSession:true` capability is process-local** — `initialize` advertises `agentCapabilities.loadSession=true`, but `session/load` with an id from a previous cursor-agent process returns `-32602 Session not found`. Session IDs do not survive restarts, so pool-eviction and pod-restart both force a fresh session.
--- a/src/acp/connection.rs
+++ b/src/acp/connection.rs
@@ -89,6 +89,15 @@ pub enum ContentBlock {
     Image { media_type: String, data: String },
 }
 
+/// Outcome of a silent probe turn used to validate a model switch.
+#[derive(Debug, Clone, Default)]
+pub struct ProbeResult {
+    /// Concatenated `agent_message_chunk.content.text` from the probe turn.
+    pub buffer: String,
+    /// True if the probe exceeded the timeout and was cancelled.
+    pub timed_out: bool,
+}
+
 impl ContentBlock {
     pub fn to_json(&self) -> Value {
         match self {
@@ -460,6 +469,106 @@ impl AcpConnection {
     pub async fn prompt_done(&mut self) {
         *self.notify_tx.lock().await = None;
         self.last_active = Instant::now();
+    }
+
+    /// Send a probe prompt and collect the agent's streaming text into an
+    /// in-memory buffer without exposing it to the caller's session.
+    ///
+    /// Used by `/models` to validate that a freshly-switched model actually
+    /// routes. On timeout, sends `session/cancel` to free the stream. The
+    /// subscriber channel is reset before returning so a subsequent real
+    /// prompt picks up a clean notify pipe.
+    pub async fn probe_prompt(
+        &mut self,
+        text: &str,
+        timeout_secs: u64,
+    ) -> Result<ProbeResult> {
+        self.last_active = Instant::now();
+        let session_id = self
+            .acp_session_id
+            .as_ref()
+            .ok_or_else(|| anyhow!("no session"))?
+            .clone();
+
+        let (tx, mut rx) = mpsc::unbounded_channel();
+        *self.notify_tx.lock().await = Some(tx);
+
+        let id = self.next_id();
+        let req = JsonRpcRequest::new(
+            id,
+            "session/prompt",
+            Some(json!({
+                "sessionId": session_id,
+                "prompt": [{"type": "text", "text": text}],
+            })),
+        );
+        let data = serde_json::to_string(&req)?;
+        let (resp_tx, _resp_rx) = oneshot::channel();
+        self.pending.lock().await.insert(id, resp_tx);
+
+        if let Err(e) = self.send_raw(&data).await {
+            self.pending.lock().await.remove(&id);
+            *self.notify_tx.lock().await = None;
+            return Err(e);
+        }
+
+        let mut buffer = String::new();
+        let collect = async {
+            while let Some(msg) = rx.recv().await {
+                if msg.id == Some(id) {
+                    break;
+                }
+                if msg.method.as_deref() != Some("session/update") {
+                    continue;
+                }
+                let Some(params) = msg.params.as_ref() else {
+                    continue;
+                };
+                let Some(update) = params.get("update") else {
+                    continue;
+                };
+                let kind = update.get("sessionUpdate").and_then(|v| v.as_str());
+                if kind != Some("agent_message_chunk") {
+                    continue;
+                }
+                if let Some(t) = update
+                    .get("content")
+                    .and_then(|c| c.get("text"))
+                    .and_then(|t| t.as_str())
+                {
+                    buffer.push_str(t);
+                }
+            }
+        };
+
+        let mut timed_out = false;
+        if tokio::time::timeout(std::time::Duration::from_secs(timeout_secs), collect)
+            .await
+            .is_err()
+        {
+            timed_out = true;
+            let cancel = serde_json::to_string(&json!({
+                "jsonrpc": "2.0",
+                "method": "session/cancel",
+                "params": {"sessionId": session_id},
+            }))?;
+            let _ = self.send_raw(&cancel).await;
+            // Drain until the prompt response arrives so the next turn starts clean.
+            let _ = tokio::time::timeout(std::time::Duration::from_secs(3), async {
+                while let Some(msg) = rx.recv().await {
+                    if msg.id == Some(id) {
+                        break;
+                    }
+                }
+            })
+            .await;
+        }
+
+        *self.notify_tx.lock().await = None;
+        self.pending.lock().await.remove(&id);
+        self.last_active = Instant::now();
+
+        Ok(ProbeResult { buffer, timed_out })
     }
 
     /// Return a clone of the stdin handle for lock-free cancel.
--- a/src/acp/connection.rs
+++ b/src/acp/connection.rs
@@ -89,15 +89,6 @@ pub enum ContentBlock {
     Image { media_type: String, data: String },
 }
 
-/// Outcome of a silent probe turn used to validate a model switch.
-#[derive(Debug, Clone, Default)]
-pub struct ProbeResult {
-    /// Concatenated `agent_message_chunk.content.text` from the probe turn.
-    pub buffer: String,
-    /// True if the probe exceeded the timeout and was cancelled.
-    pub timed_out: bool,
-}
-
 impl ContentBlock {
     pub fn to_json(&self) -> Value {
         match self {
@@ -469,106 +460,6 @@ impl AcpConnection {
     pub async fn prompt_done(&mut self) {
         *self.notify_tx.lock().await = None;
         self.last_active = Instant::now();
-    }
-
-    /// Send a probe prompt and collect the agent's streaming text into an
-    /// in-memory buffer without exposing it to the caller's session.
-    ///
-    /// Used by `/models` to validate that a freshly-switched model actually
-    /// routes. On timeout, sends `session/cancel` to free the stream. The
-    /// subscriber channel is reset before returning so a subsequent real
-    /// prompt picks up a clean notify pipe.
-    pub async fn probe_prompt(
-        &mut self,
-        text: &str,
-        timeout_secs: u64,
-    ) -> Result<ProbeResult> {
-        self.last_active = Instant::now();
-        let session_id = self
-            .acp_session_id
-            .as_ref()
-            .ok_or_else(|| anyhow!("no session"))?
-            .clone();
-
-        let (tx, mut rx) = mpsc::unbounded_channel();
-        *self.notify_tx.lock().await = Some(tx);
-
-        let id = self.next_id();
-        let req = JsonRpcRequest::new(
-            id,
-            "session/prompt",
-            Some(json!({
-                "sessionId": session_id,
-                "prompt": [{"type": "text", "text": text}],
-            })),
-        );
-        let data = serde_json::to_string(&req)?;
-        let (resp_tx, _resp_rx) = oneshot::channel();
-        self.pending.lock().await.insert(id, resp_tx);
-
-        if let Err(e) = self.send_raw(&data).await {
-            self.pending.lock().await.remove(&id);
-            *self.notify_tx.lock().await = None;
-            return Err(e);
-        }
-
-        let mut buffer = String::new();
-        let collect = async {
-            while let Some(msg) = rx.recv().await {
-                if msg.id == Some(id) {
-                    break;
-                }
-                if msg.method.as_deref() != Some("session/update") {
-                    continue;
-                }
-                let Some(params) = msg.params.as_ref() else {
-                    continue;
-                };
-                let Some(update) = params.get("update") else {
-                    continue;
-                };
-                let kind = update.get("sessionUpdate").and_then(|v| v.as_str());
-                if kind != Some("agent_message_chunk") {
-                    continue;
-                }
-                if let Some(t) = update
-                    .get("content")
-                    .and_then(|c| c.get("text"))
-                    .and_then(|t| t.as_str())
-                {
-                    buffer.push_str(t);
-                }
-            }
-        };
-
-        let mut timed_out = false;
-        if tokio::time::timeout(std::time::Duration::from_secs(timeout_secs), collect)
-            .await
-            .is_err()
-        {
-            timed_out = true;
-            let cancel = serde_json::to_string(&json!({
-                "jsonrpc": "2.0",
-                "method": "session/cancel",
-                "params": {"sessionId": session_id},
-            }))?;
-            let _ = self.send_raw(&cancel).await;
-            // Drain until the prompt response arrives so the next turn starts clean.
-            let _ = tokio::time::timeout(std::time::Duration::from_secs(3), async {
-                while let Some(msg) = rx.recv().await {
-                    if msg.id == Some(id) {
-                        break;
-                    }
-                }
-            })
-            .await;
-        }
-
-        *self.notify_tx.lock().await = None;
-        self.pending.lock().await.remove(&id);
-        self.last_active = Instant::now();
-
-        Ok(ProbeResult { buffer, timed_out })
     }
 
     /// Return a clone of the stdin handle for lock-free cancel.
--- a/src/acp/pool.rs
+++ b/src/acp/pool.rs
@@ -1,4 +1,4 @@
-use crate::acp::connection::AcpConnection;
+use crate::acp::connection::{AcpConnection, ProbeResult};
 use crate::acp::protocol::ConfigOption;
 use crate::config::AgentConfig;
 use anyhow::{anyhow, Result};
@@ -268,6 +268,33 @@ impl SessionPool {
         };
         let mut conn = conn.lock().await;
         conn.set_config_option(config_id, value).await
+    }
+
+    /// Seconds to wait for a silent probe turn before declaring the switched
+    /// model unavailable. Sourced from [`AgentConfig::probe_timeout_secs`].
+    pub fn probe_timeout_secs(&self) -> u64 {
+        self.config.probe_timeout_secs
+    }
+
+    /// Send a silent probe prompt and return what the agent streamed back.
+    /// Used by `/models` to validate a fresh model selection without surfacing
+    /// the probe turn to Discord.
+    pub async fn probe_session(
+        &self,
+        thread_id: &str,
+        text: &str,
+        timeout_secs: u64,
+    ) -> Result<ProbeResult> {
+        let conn = {
+            let state = self.state.read().await;
+            state
+                .active
+                .get(thread_id)
+                .cloned()
+                .ok_or_else(|| anyhow!("no connection for thread {thread_id}"))?
+        };
+        let mut conn = conn.lock().await;
+        conn.probe_prompt(text, timeout_secs).await
     }
 
     /// Cancel the current in-flight operation for a session.
--- a/src/acp/pool.rs
+++ b/src/acp/pool.rs
@@ -1,4 +1,4 @@
-use crate::acp::connection::{AcpConnection, ProbeResult};
+use crate::acp::connection::AcpConnection;
 use crate::acp::protocol::ConfigOption;
 use crate::config::AgentConfig;
 use anyhow::{anyhow, Result};
@@ -268,33 +268,6 @@ impl SessionPool {
         };
         let mut conn = conn.lock().await;
         conn.set_config_option(config_id, value).await
-    }
-
-    /// Seconds to wait for a silent probe turn before declaring the switched
-    /// model unavailable. Sourced from [`AgentConfig::probe_timeout_secs`].
-    pub fn probe_timeout_secs(&self) -> u64 {
-        self.config.probe_timeout_secs
-    }
-
-    /// Send a silent probe prompt and return what the agent streamed back.
-    /// Used by `/models` to validate a fresh model selection without surfacing
-    /// the probe turn to Discord.
-    pub async fn probe_session(
-        &self,
-        thread_id: &str,
-        text: &str,
-        timeout_secs: u64,
-    ) -> Result<ProbeResult> {
-        let conn = {
-            let state = self.state.read().await;
-            state
-                .active
-                .get(thread_id)
-                .cloned()
-                .ok_or_else(|| anyhow!("no connection for thread {thread_id}"))?
-        };
-        let mut conn = conn.lock().await;
-        conn.probe_prompt(text, timeout_secs).await
     }
 
     /// Cancel the current in-flight operation for a session.
--- a/src/config.rs
+++ b/src/config.rs
@@ -164,6 +164,11 @@ pub struct AgentConfig {
     pub working_dir: String,
     #[serde(default)]
     pub env: HashMap<String, String>,
+    /// Upper bound (seconds) on how long `/models` waits for a probe turn
+    /// after switching. Timeout is treated as "model unavailable" and
+    /// triggers the auto-fallback to Auto path. Default 15.
+    #[serde(default = "default_probe_timeout_secs")]
+    pub probe_timeout_secs: u64,
 }
 
 #[derive(Debug, Deserialize)]
@@ -224,6 +229,7 @@ fn default_working_dir() -> String { "/tmp".into() }
 fn default_max_sessions() -> usize { 10 }
 fn default_ttl_hours() -> u64 { 4 }
 fn default_true() -> bool { true }
+fn default_probe_timeout_secs() -> u64 { 15 }
 
 fn emoji_queued() -> String { "👀".into() }
 fn emoji_thinking() -> String { "🤔".into() }
--- a/src/config.rs
+++ b/src/config.rs
@@ -164,11 +164,6 @@ pub struct AgentConfig {
     pub working_dir: String,
     #[serde(default)]
     pub env: HashMap<String, String>,
-    /// Upper bound (seconds) on how long `/models` waits for a probe turn
-    /// after switching. Timeout is treated as "model unavailable" and
-    /// triggers the auto-fallback to Auto path. Default 15.
-    #[serde(default = "default_probe_timeout_secs")]
-    pub probe_timeout_secs: u64,
 }
 
 #[derive(Debug, Deserialize)]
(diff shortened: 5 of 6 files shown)


shaun-agent:
## OpenAB PR Screening

This is a manually uploaded screening report from the OpenAB project-screening flow for context collection and reviewer handoff.

- Title: feat(cursor): wire /models /agents /cancel to cursor-agent configOptions
- Source: https://github.com/openabdev/openab/pull/492
- Generated at: 2026-04-21T07:13:37Z

<details>

## Intent

pr `#492` is trying to finish the cursor-specific user experience around discord slash commands without turning openab into a cursor workaround layer.

the concrete problem is twofold. first, cursor-backed channels currently miss the intended `/models`, `/agents`, and `/cancel` control path because the handlers assume a session already exists and can run out of time against discord's 3 second interaction deadline. second, cursor-agent exposes broken ACP behavior around model selection, so the pr also tightens the docs to explain what openab will and will not do when the backend lies.

assumption: deploy-time cursor support already exists in-tree through the existing cursor image, helm values, and docs, so this pr is intentionally about discord interaction flow and policy boundary, not first-time packaging.

## Feat

this is a feature and interaction-hardening pass centered in `src/discord.rs`, with supporting docs and config example updates.

in plain terms, it makes cursor sessions behave like a real first-class ACP backend for discord controls:

- slash commands defer immediately so slow or cold session startup does not blow the interaction deadline
- the handler creates or restores the pooled ACP session before reading cached `configOptions`
- `/agents` works for both `category=agent` and `category=mode`
- `/models` respects discord's 25-option select limit with a plain in-order truncation
- `/cancel` becomes usable against an active cursor session through the same deferred-response pattern
- docs now state that `--model auto` is load-bearing and that openab passes cursor soft rejects through instead of probing or auto-falling back

notably, the ACP transport, pool internals, and config parsing layer stay untouched. that is a good boundary.

## Who It Serves

the primary beneficiary is discord users running openab against `cursor-agent acp`, especially users on a fresh channel session who previously hit the "start a conversation first" dead-end or a silent timeout.

secondarily it serves maintainers and deployers. the docs now make the startup model invariant and upstream cursor-agent defects explicit, which reduces false debugging loops where openab gets blamed for cursor's bad protocol behavior.

## Rewritten Prompt

implement cursor-compatible discord config controls without adding cursor-specific heuristics to the ACP layer.

acceptance criteria:

- `/models`, `/agents`, and `/cancel` must call `defer_ephemeral` before any potentially slow work
- config commands must call `pool.get_or_create(thread_key)` before reading `configOptions`, so they work on cold-start sessions and after pool eviction
- config select handling must re-run `get_or_create` before `set_config_option`
- `build_config_select` must treat `agent` and `mode` as aliases and cap string select menus to 25 options in order
- successful config changes must be acknowledged via `edit_response`
- failures to create a session or set an option must surface a direct error in the interaction response
- `docs/cursor.md` and `config.toml.example` must explain that `--model auto` is required for true auto routing
- openab must not probe, string-match, or auto-fallback after `set_config_option`; upstream cursor-agent ACP defects remain documented in `#493`

validation:

- manual e2e against a live cursor-agent deployment for `/models`, `/agents`, and `/cancel`
- verify cold-start command use works before the first normal prompt
- verify a pool-evicted session can still switch config after the menu interaction

if issue `#489` still has deployer ergonomics left over, split them into a follow-up instead of growing this pr sideways.

## Merge Pitch

this is worth merging because it closes a real product gap for cursor-backed deployments with relatively tight scope. users get the same slash-command control surface other ACP backends already expose, and the implementation does it in the right layer: discord handler and docs, not protocol-shaping hacks.

risk is moderate but localized. the code changes sit in interaction timing and pooled session bootstrapping, which can regress command UX if wrong, but they do not alter the ACP core. the likely reviewer concern is not the handler logic itself. it is whether this pr is claiming to close more than it actually does, because `#489` originally mixed UI wiring, deploy config, and a probe-based workaround that the project has since rejected.

## Best-Practice Comparison

**OpenClaw:** the relevant principle is boundary discipline. route work explicitly, surface failures honestly, and do not smuggle backend-specific recovery rules into the wrong layer. this pr matches that better than the earlier probe-and-fallback direction. durable job persistence and retry/backoff are not directly relevant here because this is an interactive discord control path, not a scheduled execution pipeline.

**Hermes Agent:** the relevant principle is making state transitions explicit before acting. the defer-first plus `get_or_create` flow is a small but meaningful version of that: establish session state before assuming cached config exists. file locking, atomic writes, and fresh-session scheduling are not relevant to this path.

net: the useful lesson from both reference systems is not their runtime architecture. it is their refusal to paper over undefined state with wishful logic. `#492` is stronger after removing the proactive probe because it keeps openab's contract clean.

## Implementation Options

1. merge this pr as the narrow discord-and-docs slice, and split any remaining cursor deployer ergonomics into a separate follow-up.

2. broaden this pr slightly to also add or normalize default cursor deployment artifacts if maintainers still think `#489` requires a repo-native config entry point beyond the existing helm/docs path.

3. reintroduce an openab-side probe, entitlement filter, or auto-fallback layer to hide cursor-agent's broken model semantics from users.

## Comparison Table

| --- | --- | --- | --- | --- | --- | --- |
| 1. merge as a narrow slice | high | low | high | high | high | very high |
| 2. widen scope to include deploy artifacts | medium | medium | medium-high | medium | high | medium |
| 3. add openab-side workaround logic | low | high | medium | low | medium-high in the short term | low |

## Recommendation

take option 1.

the pr is strongest when treated as a focused interaction-layer improvement plus documentation cleanup. it fixes a real user-facing gap, keeps the ACP boundary honest, and avoids merging policy-confused workaround code just because cursor-agent is sloppy. if maintainers still want a dedicated `config/cursor.toml`, compose profile, or similar deploy-time artifact, do that as a follow-up and do not use it to block the handler fix.

the one thing to watch in review is scope accounting: if `#489` is meant to cover more than this pr actually ships, close only the part that is now done and leave the rest as explicit follow-up work.

</details>

masami-agent (approved):
## PR Review: #492

### Summary
- **Problem**: Discord slash commands `/models`, `/agents`, `/cancel` did not work against cursor-agent because the handlers assumed a session already existed and used `create_response` which could exceed Discord's 3-second interaction deadline on cold starts.
- **Approach**: Switch all three handlers to `defer_ephemeral` + `edit_response`, add a pre-flight `get_or_create` to ensure a session (and its cached `configOptions`) exists, alias `agent`/`mode` categories for kiro-cli/cursor-agent compatibility, and cap the select menu at Discord's 25-option limit with a plain `take(25)`.
- **Risk level**: Low

### Core Assessment
1. Problem clearly stated: ✅ — linked to #489, thorough PR description with revision history
2. Approach appropriate: ✅ — minimal changes scoped to discord.rs, no ACP layer modifications
3. Alternatives considered: ✅ — v1 (probe+fallback) → v2 (trust-the-protocol) → v3 (simplified truncation), well-documented iteration
4. Best approach for now: ✅ — trust-the-protocol is the right call; upstream bugs belong upstream (#493)

### Findings

**`build_config_select` — category aliasing and truncation**

The alias table is clean and handles the kiro-cli (`agent`) vs cursor-agent (`mode`) divergence correctly. The `take(25)` truncation is ACP-agnostic — no vendor literals, no assumptions about which option is the fallback. Good simplification from v3.

**`handle_config_command` — defer + get_or_create + edit**

The `defer_ephemeral` → `get_or_create` → `get_config_options` → `edit_response` sequence is correct. `get_or_create` handles both missing-session and evicted-session cases, and `session/new` populates `config_options` (confirmed in `connection.rs:358`), so the subsequent `get_config_options` call will have data. The error message on `get_or_create` failure is clear and actionable.

**`handle_cancel_command` — defer + cancel**

Using `defer_ephemeral` here is consistent with the other handlers. `cancel_session` uses pre-stored `cancel_handles` and doesn't require a live session lock, so the "no session for thread" error on missing session is fine UX — it tells the user there's nothing to cancel.

**`handle_config_select` — defensive get_or_create**

The defensive `get_or_create` before `set_config_option` is a good catch — the pool could evict the session between the initial `/models` interaction and the user's selection. This closes a real race window.

**Docs (`cursor.md`)**

The `--model auto` startup table is a valuable addition — this is a non-obvious gotcha. The "Soft rejects from unroutable models" section clearly explains the trust-the-protocol semantics and links to #493. Known Limitations is well-structured.

### Review Summary

#### 🔴 Blockers
(none)

#### 💬 Questions
(none)

#### 🔧 Suggested Changes
(none)

#### ℹ️ Info
- In `handle_config_select`, the `display_name` extraction changed from `.name.as_str()` (borrowing) to `.name.clone()` (owning). Functionally identical since the value is immediately consumed by `format!()`, but worth noting as a minor style divergence from the original. Not worth changing.
- The commit history (v1 → v2 → v3) is well-structured and tells a clear story. The squash merge will collapse it, but the PR description preserves the rationale nicely.

#### ⚪ Nits
(none)

### Verdict
APPROVE — clean, well-scoped PR with no blockers. The defer+edit pattern, pre-flight session creation, and trust-the-protocol approach are all correct. Docs are thorough.

obrutjack (approved):
LGTM. Clean defer-first pattern, correct category aliasing, ACP layer untouched. Two minor suggestions (success-path error logging, alias comment) — non-blocking. Ship it.

shaun-agent (approved):
no blockers from my pass. the final revision keeps the fix in the discord handler layer: defer/edit interactions, get_or_create before config reads, agent/mode aliasing, and an ACP-agnostic 25-option cap.\n\nCI is green. local cargo test was not runnable in this pod because there is no C linker (cc), so build validation is from CI.

chaodu-agent (commented):
🔴 **Needs Rebase + Rework** — Good improvements but stale against main; regresses pagination.

<details>

| Field | Value |
|-------|-------|
| State | OPEN |
| Mergeable | **CONFLICTING** |
| Created | 2026-04-20 (11 days ago) |
| Changes | +152 / -58 across 3 files |
| Closes | #489 |

**Main branch status**: Main now has pagination support (PR #687) for `/models` and `/agents` select menus. The PR's `take(25)` approach unknowingly regresses this.
</details>

<details>

**1. What problem does it solve?**
`/models`, `/agents`, `/cancel` slash commands didn't work reliably against `cursor-agent` ACP sessions. Three issues: (a) "start a conversation first" dead-end, (b) Discord 3s interaction deadline exceeded on cold-start, (c) cursor-agent uses `category=mode` while kiro-cli uses `category=agent`.

**2. How does it solve it?**
- **defer_ephemeral → edit_response pattern**: Defers immediately (buys 15 min), then does slow work, then edits. Solves 3s deadline.
- **`pool.get_or_create()` pre-flight**: Ensures active ACP session before reading configOptions.
- **agent/mode alias table**: Treats `"agent"` and `"mode"` as equivalent categories.
- **`take(25)` truncation**: Caps select menu at Discord's 25-option limit.
- **docs/cursor.md overhaul**: Documents `--model auto` startup semantics and known limitations.

**3. What was considered?**
Three iterations documented: v1 had proactive probe (removed), v2 trust-the-protocol, v3 removed vendor-specific pinning. Good iterative refinement.

**4. Is it the best approach?**
The defer + get_or_create + alias approach is sound. However, the `take(25)` flat truncation regresses main's pagination support.
</details>

<details>

🟢 **INFO**
- **defer_ephemeral pattern** — Correct fix for Discord's 3s interaction deadline. Not on main yet; genuine improvement.
- **get_or_create pre-flight** — Eliminates UX dead-end. Well-placed defensive code.
- **agent/mode alias** — Clean solution for cross-backend category mismatch.
- **Trust-the-protocol docs** — cursor.md additions are excellent.
- **Thorough test plan** — Manual E2E against real cursor-agent pod.

🟡 **NIT**
1. Use existing `SELECT_MENU_PAGE_SIZE` constant after rebase (don't introduce duplicate `DISCORD_SELECT_MAX`).
2. `aliases` slice allocation on every call — consider simple `||` condition instead.

🔴 **SUGGESTED CHANGES**
1. **Merge conflicts — rebase required**. Main's #687 rewrote imports and added pagination infrastructure.
2. **Pagination regression** — Main's `build_config_select` takes a `page` parameter and supports pagination. The PR replaces this with flat `take(25)` that silently drops options. After rebase, integrate defer + get_or_create + alias into existing pagination-aware code.
3. **handle_config_select defer/edit_response flow** — Verify that `comp.defer()` + `comp.edit_response()` produces the intended UX for component interactions.
</details>

*Reviewed by 超渡法師 🔃 chaodu Backlog triage*

chaodu-agent (commented):
## 🔴 Needs Rebase + Rework — Valuable improvements, but pagination regression must be fixed

<details>
<summary>📋 Baseline Check (Step 0)</summary>

Checked `main` branch first. Main now has:
- `SELECT_MENU_PAGE_SIZE = 25` constant
- `build_config_select` takes a `page` parameter with full pagination
- `handle_pagination` for ◀/▶ button navigation
- `build_pagination_buttons` helper

Main does **NOT** have: `defer_ephemeral` pattern, `get_or_create` pre-flight, or agent/mode alias. These are net-new value from this PR.

**Critical finding**: The PR's `take(25)` flat truncation **regresses** main's existing pagination support.

</details>

<details>
<summary>🧭 Four-Question Framework</summary>

### 1. What problem does this solve?
Multiple UX pain points with cursor-agent Discord slash commands:
- Discord 3-second interaction deadline causes failures on cold-start
- Users hit "start a conversation first" dead-end when no conversation exists
- cursor-agent uses "mode" terminology while kiro-cli uses "agent" — confusing for users
- docs/cursor.md needed updating for `--model auto` semantics

### 2. How does it solve it?
Changes across 3 files (+152/-58):
- `defer_ephemeral` → `edit_response` pattern to handle Discord's 3s deadline
- `get_or_create` pre-flight check to auto-create conversations
- Agent/mode category alias so both terminologies work
- docs/cursor.md overhaul with known limitations section

### 3. What was considered?
The PR introduces `DISCORD_SELECT_MAX = 25` as a new constant, but main already has `SELECT_MENU_PAGE_SIZE = 25` serving the same purpose. The `take(25)` approach was likely written before main gained full pagination support.

The `defer_ephemeral`, `get_or_create`, and alias features are solid improvements. However, the select menu handling needs rework to integrate with main's pagination system rather than replacing it with flat truncation.

</details>

<details>
<summary>🚦 Traffic Light</summary>

🟢 **INFO** — `defer_ephemeral` → `edit_response` pattern is the correct solution for Discord's 3-second interaction deadline. Well implemented.

🟢 **INFO** — `get_or_create` pre-flight eliminates a real UX dead-end. Good improvement.

🟢 **INFO** — Agent/mode alias is a thoughtful cross-tool consistency fix.

🟢 **INFO** — docs/cursor.md overhaul with `--model auto` semantics and known limitations is valuable.

🔴 **SUGGESTED CHANGES** — PR is in **CONFLICTING** state. Needs rebase against `main`.

🔴 **SUGGESTED CHANGES** — **Pagination regression**: Main's `build_config_select` now takes a `page` parameter with full ◀/▶ pagination. This PR's `take(25)` approach drops that support entirely. After rebase, please integrate with the existing pagination system instead of truncating.

🟡 **NIT** — After rebase, use the existing `SELECT_MENU_PAGE_SIZE` constant instead of introducing `DISCORD_SELECT_MAX`. Same value, one source of truth.

🟡 **NIT** — The `aliases` slice allocation on every call is unnecessary overhead. A simple `category == "agent" || category == "mode"` condition would be cleaner and zero-alloc.

</details>

---
*Review by 超渡法師 🙏 — the net-new features are 🔥, just need to harmonize with main's pagination after rebase!*

chaodu-agent:
## 超渡法師 Review — PR #492

Discord slash commands `/models`, `/agents`, `/cancel` did not work against `cursor-agent acp` because: (a) cursor-agent uses `category=mode` instead of `agent`, (b) cursor-agent returns 26 models (exceeding Discord's 25-option limit), and (c) the old `create_response` pattern could not survive a cold-start session spawn within Discord's 3-second interaction deadline.

### 2. How does it solve it?

- **`defer_ephemeral` → `get_or_create` → `edit_response`** pattern buys time for cold-start session spawns. Clean and correct.
- **`agent`/`mode` category alias** — handles the kiro-cli vs cursor-agent naming difference. Per ACP spec the reserved value is `"mode"`, so this is correct.
- **`take(25)`** — simple, ACP-agnostic truncation. No vendor literals.
- **Trust-the-protocol** — `setOk:true` is taken at face value; upstream soft-reject bugs are documented and tracked in #493.
- **Docs** — `docs/cursor.md` updated with startup `--model auto` table, soft-reject explanation, and Known Limitations section.

### 3. What was considered?

Three iterations documented in commit history:
- v1: proactive per-model probe + auto-fallback (removed per maintainer feedback)
- v2: current-value + `default[]` pinning during truncation (removed — `default[]` is a cursor-specific literal)
- v3: plain `take(25)` — ACP-agnostic, no vendor assumptions

The core changes (defer pattern, alias, take(25), trust-the-protocol) are all correct. However, the PR was branched before `main` gained pagination support, and the diff **regresses** two features that now exist on `main`.

---

### Traffic Light

🟢 **INFO — Well done**
- `defer_ephemeral` + `get_or_create` + `edit_response` is the correct pattern for Discord interactions that may need >3s. This fixes a real UX dead-end.
- `agent`/`mode` alias table is clean and extensible.
- `take(25)` is the right call — no vendor-specific pinning, no complexity.
- `docs/cursor.md` Known Limitations section is thorough and links to #493.
- Commit history shows good iteration discipline (probe removed, pinning removed).

🔴 **SUGGESTED CHANGES**

1. **Pagination regression** — `main` now has `build_config_components()`, `build_pagination_buttons()`, and `handle_pagination()` that support ◀/▶ page navigation for >25 options. The PR replaces this with a flat `take(25)` that silently drops option #26. The PR needs to be rebased onto current `main` and integrated with the pagination system instead of replacing it.

2. **Current-value-first sorting regression** — `main`'s `build_config_select()` sorts the current selection to position 0 so it always appears on page 0. The PR's version iterates options in original order, meaning the current selection could be on any page (or truncated away at position >25). After rebase, the PR should preserve this sorting.

3. **`handle_config_command` missing `defer_ephemeral`** — The PR correctly adds `defer_ephemeral` to `handle_config_command`, `handle_cancel_command`, and `handle_config_select`. But `main`'s `handle_config_command` uses `create_response` (no defer). The defer pattern is the right approach — but after rebase, ensure `handle_pagination` also gets the same treatment if needed.

4. **`handle_config_select` drops `UpdateMessage` for `edit_response`** — `main` uses `CreateInteractionResponse::UpdateMessage` for the select handler (which is the correct Discord pattern for component interactions — it updates the original ephemeral message in-place). The PR switches to `defer` + `edit_response`, which works but creates a visible "thinking..." state. For the select handler specifically, `UpdateMessage` is preferred because the session is already active (the user just interacted with the menu). Consider keeping `UpdateMessage` for `handle_config_select` and only using `defer` + `edit_response` for `handle_config_command` (where cold-start is possible).

🟡 **NIT**
- `docs/cursor.md:91` — "gpt-5.4/gpt-5.3-codex-spark" model names may become stale quickly. Consider phrasing as "some models on certain accounts" without specific model names, or add a date stamp.

---

<details>
<summary>Verdict</summary>

🔴 **Changes requested** — The core approach is sound, but the PR needs a rebase onto current `main` to integrate with the pagination system (`build_config_components`, `build_pagination_buttons`, `handle_pagination`) instead of regressing it. The `agent`/`mode` alias and `defer_ephemeral` pattern should be preserved during rebase.

</details>

chaodu-agent:
## CHANGES REQUESTED 🔴 — Rebase required (168 commits behind, merge conflicts)

The code changes are solid — three meaningful improvements to the Discord slash command flow — but the branch cannot merge in its current state. Please rebase onto `main` and resolve conflicts before re-requesting review.

<details>
<summary>四問框架 Analysis</summary>

### 1. What problem does this solve?

Discord slash commands (`/models`, `/agents`, `/cancel`) had two UX issues against `cursor-agent`:

1. **3-second deadline**: `CreateInteractionResponse` must be sent within 3s of the interaction. Cold-starting an ACP session (spawning `cursor-agent acp`) can exceed this, causing a "This interaction failed" error.
2. **"Start a conversation first" dead-end**: If no session existed (user never @mentioned the bot), `/models` returned "No model options available" with no way to recover.
3. **Category mismatch**: `cursor-agent` uses `category: "mode"` while `kiro-cli` uses `category: "agent"` — `/agents` returned empty against cursor-agent.

### 2. How does it solve them?

Three changes in `src/discord.rs`:

- **`defer_ephemeral` + `EditInteractionResponse`**: Buys time by deferring the response, then editing it after the ACP call completes. Applied to `/models`, `/agents`, `/cancel`, and the config select handler.
- **`get_or_create` pre-flight**: Before fetching `configOptions`, ensures an active ACP session exists. Handles both never-started and pool-evicted sessions.
- **`agent`/`mode` aliasing**: `build_config_select` now treats `"agent"` and `"mode"` as aliases, so `/agents` works against both kiro-cli and cursor-agent.

Additionally:
- **Pagination removed**: `build_pagination_buttons`, `build_config_components`, `handle_pagination` all deleted. Replaced with a plain `take(25)` in `build_config_select`. Since Discord caps StringSelectMenu at 25 options and cursor-agent returns 26 on Pro, only the last one is dropped.

### 3. What was considered?

The PR description documents three iterations:
- **v1**: Proactive per-model probe (silent ping to detect soft-rejects) + auto-fallback to `default[]`
- **v2**: Probe removed per maintainer feedback — trust-the-protocol approach
- **v3**: Truncation pinning (current + `default[]` always included) also removed — plain `take(25)`

Good evolution. Each simplification is well-justified.

The defer + get_or_create + aliasing approach is correct. Two concerns:

1. **Pagination removal is a regression for future backends**: cursor-agent returns 26 models (one dropped), but a future backend could return 50+. The pagination code was already working. Removing it saves ~60 lines but loses functionality. Consider keeping it or at minimum documenting the 25-option hard cap as a known limitation.
2. **Aliasing is hardcoded**: The `agent`/`mode` alias table is a `match` block. If a third backend uses yet another category name, this needs another code change. A config-driven alias map would be more extensible, but this is a NIT — the current approach is fine for two known backends.

</details>

<details>

### 🔴 SUGGESTED CHANGES

1. **Rebase onto main** — 168 commits behind, merge state is `CONFLICTING`. The diff cannot be reviewed for correctness against current main until conflicts are resolved. This is the only blocker.

2. **Missing Prior Art section** — Per [RFC: PR Contribution Guidelines](https://github.com/openabdev/openab/blob/main/docs/steering/rfc-pr-guidelines.md), PRs should include research on how OpenClaw and Hermes Agent handle the same problem (Discord interaction deadline management, config option menus). The PR description is otherwise excellent — adding this section would complete it.

### 🟡 NIT

3. **Pagination removal trade-off** — The removed pagination was functional and handled >25 options gracefully. `take(25)` silently drops options. Consider either (a) keeping pagination, or (b) adding a log warning when options are truncated so operators know items were dropped:
   ```rust
   if opt.options.len() > DISCORD_SELECT_MAX {
       tracing::warn!(category, total = opt.options.len(), shown = DISCORD_SELECT_MAX, "config options truncated to Discord limit");
   }
   ```

4. **No tests for aliasing logic** — The `agent`/`mode` alias matching in `build_config_select` is new logic with no unit test. A small test with `category: "mode"` input matched by `"agent"` lookup would prevent regressions.

5. **`handle_config_select` double `get_or_create`** — The select interaction can only arrive if the user already has an active session (they just saw the menu). The `get_or_create` call is defensive but adds latency. Consider documenting why it is needed (pool eviction between interactions) in a code comment — the current comment is good but could note the expected frequency.

### 🟢 INFO

6. **Excellent PR description** — The revision history (v1→v2→v3) is a model for how to document design evolution. The test plan with specific E2E scenarios against a real cursor-agent pod is thorough.

7. **Docs changes are high quality** — The startup default table in `docs/cursor.md` is immediately useful. The soft-reject documentation and upstream ACP compliance gaps section (linking to #493) are exactly the kind of operational knowledge that saves debugging time.

8. **Trust-the-protocol principle** — The decision to not work around cursor-agent's `setOk:true` for unroutable models is architecturally correct. Workarounds in the gateway layer create coupling; surfacing the upstream bug directly is the right call.

9. **`config.toml.example` comment** — Small but valuable: documenting that `--model auto` is load-bearing prevents a common misconfiguration.

</details>

---

**Action needed**: Rebase onto `main`, resolve conflicts, then re-request review. The code quality is good — once the branch is current, this should be a quick re-review.

chaodu-agent:
## 🔃 Backlog Sweep — 2026-05-03

**Status: 🔴 Stale — rebase required**

No new commits since last review (2026-05-03). Branch still has merge conflicts against `main`. Previous review findings remain valid:

1. **Rebase required** — merge conflicts prevent merge
2. **Pagination regression** — `main` now has pagination support that this PR removes
3. **`agent`/`mode` aliasing** — good approach, needs integration with current `main`

The core changes (defer_ephemeral, get_or_create pre-flight, category aliasing) are solid. Waiting on contributor to rebase and address review feedback.

**Label**: `closing-soon` recommended if no activity within 14 days.

chaodu-agent:
Closing due to staleness — merge conflicts, no contributor activity. The feature can be re-proposed in a fresh PR against current `main`.

Issue #287, opened 2026-04-13
feat: inline text/plain attachments into prompt (long-message support)


ruan330:
### Description

Add support for `text/plain` file attachments by inlining them into the ACP prompt as text content blocks.

Discord automatically converts messages longer than 2,000 characters into a `.txt` file attachment. When this happens, the agent receives an empty prompt — the actual content is in the attachment, which openab currently skips.

### Use Case

Users pasting long text (logs, articles, code) for summarization or analysis. This is the most common non-image attachment scenario in practice.

Current attachment routing:

```
Discord message with attachment
         │
         ▼
    content_type?
    ├── audio/*    → download → STT transcribe → prompt text  ✅
    ├── image/*    → download → resize → base64 → ImageContent ✅
    └── */*        → silently skipped                          ❌
             ▲
             └── text/plain lands here (.txt from long messages)
```

### Proposed Solution

Add one branch to the existing attachment routing: detect `text/plain`, download the content, and inline it as a text content block — the same pattern STT uses for audio.

```
Discord message with attachment
         │
         ▼
    content_type?
    ├── audio/*      → download → STT transcribe → prompt text  ✅
    ├── text/plain   → download → decode UTF-8   → prompt text  ✅  ← NEW
    ├── image/*      → download → resize → base64 → ImageContent ✅
    └── */*          → skipped (future: #254)
```

No external API needed (unlike STT). Pure HTTP download + UTF-8 decode.

**Implementation notes:**

- Check `content_type == "text/plain"` (Discord sets this for auto-generated `.txt`)
- Size guard: skip files > 128 KB to avoid flooding the context window (128 KB ≈ ~130k chars ≈ well within model limits; configurable later)
- Inject as `[Attached text: {filename} ({size})]\n{content}` so the agent knows it came from an attachment
- ~15 lines of code, mirrors the STT download pattern

**Related:**

- #254 — General non-image file attachment support (download to disk)
- #225 — Voice message STT (same routing pattern)
- #158 — Image attachment support

ruan330:
Reopening — I filed this on 2026-04-13 but the issue was closed on 2026-04-15 without an upstream fix, and the functionality is still absent from `src/discord.rs` (verified against today's `main`). Apologies for the earlier confusion — the closing reference came from a fork-only commit that never made it upstream as a PR.

[...]

PR #501, opened 2026-04-21
feat(discord): inline text/plain attachments into prompt

ruan330:
## Summary

Adds text/plain attachment inline support for Discord. When Discord
auto-converts a message longer than 2,000 characters into a .txt
attachment (the single most common non-image attachment in practice),
the broker now downloads the content, UTF-8 decodes it, and injects
it as a text content block tagged
`[Attached text: <filename> (<size> bytes)]` at the front of
`content_blocks`. The agent (`claude-agent-acp` and any other
ACP-compatible CLI) processes it identically to inline chat text.

Reopens #287.

## Problem

Before this change, messages over 2,000 characters silently vanished:
Discord packaged the content as a `.txt` attachment, the broker's
attachment dispatcher had no `text/plain` branch, and the agent saw
an empty prompt. Pasted logs, stack traces, and long-form articles
just disappeared.

## Design

- New `download_text_attachment(url, filename, size, auth_token) -> Option<String>`
  in `src/media.rs`, structurally parallel to the existing
  `download_and_transcribe` (STT audio) and `download_and_encode_image`
  (image) entry points.
- Dispatched from `src/discord.rs` when `content_type.starts_with("text/")`.
- Body size capped at 128 KB. Rationale: ~20× the 2,000-char Discord
  threshold, generous for logs and stack traces, while keeping a
  bounded share of the agent's context window.

## Why a single streaming gate (no Content-Length check)

An earlier revision had a three-layer defense (metadata →
Content-Length → streaming). During internal adversarial review, the
extra layers surfaced as test-matrix complexity (a "which layer
caught it?" distinction) without materially improving safety: the
streaming cap alone handles honest origins, hostile origins that lie
about `size`, and chunked/unknown-Content-Length responses. Simplified
to a single authoritative body-size gate.

The Discord metadata `size` hint is still checked first — purely as a
cheap fail-fast to skip the HTTP round-trip for obviously oversized
attachments — but it is not trusted as the enforcement layer.

## Tests (new `[dev-dependencies]` entry: `wiremock = "0.6"`)

Three cases:

1. **Happy path**: body < 128 KB returns `Some(content)`.
2. **Metadata short-circuit**: `size > 128 KB` short-circuits before
   any HTTP call. Verified with a `MockServer` + `Mock::expect(0)` so
   any regression that drops the metadata guard and delegates to the
   streaming cap is caught as unwanted outbound traffic — even though
   it would still return `None`.
3. **Oversized body**: metadata claims 1 KB, server serves 200 KB →
   returns `None`. The streaming cap is the only production
   body-size gate, so a regression that removes it fails this test
   directly.

Local results: `cargo test --release` 89 pass, `cargo clippy
--all-targets -- -D warnings` clean.

## Scope notes

- **Discord only**. Slack's attachment handler in `src/slack.rs`
  currently has no text-attachment branch either. This PR does not
  wire Slack — happy to leave that to a Slack-using contributor as a
  separate feature PR.
- `download_and_transcribe` (STT audio, no body-level cap at all)
  and `download_and_encode_image` (image, post-download check only —
  already buffered before rejecting) share the same
  metadata-trust-then-full-buffer pattern fixed here. Happy to follow
  up with dedicated PRs to apply the same streaming pattern to those
  if there's interest.

## Behavior on edge cases

- Non-UTF-8 bytes → log `tracing::warn`, skip.
- HTTP non-2xx → log `tracing::error`, skip.
- Body > 128 KB → log `tracing::warn`, skip.
- No hard failures; all skips preserve the user's regular prompt if
  there is one.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -83,6 +83,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "assert-json-diff"
+version = "2.0.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "47e4f2b81832e72834d7518d8487a0396a28cc408186a2e8854c0f98011faf12"
+dependencies = [
+ "serde",
+ "serde_json",
+]
+
+[[package]]
 name = "async-trait"
 version = "0.1.89"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -279,6 +289,24 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "d7a1e2f27636f116493b8b860f5546edb47c8d8f8ea73e1d2a20be88e28d1fea"
 
 [[package]]
+name = "deadpool"
+version = "0.12.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "0be2b1d1d6ec8d846f05e137292d0b89133caf95ef33695424c09568bdd39b1b"
+dependencies = [
+ "deadpool-runtime",
+ "lazy_static",
+ "num_cpus",
+ "tokio",
+]
+
+[[package]]
+name = "deadpool-runtime"
+version = "0.1.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "092966b41edc516079bdf31ec78a2e0588d1d0c08f78b91d8307215928642b2b"
+
+[[package]]
 name = "deranged"
 version = "0.5.8"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -351,6 +379,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "fnv"
+version = "1.0.7"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "3f9eec918d3f24069decb9af1554cad7c880e2da24a9afd88aca000531ab82c1"
+
+[[package]]
 name = "foldhash"
 version = "0.1.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -373,6 +407,7 @@ checksum = "8b147ee9d1f6d097cef9ce628cd2ee62288d963e16fb287bd9286455b241382d"
 dependencies = [
  "futures-channel",
  "futures-core",
+ "futures-executor",
  "futures-io",
  "futures-sink",
  "futures-task",
@@ -394,6 +429,17 @@ name = "futures-core"
 version = "0.3.32"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "7e3450815272ef58cec6d564423f6e755e25379b217b0bc688e295ba24df6b1d"
+
+[[package]]
+name = "futures-executor"
+version = "0.3.32"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "baf29c38818342a3b26b5b923639e7b1f4a61fc5e76102d4b1981c6dc7a7579d"
+dependencies = [
+ "futures-core",
+ "futures-task",
+ "futures-util",
+]
 
 [[package]]
 name = "futures-io"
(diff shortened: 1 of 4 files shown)


shaun-agent:
## OpenAB PR Screening
- Title: feat(discord): inline text/plain attachments into prompt
- Source: https://github.com/openabdev/openab/pull/501
- Generated at: 2026-04-21T05:01:28.007Z
- Discord thread: https://discord.com/channels/1488041051187974246/1496012822763671693
<details>
## Intent

PR #501 fixes a Discord-specific prompt loss path: when a user sends text longer than Discord’s 2,000-character message limit, Discord often converts it into a `.txt` attachment, and OpenAB currently drops that content instead of passing it to the agent.

The user-visible problem is severe but narrow: pasted logs, stack traces, long prompts, or article-length text can arrive as an apparently valid Discord message, yet the agent receives little or no usable prompt content.

## Feat

This is a feature-sized bug fix for the Discord broker.

Behaviorally, it teaches the Discord ingestion path to treat `text/plain` attachments as prompt input by downloading small text attachments, decoding them as UTF-8, and prepending them to the message content blocks with a clear attachment label. The feature is scoped to Discord only, with bounded size handling and skip-on-error behavior.

## Who It Serves

The primary beneficiary is Discord end users sending long text, especially operators and developers pasting logs, traces, config snippets, or long-form instructions.

Secondarily, it helps maintainers and reviewers by removing a silent-failure class that otherwise looks like agent unreliability.

## Rewritten Prompt

Implement Discord support for inlining small `text/plain` attachments into the agent prompt.

Add a media helper that downloads text attachments with a hard streamed body cap of 128 KB, rejects non-UTF-8 or non-2xx responses, and returns `None` on failure without breaking normal message handling. In the Discord attachment dispatcher, route `text/*` attachments through this helper and prepend successful results to `content_blocks` using a labeled text block that includes filename and byte size. Add tests for: successful inline of a small text attachment, metadata-based short-circuit for oversized attachments, and enforcement of the streamed size cap when server metadata is misleading. Keep scope limited to Discord.

## Merge Pitch

This is worth advancing because it fixes a real input-loss bug in one of the most common Discord workflows without changing the broader agent contract. The implementation is bounded, low-surface-area, and easy to reason about.

The main reviewer concern will be prompt-shaping risk: whether prepending attachment text changes conversational ordering, context pressure, or abuse surface. The risk profile is moderate but controlled because the PR caps body size, skips invalid content, and avoids hard-failing the message path.

## Best-Practice Comparison

Against OpenClaw, the most relevant principle is explicit delivery shaping: user input should be normalized at the gateway/broker boundary before execution. This PR aligns with that idea by making attachment-derived text explicit prompt input rather than leaving it implicit or dropped. Durable job persistence, retry orchestration, and isolated executions are mostly not in scope here because this PR is about ingestion, not scheduling or run lifecycle.

Against Hermes Agent, the relevant principle is self-contained prompt construction per run. This PR moves Discord text attachments into the prompt assembly step so each run sees the actual user payload. File locking, atomic state writes, and daemon tick mechanics are not directly relevant. The stronger shared principle is “make scheduled or brokered work self-sufficient at execution time”; this PR does that for Discord message content.

A partial gap remains versus both reference systems: the handling is channel-specific and media-type-specific rather than a more unified attachment normalization layer.

## Implementation Options

1. Conservative option: keep the PR nearly as-is, with Discord-only `text/plain` inlining and the 128 KB streamed cap.
2. Balanced option: extract a generic “text attachment to prompt block” helper in shared media handling, then use it from Discord now while shaping the API so Slack can adopt the same path later.
3. Ambitious option: introduce a unified attachment normalization pipeline across brokers that classifies attachments into prompt-ready text, transcribable media, image inputs, or ignored payloads, all with consistent streaming limits, logging, and prompt annotations.

## Comparison Table

|---|---|---|---|---|---|---|
| Conservative: Discord-only inline text attachment support | High | Low | Medium-High | Medium | High for Discord users | Strong |
| Balanced: shared text-attachment helper with Discord adoption first | Medium-High | Medium | High | High | High for Discord now, better future reuse | Best |
| Ambitious: cross-broker attachment normalization pipeline | Low | High | High if completed well | High long-term, lower short-term | Highest eventual impact | Weak for immediate merge |

## Recommendation

Take the balanced path.

The core behavior in this PR is correct and worth merging, but the best next-step framing for discussion is slightly broader than “Discord special case.” Keep the immediate code change Discord-scoped if that reduces merge friction, but shape the media helper and tests so the same pattern can be reused by Slack and by future fixes to audio/image download limits. That gives Masami or Pahud a mergeable unit now without locking OpenAB into another one-off broker branch.

If needed, split follow-up into two steps: merge Discord text attachment support first, then open a separate hardening pass for shared attachment download limits and Slack parity.
</details>

ruan330:
Closing as a duplicate of #291 — which I hadn't noticed when I
filed #287 and opened this. Apologies for the churn.

Leaving one small thought on #291 in case it's useful, but not
blocking anything there. Also closing #287 as a duplicate of #161.

PR #505, opened 2026-04-21
fix(discord): detect threads when allow_all_channels=true

### LINKED ISSUE #504: fix(discord): allow_all_channels=true skips thread detection, breaking in-thread follow-ups
### Description

When `allow_all_channels = true`, the bot does not respond to thread follow-up messages (with or without @mention). The same bot with `allow_all_channels = false` and explicit channel IDs works correctly.

### Steps to Reproduce

1. Deploy with `allow_all_channels = true` and `allow_user_messages = "multibot-mentions"`
2. @mention the bot in a channel — bot creates thread and responds ✅
3. Send a follow-up message in the thread without @mention — bot ignores it ❌

### Expected Behavior

Bot responds to thread follow-ups in `multibot-mentions` / `involved` mode regardless of `allow_all_channels` setting.

### Root Cause

In `src/discord.rs`, thread detection is gated behind `!in_allowed_channel`:

```rust
let (in_thread, bot_owns_thread) = if !in_allowed_channel {
    // ... thread detection via to_channel() ...
    (parent_allowed, owned)
} else {
    (false, false)  // ← BUG: skips thread detection entirely
};
```

When `allow_all_channels = true`, `in_allowed_channel = true` for all messages, so the else branch always runs, setting `in_thread = false`. The `MultibotMentions` and `Involved` handlers then see `in_thread = false` and return early.

### Proposed Fix

When `in_allowed_channel = true`, still detect if the message is in a thread:

```rust
} else {
    match msg.channel_id.to_channel(&ctx.http).await {
        Ok(Channel::Guild(gc)) if gc.parent_id.is_some() => {
            let owned = gc.owner_id.is_some_and(|oid| oid == bot_id);
            (true, owned)
        }
        _ => (false, false),
    }
};
```

chaodu-agent:
### Description

When `allow_all_channels = true`, thread detection was skipped entirely, setting `in_thread = false` for all messages. This caused `MultibotMentions` and `Involved` modes to ignore thread follow-ups — the bot would respond to the initial @mention but not to any subsequent messages in the thread.

### Root Cause

The `else` branch of `if !in_allowed_channel` returned `(false, false)` unconditionally, skipping the `to_channel()` call needed to detect threads.

### Fix

When `in_allowed_channel = true`, still call `to_channel()` to check if the message is in a thread (has `parent_id`).

### Validation

- `cargo check` ✅
- `cargo test` — 75/75 passed ✅
- Tested on OrbStack: `allow_all_channels=false` with explicit channels responds to thread follow-ups; same bot with `allow_all_channels=true` did not — this fix resolves it.

Closes #504

### DIFF
--- a/src/discord.rs
+++ b/src/discord.rs
@@ -344,6 +344,8 @@ impl EventHandler for Handler {
 
         // Thread detection: check if the message is in a thread whose parent
         // is an allowed channel, and whether the bot owns that thread.
+        // When in_allowed_channel is true, we still need to detect threads
+        // so that MultibotMentions/Involved modes work correctly (#504).
         let (in_thread, bot_owns_thread) = if !in_allowed_channel {
             match msg.channel_id.to_channel(&ctx.http).await {
                 Ok(serenity::model::channel::Channel::Guild(gc)) => {
@@ -371,7 +373,13 @@ impl EventHandler for Handler {
                 }
             }
         } else {
-            (false, false)
+            match msg.channel_id.to_channel(&ctx.http).await {
+                Ok(serenity::model::channel::Channel::Guild(gc)) if gc.parent_id.is_some() => {
+                    let owned = gc.owner_id.is_some_and(|oid| oid == bot_id);
+                    (true, owned)
+                }
+                _ => (false, false),
+            }
         };
 
         if !in_allowed_channel && !in_thread {


chaodu-agent:
Note: the `!in_allowed_channel` branch and the new `else` branch both call `to_channel()` with nearly identical logic. A follow-up refactor could unify them into a single `to_channel()` call outside the `if-else`:

```rust
let (in_thread, bot_owns_thread) = match msg.channel_id.to_channel(&ctx.http).await {
    Ok(Channel::Guild(gc)) if gc.parent_id.is_some() => {
        let parent_allowed = in_allowed_channel
            || self.allow_all_channels
            || gc.parent_id.is_some_and(|pid| self.allowed_channels.contains(&pid.get()));
        let owned = gc.owner_id.is_some_and(|oid| oid == bot_id);
        (parent_allowed, owned)
    }
    _ => (false, false),
};
```

Keeping this PR minimal for the bug fix — refactor can follow separately.

thepagent (approved):

PR #528, opened 2026-04-22
fix: gate bot turn warning behind allowed_channels check

### LINKED ISSUE #529: Bug: Bot turn warning posted to threads where bot is not allowed
## Summary

Bots post `⚠️ Bot turn limit reached (20/20)` warnings in threads/channels they are **not allowed** to participate in.

## Root Cause

In `discord.rs`, the `message()` handler runs bot turn counting and warning dispatch **before** the `allowed_channels` check:

```
message() execution order:

  ① Bot turn counting + say(warning)   ← runs first, no channel check
  ② Channel gating (allowed_channels)  ← too late, ① already returned
```

Every bot instance subscribes to **all** Discord gateway events regardless of its `allowed_channels` config. When a bot message arrives:

1. `classify_bot_message()` increments the per-thread counter
2. If the counter hits the soft limit, `msg.channel_id.say(warning)` fires immediately
3. The handler `return`s — never reaches the `allowed_channels` check at line 301

This means a bot configured to only operate in Thread B will still post warnings in Thread A if Thread A hits the bot turn limit.

## Reproduction

1. Configure Bot Y with `allowed_channels` that does **not** include Thread A
2. Have multiple bots ping-pong in Thread A until 20 bot messages
3. Bot Y posts `⚠️ Bot turn limit reached (20/20)` in Thread A despite never participating there

## Expected Behavior

Only bots that are allowed in a channel/thread should post the warning message. Counting can still happen globally (to preserve the #483 invariant), but the `say()` must be gated.

## Related

- PR #528 — proposed fix
- #483 — bot turn counting runs before self-check (intentional, counting invariant should be preserved)

chaodu-agent:
## Problem

Fixes #529

Bot turn counting runs before channel gating (`allowed_channels` check) in `discord.rs`. When a thread hits the soft limit (20/20), the `WarnAndStop` branch calls `msg.channel_id.say()` **before** the handler reaches the channel permission check — then immediately `return`s.

This means bots that are **not allowed** in a channel/thread still post `⚠️ Bot turn limit reached (20/20)` warnings there, because they receive all Discord gateway events regardless of their `allowed_channels` config.

```
message() execution order:

  ① Bot turn counting + say(warning)   ← runs first, no channel check
  ② Channel gating (allowed_channels)  ← too late, ① already returned
```

## Fix

Add an `allowed_here` check before `say()` in the `WarnAndStop` branch:

```rust
let ch = msg.channel_id.get();
let allowed_here = self.allow_all_channels
    || self.allowed_channels.contains(&ch);
if msg.author.id != bot_id && allowed_here {
    let _ = msg.channel_id.say(&ctx.http, &user_message).await;
}
```

**What stays the same:**
- Bot turn counting is unchanged — ALL bot messages still count toward the per-thread limit (preserves #483 invariant)
- `SilentStop` / `return` behavior unchanged
- `allow_all_channels = true` bots are unaffected

**What changes:**
- Bots not allowed in a channel no longer post warning messages there

## Out of scope

Dedup (multiple allowed bots each posting their own warning in the same thread) is a separate issue to track independently.

### DIFF
--- a/src/discord.rs
+++ b/src/discord.rs
@@ -276,7 +276,33 @@ impl EventHandler for Handler {
                                 "soft bot turn limit reached",
                             ),
                         }
-                        if msg.author.id != bot_id {
+                        // Only post the warning if this bot is allowed in the channel/thread.
+                        // Bot turn counting intentionally runs before channel gating so ALL
+                        // bot messages are counted, but the *warning message* must respect
+                        // channel permissions — otherwise bots that never participated in a
+                        // thread will spam it with warnings.
+                        //
+                        // Must match the full thread allowlist semantics: a thread is allowed
+                        // if its own channel_id OR its parent_id is in allowed_channels.
+                        let ch = msg.channel_id.get();
+                        let mut allowed_here = self.allow_all_channels
+                            || self.allowed_channels.contains(&ch);
+                        if !allowed_here {
+                            // Thread channel_id won't be in allowed_channels directly —
+                            // check parent_id via to_channel(). Only called on the
+                            // WarnAndStop path (once per soft/hard limit hit), not on
+                            // every bot message.
+                            if let Ok(serenity::model::channel::Channel::Guild(gc)) =
+                                msg.channel_id.to_channel(&ctx.http).await
+                            {
+                                if gc.parent_id.is_some_and(|pid| {
+                                    self.allowed_channels.contains(&pid.get())
+                                }) {
+                                    allowed_here = true;
+                                }
+                            }
+                        }
+                        if msg.author.id != bot_id && allowed_here {
                             let _ = msg.channel_id.say(&ctx.http, &user_message).await;
                         }
                         return;


shaun-agent:
## OpenAB PR Screening
- Title: fix: gate bot turn warning behind allowed_channels check
- Source: https://github.com/openabdev/openab/pull/528
- Generated at: 2026-04-22T16:31:49.436Z
- Discord thread: https://discord.com/channels/1488041051187974246/1496548852772114453
<details>
## Intent

PR #528 fixes a Discord handler leak where bot-turn soft-limit warnings are emitted before channel authorization is checked. The user-visible problem is that a bot configured to stay out of certain channels or threads can still post `Bot turn limit reached` warnings there, which breaks operator expectations around `allowed_channels` and creates unwanted bot noise in disallowed spaces.

## Feat

This is a bug fix to Discord message handling. In plain terms: keep counting bot turns exactly as today, but only send the soft-limit warning if the bot is actually allowed to speak in that channel/thread.

## Who It Serves

The primary beneficiaries are Discord deployers and server operators configuring `allowed_channels`. Secondarily, Discord end users benefit because disallowed threads stop receiving stray system-like warning messages from bots that should be silent there.

## Rewritten Prompt

Update the Discord message handler so soft-limit warning delivery respects channel gating. Preserve the existing per-thread bot-turn counting invariant for all messages, including messages in disallowed channels, but prevent `WarnAndStop` from calling `say()` unless `allow_all_channels` is enabled or the current channel ID is present in `allowed_channels`. Keep `SilentStop` and early-return behavior unchanged. Add or update tests to prove that:
- turn counting still advances in disallowed channels,
- warning messages are suppressed in disallowed channels,
- warning messages still appear in allowed channels,
- `allow_all_channels = true` behavior is unchanged.

## Merge Pitch

This is a small, targeted correctness fix with low behavioral risk. It aligns runtime behavior with operator configuration without disturbing the existing counting semantics introduced for bot-turn limiting. The main reviewer concern will be whether the patch accidentally weakens the “all bot messages count” invariant from prior work; that should be answered explicitly with tests or a clear code-path explanation.

## Best-Practice Comparison

OpenClaw-relevant principles:
- Explicit delivery routing is directly relevant. This PR moves warning emission closer to that principle by ensuring the bot only delivers into channels it is configured to serve.
- Retry/backoff, durable persistence, isolated executions, and run logs are not materially relevant here because this is an in-memory message-handler guard, not a scheduler or job-runner change.

Hermes Agent-relevant principles:
- Self-contained prompts for scheduled tasks are not relevant.
- Atomic writes and file locking are not relevant.
- The closest fit is the general principle of enforcing execution boundaries early and predictably. This PR improves boundary handling, but only for one warning path rather than as a broader architectural rule.

Overall, the proposed direction matches a basic best practice shared by both systems: decide whether delivery is allowed before emitting side effects. It does not touch the more advanced durability or scheduling patterns from either reference system, and it does not need to.

## Implementation Options

### Option 1: Minimal local guard

Add an `allowed_here` check exactly in the `WarnAndStop` branch before `say()`, leaving the rest of the handler order intact.

Pros:
- Fastest to ship
- Lowest regression risk
- Preserves current control flow and invariants with minimal code churn

Cons:
- Fixes only this one side-effect path
- Leaves the broader handler ordering slightly brittle if future side effects are added before gating

### Option 2: Centralized delivery gating

Extract a helper or small utility that determines whether the bot may emit a message in the current channel, and require warning paths to go through it.

Pros:
- Still small, but more maintainable than a one-off conditional
- Reduces the chance of future permission-gating mistakes on other reply/warning paths
- Makes reviewer intent clearer

Cons:
- Slightly more code movement
- Still does not fully restructure handler flow

### Option 3: Reorder handler into “count first, side effects after authorization”

Refactor the message path so state/accounting decisions happen first, channel authorization is computed once, and all outward side effects are routed through an authorized-delivery phase.

Pros:
- Best long-term correctness model
- Makes future side-effect bugs less likely
- Easier to reason about in review

Cons:
- Larger change surface
- Higher regression risk for a bug that is otherwise narrowly scoped
- Harder to merge quickly

## Comparison Table

|---|---|---|---|---|---|---|
| 1. Minimal local guard | High | Low | Medium | Medium-Low | High for affected operators | Strong |
| 2. Centralized delivery gating | Medium-High | Medium | High | High | High for affected operators | Best balance |
| 3. Reorder handler flow | Medium-Low | High | High | High | High but indirect | Weaker near-term fit |

## Recommendation

Recommend Option 2: centralized delivery gating, or Option 1 if the goal is to merge the narrowest safe fix immediately.

For current merge discussion, the strongest practical path is: accept the narrow behavioral goal from this PR, but frame it as “warning delivery must respect channel authorization while counting remains global.” If the existing patch is truly only a localized guard and has tests proving the invariant, it is mergeable. If maintainers want one extra step of hardening without broad refactor risk, a small helper-based version is the best follow-up because it makes future reply paths less likely to repeat the same mistake.
</details>

thepagent (approved):

PR #560, opened 2026-04-24
fix(slack): use thread_ts key for multibot_threads cache lookup in handle_message

### LINKED ISSUE #556: bug(slack): streaming not disabled in multi-bot threads — wrong cache key in handle_message
### Description

`src/slack.rs:1120` — `other_bot_present` is always `false` because the `multibot_threads` cache is queried with the wrong key.

The cache is **written** in `note_other_bot_in_thread` (line 91) keyed by `thread_ts`:
```rust
cache.entry(thread_ts.to_string()).or_insert_with(tokio::time::Instant::now);
```

It is **correctly read** in `bot_participated_in_thread` (line 228) using `thread_ts`:
```rust
cache.get(thread_ts).is_some_and(|ts| ts.elapsed() < self.session_ttl)
```

But `handle_message` at line 1120 diverges and uses `channel_id` — a key that never exists in the cache:
```rust
// Bug: always false — channel_id is never a key in multibot_threads
cache.contains_key(&thread_channel.channel_id)
```

Since `use_streaming()` (line 419) returns `!other_bot_present`, streaming is permanently ON for all threads regardless of whether another bot is present. Regresses PR #535.

**Suggested fix:**
```rust
let other_bot_present = {
    let cache = adapter.multibot_threads.lock().await;
    thread_channel.thread_id.as_deref()
        .is_some_and(|ts| cache.get(ts).is_some_and(|inst| inst.elapsed() < adapter.session_ttl))
};
```

> Uses `.get(ts).is_some_and(|inst| inst.elapsed() < session_ttl)` to match the TTL-aware pattern at line 228, rather than bare `contains_key`.

### Steps to Reproduce

**Precondition:** Two bots (`openab-bot-1`, `openab-bot-2`) running in the same workspace, both members of the test channel.

1. @mention `openab-bot-1` in a channel — confirm it replies (`edited: true` in `conversations.replies` is expected at this point; only bot-1 is in the thread).
2. In the **same thread**, @mention `openab-bot-2` — get a reply. This calls `note_other_bot_in_thread` and populates the cache with the `thread_ts`.
3. In that **same thread**, @mention `openab-bot-1` again.
4. Inspect `openab-bot-1`'s reply via `conversations.replies`.

### Expected Behavior

`openab-bot-1`'s reply in step 4 should have no `edited` field — streaming must be disabled once another bot has posted in the thread.

### Environment

- Version: `openab-0.8.1-beta.3`
- Platform: Slack
- Confirmed reproduction: thread `1777050262.713699` in `#all-chuangwangfu` — `bot-1` reply after `bot-2` posted still returned `edited: true`
- Tested with: two live bot instances against Slack API

antigenius0910:
## Summary

Fixes the wrong cache key used in `handle_message` when computing `other_bot_present`, which caused streaming to never be disabled in multi-bot threads.

- `multibot_threads` cache is written with `thread_ts` as key (`note_other_bot_in_thread`, line 91)
- `bot_participated_in_thread` (line 228) reads it correctly with `thread_ts`
- `handle_message` (line 1120) was reading with `channel_id` — a key that never exists — so `other_bot_present` was permanently `false` and `use_streaming()` always returned `true`

**Fix:** use `thread_channel.thread_id` with the same TTL-aware pattern as line 228.

Fixes openabdev/openab#556

Discord Discussion URL: https://discord.com/channels/1491295327620169908/1496961096182006023/1497313281000603819

## Test plan

E2E verified on fork branch `bugfix/556-multibot-cache-key` with two live bot instances against Slack API:

- [x] bot-1 replies in a fresh thread → `edited: true` (streaming ON — correct, only bot-1 in thread)
- [x] bot-2 posts in same thread → `note_other_bot_in_thread` populates cache with `thread_ts`
- [x] bot-1 replies again in same thread → **no `edited` field** (streaming OFF — fix confirmed)

### DIFF
--- a/src/slack.rs
+++ b/src/slack.rs
@@ -1117,7 +1117,8 @@ async fn handle_message(
     let adapter_dyn: Arc<dyn ChatAdapter> = adapter.clone();
     let other_bot_present = {
         let cache = adapter.multibot_threads.lock().await;
-        cache.contains_key(&thread_channel.channel_id)
+        thread_channel.thread_id.as_deref()
+            .is_some_and(|ts| cache.get(ts).is_some_and(|inst| inst.elapsed() < adapter.session_ttl))
     };
     if let Err(e) = router
         .handle_message(&adapter_dyn, &thread_channel, &sender_json, &prompt, extra_blocks, &trigger_msg, other_bot_present)
--- a/test.txt
+++ b/test.txt
@@ -1,0 +1,1 @@
+test file
--- a/test.txt
+++ b/test.txt
@@ -1,1 +1,0 @@
-test file


shaun-agent:
## OpenAB PR Screening
- Title: fix(slack): use thread_ts key for multibot_threads cache lookup in handle_message
- Source: https://github.com/openabdev/openab/pull/560
- Generated at: 2026-04-24T20:32:22.010Z
- Discord thread: https://discord.com/channels/1488041051187974246/1497334261278052404
<details>
## Intent

This PR fixes a Slack thread-state bug in multi-bot conversations. The visible problem is that once a second bot joins a Slack thread, OpenAB should stop using streaming edits for follow-up replies, but `handle_message` was checking `multibot_threads` with `channel_id` instead of the thread key (`thread_ts`), so the lookup always missed and streaming stayed enabled when it should have been disabled.

## Feat

This is a bug fix in Slack thread handling. In plain language: make multi-bot thread detection actually work inside `handle_message` by reading the cache with the same `thread_ts` key used everywhere else, so reply behavior matches the intended non-streaming fallback in shared bot threads.

## Who It Serves

The primary beneficiary is Slack end users in channels where more than one bot can participate in the same thread. It also helps maintainers and runtime operators by restoring consistent, predictable thread behavior instead of a silent logic mismatch.

## Rewritten Prompt

Fix Slack multi-bot thread detection in `src/slack.rs` so `handle_message` looks up `multibot_threads` by the thread identifier (`thread_ts` / thread id), not `channel_id`. Reuse the same TTL-aware cache access pattern already used by `bot_participated_in_thread`, and verify that when a second bot has been recorded in a thread, subsequent replies in that thread disable streaming. Remove any stray non-product files from the PR unless they are intentionally required for testing.

## Merge Pitch

This is a small, high-confidence correctness fix for a user-visible Slack behavior bug. The risk is low because it aligns one incorrect read path with the existing write path and another already-correct read path; the main reviewer concern will be whether the change is fully covered by a regression test and whether `test.txt` is accidental noise that should be dropped before merge.

## Best-Practice Comparison

OpenClaw principles that matter here:
- Explicit delivery routing is relevant: thread-scoped behavior should be keyed by the actual thread identifier, not a broader channel identifier.
- Retry/backoff, durable job persistence, isolated executions, and run logs are mostly not relevant for this narrow in-memory cache bug.

Hermes Agent principles that matter here:
- Atomic writes and file locking are not directly relevant because this PR is not about persisted scheduler state.
- The relevant principle is consistency of state access: the same unit of state should be written and read with the same key shape.
- The “fresh session per scheduled run” and self-contained scheduled prompts are not relevant to this Slack runtime path.

Net: this PR is less about distributed-system durability and more about basic state-key correctness. It aligns with both reference systems only in the sense that state ownership and lookup keys must be explicit and consistent.

## Implementation Options

### Option 1: Minimal key fix

Change only the `handle_message` lookup to use `thread_ts` / thread id with the same TTL-aware read pattern used elsewhere. Keep behavior and cache structure unchanged.

### Option 2: Balanced cleanup around thread cache access

Fix the key mismatch and also refactor `multibot_threads` access behind a small helper so reads and writes all go through one thread-keyed API. Add a focused regression test for the “second bot disables streaming” path.

### Option 3: Stronger thread-state model

Replace the ad hoc cache access with a dedicated thread participation state object or service that owns bot participation checks, expiration, and streaming eligibility decisions. Add broader tests for fresh thread, single-bot thread, multi-bot thread, and TTL expiry behavior.

## Comparison Table

|---|---|---|---|---|---|---|
| 1. Minimal key fix | Fastest | Low | Medium | Medium | High immediate fix | Strong |
| 2. Helper + regression test | Fast | Medium | High | High | High immediate fix | Best |
| 3. Thread-state model | Slowest | High | Highest long-term | Highest long-term | High but delayed | Weak for this PR size |

## Recommendation

Option 2 is the right path for merge discussion. The core bug is small, but this is exactly the kind of bug that reappears when state reads and writes are duplicated across a file, so a tiny helper plus one regression test gives a much better merge artifact than a one-line patch alone.

If the contributor wants to keep this PR narrowly scoped, split it into:
1. the key fix plus test as the mergeable change,
2. any broader cleanup as follow-up.

Also, `test.txt` should be treated as suspicious unless there is a clear reason to keep it.
</details>

PR #622, opened 2026-04-28
test(gateway): add coverage for hybrid LINE reply/push dispatch

### LINKED ISSUE #620: test(gateway): add coverage for hybrid LINE reply/push dispatch
Context

PR #608 implements the hybrid LINE reply/push strategy in `gateway/src/main.rs`, and the code has been reviewed and accepted. What is still missing is focused gateway-level test coverage for the dispatch behavior.

Why

The current confidence comes primarily from code review plus existing build/test signals, not from automated tests around the new reply-token cache and fallback control flow. This area is correctness-sensitive because it decides when to use the free Reply API, when to fall back to Push API, and when *not* to fall back to avoid duplicate delivery.

Requested tests

- cache hit: uses LINE Reply API and does not call Push API
- cache miss: falls back to Push API
- expired cached token: falls back to Push API
- Reply API 400 with invalid/expired reply token semantics: falls back to Push API
- Reply API 5xx: does not fall back to Push API to avoid duplicate delivery
- Reply API network/timeout error: does not fall back to Push API to avoid duplicate delivery

Suggested scope

- Extract the LINE dispatch branch into a testable helper if needed
- Keep the tests local to the gateway crate / module
- Prefer explicit assertions about whether Reply or Push was attempted

Related

- PR #608
- PR #619

chaodu-agent:

Issue #620 — the hybrid LINE reply/push strategy from PR #608 lacks focused test coverage for the dispatch decision logic. This is correctness-sensitive because it decides when to use the free Reply API, when to fall back to Push API, and when *not* to fall back to avoid duplicate delivery.

Closes #620

## Proposed Solution

1. **Extract `dispatch_line_reply()`** — pulled the LINE dispatch branch out of the WebSocket recv_task closure into a standalone async function with an `api_base` parameter for testability.

2. **6 test cases** using `wiremock` for HTTP mocking with scoped expectations (`expect(N)` auto-verified on drop):

| Test | Scenario | Expected |
|------|----------|----------|
| `cache_hit_uses_reply_api` | Valid cached token | Reply API ✓, Push API ✗ |
| `cache_miss_uses_push_api` | No token in cache | Reply API ✗, Push API ✓ |
| `expired_token_uses_push_api` | Token older than TTL | Reply API ✗, Push API ✓ |
| `reply_400_invalid_token_falls_back_to_push` | Reply returns 400 "Invalid reply token" | Reply API ✓, Push API ✓ |
| `reply_5xx_does_not_fallback` | Reply returns 500 | Reply API ✓, Push API ✗ |
| `reply_network_error_does_not_fallback` | Connection refused | Reply API attempted, Push API ✗ |

## Why this approach?

- Extracting the function with an `api_base` parameter is the minimal refactor needed to make the logic testable without changing behavior.
- `wiremock` scoped mocks verify both positive (was called) and negative (was NOT called) expectations, which is critical for the "no fallback on 5xx/network error" safety invariant.

## Test Plan

- [x] `cargo build` passes
- [x] `cargo test` passes — all 6 tests green
- [x] No new warnings introduced (existing warnings are pre-existing: unused reply_tx/reply_rx, dead code is_forum)


chaodu-agent:
Superseded by #623 — rebased onto a clean branch from main to remove unrelated diff noise.

PR #654, opened 2026-04-30
fix(helm): add missing gateway permission fields in ConfigMap

iamninihuang:
## What problem does it solve?
This PR fixes a bug where gateway permissions (`allowedUsers`, `allowAllUsers`, etc.)
were missing from the Helm Chart template.
It also improves robustness in NOTES.txt.\n\n

## How does it solve it?
1. Added missing fields to `configmap.yaml`.
2. Added ID validation.
3. Updated `NOTES.txt` to handle list-type commands safely.

### DIFF
--- a/charts/openab/templates/NOTES.txt
+++ b/charts/openab/templates/NOTES.txt
@@ -24,22 +24,22 @@ Agents deployed:
     Required scopes: app_mentions:read, chat:write, channels:history, groups:history, channels:read, groups:read, reactions:write, files:read
 {{- end }}
 
-{{- if eq $cfg.command "kiro-cli" }}
+{{- if eq (toString $cfg.command) "kiro-cli" }}
     Authenticate:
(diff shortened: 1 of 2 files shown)


chaodu-agent (changes requested):
## 🔴 Duplicate `botUsername` block

The PR adds a new `botUsername` block at the end of the new permission fields (lines 145-147 in the PR diff), but one already exists at lines 128-130 on `main`. After merge, the rendered TOML will contain two `bot_username` keys under `[gateway]`, which will cause a TOML parse error or unexpected override.

**Fix:** Remove the duplicate block:
```diff
    allowed_users = {{ ($cfg.gateway).allowedUsers | default list | toJson }}
-    {{- if ($cfg.gateway).botUsername }}
-    bot_username = {{ ($cfg.gateway).botUsername | toJson }}
-    {{- end }}
    {{- end }}
```

I've pushed the fix to `chaodu-agent/openab:fix/helm-configmap-v2` for reference.

Everything else in this PR looks solid — the `toString` wrapping in NOTES.txt and the ID validation with `regexMatch` are both good improvements. 👍

thepagent (approved):

PR #655, opened 2026-04-30
feat(gateway): support inbound media and file attachments

iamninihuang:

Currently, the Custom Gateway adapter only supports receiving plain text (`GwContent.text`). If a user sends an image, file, or voice message via a custom gateway (like LINE, Telegram, etc.), the bot ignores it because the schema lacks attachment support.

## How does it solve it?

This PR introduces comprehensive media support for the Gateway adapter, mirroring the logic already present in Discord and Slack adapters:

1. **Schema Update**: Expands `GwContent` with an optional `attachments: Vec<GwAttachment>` array.
2. **Image Processing**: Automatically downloads, compresses, and base64 encodes image attachments using `media::download_and_encode_image`.
3. **Text File Inlining**: Supports inline reading of text-based files (with a 1MB total size limit and a 5 file cap) via `media::download_and_read_text_file`.
4. **Voice Messages (STT)**: If `stt.enabled` is true, it transcribes audio attachments using `media::download_and_transcribe`. To support this, `stt_config` is now propagated to the Gateway params.
5. **Fallback**: Logs an info message and passes a text block for unsupported file types (like PDFs) so the agent knows a file was uploaded.

## Is this the best approach?

Yes. It leverages the existing, battle-tested functions in `src/media.rs` that the Discord and Slack adapters rely on, ensuring consistent memory bounds (e.g., resizing images to 1200px max, 10MB limits, 1MB text caps) and consistent `ContentBlock` structures across all platforms.

### DIFF
--- a/src/gateway.rs
+++ b/src/gateway.rs
@@ -1,4 +1,7 @@
 use crate::adapter::{AdapterRouter, ChannelRef, ChatAdapter, MessageRef, SenderContext};
+use crate::acp::ContentBlock;
+use crate::config::SttConfig;
+use crate::media;
 use anyhow::Result;
 use async_trait::async_trait;
 use futures_util::{SinkExt, StreamExt};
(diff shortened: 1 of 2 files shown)


thepagent:
Close in favor of https://github.com/openabdev/openab/pull/652
Demo https://discord.com/channels/1491295327620169908/1491365150664560881/1499490944045547562

PR #670, opened 2026-05-01
fix: env_clear() — make [agent].env a true whitelist

### LINKED ISSUE #669: Security: agent subprocess inherits all OAB env vars (including DISCORD_BOT_TOKEN)
## Summary

`AcpConnection::spawn()` in `connection.rs` does not call `env_clear()` before spawning the agent subprocess. As a result, the child process **inherits all environment variables from the OAB parent process**, including sensitive credentials like `DISCORD_BOT_TOKEN`, `SLACK_BOT_TOKEN`, etc.

The `[agent].env` config is additive — it adds variables on top of the inherited environment, not a whitelist replacement.

## Impact

A user can ask the agent to run `env` or `echo $DISCORD_BOT_TOKEN` and the agent will return the value. The agent could also be tricked via prompt injection to exfiltrate these credentials through arbitrary outbound requests (e.g. `curl`).

## Reproduction

1. Set `DISCORD_BOT_TOKEN` as an env var for the OAB process
2. Ask the agent: "run `env` and show me the output"
3. The agent returns all OAB env vars including `DISCORD_BOT_TOKEN`

## Fix

Add `cmd.env_clear()` before injecting `[agent].env` values, and explicitly pass only `HOME` and `PATH` as baseline variables:

```rust
cmd.env_clear();
cmd.env("HOME", working_dir);
cmd.env("PATH", std::env::var("PATH").unwrap_or_default());
for (k, v) in env {
    cmd.env(k, expand_env(v));
}
```

This ensures the agent subprocess only sees credentials explicitly listed in `[agent].env`.

chaodu-agent:
## Summary

Fixes #669

Agent subprocess inherited **all** OAB process env vars because `spawn()` never called `env_clear()`. This means `DISCORD_BOT_TOKEN`, `SLACK_BOT_TOKEN`, and other OAB credentials were visible to the agent — a user could ask the agent to `echo $DISCORD_BOT_TOKEN` and get the real value back.

## Before vs After

| | Before | After |
|---|---|---|
| Agent sees OAB credentials (`DISCORD_BOT_TOKEN`, etc.) | ✅ Yes — full inheritance | ❌ No — `env_clear()` |
| Agent sees `[agent].env` values | ✅ Yes | ✅ Yes (whitelist) |
| Agent sees `HOME`, `PATH` | ✅ Yes (inherited) | ✅ Yes (explicit) |
| Agent sees all other host env vars | ✅ Yes — everything leaked | ❌ No |
| Warning when `[agent].env` is set | ❌ No | ✅ Log warning with key names |

## Changes

1. Added `cmd.env_clear()` in `AcpConnection::spawn()` — agent starts with a clean environment
2. Only `HOME` and `PATH` are passed as baseline
3. `[agent].env` becomes a true whitelist — only explicitly listed env vars are passed through
4. Log warning at startup when `[agent].env` is configured, listing the key names being exposed
5. Added security warning in `config.toml.example` advising OAuth login over env var API keys

## Impact per backend

All supported backends use OAuth/file-based auth and work without `[agent].env`:

| Backend | Auth method |
|---------|------------|
| Kiro CLI | `~/.kiro/` credentials |
| Claude Code | `claude login` (OAuth) |
| Codex | `codex login` (OAuth) |
| Gemini CLI | `gemini login` (OAuth) |
| Copilot | `gh auth login` (OAuth) |
| OpenCode | `opencode auth login` |

`[agent].env` remains available for users who need to pass custom env vars, with a startup warning about the exfiltration risk.

## What was tested

- Code review: standard `std::process::Command` API
- Local build not possible (no C linker in this environment) — CI will verify

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -508,15 +508,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "getopts"
-version = "0.2.24"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "cfe4fbac503b8d1f88e6676011885f34b7174f46e59956bba534ba83abded4df"
-dependencies = [
- "unicode-width",
-]
-
-[[package]]
 name = "getrandom"
 version = "0.2.17"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1055,7 +1046,7 @@ checksum = "384b8ab6d37215f3c5301a95a4accb5d64aa607f1fcb26a11b5303878451b4fe"
 
 [[package]]
 name = "openab"
-version = "0.8.1"
+version = "0.8.2"
 dependencies = [
  "anyhow",
  "async-trait",
@@ -1243,17 +1234,9 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "7c3a14896dfa883796f1cb410461aef38810ea05f2b2c33c5aded3649095fdad"
 dependencies = [
  "bitflags",
- "getopts",
  "memchr",
- "pulldown-cmark-escape",
  "unicase",
 ]
-
-[[package]]
-name = "pulldown-cmark-escape"
-version = "0.11.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "007d8adb5ddab6f8e3f491ac63566a7d5002cc7ed73901f72057943fa71ae1ae"
 
 [[package]]
 name = "pxfm"
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -508,6 +508,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "getopts"
+version = "0.2.24"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "cfe4fbac503b8d1f88e6676011885f34b7174f46e59956bba534ba83abded4df"
+dependencies = [
+ "unicode-width",
+]
+
+[[package]]
 name = "getrandom"
 version = "0.2.17"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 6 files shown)


chaodu-agent (commented):
Blocking finding: `cmd.env("HOME", working_dir)` in `src/acp/connection.rs` introduces a compatibility regression for configurations where `working_dir` is only a workspace path, not the agent home directory. On `main`, `working_dir` is documented as the agent working directory in `README.md`, and its default remains `/tmp` in `src/config.rs`. Before this PR, child processes inherited the real HOME, so locally authenticated CLIs could still find `~/.codex`, `~/.claude`, `~/.config/gh`, etc. After this change, any config that leaves `working_dir = "/tmp"` (or points it somewhere other than the auth home) will lose access to those existing credentials and can fail even though no secrets are being leaked anymore. `env_clear()` itself looks right; the risky part is repurposing `working_dir` as `HOME`.

chaodu-agent (commented):
Re-review on latest head (`a6ceeee`, `de41c50`): the Windows baseline env additions look good, but my earlier blocking concern is still unresolved because the runtime behavior did not change. `HOME` is still forcibly set to `working_dir`; the new comment clarifies that this is intentional, but it does not remove the compatibility regression for configs where `working_dir` is only a workspace path or is left at the default `/tmp`. So my verdict is unchanged: `env_clear()` is good, Windows baseline vars are good, but repurposing `working_dir` as `HOME` still breaks previously working auth setups unless users also discover and override it via `[agent].env`.

chaodu-agent (commented):
Re-review on latest head (`51bc256`): this resolves my previous blocker. The child process now preserves the real `HOME`/`USERPROFILE` from the parent environment, falling back to `working_dir` only when those vars are unset, so existing OAuth/auth file lookup paths (`~/.codex`, `~/.claude`, `~/.config/gh`, etc.) keep working. `env_clear()` still provides the intended credential-leakage fix, and the Windows baseline env additions remain correct. LGTM from my side.

thepagent (approved):

PR #705, opened 2026-05-03
fix(gateway): prevent feishu bot self-echo message loop

wangyuyan-agent:
## Summary

Fixes a message loop where the Feishu bot's own replies are pushed back via WebSocket and processed as new inbound messages.

```
User sends message
  --> Gateway receives via WebSocket
  --> OAB processes, agent replies
  --> Gateway sends reply via REST API
  --> Feishu pushes bot reply back via WebSocket  <-- problem
  --> Gateway processes it as new message
  --> OAB processes again --> infinite loop
```

### Root Cause

Feishu WebSocket pushes ALL messages in a chat back to the subscribing app — including messages the app itself sent. Unlike Discord (where `author.bot=true`), Feishu marks these with `sender_type="user"` and the sender open_id is the original user, not the bot. No existing filter (sender_type, bot_open_id, allowed_users) can catch this.

### Fix

`send_text_message` now extracts the `message_id` from the Feishu API response and inserts it into the dedupe cache. When the echo event arrives via WebSocket, the dedupe check catches it.

## Changes

| File | Change | Description |
|------|--------|-------------|
| `gateway/src/adapters/feishu.rs` | MOD | `send_text_message` returns `Option<String>` (message_id); `handle_reply` stores sent IDs in dedupe cache; tests updated |

## Testing

- 64 tests passed (2 updated for new return type)
- Live verified: bot replies once, no loop
- Restart verified: no stale replay (ACK still works)

## Discord Discussion URL

https://discord.com/channels/1491295327620169908/1500160821567684660

### DIFF
--- a/gateway/src/adapters/feishu.rs
+++ b/gateway/src/adapters/feishu.rs
@@ -831,14 +831,14 @@ async fn resolve_user_name(
 // ---------------------------------------------------------------------------
 
 /// Send a text message to a feishu chat_id.
-/// Returns true on success.
+/// Returns the sent message_id on success (for self-echo dedupe), None on failure.
 pub async fn send_text_message(
     client: &reqwest::Client,
     api_base: &str,
     token: &str,
     chat_id: &str,
     text: &str,
-) -> bool {
+) -> Option<String> {
     let url = format!(
         "{}/open-apis/im/v1/messages?receive_id_type=chat_id",
         api_base
@@ -860,18 +860,23 @@ pub async fn send_text_message(
     {
         Ok(resp) => {
             if resp.status().is_success() {
-                info!(chat_id = %chat_id, "feishu message sent");
-                true
+                let resp_body: serde_json::Value = resp.json().await.unwrap_or_default();
+                let msg_id = resp_body
+                    .pointer("/data/message_id")
+                    .and_then(|v| v.as_str())
+                    .map(|s| s.to_string());
+                info!(chat_id = %chat_id, message_id = ?msg_id, "feishu message sent");
+                msg_id
             } else {
                 let status = resp.status();
                 let text = resp.text().await.unwrap_or_default();
                 tracing::error!(status = %status, body = %text, "feishu send message failed");
-                false
+                None
             }
         }
         Err(e) => {
             tracing::error!(err = %e, "feishu send message request failed");
-            false
+            None
         }
     }
 }
@@ -1002,12 +1007,17 @@ pub async fn handle_reply(
     let text = &reply.content.text;
     let limit = adapter.config.message_limit;
 
-    // Split long messages
+    // Split long messages; store sent message_ids in dedupe to prevent
+    // self-echo (Feishu pushes bot's own messages back via WebSocket)
     if text.len() <= limit {
-        send_text_message(&adapter.client, &api_base, &token, &reply.channel.id, text).await;
+        if let Some(msg_id) = send_text_message(&adapter.client, &api_base, &token, &reply.channel.id, text).await {
+            adapter.dedupe.is_duplicate(&msg_id);
+        }
     } else {
         for chunk in split_text(text, limit) {
-            send_text_message(&adapter.client, &api_base, &token, &reply.channel.id, chunk).await;
+            if let Some(msg_id) = send_text_message(&adapter.client, &api_base, &token, &reply.channel.id, chunk).await {
+                adapter.dedupe.is_duplicate(&msg_id);
+            }
         }
     }
 }
@@ -1400,14 +1410,15 @@ mod tests {
             .respond_with(ResponseTemplate::new(200).set_body_json(serde_json::json!({
                 "code": 0,
                 "msg": "success",
+                "data": {"message_id": "om_test123"}
             })))
             .expect(1)
             .mount(&server)
             .await;
 
         let client = reqwest::Client::new();
-        let ok = send_text_message(&client, &server.uri(), "t-tok", "oc_chat1", "hello").await;
-        assert!(ok);
+        let msg_id = send_text_message(&client, &server.uri(), "t-tok", "oc_chat1", "hello").await;
+        assert_eq!(msg_id.as_deref(), Some("om_test123"));
     }
 
     #[tokio::test]
@@ -1421,8 +1432,8 @@ mod tests {
             .await;
 
         let client = reqwest::Client::new();
-        let ok = send_text_message(&client, &server.uri(), "t-tok", "oc_chat1", "hello").await;
-        assert!(!ok);
+        let msg_id = send_text_message(&client, &server.uri(), "t-tok", "oc_chat1", "hello").await;
+        assert!(msg_id.is_none());
     }
 
     // --- Split text tests ---
--- a/gateway/src/adapters/feishu.rs
+++ b/gateway/src/adapters/feishu.rs
@@ -860,11 +860,16 @@ pub async fn send_text_message(
     {
         Ok(resp) => {
             if resp.status().is_success() {
-                let resp_body: serde_json::Value = resp.json().await.unwrap_or_default();
-                let msg_id = resp_body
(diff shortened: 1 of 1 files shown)


shaun-agent:
## OpenAB PR Screening
- Title: fix(gateway): prevent feishu bot self-echo message loop
- Source: https://github.com/openabdev/openab/pull/705
- Generated at: 2026-05-03T00:31:39.960Z
- Discord thread: https://discord.com/channels/1488041051187974246/1500293515383406632
<details>
## Intent

Prevent the Feishu gateway adapter from re-processing the bot’s own outbound replies as fresh inbound messages. The user-visible problem is an infinite or repeated reply loop in Feishu chats after the bot answers once.

## Feat

This is a gateway bug fix. The behavior change is: when OpenAB sends a Feishu message, the gateway records the returned `message_id` in the existing dedupe path so the later WebSocket echo of that same message is ignored.

## Who It Serves

Primary beneficiaries: Feishu users and gateway operators.

Secondary beneficiaries: reviewers and maintainers, because the fix narrows a noisy production failure mode to a localized adapter change.

## Rewritten Prompt

Fix Feishu self-echo loops in `gateway/src/adapters/feishu.rs`.

Feishu WebSocket events include messages sent by the bot itself, and those echo events are not distinguishable through `sender_type`, `bot_open_id`, or `allowed_users`. Update the Feishu send path so `send_text_message` extracts the Feishu API `message_id` from successful REST responses and returns it to the reply handler. When a bot reply is sent successfully, insert that `message_id` into the gateway dedupe cache before the WebSocket echo can be processed.

Add or update tests covering:

- successful extraction of outbound Feishu `message_id`
- dedupe insertion after sending a reply
- existing behavior when the API response has no usable `message_id`
- no regression to inbound user message handling

Keep the fix scoped to Feishu unless a small shared helper clearly matches existing gateway patterns.

## Merge Pitch

This is worth advancing because it fixes a concrete production loop in the Feishu adapter with a narrow, testable change. The risk profile is low if the dedupe cache is already the accepted mechanism for suppressing duplicate message events.

Likely reviewer concern: whether an in-memory dedupe entry is sufficient, especially across restarts or delayed WebSocket delivery. The PR body claims restart behavior was verified because Feishu ACK handling prevents stale replay, but reviewers should confirm the event ordering and TTL assumptions.

## Best-Practice Comparison

OpenClaw principles that are relevant:

- Explicit delivery routing: relevant. The gateway should know which outbound messages it produced and suppress matching inbound echoes.
- Retry/backoff and run logs: partially relevant. Structured logging around “sent Feishu message id inserted into dedupe” and “echo dropped by dedupe” would make production debugging easier.
- Durable job persistence: only lightly relevant. This is not scheduled work, but persistence would matter if Feishu can replay bot echoes after gateway restart.

OpenClaw principles that are not central here:

- Gateway-owned scheduling
- Isolated executions

Hermes Agent principles that are relevant:

- Atomic writes for persisted state: relevant only if the dedupe cache becomes durable later.
- Fresh session per scheduled run and self-contained prompts: not relevant to message echo suppression.
- Gateway daemon tick model and file locking: not relevant unless OpenAB later models inbound delivery as persisted gateway jobs.

Overall, the PR aligns with the “gateway owns delivery correctness” principle, but it uses the existing lightweight dedupe mechanism rather than introducing durable delivery state.

## Implementation Options

Option 1: Conservative Feishu-only dedupe patch

Accept the current direction: make `send_text_message` return `Option<String>`, extract the outbound `message_id`, and insert it into the existing dedupe cache from `handle_reply`.

This is the fastest path and keeps the blast radius low.

Option 2: Balanced outbound echo-suppression helper

Keep the Feishu behavior but formalize it behind a small helper such as `record_outbound_message_for_dedupe(adapter, message_id)` or a Feishu-local `mark_sent_message_id`. Add structured logs and focused tests around send response parsing, dedupe insertion, and echo drop behavior.

This keeps the merge small while making the intent clearer for future adapters.

Option 3: Ambitious durable gateway delivery ledger

Introduce a gateway-level outbound message ledger with TTL-backed persistence. Store outbound platform message IDs, delivery channel, adapter, timestamp, and suppression status. Use it to dedupe echoes across restarts and to support observability for outbound/inbound correlation.

This is more robust but much larger than the PR’s immediate bug fix.

Option 4: Adapter-level sender identity filtering

Try to identify Feishu bot echoes through sender metadata, app identity, or message content correlation rather than message ID dedupe.

The PR body suggests this does not work reliably because Feishu reports bot replies with misleading sender fields, so this should only be used as a supplement, not the primary fix.

## Comparison Table

| Conservative Feishu-only dedupe patch | High | Low | Good for immediate echoes | Good if tests are clear | High | Strong |
| Balanced outbound echo-suppression helper | Medium | Medium-low | Good, easier to observe | Stronger long-term shape | High | Strongest |
| Durable gateway delivery ledger | Low | High | Highest | Good if gateway delivery is becoming a larger concern | High | Premature for this PR |
| Adapter-level sender identity filtering | Medium | Medium | Weak for Feishu | Fragile | Medium | Poor as primary fix |

## Recommendation

Advance this item using Option 2 if the PR can absorb a small clarity pass; otherwise Option 1 is mergeable if tests already prove the critical path.

The right next step for Masami or Pahud is to review the existing dedupe cache semantics: TTL, insertion timing, restart behavior, and whether missing `message_id` responses fail safely. Do not expand this PR into a durable delivery ledger unless reviewers find evidence that Feishu regularly replays self-echo events after gateway restart.
</details>

chaodu-agent:
## 🟢 LGTM — Clean, minimal fix for Feishu self-echo loop

**Verdict:** This PR correctly prevents the infinite message loop caused by Feishu WebSocket pushing the bot's own sent messages back as inbound events.

### Traffic Light

🟢 **INFO — Correct root cause analysis.** Feishu marks bot-sent messages with `sender_type="user"` and the original user's open_id, so no existing filter (sender_type, bot_open_id, allowed_users) can catch the echo. The dedupe-based approach is the right call.

🟢 **INFO — Minimal, surgical change.** Single file, +24/-13. Reuses the existing `DedupeCache` instead of introducing new data structures. The outbound `message_id` is inserted into the same cache that the inbound path already checks — elegant.

🟢 **INFO — Graceful degradation.** If the Feishu API returns 200 but no `message_id` in the response (API version change, etc.), `msg_id` is `None`, dedupe is skipped, and behavior falls back to pre-PR state. No regression risk.

🟢 **INFO — Tests updated.** Both success and failure test cases updated for the new `Option<String>` return type. Mock response includes `data.message_id`.

🟡 **NIT — Silent JSON parse failure.** `resp.json().await.unwrap_or_default()` will silently produce an empty `Value` if the 200 response body is not valid JSON. Consider adding a `warn!` log for this edge case so it's visible in production. Not blocking — Feishu API should always return valid JSON on 200.

<details>
<summary>Baseline verification</summary>

Verified on `main` before reviewing the diff:
- `send_text_message` returns `bool`, `handle_reply` ignores the return value
- `DedupeCache` uses `Mutex<HashMap<String, Instant>>` with TTL + `max_size=10000`
- `is_duplicate()` inserts on first call (returns `false`), returns `true` on subsequent calls within TTL
- Inbound path already dedupes by both `event_id` and `message_id`
- `send_text_message` is only called in `feishu.rs` (handle_reply + 2 tests)
- The PR's approach of inserting sent `message_id`s into the same dedupe cache is sound — when the echo arrives, `is_duplicate(message_id)` returns `true` and the message is dropped

</details>

chaodu-agent (dismissed):
✅ Approved — four-monk consensus: clean, minimal fix that reuses the existing DedupeCache to prevent Feishu self-echo loop. No blocking findings.

Follow-up suggestions (non-blocking, can be tracked in separate issues):
- Add `warn!` log when 200 response has no `message_id`
- Consider `mark_seen()` alias for `is_duplicate()` to clarify intent (naming debt)
- E2E self-echo regression test coverage

thepagent (approved):

PR #726, opened 2026-05-04
feat(gateway): support images and audio for LINE/Telegram

### LINKED ISSUE #690: feat(gateway): support images and audio for LINE/Telegram
## Summary

The Custom Gateway currently only supports text messages. Add support for sending and receiving images/audio on webhook-based platforms (LINE, Telegram).

| Direction | Text | Images | Audio/Voice |
|-----------|------|--------|-------------|
| Inbound (user → bot) | ✅ | ❌ | ❌ |
| Outbound (bot → user) | ✅ | ❌ | ❌ |

## Use Case

Users on LINE and Telegram want to send photos/screenshots to the AI agent for analysis (e.g. "what's in this image?", "review this architecture diagram", "debug this error screenshot"). Currently the gateway silently drops non-text messages — the agent never sees them. This blocks image-understanding workflows that already work on Discord.

## Recommended Approach: Gateway Media Proxy

Gateway downloads media from platform APIs (which require auth), serves it at a local HTTP endpoint. OAB core fetches from that URL — same pattern as Discord CDN, zero core changes needed.

```
LINE/Telegram webhook (image message)
  → gateway downloads via platform API (auth required)
  → gateway stores in memory with UUID key + 2-min TTL
  → gateway serves at: GET http://gateway:8080/media/<uuid>
  → WS message to OAB: { attachments: [{ type: "image", url: "http://gateway:8080/media/<uuid>" }] }
  → OAB core downloads from gateway URL (no auth, internal network)
  → same flow as Discord CDN images through media.rs
```

### Why this approach

- Gateway already listens on :8080 — adding `/media/<uuid>` route is trivial
- OAB core `media.rs` already downloads images from URLs — zero code change
- No shared volumes, no S3, no external dependencies
- Works for all gateway platforms (LINE, Telegram, Teams)

<details>
<summary>Implementation sketch (~50 lines in gateway)</summary>

```rust
struct MediaEntry {
    data: Vec<u8>,
    content_type: String,
    created_at: Instant,
}
type MediaStore = Arc<RwLock<HashMap<String, MediaEntry>>>;

// On inbound image:
//   1. Download from platform API (LINE Content API, Telegram getFile)
//   2. Store: media_store.insert(uuid, MediaEntry { data, content_type, now })
//   3. Send WS message with url: "http://gateway:8080/media/<uuid>"

// GET /media/<uuid> handler:
//   → return bytes + content-type
//   → 404 if expired/not found

// Background eviction: every 30s, remove entries older than 2 minutes
```

</details>

<details>
<summary>Why not base64 over WebSocket?</summary>

- 33% size overhead (3MB photo → 4MB payload)
- Large WS frames cause backpressure issues

</details>

<details>
<summary>Why not pass platform URL directly to core?</summary>

- LINE: no public URL — requires `Authorization: Bearer` header with channel access token
- Telegram: download URL contains bot token (secret leak concern)
- Discord CDN works because URLs are public (no auth needed)

</details>

<details>
<summary>Prior art: OpenClaw media store</summary>

OpenClaw uses a local filesystem store (`~/.openclaw/media/inbound/<uuid>`) with 2-min TTL and `media://` URI scheme. Their approach assumes co-located processes (same pod/shared volume). Our separated gateway architecture makes the HTTP proxy more appropriate, but the TTL cleanup and size-limit patterns are worth adopting.

</details>

### Inbound

| Platform | Image source | Auth required |
|----------|-------------|---------------|
| LINE | `GET https://api-data.line.me/v2/bot/message/{id}/content` | `Authorization: Bearer {token}` |
| Telegram | `GET https://api.telegram.org/file/bot<token>/<path>` | Token in URL |
| Teams | Attachment URL in activity | Bearer token |

### Outbound (future)

| Platform | Method | Requirement |
|----------|--------|-------------|
| LINE | Image message type | Public HTTPS URL for image |
| Telegram | `sendPhoto` API | Supports direct file upload |
| Teams | Adaptive Card with image | Public URL or hosted attachment |

## Scaling Path

Start with in-memory HashMap (sufficient for low-volume usage). If memory pressure becomes an issue, swap to temp files — same API surface, no protocol change.

## Suggested Implementation Order

1. **Inbound images** — highest value (user sends photo → agent sees it)
2. **Inbound audio/voice** — STT pipeline (similar to Discord voice messages)
3. **Outbound images** — agent sends image back to user

iamninihuang:
## Summary

Added support for incoming images and audio/voice messages for LINE and Telegram adapters.

The Custom Gateway currently only supports text messages. This PR implements a **Media Proxy** that downloads media from platform APIs (which require auth) and serves it at a temporary local HTTP endpoint. This allows OpenAB Core and AI agents to fetch and analyze media via standard URLs, mirroring the image-understanding workflow already present in the Discord adapter.

## Features

- **Inbound Media Proxy**: Automatic download of images and audio from LINE/Telegram.
- **MediaStore**: In-memory caching of media with configurable TTL (default 300s).
- **Schema Update**: Added `attachments` field to `GatewayEvent` to propagate media URLs to the agent.
- **STT Support**: Works with OpenAB's built-in speech-to-text pipeline.

## Testing

| Scenario | Result |
|----------|--------|
| LINE: Send image → Gateway downloads & AI analyzes | PASS |
| Telegram: Send image → Gateway downloads & AI analyzes | PASS |
| LINE: Send audio/voice → Gateway downloads & STT works | PASS |
| Telegram: Send audio/voice → Gateway downloads & STT works | PASS |
| `cargo test` — 96 passed | PASS |
| `cargo clippy` — 0 warnings | PASS |

## Breaking Changes

None. Added optional `attachments` field to the gateway event schema. New environment variables `GATEWAY_MEDIA_BASE_URL` and `GATEWAY_MEDIA_STORE_TTL` have safe defaults.

## Discord Discussion URL

https://discord.com/channels/1491295327620169908/1500160821567684660/1499859716409393172

Closes #690

### DIFF
--- a/docs/config-reference.md
+++ b/docs/config-reference.md
@@ -82,8 +82,9 @@ Custom Gateway adapter for platforms like Telegram, LINE, Feishu/Lark, and Googl
 
 | Variable | Description | Default |
 |----------|-------------|---------|
-| `GATEWAY_MEDIA_BASE_URL` | Public base URL for media proxy (e.g. `https://gateway.example.com`). | `http://127.0.0.1:9090` |
+| `GATEWAY_MEDIA_BASE_URL` | Public base URL for media proxy (e.g. `https://gateway.example.com`). | `http://localhost:8080` |
 | `GATEWAY_MEDIA_STORE_TTL` | Seconds to keep media in memory before expiration. | `300` |
+| `GATEWAY_MEDIA_STORE_MAX_ENTRIES` | Maximum number of media items to keep in memory. | `1000` |
 
 
 ---
--- a/docs/config-reference.md
+++ b/docs/config-reference.md
@@ -78,6 +78,14 @@ Custom Gateway adapter for platforms like Telegram, LINE, Feishu/Lark, and Googl
 | `allow_all_users` | bool \| omit | auto-detect | `true` = any user; `false` = only `allowed_users`. Omitted = inferred from list. |
 | `allowed_users` | string[] | `[]` | User IDs to allow. Only checked when `allow_all_users` resolves to false. |
 
+### Gateway Environment Variables
+
+| Variable | Description | Default |
+|----------|-------------|---------|
+| `GATEWAY_MEDIA_BASE_URL` | Public base URL for media proxy (e.g. `https://gateway.example.com`). | `http://127.0.0.1:9090` |
+| `GATEWAY_MEDIA_STORE_TTL` | Seconds to keep media in memory before expiration. | `300` |
+
+
 ---
 
 ## `[agent]`
--- a/docs/line.md
+++ b/docs/line.md
@@ -84,6 +84,7 @@ In the LINE Developers Console → **Messaging API** tab → scan the QR code wi
 - **1:1 chat** — send a message to the bot, get an AI agent response
 - **Group chat** — add the bot to a group, it responds to all messages
 - **Webhook signature validation** — HMAC-SHA256 via `LINE_CHANNEL_SECRET`
+- **Media Support** — Support for incoming images and audio/voice messages via media proxy.
 
 ### Not Supported (LINE API limitations)
 
--- a/docs/telegram.md
+++ b/docs/telegram.md
@@ -168,6 +168,10 @@ explain VPC peering              ← ignored in groups
 
 DMs and replies within forum topics always trigger the agent (no @mention needed).
 
+### Media support
+
+Incoming images and audio/voice messages are supported via media proxy. The gateway downloads the media and provides a temporary URL to the agent for analysis.
+
 ### Emoji reactions
 
 The bot shows status reactions on your message as the agent works:
--- a/gateway/src/adapters/feishu.rs
+++ b/gateway/src/adapters/feishu.rs
@@ -189,12 +189,14 @@ mod event_types {
         pub event: Option<FeishuEventBody>,
         pub challenge: Option<String>,
         #[serde(rename = "type")]
+        #[allow(dead_code)]
         pub event_type_field: Option<String>,
     }
 
     #[derive(Debug, Deserialize)]
     pub struct FeishuEventHeader {
         pub event_id: Option<String>,
+        #[allow(dead_code)]
         pub event_type: Option<String>,
     }
 
@@ -231,6 +233,7 @@ mod event_types {
     pub struct FeishuMention {
         pub key: Option<String>,
         pub id: Option<FeishuMentionId>,
+        #[allow(dead_code)]
         pub name: Option<String>,
     }
 
@@ -781,6 +784,7 @@ pub async fn start_websocket(
 }
 
 /// Single WebSocket connection lifecycle.
+#[allow(clippy::too_many_arguments)]
 async fn ws_connect_loop(
     token_cache: &Arc<FeishuTokenCache>,
     bot_open_id_store: &Arc<RwLock<Option<String>>>,
@@ -848,7 +852,7 @@ async fn ws_connect_loop(
                                     ack.payload = Some(b"{\"code\":200}".to_vec());
                                     let ack_bytes = ack.encode_to_vec();
                                     let _ = ws_tx.send(
-                                        tokio_tungstenite::tungstenite::Message::Binary(ack_bytes.into())
+                                        tokio_tungstenite::tungstenite::Message::Binary(ack_bytes)
                                     ).await;
                                 }
                             }
@@ -869,6 +873,7 @@ async fn ws_connect_loop(
 }
 
 /// Process a single WebSocket text message.
+#[allow(clippy::too_many_arguments)]
 async fn handle_ws_message(
     text: &str,
     bot_open_id_store: &Arc<RwLock<Option<String>>>,
@@ -1080,8 +1085,8 @@ fn markdown_to_post(md: &str) -> serde_json::Value {
         let line = raw_lines[li];
         // Detect fenced code block
         let trimmed = line.trim_start();
-        if trimmed.starts_with("```") {
-            let lang = trimmed[3..].trim().to_string();
+        if let Some(rest) = trimmed.strip_prefix("```") {
+            let lang = rest.trim().to_string();
             let mut code = String::new();
             li += 1;
             while li < raw_lines.len() {
@@ -1154,8 +1159,8 @@ fn parse_inline(line: &str) -> Vec<serde_json::Value> {
                     }
                     if close_ticks == ticks {
                         // Found matching close — content between is literal
-                        for j in i..end {
-                            buf.push(chars[j]);
+                        for ch in chars.iter().take(end).skip(i) {
+                            buf.push(*ch);
                         }
                         i = end + close_ticks;
                         break 'outer;
@@ -1167,8 +1172,8 @@ fn parse_inline(line: &str) -> Vec<serde_json::Value> {
             }
             if end >= len {
                 // No matching close — treat backticks as literal
-                for j in i..len {
-                    buf.push(chars[j]);
+                for ch in chars.iter().skip(i) {
+                    buf.push(*ch);
                 }
                 i = len;
             }
@@ -1194,8 +1199,8 @@ fn parse_inline(line: &str) -> Vec<serde_json::Value> {
                     }
                     if close_run == run {
                         // Found matching close — strip both, keep inner text
-                        for j in after..scan {
-                            buf.push(chars[j]);
+                        for ch in chars.iter().take(scan).skip(after) {
+                            buf.push(*ch);
                         }
                         i = scan + close_run;
                         found_close = true;
--- a/gateway/src/adapters/feishu.rs
+++ b/gateway/src/adapters/feishu.rs
@@ -1356,12 +1356,14 @@ pub async fn download_feishu_image(
     let ext = if mime == "image/gif" { "gif" } else { "jpg" };
     Some(crate::schema::Attachment {
         attachment_type: "image".into(),
-        filename: format!("{}.{}", image_key, ext),
-        mime_type: mime,
-        data,
-        size: compressed.len() as u64,
+        url: None,
+        filename: Some(format!("{}.{}", image_key, ext)),
+        mime_type: Some(mime),
+        data: Some(data),
+        size: Some(compressed.len() as u64),
     })
 }
+
 
 /// Download a Feishu file by message_id + file_key → base64 Attachment (text files only).
 pub async fn download_feishu_file(
@@ -1418,12 +1420,14 @@ pub async fn download_feishu_file(
     let data = base64::engine::general_purpose::STANDARD.encode(text.as_bytes());
     Some(crate::schema::Attachment {
         attachment_type: "text_file".into(),
-        filename: file_name.to_string(),
-        mime_type: "text/plain".into(),
-        data,
-        size: bytes.len() as u64,
+        url: None,
+        filename: Some(file_name.to_string()),
+        mime_type: Some("text/plain".into()),
+        data: Some(data),
+        size: Some(bytes.len() as u64),
     })
 }
+
 
 /// Send a post (rich text) message to a feishu chat_id.
 /// Returns the sent message_id on success, None on failure.
--- a/gateway/src/adapters/googlechat.rs
+++ b/gateway/src/adapters/googlechat.rs
@@ -64,6 +64,7 @@ pub struct GoogleChatSpace {
     pub name: String,
     #[serde(rename = "type")]
     pub space_type: Option<String>,
+    #[allow(dead_code)]
     pub space_type_renamed: Option<String>,
 }
 
--- a/gateway/src/adapters/line.rs
+++ b/gateway/src/adapters/line.rs
@@ -125,10 +125,17 @@ pub async fn webhook(
                         {
                             let mut store =
                                 state.media_store.lock().unwrap_or_else(|e| e.into_inner());
-                            store.insert(
-                                uuid.clone(),
-                                (data.to_vec(), mime, std::time::Instant::now()),
-                            );
+                            if store.len() >= state.media_max_entries {
+                                warn!(
+                                    size = store.len(),
+                                    "media store full, skipping LINE media proxy"
+                                );
+                            } else {
+                                store.insert(
+                                    uuid.clone(),
+                                    (data.to_vec(), mime, std::time::Instant::now()),
+                                );
+                            }
                         }
                         attachments.push(Attachment {
                             attachment_type: msg.message_type.clone(),
--- a/gateway/src/adapters/line.rs
+++ b/gateway/src/adapters/line.rs
@@ -105,6 +105,24 @@ pub async fn webhook(
                     if !r.status().is_success() {
                         warn!(status = %r.status(), id = %msg.id, "failed to download LINE media");
                     } else {
+                        // Issue #690 review fix: Check file size before downloading
+                        let content_length = r
+                            .headers()
+                            .get("content-length")
+                            .and_then(|v| v.to_str().ok())
+                            .and_then(|s| s.parse::<u64>().ok())
+                            .unwrap_or(0);
+
+                        if content_length > state.media_max_file_size {
+                            warn!(
+                                size = content_length,
+                                max = state.media_max_file_size,
+                                id = %msg.id,
+                                "LINE media too large, skipping"
+                            );
+                            continue;
+                        }
+
                         let mime = r
                             .headers()
                             .get("content-type")
@@ -161,9 +179,18 @@ pub async fn webhook(
                     }
                 }
             } else if msg.message_type != "text" {
+                // Issue #690 review fix: Warn when media message is dropped due to missing access_token
+                warn!(
+                    msg_type = %msg.message_type,
+                    "LINE media message dropped (access_token not configured)"
+                );
                 continue;
             }
         } else if msg.message_type != "text" {
+            warn!(
+                msg_type = %msg.message_type,
+                "LINE media message dropped (access_token not configured)"
+            );
             continue;
         }
 
--- a/gateway/src/adapters/line.rs
+++ b/gateway/src/adapters/line.rs
@@ -104,8 +104,9 @@ pub async fn webhook(
                 if let Ok(r) = resp {
                     if !r.status().is_success() {
                         warn!(status = %r.status(), id = %msg.id, "failed to download LINE media");
-                    } else {
-                        // Issue #690 review fix: Check file size before downloading
+                        continue;
+                    }
+                    // Issue #690 review fix: Check file size before downloading
                         let content_length = r
                             .headers()
                             .get("content-length")
--- a/gateway/src/adapters/line.rs
+++ b/gateway/src/adapters/line.rs
@@ -90,13 +90,65 @@ pub async fn webhook(
         let Some(ref msg) = event.message else {
             continue;
         };
-        if msg.message_type != "text" {
+
+        let mut attachments = vec![];
+        let mut text = msg.text.clone().unwrap_or_default();
+
+        // Handle Image/Audio attachments (Issue #690 Phase 1)
+        if let Some(ref access_token) = state.line_access_token {
+            if msg.message_type == "image" || msg.message_type == "audio" {
+                let url = format!("https://api-data.line.me/v2/bot/message/{}/content", msg.id);
+                let client = reqwest::Client::new();
+                let resp = client
+                    .get(url)
+                    .bearer_auth(access_token)
+                    .send()
+                    .await;
+
+                if let Ok(r) = resp {
+                    if !r.status().is_success() {
+                        warn!(status = %r.status(), id = %msg.id, "failed to download LINE media");
+                    } else {
+                        let mime = r
+                        .headers()
+                        .get("content-type")
+                        .and_then(|v| v.to_str().ok())
+                        .unwrap_or(if msg.message_type == "image" {
+                            "image/jpeg"
+                        } else {
+                            "audio/x-m4a"
+                        })
+                        .to_string();
+
+                    if let Ok(data) = r.bytes().await {
+                        let uuid = uuid::Uuid::new_v4().to_string();
+                        {
+                            let mut store =
+                                state.media_store.lock().unwrap_or_else(|e| e.into_inner());
+                            store.insert(
+                                uuid.clone(),
+                                (data.to_vec(), mime, std::time::Instant::now()),
+                            );
+                        }
+                        attachments.push(Attachment {
+                            attachment_type: msg.message_type.clone(),
+                            url: format!("{}/media/{}", state.public_url, uuid),
+                        });
+                        if text.is_empty() {
+                            text = format!("[{}]", msg.message_type);
+                        }
+                        info!(id = %msg.id, uuid = %uuid, "proxied LINE inbound media");
+                    }
+                }
+            }
+            } else if msg.message_type != "text" {
+                continue;
+            }
+        } else if msg.message_type != "text" {
             continue;
         }
-        let Some(ref text) = msg.text else {
-            continue;
-        };
-        if text.trim().is_empty() {
+
+        if text.trim().is_empty() && attachments.is_empty() {
             continue;
         }
 
@@ -138,7 +190,7 @@ pub async fn webhook(
             .and_then(|s| s.user_id.as_deref())
             .unwrap_or("unknown");
 
-        let gateway_event = GatewayEvent::new(
+        let mut gateway_event = GatewayEvent::new(
             "line",
             ChannelInfo {
                 id: channel_id.clone(),
@@ -151,10 +203,11 @@ pub async fn webhook(
                 display_name: user_id.into(),
                 is_bot: false,
             },
-            text,
+            &text,
             &msg.id,
             vec![],
         );
+        gateway_event.attachments = attachments;
 
         // Cache the reply token for hybrid Reply/Push dispatch
         if let Some(ref reply_token) = event.reply_token {
--- a/gateway/src/adapters/line.rs
+++ b/gateway/src/adapters/line.rs
@@ -158,7 +158,7 @@ pub async fn webhook(
                             if proxied {
                                 attachments.push(Attachment {
                                     attachment_type: msg.message_type.clone(),
-                                    url: format!("{}/media/{}", state.public_url, uuid),
+                                    url: Some(format!("{}/media/{}", state.public_url, uuid)),
                                     mime_type: Some(mime),
                                     filename: Some(format!(
                                         "line-{}.{}",
@@ -170,6 +170,7 @@ pub async fn webhook(
                                         }
                                     )),
                                     size: Some(size),
+                                    data: None,
                                 });
                                 if text.is_empty() {
                                     text = format!("[{}]", msg.message_type);
@@ -178,7 +179,6 @@ pub async fn webhook(
                             }
                         }
                     }
-                }
             } else if msg.message_type != "text" {
                 // Issue #690 review fix: Warn when media message is dropped due to missing access_token
                 warn!(
--- a/gateway/src/adapters/line.rs
+++ b/gateway/src/adapters/line.rs
@@ -99,55 +99,67 @@ pub async fn webhook(
             if msg.message_type == "image" || msg.message_type == "audio" {
                 let url = format!("https://api-data.line.me/v2/bot/message/{}/content", msg.id);
                 let client = reqwest::Client::new();
-                let resp = client
-                    .get(url)
-                    .bearer_auth(access_token)
-                    .send()
-                    .await;
+                let resp = client.get(url).bearer_auth(access_token).send().await;
 
                 if let Ok(r) = resp {
                     if !r.status().is_success() {
                         warn!(status = %r.status(), id = %msg.id, "failed to download LINE media");
                     } else {
                         let mime = r
-                        .headers()
-                        .get("content-type")
-                        .and_then(|v| v.to_str().ok())
-                        .unwrap_or(if msg.message_type == "image" {
-                            "image/jpeg"
-                        } else {
-                            "audio/x-m4a"
-                        })
-                        .to_string();
-
-                    if let Ok(data) = r.bytes().await {
-                        let uuid = uuid::Uuid::new_v4().to_string();
-                        {
-                            let mut store =
-                                state.media_store.lock().unwrap_or_else(|e| e.into_inner());
-                            if store.len() >= state.media_max_entries {
-                                warn!(
-                                    size = store.len(),
-                                    "media store full, skipping LINE media proxy"
-                                );
+                            .headers()
+                            .get("content-type")
+                            .and_then(|v| v.to_str().ok())
+                            .unwrap_or(if msg.message_type == "image" {
+                                "image/jpeg"
                             } else {
-                                store.insert(
-                                    uuid.clone(),
-                                    (data.to_vec(), mime, std::time::Instant::now()),
-                                );
+                                "audio/x-m4a"
+                            })
+                            .to_string();
+
+                        if let Ok(data) = r.bytes().await {
+                            let uuid = uuid::Uuid::new_v4().to_string();
+                            let size = data.len() as u64;
+                            let proxied = {
+                                let mut store =
+                                    state.media_store.lock().unwrap_or_else(|e| e.into_inner());
+                                if store.len() >= state.media_max_entries {
+                                    warn!(
+                                        size = store.len(),
+                                        "media store full, skipping LINE media proxy"
+                                    );
+                                    false
+                                } else {
+                                    store.insert(
+                                        uuid.clone(),
+                                        (data.to_vec(), mime.clone(), std::time::Instant::now()),
+                                    );
+                                    true
+                                }
+                            };
+                            if proxied {
+                                attachments.push(Attachment {
+                                    attachment_type: msg.message_type.clone(),
+                                    url: format!("{}/media/{}", state.public_url, uuid),
+                                    mime_type: Some(mime),
+                                    filename: Some(format!(
+                                        "line-{}.{}",
+                                        msg.id,
+                                        if msg.message_type == "image" {
+                                            "jpg"
+                                        } else {
+                                            "m4a"
+                                        }
+                                    )),
+                                    size: Some(size),
+                                });
+                                if text.is_empty() {
+                                    text = format!("[{}]", msg.message_type);
+                                }
+                                info!(id = %msg.id, uuid = %uuid, "proxied LINE inbound media");
                             }
                         }
-                        attachments.push(Attachment {
-                            attachment_type: msg.message_type.clone(),
-                            url: format!("{}/media/{}", state.public_url, uuid),
-                        });
-                        if text.is_empty() {
-                            text = format!("[{}]", msg.message_type);
-                        }
-                        info!(id = %msg.id, uuid = %uuid, "proxied LINE inbound media");
                     }
                 }
-            }
             } else if msg.message_type != "text" {
                 continue;
             }
@@ -161,33 +173,27 @@ pub async fn webhook(
 
         let source = event.source.as_ref();
         let (channel_id, channel_type) = match source {
-            Some(s) if s.source_type == "group" => {
-                match s.group_id.as_deref() {
-                    Some(id) if !id.is_empty() => (id.to_string(), "group".to_string()),
-                    _ => {
-                        warn!("LINE group event missing groupId, skipping");
-                        continue;
-                    }
+            Some(s) if s.source_type == "group" => match s.group_id.as_deref() {
+                Some(id) if !id.is_empty() => (id.to_string(), "group".to_string()),
+                _ => {
+                    warn!("LINE group event missing groupId, skipping");
+                    continue;
                 }
-            }
-            Some(s) if s.source_type == "room" => {
-                match s.room_id.as_deref() {
-                    Some(id) if !id.is_empty() => (id.to_string(), "room".to_string()),
-                    _ => {
-                        warn!("LINE room event missing roomId, skipping");
-                        continue;
-                    }
+            },
+            Some(s) if s.source_type == "room" => match s.room_id.as_deref() {
+                Some(id) if !id.is_empty() => (id.to_string(), "room".to_string()),
+                _ => {
+                    warn!("LINE room event missing roomId, skipping");
+                    continue;
                 }
-            }
-            Some(s) => {
-                match s.user_id.as_deref() {
-                    Some(id) if !id.is_empty() => (id.to_string(), "user".to_string()),
-                    _ => {
-                        warn!("LINE user event missing userId, skipping");
-                        continue;
-                    }
+            },
+            Some(s) => match s.user_id.as_deref() {
+                Some(id) if !id.is_empty() => (id.to_string(), "user".to_string()),
+                _ => {
+                    warn!("LINE user event missing userId, skipping");
+                    continue;
                 }
-            }
+            },
             None => {
                 warn!("LINE event missing source, skipping");
                 continue;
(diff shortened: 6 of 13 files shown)


shaun-agent:
## OpenAB PR Screening
- Title: feat(gateway): support images and audio for LINE/Telegram
- Source: https://github.com/openabdev/openab/pull/726
- Generated at: 2026-05-04T06:33:12.853Z
- Discord thread: https://discord.com/channels/1488041051187974246/1500746791824654479
<details>
## Intent

PR #726 aims to let the custom gateway handle non-text inbound messages from LINE and Telegram. The current operator-visible gap is that those adapters only pass text into OpenAB, so images, audio clips, and voice messages sent by users cannot be analyzed by agents or routed through speech-to-text.

The PR proposes solving this by downloading protected platform media through the gateway, exposing it through temporary local HTTP URLs, and adding those URLs to gateway events as attachments.

## Feat

This is a feature with a schema extension.

Behavioral change:

- LINE image messages become gateway events with fetchable attachment URLs.
- Telegram photo/image messages become gateway events with fetchable attachment URLs.
- LINE and Telegram audio or voice messages can flow into OpenAB’s STT path.
- Gateway events gain an optional `attachments` field.
- Gateway runtime gains a temporary in-memory media proxy/cache with TTL configuration.

## Who It Serves

Primary beneficiaries:

- LINE and Telegram end users who expect agents to understand images and voice/audio.
- Agent runtime operators who need platform media normalized into URL-based inputs.
- Maintainers working toward feature parity across Discord, LINE, and Telegram adapters.

Secondary beneficiaries:

- Reviewers, because a shared attachment shape reduces adapter-specific handling in downstream code.
- Deployers, assuming the media proxy behavior is documented clearly enough to configure safely.

## Rewritten Prompt

Implement inbound media support for LINE and Telegram in the gateway.

Add an optional `attachments` field to `GatewayEvent` using a schema compatible with existing agent media ingestion. For LINE and Telegram inbound image, audio, and voice messages, download the platform-protected media using the adapter’s authenticated API client, store it temporarily in a gateway-owned media store with configurable TTL, and expose it through a local HTTP media endpoint. The generated attachment URLs must be passed through gateway events so OpenAB Core and agents can fetch images and audio using standard URLs.

Include configuration for media base URL and TTL with safe defaults, document the new settings, and add tests or focused validation for:

- Text-only events remain unchanged.
- LINE image events produce valid attachments.
- Telegram image events produce valid attachments.
- LINE audio or voice events produce STT-compatible attachments.
- Telegram audio or voice events produce STT-compatible attachments.
- Expired media is no longer served.
- Missing or failed media downloads produce useful logs without crashing the gateway.

Avoid unrelated adapter refactors unless required by the shared attachment schema.

## Merge Pitch

This is worth advancing because it closes a real parity gap: Discord already has an image-understanding path, while LINE and Telegram currently lose media context. Normalizing protected platform media into temporary gateway URLs is a practical bridge between chat platforms and agent media analysis.

Risk profile is moderate. The main reviewer concerns should be around serving local media safely, memory-only cache behavior under load, URL exposure, TTL cleanup, and whether the schema shape matches downstream agent expectations. The feature is mergeable if the proxy is clearly scoped, configurable, and tested for expiry and failure paths.

## Best-Practice Comparison

- Gateway-owned scheduling: not directly relevant; this is request/event-time media handling, not scheduled jobs.
- Durable job persistence: mostly not relevant unless media processing becomes asynchronous later.
- Isolated executions: relevant in spirit. Media downloads and serving should not let one platform request destabilize the gateway.
- Explicit delivery routing: relevant. Attachments should be explicitly attached to the event payload rather than inferred downstream.
- Retry/backoff and run logs: partially relevant. Media download failures should have structured logs; retries may be useful but should not block the initial merge unless platform APIs are flaky.

- Gateway daemon tick model: not directly relevant.
- File locking to prevent overlap: not relevant for in-memory cache unless later moved to disk.
- Atomic writes for persisted state: not relevant unless the media store becomes persistent.
- Fresh session per scheduled run: not relevant.
- Self-contained prompts for scheduled tasks: not relevant.
- Atomicity principle is still useful: attachment metadata should only be emitted when the gateway has successfully stored or can serve the media.

The relevant best-practice takeaway is that the gateway should own media normalization explicitly, keep temporary state bounded, log failures, and avoid leaking platform-authenticated URLs directly to agents.

## Implementation Options

Option 1: Conservative adapter-local support

Implement LINE and Telegram media downloads directly inside each adapter, add optional `attachments`, and serve cached media through a simple in-memory gateway endpoint.

Pros: fastest path, minimal architectural churn, likely matches the PR shape.  
Cons: duplication between adapters, media behavior may become inconsistent as more platforms add support.

Option 2: Balanced shared media proxy

Introduce a small shared `MediaStore` and media-proxy service used by LINE and Telegram adapters. Adapters are responsible only for identifying media and calling a common helper to download, cache, and emit attachment metadata.

Pros: keeps adapter code focused, easier to test expiry and serving behavior, sets a reusable pattern for future platforms.  
Cons: slightly more review surface than adapter-local changes.

Option 3: Ambitious durable media pipeline

Create a gateway media subsystem with pluggable storage, durable metadata, background cleanup, configurable size limits, richer content-type handling, and future support for async processing or retries.

Pros: strongest long-term reliability and scale story.  
Cons: larger design burden, slower to merge, likely too much for the current PR unless OpenAB already needs durable media retention.

Option 4: Downstream-only media handling

Pass platform media identifiers to OpenAB Core or agents and let them fetch media using platform credentials.

Pros: avoids running a media-serving endpoint in the gateway.  
Cons: leaks platform-specific auth concerns downstream, weakens adapter normalization, and makes agents responsible for platform APIs.

## Comparison Table

| 1. Adapter-local support | High | Low | Medium | Medium-low | High | Good if scoped tightly |
| 2. Shared media proxy | Medium | Medium | High | High | High | Best |
| 3. Durable media pipeline | Low | High | High | Medium-high | High | Premature unless scale issues exist |
| 4. Downstream-only handling | Medium | Medium | Low-medium | Low | Medium | Poor |

## Recommendation

Advance the PR toward Option 2: a shared gateway-owned media proxy with a bounded in-memory store, explicit attachment schema, TTL expiry, and clear docs.

That is the right merge discussion target because it solves the user-facing media gap without turning this PR into a broader storage or job-processing redesign. The next reviewer should focus on whether the current implementation already has a clean shared `MediaStore`, whether expiry and failure behavior are tested, and whether LINE/Telegram adapter changes are limited to platform-specific media discovery and authenticated download.

If the PR currently mixes this with unrelated docs or adapter cleanup, split only the unrelated material. Keep the schema, media proxy, LINE support, Telegram support, and config docs together because they form one coherent feature.
</details>

eric-yh-huang:
@copilot Please re-review. All issues resolved.

chaodu-agent:
CHANGES REQUESTED 🔴 — Doc default mismatch and undocumented behavior change need fixing before merge. Media proxy architecture is solid and follows #690 closely.

<details>
<summary>1️⃣ What problem does this solve?</summary>

The Custom Gateway only supports text messages. Users on LINE and Telegram who send images or voice messages get silently dropped — the agent never sees them. This blocks image-understanding and STT workflows that already work on Discord.

This PR implements the **inbound media proxy** approach from #690: download media from platform APIs (which require auth), cache in-memory with TTL, serve at a public `/media/:uuid` endpoint so OAB core can fetch via standard URL.
</details>

<details>
<summary>2️⃣ How does it solve it?</summary>

**Architecture** (matches #690 recommendation):
- `MediaStore` = `Arc<Mutex<HashMap<String, (Vec<u8>, String, Instant)>>>` — UUID → (bytes, MIME, created_at)
- LINE: downloads from `api-data.line.me/v2/bot/message/{id}/content` with bearer auth
- Telegram: two-step — `getFile` → download from `/file/bot<token>/<path>`
- `GET /media/:uuid` handler serves cached bytes with correct content-type
- Background sweep every 60s removes entries older than `GATEWAY_MEDIA_STORE_TTL` (default 300s)
- `Attachment` struct added to `GatewayEvent` schema (`skip_serializing_if = "Vec::is_empty"` — backward compatible)
- Clippy fixes in `feishu.rs` and `googlechat.rs` (dead_code, too_many_arguments, iterator patterns)

**New env vars**: `GATEWAY_MEDIA_BASE_URL`, `GATEWAY_MEDIA_STORE_TTL`
</details>

<details>
<summary>3️⃣ What alternatives were considered?</summary>

Issue #690 evaluated and rejected:
- **Base64 over WebSocket** — 33% size overhead, backpressure issues
- **Pass platform URL directly** — LINE has no public URL (needs auth), Telegram URL leaks bot token
- **Shared volumes / S3** — unnecessary complexity for this use case

The PR follows the recommended in-memory HTTP proxy approach.
</details>

<details>
<summary>4️⃣ Is this the best approach? (Detailed findings)</summary>

### 🔴 SUGGESTED CHANGES

**1. Doc default mismatch (`docs/config-reference.md`)**

The doc says `GATEWAY_MEDIA_BASE_URL` defaults to `http://127.0.0.1:9090`, but the code defaults to `http://localhost:8080`:

```rust
// main.rs
let public_url = std::env::var("GATEWAY_MEDIA_BASE_URL")
    .or_else(|_| std::env::var("GATEWAY_PUBLIC_URL"))
    .unwrap_or_else(|_| "http://localhost:8080".into());
```

The gateway listens on `:8080` by default, so the code is correct. Please fix the doc to `http://localhost:8080`.

**2. Undocumented behavior change: auto-fill `reply_to` in WS handler**

The PR adds `last_event_id` tracking and auto-fill logic in `handle_oab_connection`:

```rust
// Auto-fill reply_to if empty using the last event sent to this client
if reply.reply_to.is_empty() {
    let last = last_event_id_for_recv.lock().await;
    if let Some(eid) = last.get(&reply.channel.id) {
        reply.reply_to = eid.clone();
    }
}
```

This is a behavioral change to the WS protocol — replies without `reply_to` now silently get one injected. This is **not mentioned in the PR description** and is unrelated to media support. It could affect LINE reply token dispatch (the reply_to is used to look up cached replyTokens). Please either:
- Split this into a separate PR with its own rationale, or
- Document it in the PR description and explain why it's needed for media support

### 🟡 NIT

**3. No size limit on media store**

Issue #690 mentions "If memory pressure becomes an issue, swap to temp files." Currently there's no cap on the number of entries or total bytes. A burst of large images could OOM the gateway. Consider adding a `GATEWAY_MEDIA_STORE_MAX_ENTRIES` or total byte limit (even a generous one like 100MB).

**4. `std::sync::Mutex` vs `RwLock` for MediaStore**

Issue #690's implementation sketch used `RwLock`. The current `Mutex` works fine since critical sections are short and never held across `.await`, but `RwLock` would allow concurrent reads from the `/media/:uuid` handler under load. Low priority — current approach is correct.

**5. Indentation inconsistency in `telegram.rs`**

The media download block (lines ~120-160 in the diff) has inconsistent indentation — some blocks are 4-space indented, others 8-space. This makes the nesting hard to follow. Please run `rustfmt`.

**6. Route syntax: `/media/:uuid`**

Axum 0.8 uses `/{param}` syntax, not `/:param`. The `:param` syntax was axum 0.6. This may compile (axum might still accept it) but `/{uuid}` is the canonical form. Please verify this works or update to `/media/{uuid}`.

### 🟢 INFO

- Clean architecture that closely follows the #690 recommended approach — well done
- `skip_serializing_if = "Vec::is_empty"` on attachments is the right call for backward compatibility
- TTL sweep pattern mirrors the existing reply token sweep — consistent
- Clippy fixes for `feishu.rs` and `googlechat.rs` are welcome housekeeping
- Good fallback MIME types (`image/jpeg` for LINE images, `audio/ogg` for Telegram voice)
- `unwrap_or_else(|e| e.into_inner())` for poisoned mutex is the correct pattern

</details>

chaodu-agent:
## Re-review after commits `e708d47` and `eccf4f6`

**Verdict: Approve ✅** — Contributor addressed the blocking issues. Clean media proxy implementation.

<details>
<summary>Issue resolution status</summary>

### 🔴 Issues from first review

| # | Issue | Status |
|---|-------|--------|
| 1 | Doc default mismatch (`127.0.0.1:9090` vs `localhost:8080`) | ✅ Fixed — doc now says `http://localhost:8080` |
| 2 | Undocumented `reply_to` auto-fill | ✅ Addressed — code comment explains rationale (LINE reply token dispatch). Acceptable as part of this PR since it enables media replies to work correctly. |

### 🟡 NITs from first review

| # | Issue | Status |
|---|-------|--------|
| 3 | No size limit on media store | ✅ Fixed — added `GATEWAY_MEDIA_STORE_MAX_ENTRIES` (default 1000) |
| 4 | `std::sync::Mutex` vs `RwLock` | Unchanged — acceptable, critical sections are short |
| 5 | Indentation inconsistency | Improved in latest commits |
| 6 | Route syntax `:uuid` vs `{uuid}` | ✅ Fixed — now uses `/media/{uuid}` (axum 0.8 canonical) |

### New commit: `eccf4f6` — OAB-side bridge

Clean integration: `build_attachment_blocks()` in `src/gateway.rs` converts gateway attachments into `ContentBlock`s using existing `media::download_and_encode_image` and `media::download_and_transcribe` helpers. STT config properly threaded through `GatewayParams`.

</details>

### 🟢 INFO

- Media proxy architecture closely follows #690 recommendation
- `skip_serializing_if = "Vec::is_empty"` on attachments — backward compatible
- `media_max_entries` cap prevents OOM under burst load
- Clippy fixes in `feishu.rs` / `googlechat.rs` are welcome housekeeping
- OAB-side bridge reuses existing `media.rs` helpers — no duplication

chaodu-agent (dismissed):
Approve — contributor addressed all blocking issues from first review. Media proxy architecture is solid.

wangyuyan-agent (changes requested):
## Review Summary

Solid media proxy architecture that follows #690 recommendations. Testing is comprehensive and the implementation is clean.

However, I found **3 high-severity issues** that must be fixed before merge:

---

## 🔴 High Severity — Must Fix

### 1. TTL not enforced in `media_handler`

**Issue:** Expired media is still served until the next cleanup cycle (up to 60s after expiration).

**Location:** `gateway/src/main.rs` — `media_handler`

**Current code:**
```rust
if let Some((data, mime, _)) = entry {  // ← timestamp discarded
    axum::response::Response::builder()
        .header("content-type", mime)
        .body(axum::body::Body::from(data))
        .unwrap()
}
```

**Fix:**
```rust
if let Some((data, mime, created_at)) = entry {
    if created_at.elapsed().as_secs() >= state.media_ttl {
        return axum::http::StatusCode::NOT_FOUND.into_response();
    }
    axum::response::Response::builder()
        .header("content-type", mime)
        .body(axum::body::Body::from(data))
        .unwrap()
}
```

---

### 2. Memory unbounded by size

**Issue:** `GATEWAY_MEDIA_STORE_MAX_ENTRIES=1000` limits entry count but not total bytes. 1000 × 20MB images = 20GB RAM.

**Location:** `gateway/src/adapters/line.rs` and `telegram.rs` — media download

**Additional issue:** `r.bytes().await` loads the full file into memory **before** checking if the store is full, so rejected entries still consume memory temporarily.

**Recommended fix:** Add a per-file size limit and check it before downloading:

```rust
// In main.rs, add to AppState:
pub media_max_file_size: u64,  // from GATEWAY_MEDIA_MAX_FILE_SIZE, default 10MB

// In line.rs and telegram.rs, before r.bytes().await:
let content_length = r.headers()
    .get("content-length")
    .and_then(|v| v.to_str().ok())
    .and_then(|s| s.parse::<u64>().ok())
    .unwrap_or(0);

if content_length > state.media_max_file_size {
    warn!(size = content_length, "media too large, skipping");
    continue;  // or break, depending on context
}

// Then proceed with r.bytes().await
```

This prevents OOM from a single large file and gives users a clear error message.

---

### 3. `reply_to` auto-fill can cross-contaminate channels

**Issue:** `last_event_id` is keyed by `channel.id`, but if a reply has an empty or wrong `channel.id`, it will pick up the last event from a different channel. This causes LINE reply tokens to be used for the wrong message, leading to silent failures.

**Location:** `gateway/src/main.rs` — `handle_oab_connection`

**Current code:**
```rust
if reply.reply_to.is_empty() {
    let last = last_event_id_for_recv.lock().await;
    if let Some(eid) = last.get(&reply.channel.id) {  // ← no validation
        reply.reply_to = eid.clone();
    }
}
```

**Fix:** Only auto-fill if the channel.id is non-empty and matches a known channel:

```rust
if reply.reply_to.is_empty() && !reply.channel.id.is_empty() {
    let last = last_event_id_for_recv.lock().await;
    if let Some(eid) = last.get(&reply.channel.id) {
        reply.reply_to = eid.clone();
    }
}
```

---

## 🟡 Medium Severity — Should Fix

### 4. `/media/{uuid}` endpoint has no authentication

**Issue:** Anyone who knows a UUID can download media. In typical deployments where the gateway is behind a firewall, this is low risk. However, if the gateway's port 8080 is exposed to the internet, UUIDs could leak via logs (`info!(uuid = %uuid, ...)`), error messages, or monitoring tools.

**Location:** `gateway/src/main.rs` — Router setup

**Risk assessment:**
- **Low risk** if gateway is internal-only (typical deployment)
- **Medium risk** if gateway port is publicly accessible

**Recommended mitigation (choose one):**

**Option A (deployment-level):** Document that the gateway's HTTP port should not be exposed to the public internet. Only the webhook paths need to be accessible (via reverse proxy if needed).

**Option B (code-level):** Add a shared secret query parameter:

```rust
// In AppState:
pub media_secret: String,  // from GATEWAY_MEDIA_SECRET env var

// In media_handler:
async fn media_handler(
    State(state): State<Arc<AppState>>,
    axum::extract::Path(uuid): axum::extract::Path<String>,
    axum::extract::Query(params): axum::extract::Query<HashMap<String, String>>,
) -> impl IntoResponse {
    if params.get("secret") != Some(&state.media_secret) {
        return axum::http::StatusCode::FORBIDDEN.into_response();
    }
    // ... rest of handler
}

// When constructing URLs in adapters:
url: format!("{}/media/{}?secret={}", state.public_url, uuid, state.media_secret),
```

**Note:** UUIDv4 has 122 bits of entropy, making brute-force infeasible within the 300s TTL. The main risk is UUID leakage, not guessing.

---

### 5. LINE media silently dropped when `access_token` is None

**Issue:** If `LINE_ACCESS_TOKEN` is not set, image/audio messages hit `continue` with no warning. Users get no response and no error.

**Location:** `gateway/src/adapters/line.rs`

**Fix:** Add a warning log:

```rust
} else if msg.message_type != "text" {
    warn!(msg_type = %msg.message_type, "LINE media message dropped (access_token not configured)");
    continue;
}
```

---

### 6. Internal URL leaked to LLM on download failure

**Issue:** When `media::download_and_encode_image` fails, the fallback pushes `[Image attachment URL]: http://localhost:8080/media/{uuid}` to the LLM. This exposes internal infrastructure details.

**Location:** `src/gateway.rs` — `build_attachment_blocks`

**Fix:** Use a generic message instead:

```rust
blocks.push(ContentBlock::Text {
    text: "[Image attachment failed to load]".to_string(),
});
```

Or omit the fallback entirely (just skip the attachment).

---

### 7. Mutex poisoning recovery may proceed with corrupted state

**Issue:** All `lock()` calls use `unwrap_or_else(|e| e.into_inner())`, which recovers poisoned mutexes. If a thread panics while holding the lock, subsequent operations proceed on potentially inconsistent state.

**Location:** `line.rs`, `telegram.rs`, `main.rs` — all MediaStore accesses

**Impact:** Low probability (requires a panic inside a critical section), but could lead to serving wrong media or double-counting entries.

**Recommendation:** Either:
- Accept the risk (document it), or
- Panic on poisoned mutex: `.lock().expect("media store poisoned")`

The current approach is a reasonable trade-off for availability over strict consistency.

---

## ✅ What I Like

- Clean URL-proxy architecture that decouples platform auth from core
- Backward-compatible schema with `skip_serializing_if`
- STT integration works out of the box
- Good fallback behavior (text URL if download fails)
- Comprehensive testing (LINE/Telegram image/audio, STT)

---

## Verdict

**Request changes** — Issues #1-3 are high-severity and must be fixed before merge. Issues #4-7 are medium-severity and should be addressed but are not blocking if deployment follows best practices (gateway not exposed to public internet).

Once the high-severity issues are fixed, this PR will be a solid foundation for multi-platform media support. 🚀

wangyuyan-agent (commented):
## Follow-up: Revised Severity Assessment (K8s Deployment Context)

After reviewing the Helm chart (`charts/openab/templates/gateway.yaml`), the gateway Service is `ClusterIP` — port 8080 is cluster-internal only. This changes the severity of a few items from my previous review.

---

### 🔄 Revised: Issue #4 — `/media/{uuid}` endpoint auth

**Previous assessment:** 🟡 Medium — recommended adding a shared secret query param.

**Revised assessment:** ✅ Non-issue in standard deployment. ClusterIP means the media endpoint is only reachable from within the cluster. External actors cannot access it. The Option B code I provided (secret query param) is unnecessary complexity for this deployment model and can be ignored.

Apologies for the noise.

---

### 🔴 New Issue (previously missed): `GATEWAY_PUBLIC_URL` default silently breaks media in K8s

The default `http://localhost:8080` is **actively broken** in any Kubernetes deployment. The gateway and OAB core run in separate Pods — there is no shared localhost. Media URLs generated with the default will point to the gateway container's own loopback interface, which OAB core cannot reach.

The result: media is downloaded, stored, and the attachment URL is emitted — but OAB core silently fails to fetch it. No error, no log at the OAB side, media just disappears.

**This is not a "remember to configure it" edge case — it is a guaranteed silent failure for every K8s deployment that doesn't set this variable.**

Recommended fix: either make the variable required (startup error if unset when any media-capable adapter is configured), or at minimum emit a `warn!` at startup:

```rust
if public_url.contains("localhost") || public_url.contains("127.0.0.1") {
    warn!(
        public_url = %public_url,
        "GATEWAY_PUBLIC_URL looks like a loopback address.          Media attachments will not be reachable from other pods/containers.          Set GATEWAY_PUBLIC_URL to the gateway's cluster-internal Service URL          (e.g. http://openab-gateway:8080) or its external ingress URL."
    );
}
```

The Helm chart template (`gateway.yaml`) currently does **not** inject `GATEWAY_PUBLIC_URL` — so every Helm-based deployment silently inherits the broken default. The chart should either inject the Service's cluster-internal URL automatically or surface this as a required value.

---

### 🔴 New Issue (previously missed): HTTP response status not checked before storing media

In both `line.rs` and `telegram.rs`, the download response is stored without checking `r.status().is_success()`. If the platform API returns 401, 403, or any error response, the gateway stores the error page body as the media payload and emits an attachment URL pointing to it. OAB core then receives what looks like a valid attachment but contains an HTML error page.

**line.rs** — after `if let Ok(r) = resp {`, add:
```rust
if !r.status().is_success() {
    warn!(status = %r.status(), id = %msg.id, "LINE media download failed");
    continue;
}
```

**telegram.rs** — same pattern after the file download `client.get(download_url).send().await`:
```rust
if !r.status().is_success() {
    warn!(status = %r.status(), file_id = %file_id, "Telegram media download failed");
    continue;
}
```

---

The three original 🔴 issues (#1 TTL not enforced in handler, #2 memory unbounded by size, #3 `reply_to` cross-channel contamination) stand unchanged.

iamninihuang:
經過幾次來回，發現這個 PR 不太好處理，仍然有多個地方需要修正，想請問還是先Close這個 PR，日後再戰?

wangyuyan-agent:
Don't close this — you're essentially done. 

Five rounds of fixes, all high-severity issues resolved. The only remaining item is a one-liner: a startup `warn!` when `GATEWAY_PUBLIC_URL` looks like a loopback address, so Helm users get a clear signal instead of silent media failures.

Add that and I'll approve. The hard work is already done.

wangyuyan-agent (dismissed):
All issues resolved — approved. Good work pushing this through.

chaodu-agent:
## 🟡 Mostly Good — Media proxy architecture is sound, but has code quality issues that should be addressed

**What problem does it solve?** The Custom Gateway only supports text messages. Users on LINE/Telegram cannot send images or voice messages to the AI agent. This PR adds an in-memory media proxy that downloads platform media (which requires auth) and serves it at a temporary HTTP endpoint for OAB core to fetch.

**How?** 
1. Gateway downloads media from LINE Content API / Telegram getFile API
2. Stores binary data in an in-memory `HashMap<UUID, (Vec<u8>, mime, Instant)>` with TTL sweep
3. Serves at `/media/{uuid}` endpoint
4. Adds `attachments` field to `GatewayEvent` schema
5. OAB core (`src/gateway.rs`) builds `ContentBlock`s from attachments (image → base64 encode, audio → STT transcribe)

**Considered alternatives?** The PR description references Issue #690 which proposed this "Gateway Media Proxy" approach. This avoids changes to OAB core's media handling — the gateway presents media as standard URLs, same pattern as Discord CDN.

<details>
<summary>Review Details</summary>

### 🟢 INFO — Done well

- **Security**: File size limit (`media_max_file_size`, default 10MB), max entries cap, TTL expiry — prevents OOM
- **Graceful degradation**: If media download fails, logs warning and continues (no crash)
- **Schema backward compat**: `attachments` field uses `#[serde(default, skip_serializing_if = "Vec::is_empty")]` — existing clients unaffected
- **Localhost warning**: Warns operators when `GATEWAY_MEDIA_BASE_URL` looks like loopback
- **Auto-fill `reply_to`**: Smart fallback for LINE reply token dispatch when agent omits event ID

### 🔴 SUGGESTED CHANGES

1. **Inconsistent indentation in LINE adapter** (lines ~110-150): The `content_length` check and subsequent code has extra indentation (8 spaces inside a block that should be 4). This looks like a copy-paste artifact. The Telegram adapter has the same issue. Please fix indentation to match the surrounding code.

2. **`continue` inside Telegram webhook handler**: The Telegram `webhook()` function processes a single `update.message` (not a loop), but the media download path uses `continue` (lines ~100-110 in telegram.rs diff). This would be a compile error unless there's an enclosing loop I'm not seeing. Please verify this compiles — if the `continue` is inside a `loop {}` or `for` that's not visible in the diff context, clarify; otherwise this is a bug.

3. **Duplicated media proxy logic**: The LINE and Telegram adapters have nearly identical media download → store → create attachment logic (~50 lines each). Extract a shared helper like `proxy_media(state, data, mime, msg_type, msg_id) -> Option<Attachment>` to reduce duplication and ensure consistent behavior.

### 🟡 NIT — Non-blocking

4. **`std::sync::Mutex` for `MediaStore`**: The store is accessed from async contexts. While `unwrap_or_else(|e| e.into_inner())` handles poisoning, consider using `tokio::sync::Mutex` or `parking_lot::Mutex` for consistency with the rest of the codebase. The current approach works but holds the lock across `.insert()` which is fine for small maps.

5. **Unrelated formatting changes**: ~100 lines of the diff are `cargo fmt` reformatting of `src/cron.rs` (struct initializers, function signatures). These should ideally be in a separate commit or PR to keep the diff focused. Not blocking but makes review harder.

6. **`#[allow(dead_code)]` additions in feishu.rs/googlechat.rs**: These suppress warnings for fields that are deserialized but never read. Fine as a quick fix, but consider if these fields should be used or if the structs should use `#[allow(dead_code)]` at the struct level instead of per-field.

</details>

**Verdict**: The architecture is solid and well-thought-out. Items #1 and #2 need verification/fix before merge. Item #3 is a strong suggestion for maintainability but not strictly blocking.

masami-agent:
## PR Review: #726

**Based on commit:** `804d643f9b25870324ceb7b68af1e0f438166246`

### Summary
- **Problem**: Custom Gateway (LINE/Telegram) only supports text messages; images and audio are dropped.
- **Approach**: In-memory media proxy that downloads platform media (which requires auth) and serves it at a temporary local HTTP endpoint, allowing the AI agent to fetch media via standard URLs.
- **Risk level**: Medium — new runtime state (in-memory store), new HTTP endpoint, new external API calls.

### Core Assessment
1. Problem clearly stated: ✅ (linked to #690, clear description)
2. Approach appropriate: ✅ (mirrors Discord adapter pattern, proxy is necessary because platform APIs require auth)
3. Alternatives considered: N/A (approach was discussed in Discord thread)
4. Best approach for now: ✅ (in-memory store with TTL is simple and sufficient for the current scale)

---

### Findings

## 🔴 Critical: `continue` statements outside a loop in `gateway/src/adapters/telegram.rs`

**Where:** `gateway/src/adapters/telegram.rs` lines ~121 and ~129

**What's wrong:** The `webhook` function is NOT inside a loop, yet there are two `continue` statements:
```rust
if !resp.status().is_success() {
    warn!(status = %resp.status(), id = %file_id, "Telegram getFile failed");
    continue;  // ← ERROR: not inside a loop
}
```
and
```rust
if !r.status().is_success() {
    warn!(status = %r.status(), id = %file_id, "failed to download Telegram media");
    continue;  // ← ERROR: not inside a loop
}
```

**Why it matters:** This is a compile error in Rust. It is not caught by CI because `ci.yml` only compiles the main `openab` crate (triggered by `src/**` changes), not the `gateway/` crate. This will fail when the gateway binary is actually built.

**Fix:** Replace `continue;` with early-return or restructure as `if/else`:
```rust
// Option A: just don't proceed (already inside nested if-let, so just do nothing)
if !resp.status().is_success() {
    warn!(status = %resp.status(), id = %file_id, "Telegram getFile failed");
    // Remove continue — the nested if/else structure already handles this
}
```
Or restructure the entire block to avoid deep nesting (see 🟡 suggestion below).

---

## 🔴 Critical: Inconsistent indentation suggests copy-paste error in file size check

**Where:** `gateway/src/adapters/telegram.rs` lines ~132-148, `gateway/src/adapters/line.rs` lines ~110-126

**What's wrong:** The `content_length` check block has inconsistent indentation (extra leading spaces compared to surrounding code). More importantly, in the Telegram adapter, the file size check is placed AFTER the `continue` statement on the error path, meaning if the `continue` issue is fixed by removing it, the code flow is correct. But the indentation mismatch suggests this was pasted without proper integration.

**Fix:** Fix indentation to match the surrounding code level. Run `cargo fmt` on the gateway crate.

---

## 🟡 Minor: Deep nesting in media download logic (both adapters)

**Where:** `gateway/src/adapters/telegram.rs` lines 114-195, `gateway/src/adapters/line.rs` lines 93-175

**What's wrong:** The media download logic is nested 7-8 levels deep with `if let Ok(...)` chains. This makes the code hard to read and error-prone (as evidenced by the `continue` bug above).

**Why it matters:** Maintainability. Future contributors will struggle to understand the control flow.

**Fix:** Extract the media download logic into a helper function, e.g.:
```rust
async fn download_and_proxy_media(
    state: &AppState,
    file_id: &str,
    m_type: &str,
    download_url: &str,
) -> Option<Attachment> {
    // flat early-return style
}
```

---

## 🟡 Minor: `reqwest::Client::new()` created per-request in adapters

**Where:** `gateway/src/adapters/line.rs` line 100, `gateway/src/adapters/telegram.rs` line 116

**What's wrong:** A new `reqwest::Client` is created for every incoming media message. `reqwest::Client` internally manages a connection pool, so creating a new one per request defeats connection reuse and adds overhead.

**Fix:** Store a shared `reqwest::Client` in `AppState` (or use a `LazyLock` like the main crate does in `src/media.rs`).

---

## 🟡 Minor: `content_length` check is unreliable as a size gate

**Where:** Both adapters, the `content_length` header check

**What's wrong:** `content_length` defaults to `0` when the header is missing. A value of `0` will always pass the size check, meaning files without a `Content-Length` header bypass the limit entirely. The actual bytes are then downloaded without limit (until `r.bytes().await` completes).

**Fix:** Consider also checking `data.len()` after download and discarding if over limit:
```rust
if let Ok(data) = r.bytes().await {
    if data.len() as u64 > state.media_max_file_size {
        warn!("downloaded media exceeds size limit");
        // don't store
        continue; // or return, depending on context
    }
    // ... store
}
```

---

## 🟡 Minor: `std::sync::Mutex` used for `MediaStore` in async context

**Where:** `gateway/src/main.rs` line 36 (`pub type MediaStore = Arc<std::sync::Mutex<...>>`)

**What's wrong:** Using `std::sync::Mutex` in async code can block the tokio runtime if the lock is held across an `.await` point. Currently the lock is acquired and released quickly (no await while held), so it works. But it's fragile — a future refactor could accidentally hold it across an await.

**Fix:** This is acceptable for now since the critical section is short (just `insert`/`get`). Add a comment noting this constraint, or switch to `tokio::sync::Mutex` for safety.

---

## 🟢 Info: Good design decisions

- TTL-based cleanup with background sweep is clean and simple
- `skip_serializing_if = "Vec::is_empty"` on `attachments` maintains backward compatibility
- The `Attachment` schema is well-structured and extensible
- Reusing the existing `media::download_and_encode_image` / `media::download_and_transcribe` in `src/gateway.rs` is the right call — no duplication
- The `reply_to` auto-fill with channel-scoped tracking is a nice UX improvement
- Loopback address warning at startup is helpful for deployment debugging
- The `cron.rs` changes are purely `cargo fmt` reformatting — no logic changes, clean

---

### Review Summary (Traffic Light)

🟢 **INFO**
- Well-structured feature with clear separation: gateway proxy (download + serve) ↔ core bridge (fetch + encode for LLM)
- Schema is backward-compatible (optional field, skip when empty)
- Good test coverage reported (96 tests pass, clippy clean)
- Docs updated for LINE, Telegram, and config reference

🟡 **NIT**
- Deep nesting in media download logic — extract helper functions
- `reqwest::Client::new()` per request — use shared client
- `content_length` fallback to 0 bypasses size check when header missing
- `std::sync::Mutex` in async context — acceptable but fragile

🔴 **SUGGESTED CHANGES**
- **`continue` outside loop in Telegram adapter** — this is a compile error that CI doesn't catch because the gateway crate isn't in the PR CI pipeline. Must be fixed before merge.
- **Indentation inconsistency** in file size check blocks — run `cargo fmt` on gateway crate

### Verdict
**REQUEST_CHANGES** — The `continue` statements in `telegram.rs` are invalid Rust (not inside a loop). This will fail to compile when the gateway binary is built. Fix those, run `cargo fmt` on the gateway crate, and this is ready for another look.

masami-agent:
## Follow-up: CI Blind Spot Confirmation

After a second pass, I verified the `continue` issue with `rustc` directly:

```
error[E0268]: `continue` outside of a loop
  --> <anon>:1:23
   |
   | fn main() { if true { continue; } }
   |                       ^^^^^^^^ cannot `continue` outside of a loop
```

**Why CI does not catch this:**

The `gateway/` crate is a standalone binary with its own `Cargo.toml` — it is NOT part of a Cargo workspace. The PR CI pipeline (`ci.yml`) only compiles the root `openab` crate. The gateway is only built during release (`build-gateway.yml`, triggered by `gateway-v*` tags).

This means any compile error in `gateway/src/` will pass all PR checks and only be discovered at release time.

**Suggestion:** Consider adding a `gateway-check` job to `ci.yml` (triggered on `gateway/**` path changes) that runs `cargo check` inside the `gateway/` directory. This would catch issues like this before merge.

---

_Note: The LINE adapter's `continue` statements are valid — they are inside the `for event in webhook_body.events` loop. Only the Telegram adapter has this issue (lines 121 and 129)._

iamninihuang:
原本修的差不多了，但main branch 更新造成衝突，因為attachment其它元件也有共用，不太好解，
故我打算關閉這個 PR , 會再取最新，再重新 PR。

masami-agent:
## Re-review: PR #726 (commit `d94d437`)

Thanks for addressing the feedback! The `continue` issue in `telegram.rs` is fixed — replaced with proper `if/else` branching. ✅

However, there is a new critical issue:

---

## 🔴 Critical: Unresolved merge conflict markers in `gateway/src/main.rs`

**Where:** `gateway/src/main.rs` lines 261, 276, 292

**What's wrong:** The file contains literal merge conflict markers that were not resolved:

```
<<<<<<< Updated upstream
    let listen_addr = ...
    let ws_token = ...
    // Inbound media proxy config (Issue #690)
    let public_url = ...
    if public_url.contains("localhost") || public_url.contains("127.0.0.1") {
        warn!(...);
=======
    // Load configuration from environment
    let bot_token = std::env::var("TELEGRAM_BOT_TOKEN").expect(...)
    let secret_token = ...
    let ws_token = ...
    ...
>>>>>>> Stashed changes
    }
```

The commit message says "resolve merge conflict in main.rs after stash pop" but the conflict was **not actually resolved** — the markers are still in the file. This will not compile.

**Fix:** Resolve the conflict properly. It looks like both sides need to be merged:
- The "Updated upstream" side has the media proxy config (public_url, loopback warning)
- The "Stashed changes" side has the env var loading (bot_token, secret_token, webhook paths, cloudflare tunnel)

Both are needed. Remove the conflict markers and combine the code.

---

## ✅ Fixed from previous review

- `continue` outside loop in `telegram.rs` — **fixed** (proper `if/else` now)
- Schema updated: `url` is now `Option<String>`, new `data` field added — OAB side handles this correctly

---

## 🟢 Info: Other observations on new commits

- `Content` struct now has an `attachments` field — the OAB side deserializes it but only uses `event.attachments` (top-level). This is fine for now.
- `Attachment.url` changed from `String` to `Option<String>` — OAB side uses `.as_deref().unwrap_or("")` which is correct.

---

### Verdict
**REQUEST_CHANGES** — The merge conflict markers in `gateway/src/main.rs` must be resolved. Same CI blind spot applies: the gateway crate is not compiled in PR CI, so this won't be caught automatically.

masami-agent:
Hi @iamninihuang — thanks for fixing the `continue` issue! 🎉

One more thing to resolve: `gateway/src/main.rs` still has unresolved merge conflict markers from the stash pop (lines 261–292).

Here's what's happening:

```
<<<<<<< Updated upstream       ← side A: media proxy config
    let listen_addr = ...
    let ws_token = ...
    let public_url = ...       ← new media proxy env vars
    if public_url.contains("localhost") { warn!(...); }
=======                        ← side B: original env loading
    let bot_token = ...        ← duplicates line 326 below
    let secret_token = ...     ← duplicates line 327 below
    let ws_token = ...
    let line_channel_secret = ...  ← duplicates line 341 below
    let listen_addr = ...
    let telegram_webhook_path = ...  ← duplicates line 330 below
    ...
>>>>>>> Stashed changes
```

**How to fix:** You likely want **side A only** (the "Updated upstream" block), because:
- The env vars in side B (`TELEGRAM_BOT_TOKEN`, `LINE_CHANNEL_SECRET`, webhook paths, etc.) are already loaded later in the file (lines 326–345).
- Side B also uses `.expect()` which would panic if `TELEGRAM_BOT_TOKEN` is unset, but the existing code at line 326 uses `.ok()` (graceful — Telegram is optional).
- `cloudflare_tunnel_url` (referenced at line 313) is only defined in side B, so you'll need to keep that one line: `let cloudflare_tunnel_url = std::env::var("CLOUDFLARE_TUNNEL_URL").ok();`

**Suggested resolution:**
```rust
    let listen_addr = std::env::var("GATEWAY_LISTEN").unwrap_or_else(|_| "0.0.0.0:8080".into());
    let ws_token = std::env::var("GATEWAY_WS_TOKEN").ok();

    // Inbound media proxy config (Issue #690)
    let public_url = std::env::var("GATEWAY_MEDIA_BASE_URL")
        .or_else(|_| std::env::var("GATEWAY_PUBLIC_URL"))
        .unwrap_or_else(|_| "http://localhost:8080".into());

    // Warn if public_url looks like a loopback address
    if public_url.contains("localhost") || public_url.contains("127.0.0.1") {
        warn!(
            public_url = %public_url,
            "GATEWAY_PUBLIC_URL looks like a loopback address. ..."
        );
    }

    // Optional Cloudflare Tunnel domain for logging
    let cloudflare_tunnel_url = std::env::var("CLOUDFLARE_TUNNEL_URL").ok();
```

Then delete the conflict markers (`<<<<<<<`, `=======`, `>>>>>>>`). After that, run `cargo check` inside the `gateway/` directory to verify it compiles.

Let me know if you have any questions!

iamninihuang:
原本修的差不多了，但main branch 更新造成衝突，因為attachment其它元件也有共用，不太好解，
故我打算關閉這個 PR , 會再取最新，再重新 PR。

Issue #307, opened 2026-04-13
New @mention in main channel ignored while existing session is waiting for ACP response


agent-worker-use:
### Description

When a session is blocked waiting for a long-running ACP response (e.g. subagent), new @mentions in the main channel are silently dropped. No spawning agent log appears.

### Steps to Reproduce

Start a session that triggers a long-running ACP call (e.g. subagent pipeline)
While that session is waiting for the ACP response, send a new @mention in the main channel
Observe that no new thread or session is created — the mention is silently ignored

### Expected Behavior

Broker should create a new thread + session independently, regardless of other sessions waiting for ACP responses. maxSessions: 10 should allow concurrent sessions.

### Environment

openab 0.7.2, maxSessions: 10

### Screenshots / Logs

No spawning agent log appears for the new @mention.

CHC-Agent:
## Issue at a Glance

```
Main channel: new @mention arrives
       │
       ▼
  discord.rs handler
       │
       ├── pool.get_or_create(new_thread)
       │       │
       │       ╰─► state.write().await ──► ⏳ BLOCKED
       │            (Thread A still holds write lock
       │             during stream_prompt)
       │
       └── new mention never processed → silently dropped
```

Hi @agent-worker-use,

This overlaps with #58 / #256 — the pool write lock held during prompt streaming blocks all other operations, including creating new sessions for incoming mentions.

Two PRs are in progress to fix the underlying lock scope:
- PR #257 by @clsung
- PR #183 by @ruan330

Once the pool lock is narrowed to per-connection scope, new mentions should be able to spawn sessions independently.

chaodu-agent:
## Root Cause Analysis

This is not a bug in the mention handling logic itself — the root cause is **coarse-grained locking in the session pool**, where one stalled session freezes the entire bot. Two bugs combine:

### Bug 1: Global write lock held during entire streaming (`pool.rs`)

`with_connection()` acquires a write lock on `self.connections` and holds it for the entire duration of `stream_prompt` (which can run for minutes or hours):

```rust
pub async fn with_connection(...) {
    let mut conns = self.connections.write().await;  // global write lock
    let conn = conns.get_mut(thread_id)...;
    f(conn).await  // stream_prompt runs here — lock held the whole time
}
```

Since Tokio's `RwLock` blocks new readers when a writer is waiting, even `get_or_create` (read lock) is blocked. **One thread streaming → entire pool locked → new @mentions cannot create sessions.**

[...]

PR #760, opened 2026-05-06
fix(acp): clean up pending + cancel agent on abandoned prompts

### LINKED ISSUE #732: bug(acp): recv timeout on busy agent + stale-id routing → next prompt returns `(no response)` cascade
> All file paths and line numbers below reference upstream tag **`v0.8.3-beta.2`** (commit `1fd36df`).

### Description

After the 600s `recv_timeout` in `src/adapter.rs:345` fires (`Agent stopped responding`), **subsequent prompts in the same thread return `_(no response)_`** — the thread stays soft-bricked for as long as the user keeps prompting before the agent drains the abandoned request, and in interactive use that is effectively "stuck until the connection is reset."

This is a follow-up to #470 (which added the timeout as defense-in-depth) and a concrete instance of #76's "Assumption 2" failure mode. Sibling to #307, which describes a different visible symptom (new @mentions dropped) caused by the same root family: the broker abandons waiting on an ACP request without telling the agent or cleaning up its bookkeeping.

### Relation to other issues

- **#76** — listed three assumption failures; this report is a concrete subset of "Assumption 2: prompts always complete," but specifically about the *post-timeout state corruption* that the issue body did not call out.
- **#307** — **not a duplicate.** Same family (broker abandons ACP wait without coordinating cleanup with the agent), but distinct in three ways:
  - Visible symptom: #307 = new @mention in main channel silently dropped; this report = existing thread soft-bricked, subsequent turns return `(no response)` until the user pauses long enough for the abandoned request to drain through `notify_tx == None`.
  - Code path: #307's two analyses (CHC-Agent flowchart, chaodu-agent root-cause) blame the **pool global write lock** + **unclosed notify channel on EOF**. Both have since been fixed (per-connection serialization + #470's `*sub = None` channel close), yet this bug still reproduces — proving it is a separate defect.
  - Root posture: #307's analyses treated `notify_tx` implicitly as connection-scoped (the connection's subscriber); this report argues it must be **request-scoped** — only forward responses whose `id` matches the current request. A fix at this layer would also harden #307's family of failures against future regressions.
- **#470** — explicitly framed the 600s timeout as "defense-in-depth … even if other channel-closing bugs exist." This issue documents one of those "other bugs" the timeout exposes rather than fixes.

### Root cause

The notification dispatcher in `src/acp/connection.rs:284-303` routes any inbound message that has an `id` to **the currently registered `notify_tx`** — without checking whether that `id` corresponds to the request the current subscriber is waiting on.

Combined with the timeout in `src/adapter.rs:345-354`:

1. **Prompt 1** is sent → `session_prompt` registers `notify_tx = tx1` and `pending[id=1] = oneshot`. Request `id=1` goes out on stdin.
2. Agent runs a long tool (e.g. `Bash(npm run build)`) that emits no ACP notifications for more than 10 minutes. Recv loop times out → `response_error = "Agent stopped responding"` → `prompt_done()` (sets `notify_tx = None`).
   - **`pending[id=1]` is not removed.**
   - **No `session/cancel` is sent to the agent.** The agent keeps running prompt 1 in the background.
3. **Prompt 2** is sent → registers `notify_tx = tx2`, `pending[id=2]`. Sends `id=2`.
4. Agent eventually finishes prompt 1 and emits the final response (`id=1`). Reader at `connection.rs:284` sees `id=1`, removes `pending[1]`, and forwards the message to **whichever subscriber is currently registered — now `tx2`**.
5. The `adapter.rs:356` loop reading from `tx2` sees `notification.id.is_some()` and treats it as the completion of prompt 2, breaking immediately with empty `text_buf`.
6. `final_content.is_empty()` and `response_error == None` → outputs `_(no response)_`.
7. Prompt 2's true `id=2` response arrives later. If a `tx3` is already registered (because the user sent prompt 3 before id=2 returned), it is forwarded there and shorts out prompt 3 the same way. **The cycle persists as long as each new prompt is sent before the previous prompt's true response arrives** — easy to trigger when the agent is consistently behind by one prompt.

`session_prompt()` already returns `(rx, request_id)` — but `adapter.rs:290` discards it: `let (mut rx, _) = conn.session_prompt(content_blocks).await?;`. So the subscriber has no way to filter mismatched ids.

### Steps to Reproduce

1. In any thread, send a prompt that triggers a long-running tool with no intermediate output for more than 10 minutes. The simplest reproducer is a `Bash` tool running a slow build / sleep, e.g.:
   ```
   run `bash -c 'sleep 700 && echo done'`
   ```
2. Wait for the broker to post `⚠️ Agent stopped responding`.
3. Send any new prompt in the same thread (e.g. `hi`). Observe: response is `_(no response)_`.
4. Sending further prompts faster than the agent can return each true response keeps the cascade alive (each new prompt receives the previous prompt's stale `id`). Once the user pauses long enough for the agent to drain its backlog with `notify_tx == None`, the next prompt recovers — but in interactive use the thread is functionally stuck.

### Expected Behavior

After a recv timeout, the agent process may still be alive but is considered abandoned by the broker. Subsequent prompts in the same thread should either:

- **be processed as new, independent prompts** — and in the rare case the broker does abandon a prompt (process death, hard ceiling, etc.), stale responses from the abandoned prompt must be suppressed (request-id matching) and the abandoned prompt cancelled via `session/cancel` so the agent stops doing work the user has given up on, **or**
- **fail fast with a clear error** (e.g. "session reset required") and trigger automatic session reset on the next turn.

The thread must not be silently soft-bricked into a `(no response)` cascade that is effectively stuck under any normal interactive pace.

### Environment

- `openab` `v0.8.3-beta.2` (commit `1fd36df`, latest tag at time of writing); the same code path is present back to `v0.8.2-beta.3` when #470 added the recv timeout.
- ACP backend: `claude-agent-acp` 0.29.2 (Claude Code 2.1.114)
- Channel: Discord
- Confirmed with Claude Code; likely reproduces with any ACP backend that can run a tool exceeding the 600s recv timeout without intermediate notifications, since the bug lives in broker-side `notify_tx` routing, not in any backend-specific behavior.

### Suggested Fix

**The recommended fix is all three: (A) + (B) + (C).**

- **(A)** replaces the flat 600s recv timeout with a liveness-aware loop, eliminating the typical trigger (long tools mis-classified as agent death).
- **(B)** makes the rare remaining "abandon" paths safe by cancelling the agent-side prompt and clearing the orphaned `pending` entry whenever the broker gives up.
- **(C)** is a cheap routing-layer invariant ("subscriber only accepts its own `request_id`") that prevents this family of cascade from re-emerging if any future code path forgets (B).

All three are in the broker; no agent-side change required.

#### (A) Replace flat timeout with liveness-aware loop — `src/adapter.rs:343-360`

Replace the flat 600s `tokio::time::timeout` around `rx.recv()` with a `tokio::select!` loop that distinguishes "agent process is alive but busy" from "agent process is actually stuck/dead":

```rust
let prompt_start = Instant::now();
let hard_timeout = Duration::from_secs(30 * 60); // consider making configurable in a follow-up
let liveness_check = Duration::from_secs(30);
loop {
    tokio::select! {
        msg = rx.recv() => { /* handle notification (or channel-closed → break) */ }
        _ = tokio::time::sleep(liveness_check) => {
            if !conn.alive() {
                response_error = Some("Agent process died".into());
                // apply (B) cleanup before break
                break;
            }
            if prompt_start.elapsed() > hard_timeout {
                response_error = Some("Agent exceeded hard timeout".into());
                // apply (B) cleanup before break
                break;
            }
            // alive and under ceiling → keep waiting
        }
    }
}
```

`AcpConnection::alive()` already exists at `src/acp/connection.rs:539` and is cheap (just `!reader_handle.is_finished()`). The hard ceiling acts purely as a safety net against runaway sessions, not as the primary liveness signal. `tokio::mpsc::UnboundedReceiver::recv()` is cancel-safe ([tokio docs](https://docs.rs/tokio/latest/tokio/sync/mpsc/struct.UnboundedReceiver.html#cancel-safety)), so dropping the future when the sleep arm fires is safe.

This is the same shape proposed in the closed-unmerged PR #77, restricted here to just the loop-structure change. With (A), the timeout only fires when the agent really is gone — which is what `Agent stopped responding` should mean. Legitimate long-running tools (`npm run build`, large `cargo test`, slow migrations, sequential `WebFetch`) no longer trip the 10-minute timeout at all, which is what removes the trigger for the cascade described in this report.

#### (B) Clean up at every abandon point

Whenever the loop in (A) decides to abandon the in-flight prompt (process died, hard ceiling exceeded, or any future "give up" branch), do two things before breaking:

- Call `SessionPool::cancel_session()` (already exists, lock-free via the stored stdin handle) so the agent stops working on the abandoned prompt.
- Remove the orphaned `pending[request_id]` entry so any late response is dropped at the reader instead of being routed to a future subscriber.

This is what closes the cascade in the rare case (A)'s hard ceiling does fire while the agent is still emitting late responses. Without (B), the same race described in this report can still happen at 30 min instead of 10 min.

#### (C) Defensive request-id matching — `src/adapter.rs:290`

To enforce request-scoped routing as a defense-in-depth invariant, capture the request id from `session_prompt()` and ignore any completion whose id doesn't match the current request:

```rust
let (mut rx, request_id) = conn.session_prompt(content_blocks).await?;
// ...
if let Some(notification_id) = notification.id {
    if notification_id != request_id {
        // Stale response from a previously-abandoned prompt; ignore.
        continue;
    }
    if let Some(ref err) = notification.error {
        response_error = Some(format_coded_error(err.code, &err.message));
    }
    break;
}
```

Even with (A) + (B) in place, this invariant is cheap and prevents any future code path that abandons a prompt without going through (B) from re-introducing the same routing bug. `session_prompt()` already returns `(rx, request_id)` for exactly this purpose; today's `let (mut rx, _) = ...` discards it.

---

#### Fallback ranking if you can't ship all three

Listed best to worst — each is strictly weaker than (A)+(B)+(C):

- **(A) + (B)** is the principled fix without (C)'s routing-layer invariant: timeouts fire only when the agent is gone, and when they do, both sides agree the prompt is dead. Any future "abandon" path that forgets (B)'s cleanup re-opens the cascade.
- **(B) alone** keeps the 600s flat timeout (so long tools still trip it), but at least cleans up properly when it does — eliminates the cascade without addressing the misclassification.
- **(A) alone** removes the trigger for the typical case (long tools) but leaves the underlying race in place — the cascade still reproduces if the agent ever truly hangs past the 30-min ceiling. Not recommended without (B).
- **(B) + (C) without (A)** is the smallest patch that stops the production symptom on today's code, at the cost of continuing to mis-classify long tools as agent death.
- **(C) alone** is the cheapest change and silences the visible `(no response)` symptom, but the agent still wastes tokens running prompts the broker has abandoned, and there is no `session/cancel`. Acceptable only as a stopgap.

brettchien:
Closes #732.

## Problem

_Originally reported on Discord: https://discord.com/channels/1491295327620169908/1491365158868619404/1500930306620920040_

Prompts running longer than 600 s triggered `recv_timeout` in `adapter.rs:386`. The broker printed "Agent stopped responding" and broke out of the recv loop, but:

- `pending[request_id]` was not removed
- No `session/cancel` was sent — the agent kept running
- When the agent eventually finished, it emitted a response with the original `id`. `connection.rs:286` found the stale `pending` entry and forwarded it to the **current** prompt's `notify_tx` — causing that prompt to break immediately with empty `text_buf` → `(no response)`
- The cascade repeated for every subsequent prompt until the backlog drained

## Fix (A + B + C)

**(A)** Replace flat 600 s timeout with a `tokio::select!` loop:
- recv arm — normal path
- 30 s liveness arm — checks `conn.alive()` + configurable hard ceiling
- Long-running tools no longer trip the timeout; only a dead reader or the hard ceiling abandons the prompt

**(B)** `AcpConnection::abandon_request(request_id)` — drops `pending[request_id]` and best-effort sends `session/cancel` on every abandon path

**(C)** Capture `request_id` from `session_prompt()` (was discarded as `_`); skip notifications whose `id` doesn't match — defense-in-depth if any future abandon path misses `abandon_request`

## Config

New `pool.prompt_hard_timeout_secs` (default `1800` s = 30 min). Acts as a safety net against runaway sessions while the 30 s liveness check covers the typical case.

## Testing

- [x] 238 existing tests pass (`cargo test --bin openab`)
- [x] `cargo clippy -- -D warnings` clean
- [ ] Manual repro: `bash -c 'sleep 700 && echo done'`, wait past 600 s, send follow-up → real reply (no `(no response)` cascade)
- [ ] Manual: kill agent mid-prompt → recv loop bails within ~30 s with `Agent process died`, next prompt works clean

No unit test for `abandon_request` — no test seam without a real subprocess; covered end-to-end via real ACP backends. `session/cancel` shape matches existing `cancel_session` / `reset_session` in `pool.rs` (prod-verified).

## Refs
- #76 (Assumption 2: prompts always complete)
- #307 (sibling symptom)
- #470 (introduced the 600 s timeout)

### DIFF
--- a/config.toml.example
+++ b/config.toml.example
@@ -106,6 +106,8 @@ max_sessions = 10
 session_ttl_hours = 24
 # Hard ceiling (sec) per prompt; see #732. Default: 1800 (30 min).
 # prompt_hard_timeout_secs = 1800
+# Liveness-check cadence (sec) for the recv loop; see #732. Default: 30.
+# liveness_check_secs = 30
 
 [markdown]
 tables = "code"  # "code" (default) | "bullets" | "off"
--- a/config.toml.example
+++ b/config.toml.example
@@ -104,6 +104,8 @@ working_dir = "/home/agent"
 [pool]
 max_sessions = 10
 session_ttl_hours = 24
+# Hard ceiling (sec) per prompt; see #732. Default: 1800 (30 min).
+# prompt_hard_timeout_secs = 1800
 
 [markdown]
 tables = "code"  # "code" (default) | "bullets" | "off"
--- a/src/acp/connection.rs
+++ b/src/acp/connection.rs
@@ -6,7 +6,7 @@ use serde_json::{json, Value};
 use std::collections::HashMap;
 use std::sync::atomic::{AtomicU64, Ordering};
 use std::sync::Arc;
-use tokio::io::{AsyncBufReadExt, AsyncWriteExt, BufReader};
+use tokio::io::{AsyncBufReadExt, AsyncRead, AsyncWrite, AsyncWriteExt, BufReader};
 use tokio::process::{Child, ChildStdin};
 use tokio::sync::{mpsc, oneshot, Mutex};
 use tokio::task::JoinHandle;
@@ -149,6 +149,112 @@ fn build_agent_env(
     (result, inherited)
 }
 
+/// Reader loop body: reads JSON-RPC messages from `reader`, auto-replies
+/// `session/request_permission` via `writer`, resolves pending responses,
+/// and forwards notifications + stale id-bearing messages to the active
+/// subscriber. Extracted as a free generic function so unit tests can drive
+/// it with `tokio::io::duplex()` halves instead of a real child process.
+pub(crate) async fn run_reader_loop<R, W>(
+    reader: R,
+    writer: Arc<Mutex<W>>,
+    pending: Arc<Mutex<HashMap<u64, oneshot::Sender<JsonRpcMessage>>>>,
+    notify_tx: Arc<Mutex<Option<mpsc::UnboundedSender<JsonRpcMessage>>>>,
+) where
+    R: AsyncRead + Unpin + Send + 'static,
+    W: AsyncWrite + Unpin + Send + 'static,
+{
+    let mut reader = BufReader::new(reader);
+    let mut line = String::new();
+    loop {
+        line.clear();
+        match reader.read_line(&mut line).await {
+            Ok(0) => break, // EOF
+            Ok(_) => {}
+            Err(e) => {
+                error!("reader error: {e}");
+                break;
+            }
+        }
+        let msg: JsonRpcMessage = match serde_json::from_str(line.trim()) {
+            Ok(m) => m,
+            Err(_) => continue,
+        };
+        debug!(line = line.trim(), "acp_recv");
+
+        // Auto-reply session/request_permission
+        if msg.method.as_deref() == Some("session/request_permission") {
+            if let Some(id) = msg.id {
+                let title = msg
+                    .params
+                    .as_ref()
+                    .and_then(|p| p.get("toolCall"))
+                    .and_then(|t| t.get("title"))
+                    .and_then(|t| t.as_str())
+                    .unwrap_or("?");
+
+                let outcome = build_permission_response(msg.params.as_ref());
+                info!(title, %outcome, "auto-respond permission");
+                let reply = JsonRpcResponse::new(id, outcome);
+                if let Ok(data) = serde_json::to_string(&reply) {
+                    let mut w = writer.lock().await;
+                    let _ = w.write_all(format!("{data}\n").as_bytes()).await;
+                    let _ = w.flush().await;
+                }
+            }
+            continue;
+        }
+
+        // Response (has id) → resolve pending AND forward to subscriber
+        if let Some(id) = msg.id {
+            let mut map = pending.lock().await;
+            if let Some(tx) = map.remove(&id) {
+                // Forward to subscriber so they see the completion
+                let sub = notify_tx.lock().await;
+                if let Some(ntx) = sub.as_ref() {
+                    // Clone the essential fields for the subscriber
+                    let _ = ntx.send(JsonRpcMessage {
+                        id: Some(id),
+                        method: None,
+                        result: msg.result.clone(),
+                        error: msg.error.clone(),
+                        params: None,
+                    });
+                }
+                let _ = tx.send(msg);
+                continue;
+            }
+            // Stale id (#732): pending was already abandoned. Falls through
+            // to subscriber forwarding; the adapter recv loop filters by
+            // request_id so it can't leak into the next prompt.
+            trace!(request_id = id, "stale id-bearing message after abandon");
+        }
+
+        // Notification → forward to subscriber
+        let sub = notify_tx.lock().await;
+        if let Some(tx) = sub.as_ref() {
+            let _ = tx.send(msg);
+        }
+    }
+
+    // Connection closed — resolve all pending with error
+    let mut map = pending.lock().await;
+    for (_, tx) in map.drain() {
+        let _ = tx.send(JsonRpcMessage {
+            id: None,
+            method: None,
+            result: None,
+            error: Some(crate::acp::protocol::JsonRpcError {
+                code: -1,
+                message: "connection closed".into(),
+            }),
+            params: None,
+        });
+    }
+    // Close the notify channel so rx.recv() returns None
+    let mut sub = notify_tx.lock().await;
+    *sub = None;
+}
+
 impl AcpConnection {
     pub async fn spawn(
         command: &str,
@@ -254,106 +360,12 @@ impl AcpConnection {
         let notify_tx: Arc<Mutex<Option<mpsc::UnboundedSender<JsonRpcMessage>>>> =
             Arc::new(Mutex::new(None));
 
-        let reader_handle = {
-            let pending = pending.clone();
-            let notify_tx = notify_tx.clone();
-            let stdin_clone = stdin.clone();
-            tokio::spawn(async move {
-                let mut reader = BufReader::new(stdout);
-                let mut line = String::new();
-                loop {
-                    line.clear();
-                    match reader.read_line(&mut line).await {
-                        Ok(0) => break, // EOF
-                        Ok(_) => {}
-                        Err(e) => {
-                            error!("reader error: {e}");
-                            break;
-                        }
-                    }
-                    let msg: JsonRpcMessage = match serde_json::from_str(line.trim()) {
-                        Ok(m) => m,
-                        Err(_) => continue,
-                    };
-                    debug!(line = line.trim(), "acp_recv");
-
-                    // Auto-reply session/request_permission
-                    if msg.method.as_deref() == Some("session/request_permission") {
-                        if let Some(id) = msg.id {
-                            let title = msg
-                                .params
-                                .as_ref()
-                                .and_then(|p| p.get("toolCall"))
-                                .and_then(|t| t.get("title"))
-                                .and_then(|t| t.as_str())
-                                .unwrap_or("?");
-
-                            let outcome = build_permission_response(msg.params.as_ref());
-                            info!(title, %outcome, "auto-respond permission");
-                            let reply = JsonRpcResponse::new(id, outcome);
-                            if let Ok(data) = serde_json::to_string(&reply) {
-                                let mut w = stdin_clone.lock().await;
-                                let _ = w.write_all(format!("{data}\n").as_bytes()).await;
-                                let _ = w.flush().await;
-                            }
-                        }
-                        continue;
-                    }
-
-                    // Response (has id) → resolve pending AND forward to subscriber
-                    if let Some(id) = msg.id {
-                        let mut map = pending.lock().await;
-                        if let Some(tx) = map.remove(&id) {
-                            // Forward to subscriber so they see the completion
-                            let sub = notify_tx.lock().await;
-                            if let Some(ntx) = sub.as_ref() {
-                                // Clone the essential fields for the subscriber
-                                let _ = ntx.send(JsonRpcMessage {
-                                    id: Some(id),
-                                    method: None,
-                                    result: msg.result.clone(),
-                                    error: msg.error.clone(),
-                                    params: None,
-                                });
-                            }
-                            let _ = tx.send(msg);
-                            continue;
-                        }
-                        // Stale id (#732): pending was already abandoned. Falls through
-                        // to subscriber forwarding; the adapter recv loop filters by
-                        // request_id so it can't leak into the next prompt. No automated
-                        // test seam — the path only triggers when a real subprocess
-                        // delivers a late response after abandon_request, which is
-                        // covered by manual repro against a live agent.
-                        trace!(request_id = id, "stale id-bearing message after abandon");
-                    }
-
-                    // Notification → forward to subscriber
-                    let sub = notify_tx.lock().await;
-                    if let Some(tx) = sub.as_ref() {
-                        let _ = tx.send(msg);
-                    }
-                }
-
-                // Connection closed — resolve all pending with error
-                let mut map = pending.lock().await;
-                for (_, tx) in map.drain() {
-                    let _ = tx.send(JsonRpcMessage {
-                        id: None,
-                        method: None,
-                        result: None,
-                        error: Some(crate::acp::protocol::JsonRpcError {
-                            code: -1,
-                            message: "connection closed".into(),
-                        }),
-                        params: None,
-                    });
-                }
-                // Close the notify channel so rx.recv() returns None
-                let mut sub = notify_tx.lock().await;
-                *sub = None;
-            })
-        };
+        let reader_handle = tokio::spawn(run_reader_loop(
+            stdout,
+            stdin.clone(),
+            pending.clone(),
+            notify_tx.clone(),
+        ));
 
         Ok(Self {
             _proc: proc,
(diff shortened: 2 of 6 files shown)


shaun-agent:
## OpenAB PR Screening
- Title: fix(acp): clean up pending + cancel agent on abandoned prompts
- Source: https://github.com/openabdev/openab/pull/760
- Generated at: 2026-05-06T15:31:29.327Z
- Discord thread: https://discord.com/channels/1488041051187974246/1501607147782279412
<details>
## Intent

PR #760 is trying to fix an ACP prompt lifecycle bug where long-running prompts were incorrectly treated as dead after a flat 600-second receive timeout.

The operator-visible problem is that OpenAB reports `Agent stopped responding`, but does not fully clean up the abandoned request. The stale pending request can later receive an old agent response and incorrectly deliver it into a newer prompt, causing `(no response)` cascades for subsequent prompts.

## Feat

This is a bug fix for ACP request/session handling.

It changes prompt execution behavior so long-running tools are not abandoned just because they exceed 600 seconds. Instead, OpenAB uses a liveness-check loop, abandons requests explicitly when needed, removes stale pending entries, best-effort sends `session/cancel`, and ignores mismatched response IDs as defense in depth.

It also adds a configurable hard ceiling: `pool.prompt_hard_timeout_secs`, defaulting to 1800 seconds.

## Who It Serves

Primary beneficiaries:

- Agent runtime operators
- Discord and Slack end users whose prompts depend on ACP agents
- Maintainers debugging long-running or stuck ACP sessions
- Reviewers responsible for reliability of prompt dispatch and session cleanup

## Rewritten Prompt

Fix ACP prompt abandonment so stale responses from timed-out or dead prompts cannot corrupt later prompts.

Replace the flat 600-second receive timeout with a loop that separately handles normal responses, periodic agent liveness checks, and a configurable hard timeout. When a prompt is abandoned for timeout, dead process, dropped channel, or similar failure, remove its pending request entry and best-effort send `session/cancel`. Preserve the request ID returned by `session_prompt()` and ensure notifications are only accepted when their response ID matches the active request.

Add configuration for the hard timeout with a safe default, update example config, and verify existing tests and clippy. Add targeted tests if practical; otherwise document why subprocess-backed ACP behavior requires manual or integration coverage.

## Merge Pitch

This is worth advancing because it fixes a concrete reliability failure: one abandoned ACP prompt can poison later prompts until the stale backlog clears. The fix addresses both the immediate timeout behavior and the deeper cleanup issue around pending request ownership.

Risk profile is moderate. The change touches prompt dispatch, ACP connection state, config parsing, and process/session cancellation behavior. The main reviewer concern should be whether all abandon paths consistently call the new cleanup path, and whether `session/cancel` is safe to send for every abandoned request.

## Best-Practice Comparison

Relevant OpenClaw principles:

- Gateway-owned scheduling: partially relevant. This PR improves gateway-side ownership of prompt lifecycle decisions.
- Durable job persistence: not directly relevant. The bug is in live ACP request routing, not persisted jobs.
- Isolated executions: relevant in spirit. Stale responses should not affect later prompts.
- Explicit delivery routing: highly relevant. Matching response IDs to active request IDs prevents cross-prompt delivery.
- Retry/backoff and run logs: not directly addressed. Better logs around abandon/cancel paths may be useful follow-up.

Relevant Hermes Agent principles:

- Gateway daemon tick model: partially relevant. The new 30-second liveness check resembles periodic supervision.
- File locking to prevent overlap: not relevant unless ACP sessions are persisted or shared across processes.
- Atomic writes for persisted state: not relevant.
- Fresh session per scheduled run: conceptually relevant. The fix prevents stale session output from contaminating later work, though it does not create fresh sessions per prompt.
- Self-contained prompts for scheduled tasks: not relevant to this specific PR.

Overall, the PR aligns most strongly with explicit delivery routing, isolated execution behavior, and gateway-owned lifecycle supervision.

## Implementation Options

Conservative option: minimal cleanup fix  
Keep the existing timeout shape but ensure every timeout path removes `pending[request_id]`, sends best-effort `session/cancel`, and drops mismatched response IDs. This is smaller but still leaves long-running valid tools vulnerable to the flat timeout.

Balanced option: accept this PR’s direction  
Replace the flat timeout with recv/liveness/hard-timeout supervision, add `abandon_request`, preserve request IDs, and reject mismatched notifications. This fixes the known cascade while keeping the design local to ACP prompt handling.

Ambitious option: formal ACP request lifecycle manager  
Introduce an explicit request state machine with states like `pending`, `active`, `abandoned`, `cancel_sent`, `completed`, and `failed`. Add structured run logs, metrics, integration tests with a fake ACP subprocess, and centralized cancellation semantics across prompt, reset, and session shutdown paths.

## Comparison Table

| Conservative cleanup only | High | Low | Medium | Medium | Medium | Decent, but incomplete |
| Balanced PR approach | Medium | Medium | High | Good | High | Best fit |
| Full lifecycle manager | Low | High | Very high | High if done well | High | Better as follow-up |

## Recommendation

Advance the balanced approach from this PR, with reviewer focus on abandonment coverage, cancellation safety, and response-ID routing.

This is the right next step because it fixes the observed production failure without forcing a larger ACP lifecycle redesign into the same merge. A good follow-up would be a separate PR adding a fake ACP backend or subprocess test harness so timeout, cancel, stale-response, and dead-agent cases can be covered automatically.
</details>

chaodu-agent:
## 🟢 Four-monk review — no blocking issues

**Verdict: Approve.** The 3-layer fix (select loop + abandon_request + stale id filtering) is well-structured and addresses the root cause of #732 cleanly.

### Highlights

- `tokio::select!` replaces the flat 600s timeout — long-running tools no longer get killed; only a dead reader or the hard ceiling triggers abandon
- `abandon_request()` removes `pending[request_id]` and best-effort sends `session/cancel` (confirmed: ACP spec defines this as a client notification, no `id`, no response expected)
- Stale id filtering provides defense-in-depth against any future missed abandon path
- `None => break` path is safe — reader loop already drains all pending entries on EOF

### Non-blocking NITs

| # | Note | Suggestion |
|---|------|------------|
| 1 | `LIVENESS_CHECK_INTERVAL` hardcoded at 30s | Could be configurable for operators; fine as follow-up |
| 2 | `session/cancel` response silently dropped if agent replies | Add `trace!` log when id-bearing response has no pending entry — improves observability |
| 3 | Hard timeout format shows `0m` if < 60s | Edge case; consider `as_secs()` display or min-value guard in config |

<details>
<summary>Reviewer breakdown</summary>

| Reviewer | Verdict | Extra observations |
|----------|---------|-------------------|
| 超渡法師 (Claude) | 🟢 | Baseline check confirmed leak path; verified `tokio::select!` race is safe for UnboundedReceiver |
| 普渡法師 (Claude) | 🟢 | Verified `None => break` assumption (reader drains pending on EOF); flagged format edge case |
| 擺渡法師 (Codex) | 🟢 | Confirmed `session/cancel` is spec-compliant notification; GitHub checks passing |
| 覺渡法師 (Gemini) | 🟢 | Praised select loop design; agreed on trace log suggestion |

</details>

brettchien:
## Follow-up — applied all 3 NITs

Thanks for the review. All three NITs addressed in `f323bb0`.

| # | NIT | Resolution |
|---|-----|------------|
| 1 | `LIVENESS_CHECK_INTERVAL` hardcoded at 30s | Hoisted onto `PoolConfig` as `pool.liveness_check_secs` (default 30, documented in `config.toml.example`). Wired through `AdapterRouter::new`. |
| 2 | id-bearing response has no pending entry (stale-id observability) | Added `trace!(request_id = id, "stale id-bearing message after abandon")` at the fall-through path in `acp/connection.rs`. Behaviour unchanged — the message still falls through to subscriber forwarding; the adapter recv loop's `request_id` filter remains the actual safety net. The trace just makes the stale path observable. |
| 3 | Hard timeout format shows `0m` if < 60s | Format changed from `({}m)` / `as_secs() / 60` to `({}s)` / `as_secs()`. Renders correctly for any positive ceiling (e.g. `90s` instead of `1m`). |

`cargo check && cargo clippy -- -D warnings && cargo test --bin openab` clean (238 passed).

chaodu-agent (dismissed):
## ✅ Approve — all NITs addressed

Verified commit `f323bb0` resolves the three NITs from my earlier review:

1. ✅ `liveness_check_secs` now configurable via `PoolConfig` (was hardcoded 30s)
2. ✅ `trace!(request_id = id, "stale id-bearing message after abandon")` added for observability
3. ✅ Hard timeout format uses `as_secs()` — no more `0m` edge case

The 3-layer fix (select loop + abandon_request + stale id filtering) is solid. `cargo test` + `clippy` clean. Ready to merge.

chaodu-agent:
## 🟢 Review: fix(acp): clean up pending + cancel agent on abandoned prompts

**Verdict: LGTM** — The fix is correct, well-structured, and addresses the root cause comprehensively.

### What this PR solves

The flat 600s `recv_timeout` fired "Agent stopped responding" without removing `pending[request_id]` or sending `session/cancel`. Late responses leaked into subsequent prompts causing `(no response)` cascades. The three-layer fix (A: liveness loop, B: abandon cleanup, C: stale ID filter) is robust.

### 🟢 INFO — Done well

- Root cause correctly identified and all three vectors closed
- `tokio::select!` loop is idiomatic — no spurious wakeups
- `abandon_request` is fire-and-forget safe (errors swallowed, pending removed first)
- Config defaults reasonable (30s liveness, 30min hard ceiling)
- Backward-compatible — `#[serde(default)]` means existing configs work unchanged
- Trace-level log for stale messages — good observability without spam

### 🟡 NIT — Non-blocking

1. **Liveness vs hard timeout clamp**: If `liveness_check_secs > prompt_hard_timeout_secs`, the hard timeout effectively never fires. Consider a startup warning or clamp.

2. **Cancel response noise**: `abandon_request` allocates a new request ID for `session/cancel`. The response will hit the "stale id" trace log. Harmless but a comment noting this is intentional would help future readers.

3. **`std::time::Instant` vs `tokio::time::Instant`**: `prompt_start` uses `std::time::Instant` inside async context. Fine for elapsed checks, but `tokio::time::Instant` is preferred for consistency (especially under `tokio::time::pause()` in tests).

4. **No unit test for stale-ID filter**: Acceptable given no test seam without a subprocess. A code comment noting it is covered by manual repro would help.

<details>
<summary>Baseline verification</summary>

Verified against `main`:
- `adapter.rs:386` — confirmed flat 600s `recv_timeout` with no cleanup
- `connection.rs` — confirmed `session_prompt()` returns `(rx, request_id)` but caller discarded ID as `_`
- `connection.rs:319` — confirmed pending lookup forwards to current subscriber without ID validation

All three gaps are closed by this PR.
</details>

---

*Reviewed by 超渡法師 🪬*

brettchien:
## Follow-up — applied all 4 round-2 NITs

Thanks for the second pass. All four NITs addressed across `1a65244` (NITs 1+3), `ccef788` (NIT 4 + Option A refactor), and `5183143` (NIT 2 doc clarification).

| # | NIT | Resolution | Commit |
|---|-----|------------|--------|
| 1 | Liveness vs hard timeout clamp | Added `warn!` at startup in `AdapterRouter::new` when `liveness_check_secs >= prompt_hard_timeout_secs` — the user is told the hard ceiling will only fire after the next liveness tick. Not clamped (kept the values authoritative; warning is enough to surface the misconfig). | `1a65244` |
| 2 | Cancel response hits stale-id trace — comment that this is intentional | Expanded `abandon_request` doc comment: `session/cancel` carries a fresh JSON-RPC id but is intentionally not registered in `pending`, so the agent's reply lands in the stale-id branch of `run_reader_loop` and only emits `trace!`. We never wait on the cancel response; the adapter recv loop's `request_id` filter is the actual safety net against leakage. No behavioural change. | `5183143` |
| 3 | `std::time::Instant` → `tokio::time::Instant` | `prompt_start` in `AdapterRouter::dispatch_loop` switched to `tokio::time::Instant::now()`. Plays well with `tokio::time::pause()` for future tests. | `1a65244` |
| 4 | No unit test for stale-id filter | Refactored `AcpConnection::spawn` reader loop body into a free generic `pub(crate) async fn run_reader_loop<R, W>` (bounds `AsyncRead/AsyncWrite + Unpin + Send + 'static`). Production path unchanged — `spawn()` is a single `tokio::spawn(run_reader_loop(...))` call. Added two `tokio::io::duplex()`-driven tests in `reader_loop_tests`: <br>• `stale_id_response_is_forwarded_without_pending_entry` — directly covers the #732 fall-through path<br>• `matched_id_response_resolves_pending_and_forwards` — regression guard for the resolve+forward dual branch | `ccef788` |

`cargo check && cargo clippy --bin openab -- -D warnings && cargo test --bin openab` clean (252 passed, +2 new).

chaodu-agent (approved):
## ✅ Re-approve — all round-2 NITs addressed

Verified final state (commits `1a65244` through `5183143`):

1. ✅ **Startup warning** when `liveness_check_secs >= prompt_hard_timeout_secs` — operator-visible, not a silent clamp
2. ✅ **Cancel stale-id intent documented** — `abandon_request` doc comment clarifies the intentional non-registration in `pending`
3. ✅ **`tokio::time::Instant`** — consistent with the `tokio::time::sleep` in the same `select!` arm
4. ✅ **Reader loop unit tests** — `run_reader_loop<R, W>` extracted as generic free function; 2 duplex-driven tests cover stale-id forwarding and matched-id resolve+forward

CI green (all 9 checks pass). 252 tests pass. The 3-layer fix is solid and production-ready.

*Reviewed by 超渡法師 🪬*

chaodu-agent (changes requested):
<@1490365068863606784> Review pass from 擺渡法師.

Requesting changes. I found two issues in the latest follow-up commits.

1. Blocking: `abandon_request()` now sends `session/cancel` with a JSON-RPC `id` (`src/acp/connection.rs:593-598`). ACP defines `session/cancel` as a client notification: no response is expected, and the protocol examples omit `id`. Existing OpenAB `/cancel` and reset paths in `src/acp/pool.rs` also omit `id`. With the new `id`, this is a request-shaped message, so spec-compliant behavior is no longer guaranteed and agents may send an extra response that the PR then has to route as stale. Please remove the `id` and update the doc comment back to notification semantics.

Source: https://agentclientprotocol.com/protocol/draft/schema#sessioncancel

2. Blocking/config hardening: `pool.liveness_check_secs` deserializes as a raw `u64` and is converted directly into a `Duration` used by `tokio::time::sleep` inside the recv loop (`src/config.rs:323-329`, `src/adapter.rs:172-188`, `src/adapter.rs:423-433`). If an operator sets `liveness_check_secs = 0`, the sleep arm is immediately ready and the loop can spin continuously while the prompt is still under the hard timeout. Please reject zero or clamp it to a sane minimum before constructing the router.

I could not run local Rust checks here because this container does not have `cargo` installed (`cargo: command not found`). GitHub CI for check and smoke tests is green, aside from the unrelated `check-pending` automation failure.

brettchien:
## Follow-up — applied both round-3 blocking NITs

Thanks for the third pass. Both NITs addressed in `a162061`.

### NIT 1 — `abandon_request` sending request-shaped `session/cancel`

Removed the JSON-RPC `id` field from the `session/cancel` payload in `abandon_request`. The call is now notification-style, matching the ACP spec ([protocol/draft/schema#sessioncancel](https://agentclientprotocol.com/protocol/draft/schema#sessioncancel)) and aligning with the existing `pool.rs::cancel_session` / `reset_session` convention.

The doc comment was reverted from the round-2 stale-id-rationale to plain notification semantics. The stale-id forwarding branch in `run_reader_loop` and the `request_id` filter in the adapter recv loop are kept as defense-in-depth — they're still useful for any other source of late id-bearing messages, just no longer triggered by `abandon_request` itself.

### NIT 2 — `liveness_check_secs = 0` could spin the recv loop

Added `anyhow::ensure!(config.pool.liveness_check_secs > 0, ...)` in `parse_config` (`src/config.rs`), grouped with the existing `max_buffered_messages > 0` / `max_batch_tokens > 0` validations. Misconfigured zero now fails fast at startup with a clear message; no clamp / silent rewrite.

`cargo check && cargo clippy --bin openab -- -D warnings && cargo test --bin openab` clean (252 passed).

chaodu-agent (approved):
## 🟢 Approved

All three layers of the fix are correct and well-tested:

- **(A)** `tokio::select!` liveness loop replaces the flat 600s timeout — long-running tools survive; dead agents are caught within ~30s
- **(B)** `abandon_request()` removes `pending[request_id]` + best-effort `session/cancel` (notification-style, per ACP spec)
- **(C)** Stale ID filter in the adapter recv loop — defense-in-depth against any future missed abandon path

Contributor addressed all NITs across 3 rounds:
- Configurable `liveness_check_secs` + `prompt_hard_timeout_secs`
- Startup warning when liveness >= hard timeout
- `tokio::time::Instant` for test compatibility
- Zero-value validation in `parse_config`
- `session/cancel` corrected to notification (no `id`)
- Reader loop extracted as generic `run_reader_loop` with 2 unit tests covering stale-id and matched-id paths

CI green (252 tests + all smoke tests pass). Ship it. 🚀

*Reviewed by 超渡法師 🪬*

thepagent (approved):

PR #785, opened 2026-05-10
fix(cron): translate POSIX day-of-week to match documented semantics

### LINKED ISSUE #784: bug(cron): numeric day-of-week is off by one — docs promise POSIX but cron crate uses Sun=1..Sat=7
## Description

When a cronjob uses **numeric day-of-week** values (e.g. `0 7 * * 1-5`), openab
fires on the wrong day. The docs promise POSIX semantics (Sun=0/7, Mon=1, ..., Sat=6),
so `1-5` should mean Mon–Fri. In practice it fires Sun–Thu — every numeric
day-of-week value is shifted by one day.

The timezone handling itself is correct. The bug is purely in how numeric
day-of-week values are interpreted.

## Steps to Reproduce

1. Configure a cronjob:
   ```toml
   [[jobs]]
   schedule = "0 7 * * 1-5"
   timezone = "Asia/Taipei"
   channel = "<redacted>"
   message = "weekday job"
   enabled = true
   ```
2. Wait for the next Sunday 07:00 in `Asia/Taipei` (= Saturday 23:00 UTC).
3. Observe: the job fires, even though Sunday is not in `1-5`.

Real trigger observed on 2026-05-10 (Sunday):

```
2026-05-09T23:00:01Z  INFO openab::cron: 🔔 cronjob fired
    schedule=0 7 * * 1-5  channel=<redacted>  platform=discord
    message=<redacted>  sender=openab-cron
```

UTC 23:00 on 2026-05-09 = Taipei 07:00 on 2026-05-10 (Sunday). Should not have fired.

## Expected Behavior

Per [`docs/cronjob.md`](../blob/main/docs/cronjob.md), which says:

> Standard 5-field POSIX cron, same as Linux crontab, K8s CronJob, and GitHub Actions
> `day of week (0-7, 0 and 7 = Sunday)`
> `| 0 9 * * 1-5 | Weekdays at 09:00 |`

`1-5` should fire Mon–Fri only.

## Actual Behavior

`1-5` fires Sun–Thu (entire weekday range shifted by one). Every numeric
day-of-week value is off by one:

| User writes (POSIX intent) | Actually fires on |
|---|---|
| `0` (Sun) | out-of-range — invalid |
| `1` (Mon) | Sun |
| `5` (Fri) | Thu |
| `6` (Sat) | Fri |
| `7` (Sun) | Sat |
| `1-5` (Mon–Fri) | Sun–Thu |

## Root Cause

openab uses the [`cron`](https://crates.io/crates/cron) crate (v0.16.0), which
does **not** follow POSIX day-of-week numbering. From
[`cron-0.16.0/src/time_unit/days_of_week.rs`](https://docs.rs/cron/0.16.0/src/cron/time_unit/days_of_week.rs.html):

```rust
static DAY_OF_WEEK_MAP: phf::Map<&'static str, Ordinal> = phf_map! {
    "sun" => 1,
    "mon" => 2,
    "tue" => 3,
    "wed" => 4,
    "thu" => 5,
    "fri" => 6,
    "sat" => 7,
};
// inclusive_min = 1, inclusive_max = 7
```

And in [`schedule.rs`](https://docs.rs/cron/0.16.0/src/cron/schedule.rs.html),
the weekday match uses chrono's `number_from_sunday()` (Sun=1, Mon=2, ..., Sat=7),
which matches the map above but **not POSIX**.

So `1-5` passed to the `cron` crate means Sun–Thu, not Mon–Fri.

**Note**: the timezone handling in `should_fire` is correct. `Utc::now().with_timezone(&tz)`
and the weekday check both use the same local `DateTime<Tz>`. This is not a timezone bug.

## Proposed Fix

Translate numeric day-of-week tokens in the 5th field before handing the
expression to `cron::Schedule`. This keeps `docs/cronjob.md`'s POSIX promise
intact and is a localized change to `parse_cron_expr`:

- `0` or `7` → `1` (Sunday in cron crate)
- `n` (1..=6) → `n + 1`
- Applies to each numeric token inside ranges (`1-5` → `2-6`), lists (`1,3,5` → `2,4,6`),
  and step bases (`*/2` unchanged; `1-5/2` → `2-6/2`).
- Name tokens (`Mon`, `Sun`, `Mon-Fri`, ...) pass through unchanged — the `cron` crate's
  name → ordinal map is already internally consistent.

Alternative considered: switching to a POSIX-compliant crate (e.g. `saffron`,
`croner-rust`). Larger surface area; can be re-evaluated if the translation
approach proves insufficient.

## Workaround (for existing users)

Use name-based day-of-week:

```toml
schedule = "0 7 * * Mon-Fri"
```

This works correctly because the `cron` crate's name→ordinal map is internally
consistent; only the numeric interface diverges from POSIX.

## Environment

- openab: 0.8.2
- cron crate: 0.16.0
- Platform: Discord adapter, via usercron (`~/.openab/usercron.toml`)

---

*Issue filed by my agent (OpenAB w/ Kiro CLI) under my account; content reviewed by me.*

charlie0228:

Numeric day-of-week values in cron expressions were being interpreted one day
off from the POSIX semantics the docs promise. A user-reported example:
`schedule = "0 7 * * 1-5"` with `timezone = "Asia/Taipei"` fired on Sunday
2026-05-10 at 07:00 Taipei — a Sunday that should have been excluded by
`1-5` (Mon-Fri). Every numeric day-of-week value is shifted by one day.

Root cause: the underlying `cron` crate (0.16.0) uses a non-POSIX
day-of-week numbering (`Sun=1, Mon=2, ..., Sat=7`), while
[`docs/cronjob.md`](../blob/main/docs/cronjob.md) and
[`docs/adr/basic-cronjob.md`](../blob/main/docs/adr/basic-cronjob.md) both
commit to POSIX semantics (`Sun=0/7, Mon=1, ..., Sat=6`).

Closes #784

Discord Discussion URL: https://discord.com/channels/1491295327620169908/1491365157010542652/1503051528129220799

## At a Glance

```
Before this PR
--------------
user input       openab               cron crate (0.16.0)
"0 7 * * 1-5" ─► parse_cron_expr ───► Schedule
                 (prepends "0 ")      parses "1-5" as
                                      days {1,2,3,4,5}
                                      = {Sun,Mon,Tue,Wed,Thu} ✗

After this PR
-------------
user input       openab                         cron crate
"0 7 * * 1-5" ─► parse_cron_expr                Schedule
                 ├─ translate_posix_cron_expr   parses "2-6" as
                 │  └─ translate_posix_dow_field   days {2,3,4,5,6}
                 │     "1-5" → "2-6"              = {Mon,Tue,Wed,Thu,Fri} ✓
                 └─ prepends "0 "
```

Not applicable in the full sense — this is a bug fix to existing scheduling
code, not a new design. No architectural decision is being made.

For completeness, the POSIX day-of-week convention (`0=Sun, 1=Mon, ..., 6=Sat,
7=Sun alias`) is standardized across:

- [Linux `crontab(5)`](https://man7.org/linux/man-pages/man5/crontab.5.html)
- [Kubernetes CronJob schedule](https://kubernetes.io/docs/concepts/workloads/controllers/cron-jobs/#cron-schedule-syntax)
- [GitHub Actions `schedule:`](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#schedule)
- [AWS EventBridge cron expressions](https://docs.aws.amazon.com/scheduler/latest/UserGuide/schedule-types.html) (week field 1-7=Sun-Sat with `#` modifier, slightly different but POSIX-aligned in base)

Other Rust cron crates that follow POSIX natively:

- [`saffron`](https://crates.io/crates/saffron) — uses POSIX day-of-week
- [`croner-rust`](https://crates.io/crates/croner) — uses POSIX day-of-week

openab's existing choice of the `zslayton/cron` crate is fine for every field
except day-of-week. This PR keeps the dependency and adapts the input.

## Proposed Solution

Translate the day-of-week field inside `parse_cron_expr` before handing the
expression to `cron::Schedule::from_str`:

1. Split the input into five POSIX fields. Other fields pass through untouched.
2. For the day-of-week field:
   - If it contains any ASCII letter (`Mon`, `Sun`, `Mon-Fri`), pass through —
     the cron crate's internal name-to-ordinal map is already consistent.
   - Otherwise, expand each comma-separated component (handling `*`, `a-b`,
     `a-b/s`, `*/s`, singletons) into a set of POSIX day values
     (`{0..=6}`), normalizing `7` to `0` to dedupe Sunday.
   - If the set covers all 7 days, emit `*`.
   - Otherwise emit the shifted values (`+1`) in cron-crate form, compacting
     contiguous runs into ranges for readability.

The function `parse_cron_expr` now returns `Result<Schedule, String>` so it
can surface both translation errors and underlying `cron::Error` parse errors
uniformly. All existing callers use `Display` on the error and remain
source-compatible.

Name-based day-of-week (`Mon-Fri`) continues to work unchanged. This means
the previously-required workaround (switch `1-5` to `Mon-Fri`) remains
available and safe, but is no longer necessary.

## Why this approach?

- **Honours the documented contract.** `docs/cronjob.md` and the ADR both
  explicitly specify POSIX. Users reasonably expect `1-5` to mean Mon-Fri.
- **Minimal surface.** The change is confined to `src/cron.rs` and adds no
  new dependencies.
- **Backward compatible.** Existing expressions that happened to work
  (e.g. all-`*` schedules, name-based day-of-week) continue to behave
  identically. Expressions that were silently wrong (numeric day-of-week)
  now behave as the docs promise.
- **No log noise.** Translation is silent on success. The original
  user-facing `job.schedule` string is still what gets logged; only the
  string passed to the parser is rewritten.

Tradeoffs / known limitations:

- **Mixed numeric-and-name notation** (`1,Mon`) is not supported and will
  pass through untranslated, giving the old (wrong) behaviour for the
  numeric part. This is documented in the function doc comment. Fixing
  this robustly would require a proper tokenizer, which seems excessive
  for a notation nobody writes in practice.
- **Output shape is deterministic but not always identical** to input
  shape. `1-5` becomes `2-6` (contiguous range preserved); `1,3,5` becomes
  `2,4,6` (list preserved); `*/2` becomes `1,3,5,7` (step expanded to a
  list). Semantically equivalent in every case, but if someone is debugging
  by staring at the exact string passed to the cron crate, it may differ.

## Alternatives Considered

- **Swap the `cron` crate for a POSIX-compliant one** (e.g. `saffron`,
  `croner-rust`). Rejected for this PR because the surface area is much
  larger (DST handling, `after` iterator semantics, serde derive, error
  types), and any behaviour drift would need its own investigation. The
  translation approach keeps the change localized and testable in
  isolation.
- **Document the quirk instead of fixing it.** Rejected because the failure
  mode is silent: the schedule just fires on the wrong day, with no
  validation error. Users will only discover it by observing misfires in
  production. The docs also already commit to POSIX.
- **Error out on numeric day-of-week and require name-based notation.**
  Rejected as a breaking change that would reject expressions the docs
  explicitly show as examples (`0 9 * * 1-5 | Weekdays at 09:00`).

## Validation

- [x] `rustfmt --edition 2021 src/cron.rs` — clean
- [ ] `cargo check` — not run locally (container lacks a C toolchain;
      relying on this PR's CI to validate)
- [ ] `cargo test` — not run locally for the same reason. 25 new tests
      were added; all have been hand-traced against the specification above.
- [ ] `cargo clippy` — not run locally; relying on CI

### New tests

Pure translator:
- `*`, `0`, `7`, `1`, `6` single values
- `1-5`, `3-5`, `2-6`, `0-6`, `0-7`, `1-7` ranges (including `*` collapse)
- `0,7`, `1,3,5`, `1,2,4,5` lists (incl. dedup and contiguous-run compaction)
- `*/2`, `1-5/2` steps
- Name-based `Mon-Fri`, `Mon,Wed,Fri`, `Sun` pass-through
- Error cases: `8`, `0-8`, `5-3`, empty field, `,1`, `1,`, `*/0`
- `parse_cron_expr` rejects wrong field count (already covered; reaffirmed)

Regression (against the observed #784 misfire):
- `0 7 * * 1-5` in `Asia/Taipei` must **not** match Sunday 2026-05-10 07:00
- `0 7 * * 1-5` in `Asia/Taipei` must match Monday 2026-05-11 07:00
- Friday matches, Saturday does not
- `0 7 * * 0` matches Sunday (via POSIX `0`)
- `0 7 * * 7` matches Sunday (via POSIX `7` alias)
- `0 7 * * 6` matches Saturday
- `0 7 * * Mon-Fri` still behaves correctly post-translation

### Manual testing

None — production validation is the original misfire already reported in
issue #784. The regression test encodes that exact scenario.

---

*PR opened by my agent (OpenAB w/ Kiro CLI) under my GitHub account; commit and PR
body reviewed by me before submission.*

---

## ⚠️ Breaking Change Notice

This is a **correctness fix** that changes observable behavior for any schedule using numeric day-of-week values:

| Before (broken) | After (correct) | Impact |
|---|---|---|
| `1-5` fired Sun-Thu | `1-5` fires Mon-Fri | Schedules shift to correct POSIX days |
| `0` fired on Saturday | `0` fires on Sunday | Sunday schedules now work correctly |
| `1,Mon` silently passed through | `1,Mon` returns error | Mixed notation now rejected |

### Migration Guide

- **If you used correct POSIX notation** (e.g. `1-5` intending Mon-Fri): ✅ No action needed — it now works as intended.
- **If you used a workaround** (e.g. `2-6` to get Mon-Fri because `1-5` was broken): ⚠️ Revert to the correct POSIX value (`1-5`), otherwise your schedule will shift by one day.
- **If you used name-based notation** (`Mon-Fri`): ✅ No change in behavior.
- **If you used mixed notation** (`1,Mon`): ⚠️ This now returns an error. Use either all numeric or all name-based.

### DIFF
--- a/docs/cronjob.md
+++ b/docs/cronjob.md
@@ -300,6 +300,15 @@ Config-driven cron covers the 80% use case: "send this message at this time." Fo
 
 See [Kubernetes CronJob Reference Architecture](cronjob_k8s_refarch.md) for the external scheduler approach.
 
+## Known Limitations
+
+| Limitation | Details |
+|---|---|
+| Mixed numeric/name day-of-week | `1,Mon` or `Mon,3` is not supported and will be rejected. Use either all numeric (`1-5`) or all name-based (`Mon-Fri`) notation. |
+| Wrap-around day-of-week ranges | `5-2` (Fri through Tue) is not supported. Use explicit listing instead: `5,6,0,1,2`. |
+
+> **Tip:** Name-based notation (`Mon-Fri`, `Sun`, `Mon,Wed,Fri`) is always available as an alternative to numeric day-of-week values.
+
 ## Troubleshooting
 
 | Symptom | Cause | Fix |
--- a/src/cron.rs
+++ b/src/cron.rs
@@ -91,7 +91,17 @@ fn translate_posix_dow_field(field: &str) -> Result<String, String> {
     use std::collections::BTreeSet;
 
     // Name-based notation is internally consistent in the cron crate — pass through.
-    if field.chars().any(|c| c.is_ascii_alphabetic()) {
+    // But reject mixed numeric+name notation (e.g. "1,Mon") which would leave the
+    // numeric part untranslated and silently wrong.
+    let has_alpha = field.chars().any(|c| c.is_ascii_alphabetic());
+    let has_digit = field.chars().any(|c| c.is_ascii_digit());
+    if has_alpha && has_digit {
+        return Err(format!(
+            "mixed numeric and name notation is not supported in day-of-week field: {:?}",
+            field
+        ));
+    }
+    if has_alpha {
         return Ok(field.to_string());
     }
 
@@ -148,7 +158,12 @@ fn translate_posix_dow_field(field: &str) -> Result<String, String> {
             if n > 7 {
                 return Err(format!("day-of-week value out of range (0-7): {}", n));
             }
-            vec![n]
+            if step > 1 {
+                // n/step means "from n through end-of-domain, stepping by step"
+                (n..=6).collect()
+            } else {
+                vec![n]
+            }
         };
 
         // Apply step filter, normalize 7 → 0, collect into the set.
@@ -714,6 +729,25 @@ mod tests {
             "Mon,Wed,Fri"
         );
         assert_eq!(translate_posix_dow_field("Sun").unwrap(), "Sun");
+    }
+
+    #[test]
+    fn translate_dow_step_from_singleton() {
+        // POSIX 1/2 = from Mon through Sat, step 2 = {1,3,5} = Mon,Wed,Fri -> cron crate 2,4,6
+        assert_eq!(translate_posix_dow_field("1/2").unwrap(), "2,4,6");
+    }
+
+    #[test]
+    fn translate_dow_step_from_singleton_sunday() {
+        // POSIX 0/3 = from Sun through Sat, step 3 = {0,3,6} = Sun,Wed,Sat -> cron crate 1,4,7
+        assert_eq!(translate_posix_dow_field("0/3").unwrap(), "1,4,7");
+    }
+
+    #[test]
+    fn translate_dow_rejects_mixed_notation() {
+        assert!(translate_posix_dow_field("1,Mon").is_err());
+        assert!(translate_posix_dow_field("Mon,1").is_err());
+        assert!(translate_posix_dow_field("1-Fri").is_err());
     }
 
     #[test]
--- a/src/cron.rs
+++ b/src/cron.rs
@@ -85,8 +85,8 @@ fn translate_posix_cron_expr(expr: &str) -> Result<String, String> {
 /// # Mixed numeric and name notation
 ///
 /// Mixing numeric and name tokens in the same field (e.g. `1,Mon`) is not
-/// supported and will be passed through untranslated. Use either all numeric
-/// (POSIX) or all name-based notation.
+/// supported and will return an error. Use either all numeric (POSIX) or all
+/// name-based notation.
 fn translate_posix_dow_field(field: &str) -> Result<String, String> {
     use std::collections::BTreeSet;
 
@@ -160,7 +160,9 @@ fn translate_posix_dow_field(field: &str) -> Result<String, String> {
             }
             if step > 1 {
                 // n/step means "from n through end-of-domain, stepping by step"
-                (n..=6).collect()
+                // Normalize 7 (Sunday alias) to 0 before expansion.
+                let start = if n == 7 { 0 } else { n };
+                (start..=6).collect()
             } else {
                 vec![n]
             }
@@ -741,6 +743,12 @@ mod tests {
     fn translate_dow_step_from_singleton_sunday() {
         // POSIX 0/3 = from Sun through Sat, step 3 = {0,3,6} = Sun,Wed,Sat -> cron crate 1,4,7
         assert_eq!(translate_posix_dow_field("0/3").unwrap(), "1,4,7");
+    }
+
+    #[test]
+    fn translate_dow_step_from_singleton_seven() {
+        // POSIX 7/2 = Sunday alias, same as 0/2 = {0,2,4,6} = Sun,Tue,Thu,Sat -> cron crate 1,3,5,7
+        assert_eq!(translate_posix_dow_field("7/2").unwrap(), "1,3,5,7");
     }
 
     #[test]
--- a/src/cron.rs
+++ b/src/cron.rs
@@ -13,10 +13,195 @@ use tokio::sync::Mutex;
 use tracing::{debug, error, info, warn};
 
 /// Parse a 5-field POSIX cron expression into a `Schedule`.
+///
 /// The `cron` crate expects a 6-field expression (with seconds), so we prepend "0".
-pub fn parse_cron_expr(expr: &str) -> Result<Schedule, cron::error::Error> {
-    let six_field = format!("0 {}", expr);
-    Schedule::from_str(&six_field)
+///
+/// POSIX numeric day-of-week values (0..=7, where 0 or 7 = Sunday) are translated
+/// to the `cron` crate's 1-based form (1..=7, where 1 = Sunday) before being handed
+/// to the underlying parser. Without this, numeric day-of-week values are off by one
+/// — e.g. `1-5` (Mon-Fri in POSIX) would be evaluated as Sun-Thu. See the
+/// [`translate_posix_dow_field`] doc comment for details.
+///
+/// Name-based day-of-week tokens (`Mon`, `Sun`, `Mon-Fri`, ...) are passed through
+/// unchanged — the `cron` crate's internal name-to-ordinal map is consistent.
+pub fn parse_cron_expr(expr: &str) -> Result<Schedule, String> {
+    let translated = translate_posix_cron_expr(expr)?;
+    let six_field = format!("0 {}", translated);
+    Schedule::from_str(&six_field).map_err(|e| e.to_string())
+}
+
+/// Translate a 5-field POSIX cron expression so the day-of-week field uses the
+/// numeric convention of the `cron` crate.
+///
+/// Only the 5th field (day-of-week) is rewritten; the other four fields pass
+/// through unchanged.
+fn translate_posix_cron_expr(expr: &str) -> Result<String, String> {
+    let fields: Vec<&str> = expr.split_whitespace().collect();
+    if fields.len() != 5 {
+        return Err(format!(
+            "expected 5 whitespace-separated cron fields, got {}: {:?}",
+            fields.len(),
+            expr
+        ));
+    }
+    let translated_dow = translate_posix_dow_field(fields[4])?;
+    Ok(format!(
+        "{} {} {} {} {}",
+        fields[0], fields[1], fields[2], fields[3], translated_dow
+    ))
+}
+
+/// Translate a POSIX day-of-week field to the `cron` crate's numeric form.
+///
+/// # Background
+///
+/// POSIX cron (and Linux crontab, Kubernetes CronJob, GitHub Actions) uses
+/// `0..=7` where `0` or `7` = Sunday, `1` = Monday, ..., `6` = Saturday.
+///
+/// The `cron` crate uses `1..=7` where `1` = Sunday, `2` = Monday, ..., `7` = Saturday
+/// (it matches via chrono's `Weekday::number_from_sunday()`). Without translation,
+/// every numeric day-of-week value fires one day early:
+///
+/// | POSIX intent  | Without translation (cron crate reads as) |
+/// |---------------|-------------------------------------------|
+/// | `0`, `7` (Sun) | out-of-range / Sat                       |
+/// | `1` (Mon)     | Sun                                        |
+/// | `5` (Fri)     | Thu                                        |
+/// | `1-5` (Mon-Fri) | Sun-Thu                                  |
+///
+/// # Algorithm
+///
+/// 1. If the field contains any ASCII letter (e.g. `Mon-Fri`), pass it through —
+///    the cron crate's name-to-ordinal map is internally consistent.
+/// 2. Otherwise, expand each comma-separated component into the set of POSIX
+///    day values it represents. Ranges (`a-b`) and step values (`a/s`, `a-b/s`,
+///    `*/s`) are expanded here. `7` is normalized to `0` (both = Sunday) to
+///    avoid duplication.
+/// 3. If the resulting set covers all 7 days, emit `*` for brevity.
+/// 4. Otherwise, shift each value by `+1` (POSIX `{0..=6}` → cron crate
+///    `{1..=7}`) and emit as a comma-separated list, compacting contiguous
+///    runs into ranges for readability.
+///
+/// # Mixed numeric and name notation
+///
+/// Mixing numeric and name tokens in the same field (e.g. `1,Mon`) is not
+/// supported and will be passed through untranslated. Use either all numeric
+/// (POSIX) or all name-based notation.
+fn translate_posix_dow_field(field: &str) -> Result<String, String> {
+    use std::collections::BTreeSet;
+
+    // Name-based notation is internally consistent in the cron crate — pass through.
+    if field.chars().any(|c| c.is_ascii_alphabetic()) {
+        return Ok(field.to_string());
+    }
+
+    if field.is_empty() {
+        return Err("empty day-of-week field".to_string());
+    }
+
+    let mut days: BTreeSet<u32> = BTreeSet::new();
+
+    for part in field.split(',') {
+        if part.is_empty() {
+            return Err(format!("empty component in day-of-week field: {:?}", field));
+        }
+
+        // Split off optional step: `a/s`, `a-b/s`, `*/s`.
+        let (range_part, step) = match part.split_once('/') {
+            Some((r, s)) => {
+                let step_n: u32 = s
+                    .parse()
+                    .map_err(|_| format!("invalid step value in {:?}", part))?;
+                if step_n == 0 {
+                    return Err(format!("step value cannot be zero in {:?}", part));
+                }
+                (r, step_n)
+            }
+            None => (part, 1u32),
+        };
+
+        // Expand range_part to the list of POSIX day values it represents.
+        // Values may include 7 (Sunday alias for 0); normalization happens below.
+        let raw_values: Vec<u32> = if range_part == "*" {
+            (0..=6).collect()
+        } else if let Some((a, b)) = range_part.split_once('-') {
+            let a_n: u32 = a
+                .parse()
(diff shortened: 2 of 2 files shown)


shaun-agent:
## OpenAB PR Screening
- Title: fix(cron): translate POSIX day-of-week to match documented semantics
- Source: https://github.com/openabdev/openab/pull/785
- Generated at: 2026-05-10T16:01:28.986Z
- Discord thread: https://discord.com/channels/1488041051187974246/1503064283133513828
<details>
## Intent

PR #785 fixes OpenAB cron scheduling so numeric day-of-week values match the documented POSIX contract.

The concrete user-visible bug is that schedules such as `0 7 * * 1-5`, documented as Monday through Friday, can currently fire on Sunday through Thursday because the underlying Rust `cron` crate interprets numeric weekdays as `Sun=1, Mon=2, ..., Sat=7`. This caused a reported misfire on Sunday, May 10, 2026 in `Asia/Taipei`.

## Feat

This is a bug fix.

The PR changes cron parsing so OpenAB translates numeric POSIX day-of-week values before passing expressions to the `cron` crate. Name-based weekday expressions such as `Mon-Fri` are left unchanged. The intended behavior is that OpenAB users can keep writing schedules according to the documented POSIX semantics: `0/7=Sun, 1=Mon, ..., 6=Sat`.

## Who It Serves

Primary beneficiaries are deployers and operators who rely on OpenAB scheduled jobs.

It also serves maintainers by aligning implementation with existing docs and reducing production misfires caused by silently shifted weekday schedules.

## Rewritten Prompt

Fix OpenAB cron parsing so numeric day-of-week fields follow the POSIX semantics already documented in `docs/cronjob.md` and `docs/adr/basic-cronjob.md`.

In `src/cron.rs`, adapt five-field user cron expressions before constructing `cron::Schedule`. Translate only numeric day-of-week syntax from POSIX values into the numbering expected by the current `cron` crate. Preserve name-based weekday syntax such as `Mon-Fri`. Reject invalid numeric weekday syntax with clear parse errors.

Add focused tests covering single values, ranges, lists, steps, Sunday aliases `0` and `7`, invalid values, name-based pass-through, and the regression case where `0 7 * * 1-5` in `Asia/Taipei` must not match Sunday, May 10, 2026 but must match Monday, May 11, 2026.

Keep the change localized unless testing shows the current parser integration needs a broader scheduling abstraction.

## Merge Pitch

This is worth advancing because it fixes a silent correctness bug in scheduled job execution. The current behavior can run production jobs on the wrong day while appearing valid to users and matching documented examples.

Risk profile is moderate-low if the translation is well tested: the change is localized to cron parsing, does not add dependencies, and preserves documented behavior. The main reviewer concern should be parser correctness, especially edge cases around ranges, steps, Sunday aliases, invalid syntax, and mixed numeric/name weekday expressions.

## Best-Practice Comparison

OpenClaw’s most relevant principles here are explicit scheduling semantics, reliable run behavior, and clear delivery of operator expectations. This PR does not address gateway-owned scheduling, durable job persistence, isolated execution, retry/backoff, or run logs directly. Those are broader scheduler architecture concerns outside the immediate bug.

Hermes Agent’s relevant principles are self-contained scheduled task semantics and avoiding overlapping or inconsistent scheduler state. File locking, atomic persisted state, fresh sessions per run, and daemon tick modeling are not directly implicated by this parser bug.

The strongest alignment is with both systems’ emphasis on predictable scheduled execution. A scheduler cannot be reliable if accepted schedules mean something different from the docs.

## Implementation Options

Option 1: Conservative parser adapter  
Keep the current `cron` crate and add a small POSIX day-of-week translator inside `parse_cron_expr`. Cover expected numeric syntax and leave name-based syntax untouched. This appears to be the PR’s current direction.

Option 2: Balanced scheduling wrapper  
Keep the current crate, but formalize OpenAB’s cron contract behind a dedicated parser/normalizer API with explicit tests and error types. This would make future scheduler behavior easier to audit without changing the runtime scheduling engine.

Option 3: Ambitious cron dependency migration  
Replace the current `cron` crate with a POSIX-compatible cron library such as `saffron` or `croner-rust`. This removes the need for weekday translation but requires careful review of DST behavior, iterator semantics, error handling, serde support, and any downstream scheduling differences.

Option 4: Validation-only workaround  
Reject numeric day-of-week values and require name-based weekdays. This avoids incorrect translation risk but breaks documented examples and pushes avoidable friction onto users.

## Comparison Table

| Conservative parser adapter | High | Low-Medium | Good if well tested | Acceptable | Fixes bug without user changes | Strong |
| Balanced scheduling wrapper | Medium | Medium | Strong | Strong | Same fix, cleaner future surface | Good |
| Dependency migration | Low | High | Potentially strong after review | Good if migration succeeds | Possible behavior drift | Premature |
| Validation-only workaround | High | Low | Prevents some bad schedules | Weak | Breaking/friction-heavy | Poor |

## Recommendation

Advance the conservative parser adapter, with reviewer focus on the translation tests and documented edge cases.

This is the right next step because it fixes the production-visible bug while preserving OpenAB’s documented cron contract and avoiding a larger scheduler dependency migration. For merge discussion, require CI validation and scrutinize mixed numeric/name weekday handling. If that limitation remains, document it clearly or split a follow-up issue for a fuller weekday tokenizer.

A later follow-up can consider extracting cron normalization behind a small scheduling contract module, but that should not block this bug fix.
</details>

chaodu-agent:
CHANGES REQUESTED ⚠️ — Two silent-failure bugs fixed in commit ae20770; awaiting CI green.

## What This PR Does
Translates POSIX numeric day-of-week values (0=Sun, 1=Mon, ..., 6=Sat, 7=Sun) to the `cron` crate's 1-based form before parsing, fixing an off-by-one bug where `1-5` (Mon-Fri) was interpreted as Sun-Thu. Closes #784.

## How It Works
A `translate_posix_dow_field()` function rewrites the 5th cron field: expands ranges/steps/lists into a POSIX day set, normalizes Sunday aliases (0/7), shifts +1 for the crate, and compacts back into ranges. Name-based tokens (`Mon-Fri`) pass through unchanged.

## Findings

| 1 | 🟡 | `n/step` singleton expansion bug — `1/2` produced only Monday instead of Mon,Wed,Fri | `src/cron.rs:144-151` |
| 2 | 🟡 | Mixed notation (`1,Mon`) silently passed through with incorrect numeric translation | `src/cron.rs:94-96` |
| 3 | 🟢 | Core translation logic correct, comprehensive test coverage (25+ cases) | — |
| 4 | 🟢 | Wrap-around range rejection (`5-2`) is defensible and safe | — |
| 5 | 🟢 | Return type change source-compatible with all internal callers | — |

<details>

### 🟡 F1: `n/step` singleton expansion bug (FIXED in ae20770)
The singleton branch produced `vec![n]` regardless of step value. For `1/2`, this meant only `{1}` → translated to `"2"` (Monday only). POSIX `crontab(5)` defines `n/step` as "from n through end-of-domain, stepping by step", so `1/2` should yield `{1,3,5}` (Mon,Wed,Fri).

**Fix:** When step > 1 and range_part is a singleton, expand to `(n..=6).collect()` before applying the step filter.

### 🟡 F2: Mixed notation silent pass-through (FIXED in ae20770)
Any field containing an ASCII letter triggered a fast-path `return Ok(field.to_string())`. This meant `1,Mon` would pass through with the numeric `1` untranslated (interpreted as Sunday by the crate instead of Monday).

**Fix:** Detect mixed numeric+name notation (both digits and letters present) and return `Err` with a clear message. Pure name-based notation still passes through correctly.

### 🟢 F3-F5: What's done well
- Translation logic handles all POSIX edge cases: `0`/`7` Sunday dedup, ranges, steps, lists, `*`
- Regression tests encode the exact #784 misfire scenario (Sunday 2026-05-10 in Asia/Taipei)
- `compact_ordinal_set` produces readable output (`2-6` not `2,3,4,5,6`)
- PR body is exemplary: alternatives considered, tradeoffs documented, prior art cited

</details>

<details>

- PR opened: 2026-05-10
- Main already has: `parse_cron_expr` that prepends seconds and delegates to `cron::Schedule`
- Net-new value: POSIX DOW translation layer + 25 new tests + regression coverage for #784

</details>

<details>
<summary>Reviewer Credits</summary>

- 覺渡法師: Mixed notation silent failure, Error type suggestion
- 擺渡法師: `n/step` singleton expansion bug, mixed notation confirmation
- 普渡法師: `n/step` singleton expansion bug (independent confirmation), CI concern

</details>

---

**Status:** Both 🟡 findings fixed in commit ae20770. Awaiting CI to confirm all tests pass.

chaodu-agent:
LGTM ✅ — All findings addressed. Ready for merge.

## What This PR Does
Translates POSIX numeric day-of-week values (0=Sun, 1=Mon, ..., 6=Sat, 7=Sun) to the `cron` crate's 1-based form before parsing, fixing an off-by-one bug where `1-5` (Mon-Fri) was interpreted as Sun-Thu. Closes #784.

## How It Works
A `translate_posix_dow_field()` function rewrites the 5th cron field: expands ranges/steps/lists into a POSIX day set, normalizes Sunday aliases (0/7), shifts +1 for the crate, and compacts back into ranges. Name-based tokens (`Mon-Fri`) pass through unchanged. Mixed notation (`1,Mon`) is explicitly rejected.

## Findings

| 1 | 🟡→✅ | `n/step` singleton expansion bug — `1/2` produced only Monday | Fixed in ae20770 |
| 2 | 🟡→✅ | Mixed notation silent pass-through | Fixed in ae20770 |
| 3 | 🟡→✅ | `7/step` edge case — empty range error | Fixed in 874cbe8 |
| 4 | 🟡→✅ | Stale doc comment (mixed notation) | Fixed in 874cbe8 |
| 5 | 🟢 | Core translation logic correct, comprehensive test coverage | — |
| 6 | 🟢 | Wrap-around range rejection defensible and safe | — |
| 7 | 🟢 | Return type change source-compatible with all callers | — |

<details>
<summary>Fix Summary</summary>

### Commit ae20770
- `n/step` (e.g. `1/2`) now expands `(n..=6)` before step filter → `1/2` = Mon,Wed,Fri ✓
- Mixed notation (`1,Mon`) returns `Err` instead of silent pass-through ✓

### Commit 874cbe8
- `7/step` normalizes `7→0` before expansion → `7/2` = Sun,Tue,Thu,Sat ✓
- Doc comment updated to reflect fail-closed behavior ✓
- Added `7/2` test case ✓

</details>

<details>
<summary>Reviewer Credits</summary>

| Reviewer | Findings | Final Verdict |
|----------|----------|---------------|
| 覺渡法師 (Gemini) | Mixed notation silent failure, Error type suggestion | LGTM ✅ |
| 擺渡法師 (Codex) | `n/step` singleton bug, `7/step` edge case, stale doc | LGTM ✅ |
| 普渡法師 (Claude) | `n/step` singleton bug (independent), CI concern | LGTM ✅ |

</details>

<details>

- PR opened: 2026-05-10
- Main already has: `parse_cron_expr` that prepends seconds and delegates to `cron::Schedule`
- Net-new value: POSIX DOW translation layer + 28 tests + regression coverage for #784

</details>

---

All 🟡 findings resolved. CI check green. Recommend merge.

chaodu-agent (dismissed):
All findings addressed. LGTM ✅

thepagent (dismissed):

PR #793, opened 2026-05-11
fix(media): validate Content-Type and magic bytes before sending to model

### LINKED ISSUE #776: bug(slack): unauthenticated file fetch returns HTML, gets forwarded to model as image, poisons session
## Description

When the Slack handler fetches a file URL from `files.slack.com`, it does not validate the response content-type or magic bytes before forwarding to the model API. We observed an `image/png`-labeled payload sent to Anthropic whose decoded bytes were Slack workspace login HTML (~55KB), consistent with the bot token lacking the `files:read` OAuth scope so Slack served the login page in place of the file binary. openab base64-encoded those HTML bytes, labeled them with the Slack-reported MIME, and forwarded to Anthropic, which rejected with `400 invalid_request_error "Could not process image"`.

Two failure modes:

1. **Confusing UX**: the user sees an Anthropic 400 that looks like an Anthropic-side image format problem, not a Slack auth/scope problem.
2. **Session poisoning**: the bad payload is persisted into the claude-agent-acp session JSONL on PVC. Subsequent messages in the same Slack thread resume the session and replay the bad image block as part of conversation history, so Anthropic re-rejects on every turn until the JSONL is manually deleted.

## Reproduction

1. Configure a Slack app for openab without `files:read` scope. Verify with:

   ```bash
   kubectl exec deployment/openab-claude -- sh -c \
     'curl -sS -D - -o /dev/null -X POST https://slack.com/api/auth.test \
        -H "Authorization: Bearer $SLACK_BOT_TOKEN" | grep -i "^x-oauth-scopes:"'
   ```

   The response header should not contain a comma-separated `files:read` token.

2. As a Slack user, upload any image to the bot in a thread.

3. Bot reply:

   ```
   :warning: Internal Error (code: -32603)
   Internal error: API Error: 400 {"type":"error","error":{"type":"invalid_request_error","message":"Could not process image"},"request_id":"req_011..."}
   ```

4. Add `files:read` scope to the Slack app, reinstall, rotate the token, redeploy, and verify the new scope is in `x-oauth-scopes` from step 1. Now upload another image to the **same Slack thread**: same 400.

5. Recover by deleting the session JSONL inside the pod:

   ```bash
   kubectl exec deployment/openab-claude -- bash -c \
     'grep -lE "Could not process image" /home/node/.claude/projects/-home-node/*.jsonl | xargs rm -v'
   ```

## Expected Behavior

When openab fetches a file URL from Slack and the response is not a valid image, it should:

1. **Detect the failure synchronously**, by inspecting the HTTP `Content-Type` header and/or the response body's magic bytes, before base64-encoding for the model API.
2. **Fail fast with a user-actionable error**, e.g. "I couldn't access that image — make sure the bot has the `files:read` OAuth scope." Don't forward unverified bytes to the model.
3. **Not persist the unverified payload to claude session history**, so that recovering from misconfiguration is "fix the config" without an out-of-band PVC cleanup.

End-to-end success criterion: after the operator adds `files:read` and rotates the token, uploading an image in the same Slack thread succeeds on the next message, with no manual session deletion.

## Actual Behavior

1. openab fetches `files.slack.com/...` with the bot token. We did not capture the raw HTTP response in our logs; based on the bytes that ended up in the conversation block we infer Slack returned the workspace login HTML page in place of the file binary.
2. openab does not visibly validate Content-Type or magic bytes. It reads the body, base64-encodes it, and labels it with the MIME from the Slack file metadata (`image/png` in our case).
3. openab forwards to Anthropic; Anthropic decodes, sees magic `3c 21 44 4f 43 54 59 50` (`<!DOCTYP`) instead of PNG `89 50 4e 47 0d 0a 1a 0a`, and returns 400.
4. claude-agent-acp persists the user message (containing the bad image block) into `~/.claude/projects/-home-node/<session>.jsonl` regardless of API success.
5. Subsequent messages in the same Slack thread resume that session and resend the bad image block as part of history. Anthropic 400s again.

### A. The forwarded payload was HTML, not PNG

Decoded base64 of an `image/png` payload sent to Anthropic from a poisoned session in our cluster (Anthropic request_id `req_011CarWAQQ11f3C1jzLbPkgP`):

| field | value |
|------|-------|
| claimed media_type | `image/png` |
| base64 length | 74,124 chars |
| decoded bytes | 55,592 |
| magic bytes (first 8) | `3c 21 44 4f 43 54 59 50` (`<!DOCTYP`) |
| last 120 bytes | `...slack-www-hhvm-main-iad-ya7nx4dnuum1/ 2026-05-08 21:55:25/...</body></html>` |

PNG magic should be `89 50 4e 47 0d 0a 1a 0a`. The bytes are HTML; the last-byte tail matches Slack's web-page footer. We did not capture the raw fetch response, so the "Slack returned HTML" link is inferred — sanitized fetch logs from anyone who can reproduce in their cluster would tighten this.

### B. The payload was replayed from session history on subsequent turns

Same session ID `774ec817-7ace-4e47-81a9-c875962ef720` (claude-agent-acp session JSONL on PVC):

- Line 6 (06:14 UTC, original incident): `user` message with an `image` block whose base64 decodes to the HTML above. Anthropic 400 (request_id `req_011CarWAQQ11f3C1jzLbPkgP`).
- Line 54 (~10 hours later, 08:04 UTC): user message in the same Slack thread containing **only text** ("你可看到上面的圖片了嗎?") and `sender_context` — no image block. Anthropic 400 again (request_id `req_011Carea29tJDW...`).
- Line 58 (08:06 UTC): same pattern, text-only user message ("can you read the image ?"). Anthropic 400 (request_id `req_011CarefyzaStqh7dZPAd3j5`).
- Line 60 immediately after: synthetic assistant message recording the 400. `isApiErrorMessage: true`, `apiErrorStatus: 400`, model `<synthetic>`.

i.e. text-only user turns failed with the same `Could not process image` error, which is consistent with the bad image block at line 6 being replayed from JSONL on every turn. After we deleted that JSONL on the PVC, a fresh thread/session in the same channel was unblocked.

## Suggested Fix

Two layers of defense, both worth doing:

### 1. Validate before forwarding (primary)

After fetching from Slack, before base64-encoding for the model API:

- Check the HTTP `Content-Type` header from the Slack response. Require the type to be in the model's supported set (for Anthropic Vision: `image/png`, `image/jpeg`, `image/gif`, `image/webp`). Generic `image/*` is too broad — `image/svg+xml`, `image/heic`, `image/avif`, `image/tiff` will be rejected by the model anyway.
- Check the response body's magic bytes. Whitelist the full signatures, not just the first nibble:
  - PNG: `89 50 4e 47 0d 0a 1a 0a`
  - JPEG: `ff d8 ff` (followed by `e0`/`e1`/`e2`/...)
  - GIF: `47 49 46 38 37 61` or `47 49 46 38 39 61`
  - WebP: `52 49 46 46 ?? ?? ?? ?? 57 45 42 50`
- If either check fails, surface a specific error to the user (e.g. "I couldn't access that image — does the bot have the `files:read` OAuth scope?") and skip the model call.

### 2. Don't persist unverified bytes to session history (secondary)

Persist the user-turn image block to the claude session JSONL **only after** the model call returns 200, not before. This way a 4xx from the model leaves the session in the same shape it was before the failed turn, and the operator's normal "fix config and try again" path works without manual JSONL cleanup.

This second defense matters even if (1) is in place: a real-but-corrupted upload can pass magic-byte check and still be rejected by the model, and once the bad block is in history the thread continues failing.

## Severity / Impact

- **Severity**: high for any operator who hits the misconfiguration. The user-facing error blames Anthropic; root cause is a Slack scope; once a thread is poisoned, every reply in it is broken; cluster operator intervention (pod exec, file deletion) is required to recover.
- **Surface**: any openab Slack ingest path where the bot token may not have `files:read` (which is not enforced or warned about by the chart's required-scopes section in the `helm install` NOTES — chart 0.8.2 NOTES does list `files:read` as required, but operators commonly miss it because pod startup succeeds and Slack returns 200 instead of 401 on the file fetch).
- **Discord ingest**: not tested — Discord has different auth/file URL semantics, but the same "fetched bytes forwarded without validation" code path may apply if shared.

## Regression / version scope

- Reproduced on chart 0.8.2 with image `ghcr.io/openabdev/openab-claude:latest` (digest `sha256:3d2017efbf1ab9702a9e3e7eaccb50e1147488cad04517ba50f29f63fd07d0ab` at the time of repro).
- Earlier versions not tested. Whether this is a recent regression or always present is unknown to us; if maintainers know when image ingest was introduced, that bounds the regression range.

## Workaround state

Operators can:

1. Add `files:read` to all bots' Slack apps and reinstall / rotate tokens (prevents new failures).
2. Delete poisoned session JSONLs on the PVC (recovers existing broken threads). For us, `grep -lE "Could not process image" /home/node/.claude/projects/-home-node/*.jsonl | xargs rm -v` was sufficient — we did not need to touch `thread_map.json` or `.claude/sessions/`.

We have documented the operator-side runbook at https://github.com/heyu-ai/openab-workspace/blob/main/docs/runbook.md (search "Slack 整合問題"). It covers diagnosis, remediation, and the PVC-cleanup recovery, but it is operator workaround, not an upstream fix.

## Environment

- openab chart 0.8.2
- Image digests at time of repro:
  - claude: `ghcr.io/openabdev/openab-claude@sha256:3d2017efbf1ab9702a9e3e7eaccb50e1147488cad04517ba50f29f63fd07d0ab`
  - codex: `ghcr.io/openabdev/openab-codex@sha256:4cc5f6fcf6983f57cdcd29560f96ab0d4b144c4fd4dc96d65a04f229f865a2f0`
  - gemini: `ghcr.io/openabdev/openab-gemini@sha256:f157b30aecfba8fac873a04087e836ba45b5ac164ca09f58ac9959c716cb8bd8`
- Three agents enabled (claude / codex / gemini); all three Slack apps had identical OAuth scope sets missing `files:read` before remediation
- Cluster: OrbStack-built-in K8s, single-tenant dev environment

howie:
Discord Discussion URL: https://discord.com/channels/1491295327620169908/1491969620754567270/1503586535088590868

Fixes #776.

## Root cause

When a Slack bot token lacks the `files:read` OAuth scope, Slack serves the workspace login HTML page (~55 KB) at HTTP 200 with `Content-Type: text/html` instead of the requested file binary. `download_and_encode_image` accepted this response because:

1. It never inspected the HTTP response `Content-Type` header.
2. On `resize_and_compress` failure for a body <= 1 MB it fell back to forwarding raw bytes under the Slack-reported MIME (`image/png`), bypassing any format check.

The result: a `ContentBlock::Image { media_type: "image/png", data: <base64 of HTML> }` flowed to Anthropic, which 400'd with `Could not process image`. Because claude-agent-acp persists the user message into the session JSONL _before_ the API reply, the bad block replayed on every subsequent turn until an operator manually deleted the JSONL inside the pod.

## Changes

**`src/media.rs`** (primary change)

- Add `MediaFetchError` enum: `NotAnImage` (silent skip), `UnsupportedResponseType`, `InvalidImageBody`, `SizeExceeded`, `Network`, `HttpStatus`.
- Add `validate_image_response(content_type, body)` pure helper that:
  - Rejects any response Content-Type not in `{image/png, image/jpeg, image/gif, image/webp}` (strips params, case-insensitive).
  - Sniffs magic bytes via `image::ImageReader::with_guessed_format()` (zero new dependencies) and rejects anything that doesn't decode as one of the four supported formats.
- Change `download_and_encode_image` signature from `-> Option<ContentBlock>` to `-> Result<ContentBlock, MediaFetchError>`, capturing the `Content-Type` header before consuming the response with `.bytes()`.
- Remove the <= 1 MB resize-error fallback (the direct bug path).

**`src/slack.rs`** (call site)

On validation failure, collect filenames and post one aggregated user-facing warning after the file loop:

> ":warning: I couldn't access the file(s) you shared (`photo.png`). This often means the bot is missing the `files:read` OAuth scope. Please ask an admin to reinstall the app with that scope."

All error variants (including `Network`, `HttpStatus`) surface to the user via warning message in both Slack and Discord.

**`src/discord.rs`** (call site)

Same `Result` pattern with user-facing warnings on failure (without the Slack-specific `files:read` scope hint, since Discord URLs are signed-public). Preserves the existing `is_video_file` fallback for `Err(NotAnImage)`.

## Tests

13 new unit tests in `src/media.rs::tests` for `validate_image_response`, including the exact bug reproduction and a corrupt-GIF regression:

```
validate_rejects_html_body_labeled_as_image_png
  body: b"<!DOCTYPE html>..."
  content_type: Some("image/png")
  expected: Err(InvalidImageBody { magic_prefix_hex: "3c21444f43545950" })
```

Smoke validation:

```
RUSTFLAGS='-C linker=/tmp/zigcc-wrapper' CC=/tmp/zigcc-wrapper AR=/tmp/zigar-wrapper cargo test --locked media::tests
```

Result: 24 media tests passed.

## Manual test plan (post-deploy)

1. Install with a bot token missing `files:read`. Confirm via `x-oauth-scopes` from `auth.test`.
2. Upload an image to the bot in a Slack thread.
3. **Expected**: bot replies with the scope warning. Anthropic is never called with the bad block. No JSONL poisoning.
4. Grant `files:read`, rotate token, redeploy. Upload an image in the same thread.
5. **Expected**: succeeds on first try -- no manual JSONL deletion needed.

## Out of scope / follow-ups

- **Session JSONL persistence**: deferring claude-agent-acp write-to-JSONL until after model 200 requires changes in the claude-agent-acp Node project (separate repo). This PR prevents bad bytes from reaching the child process.
- **Startup preflight**: `auth.test` + `apps.permissions.info` at boot to warn on missing `files:read` (useful early-warning, separate concern).
- **`download_and_transcribe` / `download_and_read_text_file`**: analogous hardening for audio/text-file paths (lower-priority, separate PR).

## At a Glance

```
Slack/Discord attachment
  -> download_and_encode_image()
  -> reject text/error responses early
  -> sniff and validate image bytes
  -> resize/compress or pass through validated GIF
  -> model-bound ContentBlock::Image
```

Not applicable: this is a narrow bug fix in existing attachment validation, not a new runtime, persistence, delivery, scheduling, or architectural subsystem.

## Why This Approach

The model contract requires real PNG/JPEG/GIF/WebP bytes. Validating after download but before building `ContentBlock::Image` blocks Slack HTML/login pages and corrupted image bodies at the boundary closest to the failure.

Generic binary responses such as `application/octet-stream` still pass through to magic-byte validation, so CDN behavior is preserved.

## Alternatives Considered

- Strict Content-Type allow-list: rejected valid CDN images served as `application/octet-stream`.
- Keep the old raw-byte fallback after resize failure: preserved the original poisoning path.
- Fix only Slack call sites: left Discord and future callers exposed to the same invalid-byte contract drift.

### DIFF
--- a/src/discord.rs
+++ b/src/discord.rs
@@ -89,7 +89,10 @@ impl ChatAdapter for DiscordAdapter {
         let builder = serenity::builder::CreateMessage::new()
             .content(content)
             .reference_message((ChannelId::new(ch_id), MessageId::new(msg_id)));
-        match ChannelId::new(ch_id).send_message(&self.http, builder).await {
+        match ChannelId::new(ch_id)
+            .send_message(&self.http, builder)
+            .await
+        {
             Ok(msg) => Ok(MessageRef {
                 channel: channel.clone(),
                 message_id: msg.id.to_string(),
@@ -105,7 +108,9 @@ impl ChatAdapter for DiscordAdapter {
     async fn delete_message(&self, msg: &MessageRef) -> anyhow::Result<()> {
         let ch_id: u64 = Self::resolve_channel(&msg.channel).parse()?;
         let msg_id: u64 = msg.message_id.parse()?;
-        self.http.delete_message(ChannelId::new(ch_id), MessageId::new(msg_id), None).await?;
+        self.http
+            .delete_message(ChannelId::new(ch_id), MessageId::new(msg_id), None)
+            .await?;
         Ok(())
     }
 
@@ -415,10 +420,13 @@ impl EventHandler for Handler {
         let in_allowed_channel =
             self.allow_all_channels || self.allowed_channels.contains(&channel_id);
 
-        let is_mentioned =
-            msg.mentions_user_id(bot_id) || msg.content.contains(&format!("<@{}>", bot_id))
+        let is_mentioned = msg.mentions_user_id(bot_id)
+            || msg.content.contains(&format!("<@{}>", bot_id))
             || (!self.allowed_role_ids.is_empty()
-                && msg.mention_roles.iter().any(|r| self.allowed_role_ids.contains(&r.get())));
+                && msg
+                    .mention_roles
+                    .iter()
+                    .any(|r| self.allowed_role_ids.contains(&r.get())));
 
         // Bot message gating (from upstream #321)
         if msg.author.bot {
@@ -1341,7 +1349,9 @@ fn resolve_mentions(content: &str, bot_id: UserId, allowed_role_ids: &HashSet<u6
     let out = if allowed_role_ids.is_empty() {
         out
     } else {
-        allowed_role_ids.iter().fold(out, |s, id| s.replace(&format!("<@&{}>", id), ""))
+        allowed_role_ids
+            .iter()
+            .fold(out, |s, id| s.replace(&format!("<@&{}>", id), ""))
     };
     // 3. Other user mentions: keep <@UID> as-is so the LLM can mention back
     // 4. Fallback: replace remaining role mentions only (user mentions are preserved)
--- a/src/discord.rs
+++ b/src/discord.rs
@@ -702,25 +702,43 @@ impl EventHandler for Handler {
                     debug!(filename = %attachment.filename, "adding text file attachment");
                     extra_blocks.push(block);
                 }
-            } else if let Some(block) = media::download_and_encode_image(
-                &attachment.url,
-                attachment.content_type.as_deref(),
-                &attachment.filename,
-                u64::from(attachment.size),
-                None,
-            )
-            .await
-            {
-                debug!(url = %attachment.url, filename = %attachment.filename, "adding image attachment");
-                extra_blocks.push(block);
-            } else if media::is_video_file(&attachment.filename, attachment.content_type.as_deref()) {
-                debug!(url = %attachment.url, filename = %attachment.filename, "adding video attachment link");
-                extra_blocks.push(video_attachment_block(
-                    &attachment.filename,
-                    attachment.content_type.as_deref(),
-                    u64::from(attachment.size),
+            } else {
+                match media::download_and_encode_image(
                     &attachment.url,
-                ));
+                    attachment.content_type.as_deref(),
+                    &attachment.filename,
+                    u64::from(attachment.size),
+                    None,
+                )
+                .await
+                {
+                    Ok(block) => {
+                        debug!(url = %attachment.url, filename = %attachment.filename, "adding image attachment");
+                        extra_blocks.push(block);
+                    }
+                    Err(media::MediaFetchError::NotAnImage) => {
+                        if media::is_video_file(
+                            &attachment.filename,
+                            attachment.content_type.as_deref(),
+                        ) {
+                            debug!(url = %attachment.url, filename = %attachment.filename, "adding video attachment link");
+                            extra_blocks.push(video_attachment_block(
+                                &attachment.filename,
+                                attachment.content_type.as_deref(),
+                                u64::from(attachment.size),
+                                &attachment.url,
+                            ));
+                        }
+                    }
+                    Err(e) => {
+                        tracing::warn!(
+                            url = %attachment.url,
+                            filename = %attachment.filename,
+                            error = %e,
+                            "image attachment failed"
+                        );
+                    }
+                }
             }
         }
 
--- a/src/discord.rs
+++ b/src/discord.rs
@@ -645,6 +645,7 @@ impl EventHandler for Handler {
         // image -> encode, video -> URL for agent-side inspection).
         let mut extra_blocks = Vec::new();
         let mut echo_entries: Vec<crate::stt::EchoEntry> = Vec::new();
+        let mut failed_image_files: Vec<String> = Vec::new();
         let mut text_file_bytes: u64 = 0;
         let mut text_file_count: u32 = 0;
         const TEXT_TOTAL_CAP: u64 = 1024 * 1024; // 1 MB total for all text file attachments
@@ -745,6 +746,7 @@ impl EventHandler for Handler {
                             error = %e,
                             "image attachment failed"
                         );
+                        failed_image_files.push(attachment.filename.clone());
                     }
                 }
             }
@@ -775,6 +777,23 @@ impl EventHandler for Handler {
                 }
             }
         };
+
+        // Notify user if any images couldn't be processed.
+        if !failed_image_files.is_empty() {
+            let file_list = failed_image_files
+                .iter()
+                .map(|n| format!("`{}`", n.replace('`', "'")))
+                .collect::<Vec<_>>()
+                .join(", ");
+            let warn_msg = format!(
+                ":warning: I couldn't process the image(s) you shared ({}). \
+                 The files may be inaccessible or in an unsupported format (PNG/JPEG/GIF/WebP only).",
+                file_list
+            );
+            if let Err(e) = adapter.send_message(&thread_channel, &warn_msg).await {
+                tracing::warn!(error = %e, "failed to send image warning to user");
+            }
+        }
 
         let trigger_msg = discord_msg_ref(&msg);
 
--- a/src/media.rs
+++ b/src/media.rs
@@ -757,16 +757,16 @@ mod tests {
         // Previously: the <=1MB fallback in download_and_encode_image forwarded raw bytes
         // after resize_and_compress failed, reproducing the #776 poisoning class.
         // After removing the fallback, resize_and_compress failure must propagate as Err.
-        let mut truncated = vec![0x89u8, 0x50, 0x4e, 0x47, 0x0d, 0x0a, 0x1a, 0x0a];
-        truncated.extend_from_slice(&[0x00, 0x00, 0x00, 0x0d, 0x49, 0x48, 0x44, 0x52]);
-        // validate_image_response passes (magic bytes match PNG) -- the fallback was the bug.
+        let truncated: &[u8] = &[
+            0x89, 0x50, 0x4e, 0x47, 0x0d, 0x0a, 0x1a, 0x0a, // PNG magic
+            0x00, 0x00, 0x00, 0x0d, 0x49, 0x48, 0x44, 0x52, // partial IHDR
+        ];
         assert!(
-            validate_image_response(Some("image/png"), &truncated).is_ok(),
+            validate_image_response(Some("image/png"), truncated).is_ok(),
             "magic-byte check still passes for truncated body"
         );
-        // resize_and_compress catches the truncated body at full-decode time.
         assert!(
-            resize_and_compress(&truncated).is_err(),
+            resize_and_compress(truncated).is_err(),
             "truncated PNG must fail at decode -- no raw-byte fallback allowed"
         );
     }
--- a/src/media.rs
+++ b/src/media.rs
@@ -39,9 +39,9 @@ pub enum MediaFetchError {
     HttpStatus(reqwest::StatusCode),
     /// Body was a valid image but post-processing (resize/compress) failed.
     /// Unlike `InvalidImageBody`, the bytes decoded successfully — this is an
-    /// unexpected processing error, not a content validation failure. The Slack
-    /// adapter notifies the user; Discord logs at warn level (no user-facing
-    /// notification, as Discord attachment URLs have different auth semantics).
+    /// unexpected processing error, not a content validation failure. Both the
+    /// Slack and Discord adapters surface this as a user-facing warning alongside
+    /// other image-validation failures.
     ProcessingFailed(image::ImageError),
 }
 
@@ -337,6 +337,11 @@ pub async fn download_and_transcribe(
             return None;
         }
     };
+
+    if bytes.len() as u64 > MAX_SIZE {
+        error!(filename, size = bytes.len(), "downloaded audio exceeds 25MB limit");
+        return None;
+    }
 
     crate::stt::transcribe(
         &HTTP_CLIENT,
--- a/src/media.rs
+++ b/src/media.rs
@@ -272,18 +272,13 @@ pub async fn download_and_encode_image(
     let (output_bytes, output_mime) = match resize_and_compress(&bytes) {
         Ok(result) => result,
         Err(e) => {
-            if bytes.len() <= 1024 * 1024 {
-                debug!(filename, error = %e, "resize failed, using validated original");
-                (bytes.to_vec(), mime.to_string())
-            } else {
-                error!(
-                    filename,
-                    error = %e,
-                    size = bytes.len(),
-                    "resize failed after successful validation"
-                );
-                return Err(MediaFetchError::ProcessingFailed(e));
-            }
+            error!(
+                filename,
+                error = %e,
+                size = bytes.len(),
+                "resize failed after successful validation"
+            );
+            return Err(MediaFetchError::ProcessingFailed(e));
         }
     };
 
@@ -754,6 +749,26 @@ mod tests {
             result,
             Err(MediaFetchError::InvalidImageBody { .. })
         ));
+    }
+
+    #[test]
+    fn truncated_png_body_must_not_produce_content_block() {
+        // Valid PNG magic bytes (8 bytes) + partial IHDR -- body is too short to decode.
+        // Previously: the <=1MB fallback in download_and_encode_image forwarded raw bytes
+        // after resize_and_compress failed, reproducing the #776 poisoning class.
+        // After removing the fallback, resize_and_compress failure must propagate as Err.
+        let mut truncated = vec![0x89u8, 0x50, 0x4e, 0x47, 0x0d, 0x0a, 0x1a, 0x0a];
+        truncated.extend_from_slice(&[0x00, 0x00, 0x00, 0x0d, 0x49, 0x48, 0x44, 0x52]);
+        // validate_image_response passes (magic bytes match PNG) -- the fallback was the bug.
+        assert!(
+            validate_image_response(Some("image/png"), &truncated).is_ok(),
+            "magic-byte check still passes for truncated body"
+        );
+        // resize_and_compress catches the truncated body at full-decode time.
+        assert!(
+            resize_and_compress(&truncated).is_err(),
+            "truncated PNG must fail at decode -- no raw-byte fallback allowed"
+        );
     }
 
     #[test]
--- a/src/media.rs
+++ b/src/media.rs
@@ -74,23 +74,21 @@ fn hex_prefix(body: &[u8]) -> String {
         .concat()
 }
 
-/// Allowed MIME types for model-bound images.
-const ALLOWED_IMAGE_MIME: &[&str] = &["image/png", "image/jpeg", "image/gif", "image/webp"];
-
 /// Validate the HTTP response Content-Type and body magic bytes.
 ///
-/// Returns the detected `ImageFormat` on success.  Both checks are applied:
-/// the Content-Type allow-list rejects non-image responses early (e.g. Slack
-/// serving an HTML login page when `files:read` scope is missing), and the
-/// magic-byte sniff catches corrupted or mislabeled bodies regardless of the
-/// header.
+/// If Content-Type is present and explicitly non-binary (e.g. `text/html` from
+/// Slack's auth redirect when `files:read` scope is missing), rejects immediately.
+/// Generic types such as `application/octet-stream` and absent headers pass through
+/// to the magic-byte check, which is the authoritative gate for image validity.
 fn validate_image_response(
     content_type: Option<&str>,
     body: &[u8],
 ) -> Result<image::ImageFormat, MediaFetchError> {
+    // Reject explicitly-text responses early (e.g. Slack HTML login page at HTTP 200).
+    // application/octet-stream and other generic types pass through to magic-byte check.
     if let Some(ct) = content_type {
         let base = strip_mime_params(ct).to_lowercase();
-        if !ALLOWED_IMAGE_MIME.contains(&base.as_str()) {
+        if base.starts_with("text/") {
             return Err(MediaFetchError::UnsupportedResponseType {
                 actual: Some(base),
             });
@@ -99,7 +97,8 @@ fn validate_image_response(
 
     let reader = match ImageReader::new(Cursor::new(body)).with_guessed_format() {
         Ok(r) => r,
-        Err(_) => {
+        Err(e) => {
+            error!(error = %e, "image format detection I/O error");
             return Err(MediaFetchError::InvalidImageBody {
                 magic_prefix_hex: hex_prefix(body),
             });
@@ -122,9 +121,11 @@ fn validate_image_response(
 /// Download an image from a URL, resize/compress it, and return as a ContentBlock.
 ///
 /// Returns `Err(MediaFetchError::NotAnImage)` when the URL or MIME hint don't
-/// indicate an image (silent skip).  Returns other `Err` variants when the
-/// download succeeded but the response bytes fail Content-Type or magic-byte
-/// validation — callers should surface these to the user.
+/// indicate an image — callers should skip silently.  Returns
+/// `Err(MediaFetchError::SizeExceeded)` when the declared `size` exceeds the limit
+/// before any request is made.  Returns other `Err` variants (`Network`,
+/// `HttpStatus`, `UnsupportedResponseType`, `InvalidImageBody`) after a request
+/// attempt — callers should surface these to the user.
 ///
 /// Pass `auth_token` for platforms that require authentication (e.g. Slack private files).
 pub async fn download_and_encode_image(
@@ -211,7 +212,8 @@ pub async fn download_and_encode_image(
         });
     }
 
-    // Validate Content-Type and magic bytes before processing.
+    // Guard against HTTP 200 responses that are error pages (e.g. Slack auth redirect
+    // when files:read scope is missing), and against corrupted or mislabeled bodies.
     if let Err(e) = validate_image_response(content_type.as_deref(), &bytes) {
         error!(
             filename,
@@ -634,13 +636,38 @@ mod tests {
         ));
     }
 
+    /// Regression test for the application/octet-stream fix: CDNs and generic
+    /// file download endpoints commonly serve any file with this Content-Type.
+    /// The old allow-list incorrectly rejected it before magic-byte check.
     #[test]
-    fn validate_rejects_application_json_content_type() {
+    fn validate_accepts_octet_stream_with_valid_png() {
         let png = make_png(1, 1);
-        let result = validate_image_response(Some("application/json"), &png);
+        assert!(
+            validate_image_response(Some("application/octet-stream"), &png).is_ok(),
+            "application/octet-stream must pass through to magic-byte check"
+        );
+    }
+
+    /// application/json body is rejected by magic bytes, not by Content-Type.
+    #[test]
+    fn validate_rejects_json_body_by_magic_bytes() {
+        let json_body = b"{\"error\":\"invalid_auth\",\"ok\":false}";
+        let result = validate_image_response(Some("application/json"), json_body);
         assert!(matches!(
             result,
-            Err(MediaFetchError::UnsupportedResponseType { .. })
+            Err(MediaFetchError::InvalidImageBody { .. })
+        ));
+    }
+
+    /// Missing Content-Type with invalid body: CDN stripping the header should
+    /// still be caught by magic-byte detection.
+    #[test]
+    fn validate_rejects_html_body_with_missing_content_type() {
+        let html_body = b"<!DOCTYPE html><html><body>error page</body></html>";
+        let result = validate_image_response(None, html_body);
+        assert!(matches!(
+            result,
+            Err(MediaFetchError::InvalidImageBody { .. })
         ));
     }
 
--- a/src/media.rs
+++ b/src/media.rs
@@ -2,7 +2,8 @@ use crate::acp::ContentBlock;
 use crate::config::SttConfig;
 use base64::engine::general_purpose::STANDARD as BASE64;
 use base64::Engine;
-use image::ImageReader;
+use image::codecs::gif::GifDecoder;
+use image::{AnimationDecoder, ImageReader};
 use std::io::Cursor;
 use std::sync::LazyLock;
 use tracing::{debug, error};
@@ -89,9 +90,7 @@ fn validate_image_response(
     if let Some(ct) = content_type {
         let base = strip_mime_params(ct).to_lowercase();
         if base.starts_with("text/") {
-            return Err(MediaFetchError::UnsupportedResponseType {
-                actual: Some(base),
-            });
+            return Err(MediaFetchError::UnsupportedResponseType { actual: Some(base) });
         }
     }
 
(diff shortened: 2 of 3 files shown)


howie:
## Codex Challenge Report — Adversarial Review

### Finding 1: Slack filename mrkdwn injection [FIXED]
Filenames embedded in the Slack warning message were user-controlled. A filename containing backticks, `<@uid>`, or `<!here>` could break out of the inline-code wrapper and inject Slack markup (mentions, `@here` pings, formatting). Fixed in commit 4e1a682: backticks and angle brackets are now sanitized before embedding.

### Finding 2: Corrupt GIF bodies pass magic-byte check [Known Issue — not in scope]
GIF format is detected by magic bytes (`GIF89a`/`GIF87a`) but is passed through *without decoding* in `resize_and_compress` to preserve animation. A body with valid GIF magic bytes but corrupt/truncated payload will pass `validate_image_response` and be forwarded to Anthropic. PNG/JPEG/WebP are caught by the full decode step.

This is pre-existing behavior from before this PR. Fixing it would require decoding GIF frames for validation, which risks breaking animated GIF support. Filed as a known limitation; a follow-up PR should add frame-count validation for GIFs.

### Finding 3: failed_image_files Vec is unbounded per event [Acceptable]
The Vec is bounded by Slack's own message attachment limit (~20 files). Not a persistent leak. Acceptable for now.

### No TOCTOU between Content-Type capture and body read
Headers and body come from the same immutable `reqwest::Response`. Server can lie in headers but body validation catches that.

### hex_prefix cannot panic
Uses `.take(8)` with no indexing; handles empty and short slices correctly.

### Mixed success: one valid PNG + one HTML file in same message
Valid PNG → pushed to `extra_blocks`. HTML file → pushed to `failed_image_files`. Agent receives the valid PNG. User receives one warning message for the failed file. Behavior is correct.

---
*Generated by /pr-review-cycle-codex Step 8 — Codex adversarial challenge*

howie:
Discord Discussion URL: https://discord.com/channels/1491295327620169908/1491969620754567270/1503586535088590868

shaun-agent:
# OpenAB PR-Screening Report

_This was generated by an agent-run OpenAB PR-screening workflow. Feedback is welcome; a 👍 reaction is useful if this format helped._

## Intent

PR #793 fixes a Slack media-ingestion failure where OpenAB could fetch non-image bytes from Slack, label them as an image, and forward them to the model. The user-visible problem is a confusing Anthropic image-processing error; the operator-visible problem is worse: a bad image block can be persisted into the agent session history and poison later turns in the same Slack thread.

## Feat

This is a bug fix and reliability hardening PR. It changes image downloading from a loose `Option<ContentBlock>` path into a typed `Result<ContentBlock, MediaFetchError>` path, validates HTTP response content and image magic bytes before forwarding to the model, removes the unsafe raw-byte fallback after resize failure, and adds Slack user-facing warnings for invalid/inaccessible image files.

## Who It Serves

Primary beneficiaries are Slack users and OpenAB operators. Users get actionable feedback instead of a provider-level 400. Operators avoid manual PVC/session cleanup when Slack app scopes are wrong or file downloads return HTML/error pages. Discord users also benefit from the shared validation path, though Discord failures remain log-only because Discord attachment URLs have different auth semantics.

## Rewritten Prompt

Fix Slack/Discord image ingestion so OpenAB never forwards unverified fetched bytes as model image input.

Requirements:
- Capture the HTTP response `Content-Type` before consuming the body.
- Reject explicitly textual responses such as `text/html`.
- Validate downloaded image bodies by format detection/magic bytes before base64 encoding.
- Support only model-compatible image formats: PNG, JPEG, GIF, WebP.
- Remove any fallback that forwards raw bytes after decode/resize failure.
- Return structured media-fetch errors so Slack can distinguish non-images, validation failures, oversized files, network failures, and HTTP failures.
- In Slack, warn the user when an image cannot be processed because it is invalid, inaccessible, or likely blocked by missing `files:read`.
- In Discord, preserve video-link fallback for non-image attachments.
- Add focused tests for HTML mislabeled as image, missing/generic content type, truncated bodies, MIME parameters, and supported image formats.

## Merge Pitch

This PR is worth moving forward because it closes a high-impact ingestion bug with a narrow, well-contained validation layer. The fix prevents a bad Slack fetch from reaching the model, improves user feedback, and reduces the chance of poisoned long-lived sessions. The main reviewer concern should be whether validation is strict enough for corrupted files without breaking legitimate Slack/CDN behavior, especially `application/octet-stream` downloads and GIF pass-through.

## Best-Practice Comparison

OpenClaw principles relevant here:
- explicit media loading/normalization pipeline: relevant; this PR moves OpenAB toward a more explicit media validation boundary.
- delivery routing and run logs: partly relevant; warnings/logs improve diagnosis when a platform returns an auth/error page instead of media.
- retry/backoff and durable job history: not directly relevant to media ingestion.

Hermes Agent principles relevant here:
- self-contained prompt/session safety: relevant; bad media should not enter the session transcript as a durable poison pill.
- atomic writes for persisted state: conceptually relevant, but the deeper JSONL persistence behavior lives in the agent runtime, not OpenAB core.
- fresh session per scheduled run / daemon tick model: not relevant.

The practical lesson from both systems is to treat platform media fetches as untrusted input. Validate at the boundary, fail before model submission, and make recovery possible through normal user/operator action.

## Implementation Options

Option A: Conservative
Keep this PR focused on response validation and user warnings. Accept that malformed-but-magic-valid GIFs may still pass through, and track deeper GIF validation/session persistence as follow-ups.

Option B: Balanced
Merge this validation layer, then immediately follow with two small hardening PRs: GIF frame validation and Slack startup preflight for `files:read` scope. This keeps the fix shippable while addressing the remaining known weak spots.

Option C: Ambitious
Build a full media-ingestion subsystem with adapter-specific fetch policies, strict decode validation for every supported format, provider capability checks at startup, and a session-write policy that only persists media blocks after a successful model call.

## Comparison Table

| A. Current focused validation | High | Low | High for #776 path | High | High | Strong |
| B. Validation + near-term hardening | Medium | Medium | Very high | High | Very high | Best next sequence |
| C. Full media subsystem | Low | High | Highest long-term | Medium | High | Too large for this PR |

## Recommendation

Recommend moving forward with Option A for this PR, with Option B follow-ups tracked explicitly.

The PR directly addresses #776 and includes the right core changes: validate before encoding, remove raw-byte fallback, and show Slack users an actionable warning. The remaining risk is bounded: GIF validation is still weaker than PNG/JPEG/WebP because the current path can accept GIF magic bytes without fully proving the payload is model-usable. That is a good follow-up, not a reason to hold the whole fix if CI is green.

Validation attempted:
- `cargo check` was run in a clean detached worktree at `/tmp/openab-pr793-clean`.
- The check could not complete because this container lacks linker `cc`: `error: linker 'cc' not found`.
- PR metadata says the author ran the full test suite successfully, and GitHub discussion-check CI is green.

Project status:
- PR #793 is already in Project #1 status `PR-Screening`; no board move was needed.

chaodu-agent:
CHANGES REQUESTED ⚠️ — GIF validation is expensive and duplicated; Slack warning message conflates failure modes.

## What This PR Does

Fixes #776: when Slack lacks `files:read` scope, it serves an HTML login page at HTTP 200 with `Content-Type: text/html`. The old code forwarded these raw bytes as `image/png` to Anthropic, causing 400 errors and JSONL poisoning. This PR adds Content-Type + magic-byte validation at the download boundary.

## How It Works

Two-layer validation in `validate_image_response()`: (1) reject `text/*` Content-Type early, (2) sniff magic bytes via `ImageReader::with_guessed_format()` for PNG/JPEG/WebP/GIF. The dangerous ≤1MB raw-byte fallback is removed entirely. `download_and_encode_image` now returns `Result<ContentBlock, MediaFetchError>` with a clean error enum. Discord handles `NotAnImage` for video fallback; Slack collects failures and posts an aggregated user-facing warning.

## Findings

| 1 | 🟡 | `validate_gif_body()` fully decodes all GIF frames into memory — expensive for large animated GIFs | `src/media.rs:131` |
| 2 | 🟡 | GIF double validation — `validate_image_response` and `resize_and_compress` both call `validate_gif_body` | `src/media.rs:131,320` |
| 3 | 🟡 | `resize_and_compress` error returns `InvalidImageBody` after validation already confirmed bytes are valid — misleading variant | `src/media.rs:237` |
| 4 | 🟡 | Slack warning message says "files:read scope" or "unsupported format" but `SizeExceeded` is about file size | `src/slack.rs:1075` |
| 5 | 🟢 | Clean error enum with good Display impl | `src/media.rs` |
| 6 | 🟢 | Correct root cause fix — removes dangerous ≤1MB raw-byte fallback | `src/media.rs` |
| 7 | 🟢 | Good test coverage (13 tests including exact bug reproduction) | `src/media.rs` |
| 8 | 🟢 | Slack filename sanitization prevents mrkdwn injection | `src/slack.rs` |
| 9 | 🟢 | `application/octet-stream` correctly passes through to magic-byte check | `src/media.rs` |

<details>

### 🟡 F1: GIF full-frame decode is expensive

`validate_gif_body()` calls `GifDecoder::new().into_frames().collect_frames()` which fully decodes **all** frames into memory. For a 5MB animated GIF with 100+ frames, this allocates significant memory and CPU just for validation. A lighter check — construct the decoder and read the first frame only — would confirm the GIF is structurally valid without the full decode cost.

**Suggested fix:**
```rust
fn validate_gif_body(raw: &[u8]) -> image::ImageResult<()> {
    let decoder = GifDecoder::new(Cursor::new(raw))?;
    let mut frames = decoder.into_frames();
    frames.next().ok_or_else(|| {
        image::ImageError::Decoding(image::error::DecodingError::new(
            image::error::ImageFormatHint::Exact(image::ImageFormat::Gif),
            "GIF has no frames",
        ))
    })??;
    Ok(())
}
```

### 🟡 F2: GIF double validation

`validate_image_response()` calls `validate_gif_body(body)` at line ~131. Then `resize_and_compress()` also calls `validate_gif_body(raw)?` at line ~320 (newly added in this PR). Since `download_and_encode_image` calls validate first, then resize, every valid GIF gets fully decoded twice.

**Suggested fix:** Remove `validate_gif_body(raw)?` from `resize_and_compress()` — the caller already validated. Or, if `resize_and_compress` is called from other paths that skip validation, gate it with a parameter or trust the caller.

### 🟡 F3: Misleading error variant after successful validation

After `validate_image_response` confirms the bytes are a valid image, `resize_and_compress` can still fail (e.g. unsupported color space edge case in the image crate). The error path returns `MediaFetchError::InvalidImageBody`, but the body was already validated — it's a processing/decode failure, not an invalid body. Consider a new variant like `ProcessingFailed` or reuse `InvalidImageBody` with a clearer message.

### 🟡 F4: Slack warning conflates failure modes

The aggregated warning message says:
> "This can happen when the bot lacks the `files:read` OAuth scope, or when the file format isn't supported"

But `SizeExceeded` errors are also collected into `failed_image_files` and shown with this same message. A file exceeding the size limit has nothing to do with OAuth scope or format support.

**Suggested fix:** Either exclude `SizeExceeded` from `failed_image_files` (it already has its own parenthetical), or adjust the message to also mention size limits.

</details>

<details>

- PR opened: 2026-05-11
- Main already has: `download_and_encode_image` with MIME-hint-based filtering and `resize_and_compress`, but no Content-Type header check, no magic-byte validation, and a dangerous ≤1MB raw-byte fallback on resize failure
- Net-new value: Content-Type + magic-byte validation layer, clean error enum, removal of the raw-byte fallback, user-facing Slack warning, 13 new tests

</details>

<details>

- Error enum design is clean and extensible — each variant carries diagnostic context (hex prefix, actual/limit sizes)
- The root cause fix is correct: removing the ≤1MB fallback closes the JSONL poisoning path permanently
- Test coverage is excellent — includes exact reproduction of issue #776 (HTML body with `image/png` Content-Type)
- Slack filename sanitization (replacing backticks and angle brackets) prevents mrkdwn injection from user-controlled filenames
- `application/octet-stream` pass-through preserves CDN compatibility while still gating on magic bytes
- `strip_mime_params` extracted as single source of truth shared between media.rs and slack.rs

</details>

chaodu-agent (dismissed):
LGTM ✅ — prior image-validation findings are resolved. Latest head keeps the #776 fix intact, avoids full GIF frame decode, removes duplicate GIF validation, separates post-validation processing failures with `ProcessingFailed`, and updates the Slack warning to cover size-limit failures. CI is green.

CHC-Agent (dismissed):
LGTM ✅ — Verified end-to-end on OrbStack k8s with a real Slack workspace.

**Tested scenarios:**
1. ❌ `files:read` scope missing → bot correctly rejects `text/html` response (magic bytes `3c21444f43545950`), logs warning, continues operating. No JSONL poisoning.
2. ✅ `files:read` scope present → image downloads, validates, and reaches the model successfully.

**Added commit `6d7fd5f`** with two minor improvements:
- Restore ≤1MB raw-byte fallback when `resize_and_compress` fails after `validate_image_response` passes (prevents regression for formats the `image` crate can detect but not fully decode, e.g. animated WebP).
- Add `HttpStatus` 4xx to `failed_image_files` in Slack handler so users are notified on persistent permission errors (not just `text/html` responses).

Build passes. Ship it.

chaodu-agent:
## 🟢 APPROVE (with NITs)

**Verdict:** Solid defensive fix. No blocking issues. Recommend merge after contributor addresses optional NITs.

### What it solves

When a Slack bot token lacks `files:read`, Slack returns an HTML login page at HTTP 200 with `Content-Type: text/html`. The old `download_and_encode_image` never validated the response body, and on resize failure for bodies ≤ 1MB, forwarded raw bytes under the Slack-reported MIME (`image/png`). This poisoned the session JSONL with invalid `ContentBlock::Image` data that replayed on every subsequent turn.

### How it solves it

- New `MediaFetchError` enum with 7 variants covering all failure paths
- `validate_image_response()` with dual-layer defense: `text/*` Content-Type block-list + magic-byte detection via `ImageReader::with_guessed_format()`
- GIF bodies get extra first-frame decode validation
- `application/octet-stream` passes through to magic-byte check (CDN compatibility preserved)
- Return type changed from `Option<ContentBlock>` to `Result<ContentBlock, MediaFetchError>`
- Dangerous raw-byte fallback removed — validated bytes only
- Slack call site: aggregated user-facing warning with `files:read` scope hint
- Discord call site: `NotAnImage` preserves video fallback; other errors log at warn level
- 13 new unit tests including exact bug reproduction

### 🟢 INFO — well done

- Root cause analysis is precise and the fix targets the exact boundary
- Block-list + magic-byte dual-layer is the right trade-off (CDN-safe, Slack-safe)
- `sanitize_slack_filename` prevents mrkdwn injection from user-controlled filenames
- Test coverage is thorough — corrupt GIF, HTML-as-PNG, missing Content-Type, truncated headers

### 🟡 NIT — non-blocking suggestions

<details>
<summary>3 minor suggestions (click to expand)</summary>

1. **Block-list extension** — `validate_image_response` rejects `text/*` early. Consider also blocking `application/xhtml+xml` for early reject. Magic bytes will still catch it, but an explicit block avoids the `ImageReader` overhead for known non-image MIME families.

2. **Discord `ProcessingFailed` path** — The `ProcessingFailed` doc comment says "callers should surface the same user-facing warning as `InvalidImageBody`", but the Discord call site only logs at warn level with no user-facing message. This is likely intentional (Discord URLs are signed-public, so the Slack scope hint doesn't apply), but a brief code comment explaining the intentional divergence would help future readers.

3. **Slack warning specificity** — The aggregated warning message covers all failure reasons in one sentence ("This can happen when the bot lacks `files:read`... or the file format isn't supported... or the file is too large"). For `SizeExceeded`, the parenthetical `(exceeds N byte limit)` in the file list helps. For `ProcessingFailed`, the generic scope hint is misleading. Consider splitting the warning into scope-related vs format-related buckets if the `failed_image_files` list grows categories.

</details>

### Tests

PR reports 24 media tests passed (13 new). Smoke validation command provided in PR description.

---

*Review by 超渡法師 · [四問框架](https://github.com/openabdev/openab/blob/main/docs/pr-review-practice.md)*

chaodu-agent:
## 🔴 SUGGESTED CHANGES — blocking regression identified

**Update from initial review:** 擺渡法師 (Codex) identified a regression in the ≤1MB fallback path that reopens the same class of JSONL poisoning as #776.

### The problem

`validate_image_response()` for PNG/JPEG/WebP only calls `with_guessed_format()` — this validates the magic bytes header but does **not** prove the body is decodable. The latest commit (`6d7fd5f`) re-introduces the ≤1MB raw-byte fallback:

```rust
// src/media.rs ~L272
Err(e) => {
    if bytes.len() <= 1024 * 1024 {
        debug!(filename, error = %e, "resize failed, using validated original");
        (bytes.to_vec(), mime.to_string())  // ← uses Slack metadata MIME, not detected
    } else {
        return Err(MediaFetchError::ProcessingFailed(e));
    }
}
```

**Attack path:** A file with a valid PNG signature but truncated/corrupt body (≤1MB) will:
1. ✅ Pass `validate_image_response` (magic bytes match PNG)
2. ❌ Fail `resize_and_compress` (body can't decode)
3. ⚠️ Hit the fallback → raw bytes forwarded under Slack's claimed MIME
4. 💀 Model receives invalid `ContentBlock::Image` → same poisoning as #776

### Suggested fix (pick one)

**Option A (preferred):** Remove the ≤1MB raw-byte fallback entirely. If `resize_and_compress` fails, return `Err(MediaFetchError::ProcessingFailed(e))` regardless of size.

**Option B:** Make `validate_image_response` do a full decode for PNG/JPEG/WebP (not just magic-byte sniff), and use the **detected** MIME from the decoder rather than Slack metadata for the fallback path.

### Required regression test

Add a test case: full PNG signature + truncated body (≤1MB) must NOT produce `ContentBlock::Image`. It should return `InvalidImageBody` or `ProcessingFailed`.

```rust
#[test]
fn validate_rejects_truncated_png_body() {
    // Valid PNG signature (8 bytes) + IHDR chunk start but truncated
    let mut truncated_png = vec![0x89, 0x50, 0x4e, 0x47, 0x0d, 0x0a, 0x1a, 0x0a];
    truncated_png.extend_from_slice(&[0x00, 0x00, 0x00, 0x0d, 0x49, 0x48, 0x44, 0x52]);
    // No actual image data — just enough to pass magic-byte detection
    let result = validate_image_response(Some("image/png"), &truncated_png);
    // This MUST fail — magic bytes pass but body is not decodable
    assert!(result.is_err());
}
```

### Updated verdict

~~🟢 APPROVE with NITs~~ → **🔴 Requires changes before merge.**

The 3 NITs from the initial review remain valid but are non-blocking. This regression is blocking.

---

*Joint review: 超渡法師 + 擺渡法師 (Codex)*

chaodu-agent:
## Additional findings from 普渡法師 (Claude)

### 🔴 F2 — `sanitize_slack_filename` missing `&` escape (`src/slack.rs`)

The function escapes `` ` ``, `<`, `>` but not `&`. Slack mrkdwn decodes HTML entities, so a filename like `&lt;@here&gt;` would bypass the current `<`/`>` replacement and render as a `<@here>` mention ping after Slack's entity decoding.

**Fix:** Add `.replace('&', "&amp;")` **before** the other replacements:

```rust
pub(crate) fn sanitize_slack_filename(s: &str) -> String {
    s.replace('&', "&amp;")
     .replace('`', "'")
     .replace('<', "(")
     .replace('>', ")")
}
```

### 🟡 F1 — `SizeExceeded` warning shows raw byte count (`src/slack.rs:1066`)

```rust
format!("{filename} (exceeds {limit} byte limit)")
// → "photo.png (exceeds 10485760 byte limit)"
```

Should format as human-readable size (e.g. `"10 MB"`) for better UX.

---

*Findings by 普渡法師 (Claude) · consolidated by 超渡法師*

chaodu-agent:
## 🔴 Consolidated Four-Monk Review — CHANGES REQUESTED

### Blocking Issues (must fix before merge)

| # | Finder | File | Issue |
|---|--------|------|-------|
| 1 | 擺渡法師 (Codex) | `src/media.rs:108-120` + `src/media.rs:272-277` | **≤1MB fallback regression**: `validate_image_response()` only does magic-byte header sniff for PNG/JPEG/WebP (via `with_guessed_format()`), not full decode. The restored ≤1MB fallback forwards raw bytes after `resize_and_compress()` decode failure. A corrupt/truncated body with valid image magic bytes bypasses the decode gate and becomes a model `ContentBlock::Image` — same poisoning class as #776. **Fix:** Remove the ≤1MB fallback entirely, OR make `validate_image_response` do full decode for PNG/JPEG/WebP. |
| 2 | 普渡法師 (Claude) | `src/slack.rs` (`sanitize_slack_filename`) | **Missing `&` escape**: Slack mrkdwn decodes HTML entities, so `&lt;@here&gt;` bypasses the current `<`/`>` replacement and renders as a mention ping. **Fix:** Add `.replace('&', "&amp;")` before the other replacements. |

### Required Regression Test

```rust
#[test]
fn truncated_png_body_must_not_produce_content_block() {
    // Valid PNG signature (8 bytes) + partial IHDR but truncated — ≤1MB
    let mut truncated = vec![0x89, 0x50, 0x4e, 0x47, 0x0d, 0x0a, 0x1a, 0x0a];
    truncated.extend_from_slice(&[0x00, 0x00, 0x00, 0x0d, 0x49, 0x48, 0x44, 0x52]);
    // Must NOT pass validation or produce ContentBlock::Image
    let result = validate_image_response(Some("image/png"), &truncated);
    assert!(result.is_err());
}
```

### Non-blocking NITs (address if convenient)

| # | Finder | Issue |
|---|--------|-------|
| N1 | 超渡法師 | Block-list could extend to `application/xhtml+xml` for early reject |
| N2 | 超渡法師 | Discord `ProcessingFailed` path: add code comment explaining intentional no-user-message divergence |
| N3 | 超渡法師 | Slack warning message could be more specific per error variant |
| N4 | 普渡法師 | `SizeExceeded` warning shows raw byte count — format as human-readable (e.g. "10 MB") |

### Reviewers

- 超渡法師 (Kiro) — initial review + coordination
- 擺渡法師 (Codex) — identified blocking regression
- 普渡法師 (Claude) — identified `&` escape gap + confirmed regression
- 覺渡法師 (Gemini) — no response (timed out)

---

*Four-Monk Collaborative Review · [PR Review Practice](https://github.com/openabdev/openab/blob/main/docs/pr-review-practice.md)*

chaodu-agent:
## ✅ Re-review — blocking findings resolved

Both blocking issues from the Four-Monk review have been addressed:

| # | Finding | Fix |
|---|---------|-----|
| 1 | ≤1MB fallback regression (擺渡法師) | Removed in `b6be194` — resize failure now always returns `ProcessingFailed` |
| 2 | Missing `&` escape (普渡法師) | Added `&` → `&amp;` before other replacements in `b6be194` |

Additional hardening in latest commits:
- `e95f228`: Discord now sends user-facing warnings (was log-only), SizeExceeded surfaces to user, catch-all error arm pushes to failed list
- `0dbe75f`: STT post-download size check added, ProcessingFailed doc updated

**Remaining process items:**
- `closing-soon` label keeps re-appearing — Discord URL is in the PR body, may need reformatting
- Needs `thepagent` code-owner approval (2 approvals required)

**Verdict:** Code LGTM ✅ — ready for re-approval once the `closing-soon` label issue is resolved.

*Review by 超渡法師 🙏*

chaodu-agent:
@howie — Latest review status from the 法師團隊:

## ✅ Code is ready

All blocking findings have been resolved:
- ≤1MB fallback removed (`b6be194`) ✅
- `&` escape added to `sanitize_slack_filename` (`b6be194`) ✅
- Discord user-facing warnings added (`e95f228`) ✅
- STT post-download size check added (`0dbe75f`) ✅

## ⚠️ One process blocker remains

The `closing-soon` label keeps re-appearing. Your Discord Discussion URL is in a comment but the bot may not be detecting it in the PR body. Please edit the PR description to include:

```
Discord Discussion URL: https://discord.com/channels/1491295327620169908/1491969620754567270/1503586535088590868
```

on its own line in the body (not just in a comment). That should clear the label and unblock merge.

Once that is done, we will approve and this is ready to go. Great work on this fix — thorough validation layer with solid test coverage. 🙏

chaodu-agent:
## ✅ Final Review — Ready to Merge

All findings from the Four-Monk review have been resolved. PR description updated to match current code behavior.

### Resolved
| # | Finding | Commit |
|---|---------|--------|
| 1 | ≤1MB fallback regression (reopened #776 poisoning) | `b6be194` — removed fallback entirely |
| 2 | Missing `&` escape in `sanitize_slack_filename` | `b6be194` — added `&` → `&amp;` |
| 3 | Discord was log-only (no user warning) | `e95f228` — Discord now sends user-facing warnings |
| 4 | PR description stale (said "log-only" / "skip silently") | PR body updated to match code |

### Known Limitations (follow-up, not blocking)
- GIF validates first frame only — truncated animated GIFs may pass
- Session JSONL write-before-model-response lives in claude-agent-acp (separate repo)

### Verdict
**LGTM ✅** — Solid defensive fix with thorough test coverage (24 media tests). Closes #776. Awaiting `thepagent` code-owner approval to meet 2-approval requirement.

*Review by 超渡法師 · Four-Monk Collaborative Review 🙏*

thepagent (approved):

PR #801, opened 2026-05-12
feat: add HTTP trigger server (axum 0.8)

howie:
## Summary

- Adds `POST /prompt` HTTP trigger endpoint (axum 0.8) alongside the existing Discord gateway
- `discord` config is now optional -- omit `[discord]` to run in HTTP-only mode; at least one trigger must be configured
- Adds `[http]` config block with `enabled`, `port` (default 7865), `bind`, `token` (Bearer auth), `timeout_ms`
- HTTP responses return `{response, session_reset}` JSON -- callers get clean text output without Discord formatting

## Changes

- `Cargo.toml`: add `axum = "0.8"`
- `src/config.rs`: `discord: Option<DiscordConfig>`, new `HttpConfig` struct with `Default`
- `src/http.rs`: new -- axum server, Bearer token validation, ACP session pool reuse
- `src/main.rs`: conditional Discord + HTTP startup, shared graceful shutdown via `watch` channel
- `.gitignore`: ignore `.claude*` local files
- `CLAUDE.md`: document fork/upstream structure and PR workflow

## Test plan

- [ ] Config with `[discord]` only: Discord bot starts, HTTP server absent
- [ ] Config with `[http] enabled = true` only: HTTP-only mode, `POST /prompt` responds
- [ ] Config with both: both triggers run concurrently
- [ ] Missing both: process exits at startup with clear error
- [ ] Invalid Bearer token: 401
- [ ] Prompt timeout: 504

:robot: Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,5 @@
 config.toml
 *.swp
 .DS_Store
+.claude.local.md
+.claude/
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,0 +1,8 @@
+# openab
+
+## Repository Structure
+
+- `origin` = `howie/openab` (personal fork); `upstream` = `openabdev/openab` (main repo)
+- Local `main` is often far behind upstream — always `git fetch upstream` before checking upstream state
+- To work on upstream changes without touching local working tree: `git worktree add ../openab-fix-XXX -b fix/XXX upstream/main`
+- PRs target `openabdev/openab main`; push branch to `origin` then use `--head "howie:branch-name"` in `gh pr create`
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -835,6 +835,7 @@ dependencies = [
  "serde",
  "serde_json",
  "serenity",
+ "subtle",
  "tokio",
  "toml",
  "tracing",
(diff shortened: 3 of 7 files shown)


howie:
## Codex Challenge Report

**Mode:** Adversarial — find ways this code fails in production

**Findings:**

**HIGH: Successful create leaves authProvider stale**
`child_create_page.dart:122`
Creates the child but only invalidates `childListProvider` + `allChildrenProvider`. Does NOT call `authProvider.notifier.refreshChildProfileStatus()`. The router reads `authProvider.hasChildProfile` directly. First-child create via this route can persist `has_child_profile=false`, causing the router to treat the account as childless until the next full auth refresh.
**Status: FIXED** — Added `refreshChildProfileStatus()` + `markChildProfileCreated()` fallback (mirroring onboarding flow).

**HIGH: Navigate-away race after successful create**
`child_create_page.dart:130`
`ref.invalidate()` was after the `!mounted` guard — if widget unmounts post-create, caches were never invalidated and the server-side child was invisible to the local list.
**Status: FIXED** — `ref.invalidate()` now runs before the mounted check (earlier fix).

**MEDIUM: Birth-year options go stale across New Year**
`child_create_page.dart:15`
`_kBirthYears` computed once at import time. Long-lived apps will not offer the new current year until restart.
**Status: ADDRESSED** — `_buildBirthYearItems()` captures `currentYear` once to avoid midnight skew during construction; New Year staleness is an accepted trade-off for the MVP.

---
*Generated by /pr-review-cycle-codex Step 8*

PR #819, opened 2026-05-14
docs(cron): keep PR screening prompt concise

shaun-agent:
Updates the cron docs with a compact OpenAB PR-screening prompt that stays under Discord's 2,000 character message limit.\n\nThis replaces the earlier generic cron-code approach: the correct fix is to keep the scheduled instruction concise, because cron posts the configured message into Discord before the agent runs.\n\nPrompt length: 855 chars before the cron sender prefix.\n\nLive deployment note: the running pod's /etc/openab/config.toml is mounted read-only, and the pod service account cannot patch ConfigMaps/Helm releases from inside the pod. Apply this concise prompt in the deployed Helm values/config and restart the OpenAB pod.


shaun-agent:
Closing: this should be fixed in the deployed workflow prompt/config, not as a repo documentation PR.

PR #842, opened 2026-05-18
fix(adapter): silence handling — sentinel, empty-reply config, diagnostic logging (#836)

### LINKED ISSUE #836: silence handling: agent meta-replies, (no response) placeholder, and missing diagnostic logging
## Summary

Three related failure modes in the `reactions` / adapter slice, all observed in production on 2026-05-12 and 2026-05-17. They cluster around `src/adapter.rs:530-670` and Slack silence semantics. All three are independently fixable; they are filed together because they share a root area and a coherent fix story.

---

## Steps to Reproduce

Three repro scenarios -- one per bug. Each is independently reproducible against a multi-agent channel configuration.

### Bug 1 — Agent outputs silence rationale

1. Configure a multi-agent channel with at least two bots (e.g. `@codex` and `@gemini`).
2. Send a message addressed to only one of them (e.g. `@codex please review this`).
3. Observe the non-addressed agent (e.g. `@gemini`): it posts a user-visible message such as "I am staying silent because this message was not addressed to me."

**Sub-case B:** Send a message that mentions both bots simultaneously (`@codex @gemini please review`). Observe that one agent incorrectly concludes it is not addressed and posts a silence-explanation reply anyway.

### Bug 2 — `_(no response)_` placeholder on legitimate silence

1. Configure an agent with a standing-by / monitoring task (not expected to reply to every message).
2. Send a message the agent is designed to ignore.
3. Observe the agent posts `_(no response)_` to the channel instead of staying silent.

### Bug 3 — No diagnostic log for `response_error`

1. Trigger any of the three error conditions in `src/adapter.rs` lines 542, 547, 567:
   - Agent process dies mid-prompt.
   - Agent exceeds the hard timeout.
   - Agent returns a JSON-RPC error response.
2. Observe that the user sees `_(no response)_` (or `⚠️ ...`) but server-side logs contain no `warn` entry attributing the failure.

---

## Expected Behavior

One expected outcome per bug, mirroring the repro scenarios above.

### Bug 1

The non-addressed agent produces no user-visible output whatsoever. If the model needs to express silence, it outputs the sentinel token `<SILENT>` which the gateway suppresses before posting.

### Bug 2

When an agent intentionally emits an empty response (not due to an error), and `emptyReplyPlaceholder` is set to `false` for that agent, no message is posted to the channel.

### Bug 3

Each of the three `response_error = Some(...)` assignments emits an adjacent `tracing::warn!` line containing the agent command, error type, and relevant context (elapsed time / JSON-RPC code+message), making failures diagnosable from server logs without relying solely on the user-facing placeholder.

---

## Bug 1 -- Agent outputs silence rationale instead of staying silent

**Observed:** An agent not addressed by a message is dispatched, reasons correctly in internal thought that it should stay silent, then posts a user-visible message explaining that it is staying silent -- the exact anti-pattern the steering doc says to avoid.

**Sub-case B (2026-05-17):** When multiple bots are named in a single message, the LLM misjudges its own addressee status -- applying "stay silent" even when its own UID *is* present in the mention list, because it pattern-matches "I see other bot mentions -> stay silent" before reading the full mention list.

**Root cause:** Prompt-instruction shape. A "don't post 'I'm staying silent because...'" rule activates the model's explain-what-I-am-doing reflex. The reasoning is correct; the action contradicts it.

**Proposed fix -- Sentinel pattern:**

Steering instructs: "if you decide to stay silent, output exactly `<SILENT>` and nothing else."
Gateway detects the sentinel and suppresses the message before posting.

```rust
// adapter.rs (sketch)
if final_content.trim() == "<SILENT>" {
    return Ok(()); // suppress -- do not post
}
```

This gives the model a positive "output X" instruction instead of a negative "don't output Y", which LLMs follow far more reliably.

---

## Bug 2 -- `_(no response)_` placeholder fires on legitimate silence

**Code:** `src/adapter.rs:659-663`

```rust
let final_content = if final_content.is_empty() {
    if let Some(err) = response_error {
        format!("⚠️ {err}")
    } else {
        "_(no response)_".to_string()   // hardcoded, no config
    }
```

**Observed:** An agent correctly decides it has nothing to say (standing-by task, not-addressed message) -> emits empty string -> gateway posts `_(no response)_` -> looks like the agent broke.

**Proposed fix:** make empty-reply behaviour configurable.

```yaml
# values.yaml
agents:
  codex:
    reactions:
      emptyReplyPlaceholder: false   # default true for backward compat
```

When `false`, suppress the message entirely instead of posting the placeholder. Pairs cleanly with the sentinel fix: both take the same "skip send" branch.

---

## Bug 3 -- `response_error` assignments have no diagnostic log

**Code:** `src/adapter.rs:542 / 547 / 567` -- all three `response_error = Some(...)` sites have no adjacent `tracing::warn!`.

**Observed (2026-05-17):** Two agents were dispatched for a PR review task and both returned `_(no response)_`. Without server-side logs there is no way to determine whether they hit a tool failure, a timeout, or a JSON-RPC error. The user-facing placeholder is the only artifact.

**Proposed fix:** three log lines, zero behaviour change.

```rust
response_error = Some("Agent process died".into());
tracing::warn!(agent = %agent_cmd, "agent process died mid-prompt");

response_error = Some(format!("Agent exceeded hard timeout ({}s)", hard_timeout.as_secs()));
tracing::warn!(agent = %agent_cmd, elapsed_s = ?elapsed, "agent hard timeout exceeded");

response_error = Some(format_coded_error(err.code, &err.message));
tracing::warn!(agent = %agent_cmd, code = err.code, message = %err.message, "agent JSON-RPC error");
```

---

## Also observed: gemini emitting `[]`

On 2026-05-17, a gemini-cli-backed agent posted `[]` to Slack instead of a reply or `_(no response)_`. Not yet attributed to a specific code path. Mentioned here as a fourth symptom -- without Bug 3's logging it is undiagnosable from the outside.

---

## Backward compatibility

All proposed changes are default-off or additive:

- Sentinel: new gateway check, only fires on exact `<SILENT>` string -- existing agents unaffected until steering is updated
- `emptyReplyPlaceholder: false`: opt-in suppress, default `true` preserves current behaviour
- Log lines: additive only, zero behaviour change

---

## Tests suggested

- Empty-reply path: assert no message posted when `emptyReplyPlaceholder: false`
- Sentinel-reply path: assert no message posted when agent outputs `<SILENT>`
- `response_error` log emission: assert `tracing::warn!` fires for each of the three error branches

howie:
Closes #836.

Discord Discussion URL: https://discord.com/channels/1491295327620169908/1491969620754567270/1505591332457152724

## What

Three fixes for silence/empty-reply handling. All land in `AdapterRouter::stream_prompt_blocks` (`src/adapter.rs`) — the single funnel for **every** adapter — so Discord, Slack, and all gateway-relayed adapters (Feishu, Telegram, LINE, Teams, WeChat) inherit them automatically. No per-adapter changes.

## Fix 1 — `<silent />` sentinel filter

Steering can instruct agents to output exactly `<silent />` when staying silent. The gateway detects this post-loop (not per-chunk, to avoid partial-buffer false positives), deletes any streaming placeholder, and returns without posting.

```rust
if text_buf.trim() == "<silent />" {
    info!(platform = %adapter.platform(), "agent emitted <silent /> sentinel -- suppressing reply");
    if let Some(msg) = placeholder_msg.as_ref() {
        let _ = adapter.delete_message(msg).await;
    }
    return Ok(());
}
```

Addresses Bug 1: the agent now has a positive "output X" instruction instead of a negative "don't output Y", which LLMs follow far more reliably.

## Fix 2 — `emptyReplyPlaceholder` config

New `[reactions] empty_reply_placeholder` field (default `true`, fully backward-compatible). When `false`, empty replies with no error are silently suppressed instead of posting `_(no response)_`.

```toml
# per-agent config.toml (or values.yaml -> reactions.emptyReplyPlaceholder: false)
[reactions]
empty_reply_placeholder = false
```

Helm chart: `configmap.yaml` template and `values.yaml` comment updated.

## Fix 3 — diagnostic logging

Three `tracing::warn!` lines added at the three `response_error = Some(...)` sites:

| Site | Log message |
|---|---|
| Agent process died | `agent process died mid-prompt` |
| Hard timeout exceeded | `agent hard timeout exceeded` + `elapsed_s` field |
| JSON-RPC error | `agent JSON-RPC error` + `code` + `message` fields |

Each log carries `platform` field for cross-adapter filtering. Zero behaviour change. Resolves the "two agents returned `_(no response)_` with no server-side evidence" case from 2026-05-17.

## Adapter scope

| Adapter | Benefits? |
|---|---|
| Discord native (`discord.rs`) | Yes |
| Slack native (`slack.rs`) | Yes |
| Gateway-relayed: Feishu, Telegram, LINE, Teams, WeChat | Yes — all re-enter via `AdapterRouter::handle_message` |
| Future adapters | Yes — automatically |

**No changes** to `discord.rs`, `slack.rs`, `gateway.rs`, or any `gateway/src/adapters/*.rs`. Adapters remain pure transport.

## Also adds

`docs/steering/silence.md` — agent steering template explaining `<silent />` usage and the multi-mention sub-case (Bug 1B: agent must read the full mention list before deciding silence, not pattern-match on "other bot present").

## Tests added

- `silent_sentinel_passes_through_directive_parser` — `<silent />` is NOT consumed by the `[[key:value]]` directive parser
- `silent_sentinel_with_whitespace_passes_through` — whitespace/newline variants survive intact
- `empty_reply_placeholder_defaults_to_true` — backward compat regression guard
- `empty_reply_placeholder_can_be_set_to_false` — opt-in suppression round-trips through TOML

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/charts/openab/README.md
+++ b/charts/openab/README.md
@@ -39,6 +39,7 @@ Each agent lives under `agents.<name>`.
 | `reactions.enabled` | Enable status reactions. | `true` |
 | `reactions.removeAfterReply` | Remove status reactions after the agent replies. | `false` |
 | `reactions.toolDisplay` | Tool display verbosity: `full`, `compact`, or `none`. | `"full"` |
+| `reactions.emptyReplyPlaceholder` | Post `_(no response)_` for empty successful replies. Set to `false` to suppress them. | `true` |
 | `stt.enabled` | Enable voice-message speech-to-text. | `false` |
 | `stt.apiKey` | API key for the speech-to-text provider. | `""` |
 | `stt.model` | STT model name. | `"whisper-large-v3-turbo"` |
--- a/charts/openab/templates/configmap.yaml
+++ b/charts/openab/templates/configmap.yaml
@@ -162,6 +162,7 @@ data:
     [reactions]
     enabled = {{ if hasKey ($cfg.reactions) "enabled" }}{{ ($cfg.reactions).enabled }}{{ else }}true{{ end }}
     remove_after_reply = {{ if hasKey ($cfg.reactions) "removeAfterReply" }}{{ ($cfg.reactions).removeAfterReply }}{{ else }}false{{ end }}
+    empty_reply_placeholder = {{ if hasKey ($cfg.reactions) "emptyReplyPlaceholder" }}{{ ($cfg.reactions).emptyReplyPlaceholder }}{{ else }}true{{ end }}
     {{- if ($cfg.reactions).toolDisplay }}
     {{- if not (has ($cfg.reactions).toolDisplay (list "full" "compact" "none")) }}
     {{- fail (printf "agents.%s.reactions.toolDisplay must be one of: full, compact, none — got: %s" $name ($cfg.reactions).toolDisplay) }}
--- a/charts/openab/tests/empty-reply-placeholder_test.yaml
+++ b/charts/openab/tests/empty-reply-placeholder_test.yaml
@@ -1,0 +1,17 @@
+suite: emptyReplyPlaceholder
+templates:
+  - templates/configmap.yaml
+tests:
+  - it: renders empty_reply_placeholder = false
+    set:
+      agents.kiro.reactions.emptyReplyPlaceholder: false
+    asserts:
+      - matchRegex:
+          path: data["config.toml"]
+          pattern: 'empty_reply_placeholder = false'
+
+  - it: defaults empty_reply_placeholder to true
+    asserts:
+      - matchRegex:
+          path: data["config.toml"]
+          pattern: 'empty_reply_placeholder = true'
--- a/charts/openab/values.yaml
+++ b/charts/openab/values.yaml
@@ -282,6 +282,8 @@ agents:
       # compact: show count summary (e.g. ✅ 3 · 🔧 1 tool(s))
       # full: show complete tool titles (for debugging)
       # none: hide tool lines entirely
+      # emptyReplyPlaceholder: true  # set to false to silently suppress empty replies
+      #   instead of posting "_(no response)_". Pairs with <silent /> sentinel steering.
     stt:
       enabled: false
       apiKey: ""
--- a/docs/config-reference.md
+++ b/docs/config-reference.md
@@ -179,6 +179,7 @@ Emoji reaction feedback on messages to show agent processing status.
 | `enabled` | bool | `true` | Enable/disable reaction feedback. |
 | `remove_after_reply` | bool | `false` | Remove the status reaction after the agent replies. |
 | `tool_display` | string | `"full"` | How tool calls are rendered: `"full"` (complete title), `"compact"` (count summary, e.g. `✅ 3 · 🔧 1 tool(s)`), or `"none"` (hidden). |
+| `empty_reply_placeholder` | bool | `true` | Post `_(no response)_` for empty successful replies. Set to `false` to suppress empty replies silently. |
 
 ### `[reactions.emojis]`
 
@@ -331,6 +332,7 @@ Key mapping (`values.yaml` → `config.toml`):
 | `agents.<name>.pool.sessionTtlHours` | `[pool] session_ttl_hours` |
 | `agents.<name>.reactions.enabled` | `[reactions] enabled` |
 | `agents.<name>.reactions.toolDisplay` | `[reactions] tool_display` |
+| `agents.<name>.reactions.emptyReplyPlaceholder` | `[reactions] empty_reply_placeholder` |
 | `agents.<name>.stt.apiKey` | `[stt] api_key` |
 | `agents.<name>.cronjobs[].enabled` | `[[cron.jobs]] enabled` |
 | `agents.<name>.cronjobs[].schedule` | `[[cron.jobs]] schedule` |
--- a/docs/steering/silence.md
+++ b/docs/steering/silence.md
@@ -1,0 +1,31 @@
+# Silence Steering
+
+## When to Stay Silent
+
+You may receive messages addressed to other agents, or batched messages where some
+entries are not intended for you. In those cases, staying silent is the correct
+behaviour.
+
+## How to Signal Silence
+
+**Output exactly `<silent />` and nothing else.**
+
+```
+<silent />
+```
+
+The gateway detects this sentinel and suppresses the message before posting to the
+channel. No user-visible output is produced and no placeholder is left behind.
+
+**Do not** explain or justify staying silent. Do not write things like:
+- "I'm staying silent because..."
+- "This message wasn't addressed to me."
+- "No response needed."
+
+Any text other than the exact sentinel string will be posted to the channel.
+
+## Sub-case: Multiple Bot Mentions
+
+When a message mentions several bots, read the full mention list before deciding
+whether your UID is present. Do not infer "I should stay silent" solely because
+other bot UIDs appear — check whether your own UID is also in the list.
--- a/src/adapter.rs
+++ b/src/adapter.rs
@@ -478,10 +478,14 @@ impl AdapterRouter {
 
                     let mut text_buf = String::new();
                     let mut tool_lines: Vec<ToolEntry> = Vec::new();
-
-                    if reset {
-                        text_buf.push_str("⚠️ _Session expired, starting fresh..._\n\n");
-                    }
+                    // Kept separate from text_buf so the sentinel check ("is the
+                    // agent's actual output <silent />?") is not confused by the
+                    // synthetic prelude. Prepended to final_content before send.
+                    let reset_prelude = if reset {
+                        "⚠️ _Session expired, starting fresh..._\n\n"
+                    } else {
+                        ""
+                    };
 
                     // Streaming edit: send placeholder, spawn edit loop
                     let (buf_tx, placeholder_msg) = if streaming {
@@ -587,11 +591,14 @@ impl AdapterRouter {
                                 AcpEvent::Text(t) => {
                                     text_buf.push_str(&t);
                                     if let Some(tx) = &buf_tx {
-                                        let _ = tx.send(compose_display(
-                                            &tool_lines,
-                                            &text_buf,
-                                            true,
-                                            tool_display,
+                                        let _ = tx.send(format!(
+                                            "{reset_prelude}{}",
+                                            compose_display(
+                                                &tool_lines,
+                                                &text_buf,
+                                                true,
+                                                tool_display,
+                                            )
                                         ));
                                     }
                                 }
@@ -612,11 +619,14 @@ impl AdapterRouter {
                                         });
                                     }
                                     if let Some(tx) = &buf_tx {
-                                        let _ = tx.send(compose_display(
-                                            &tool_lines,
-                                            &text_buf,
-                                            true,
-                                            tool_display,
+                                        let _ = tx.send(format!(
+                                            "{reset_prelude}{}",
+                                            compose_display(
+                                                &tool_lines,
+                                                &text_buf,
+                                                true,
+                                                tool_display,
+                                            )
                                         ));
                                     }
                                 }
@@ -640,11 +650,14 @@ impl AdapterRouter {
                                         });
                                     }
                                     if let Some(tx) = &buf_tx {
-                                        let _ = tx.send(compose_display(
-                                            &tool_lines,
-                                            &text_buf,
-                                            true,
-                                            tool_display,
+                                        let _ = tx.send(format!(
+                                            "{reset_prelude}{}",
+                                            compose_display(
+                                                &tool_lines,
+                                                &text_buf,
+                                                true,
+                                                tool_display,
+                                            )
                                         ));
                                     }
                                 }
@@ -669,9 +682,14 @@ impl AdapterRouter {
                     // Sentinel: checked post-loop — chunks may transiently match mid-stream.
                     if text_buf.trim() == "<silent />" {
                         info!(platform = %adapter.platform(), "agent emitted <silent /> sentinel -- suppressing reply");
+                        reactions.suppress().await;
                         if let Some(msg) = placeholder_msg {
                             let a = adapter.clone();
-                            tokio::spawn(async move { let _ = a.delete_message(&msg).await; });
+                            tokio::spawn(async move {
+                                if let Err(e) = a.delete_message(&msg).await {
+                                    warn!(error = ?e, "delete placeholder failed after silent sentinel");
+                                }
+                            });
                         }
                         return Ok(());
                     }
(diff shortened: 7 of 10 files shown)


shaun-agent:
## OpenAB PR Screening
- Title: fix(adapter): silence handling — sentinel, empty-reply config, diagnostic logging (#836)
- Source: https://github.com/openabdev/openab/pull/842
- Generated at: 2026-05-18T02:53:23.165Z
<details>
screened PR #842, posted the marker comment, and moved project item `PVTI_lADOEFbZWM4BUUALzgtAjzk` from `Incoming` to `PR-Screening`.

GitHub comment: https://github.com/openabdev/openab/pull/842#issuecomment-4473835421  
Project action: https://github.com/orgs/openabdev/projects/1

## Intent

This PR tries to make agent silence explicit and operationally debuggable. The visible problem is that agents can currently produce confusing empty replies like `_(no response)_`, or fail silently enough that Discord, Slack, and gateway operators have little evidence about whether the agent chose silence, died, timed out, or hit JSON-RPC failure.

## Feat

Fix work. It adds a shared adapter-level `<silent />` sentinel suppression path, a backwards-compatible `empty_reply_placeholder` reactions config option, and warning logs for agent process death, hard timeout, and JSON-RPC errors. Because the changes land in `AdapterRouter::stream_prompt_blocks`, the behavior should apply to Discord, Slack, and gateway-relayed adapters without transport-specific patches.

## Who It Serves

Primary beneficiaries are Discord users and Slack users who should stop seeing unwanted empty placeholders when silence is intentional. Secondary beneficiaries are deployers and agent runtime operators who need logs that distinguish intended silence from broken agent execution.

## Rewritten Prompt

Implement shared silence handling in the OpenAB adapter router. When a complete streamed response, after trimming, is exactly `<silent />`, delete any streaming placeholder and return without posting a reply. Add a `[reactions] empty_reply_placeholder` config field that defaults to `true`; when set to `false`, suppress empty successful replies instead of posting `_(no response)_`. Add structured warning logs with `platform` for agent process death, hard timeout, and JSON-RPC error paths. Update Helm values/templates and add focused tests for sentinel parsing and config default/override behavior. Keep transport adapters unchanged unless a shared-router integration issue requires otherwise.

## Merge Pitch

This is a narrow, high-leverage adapter fix: one shared funnel improves all current and future adapters. The risk profile is moderate-low because the default empty-reply behavior remains backwards-compatible and the sentinel requires an exact trimmed match. The likely reviewer concern is whether `<silent />` can accidentally suppress legitimate user-facing output, and whether deleting placeholders is reliable across adapter implementations.

## Best-Practice Comparison

OpenClaw relevant practices are explicit delivery routing, run logs, and isolated execution evidence. This PR moves in that direction by making silence an explicit output contract and adding failure logs, but it does not add durable run logs or retry/backoff.

Hermes Agent relevant practices are fresh scheduled runs, atomic persisted state, and self-contained scheduled prompts. Those mostly do not apply because this PR is about live adapter response handling, not scheduling. The self-contained prompt idea applies lightly: `docs/steering/silence.md` makes the silence contract clearer for agent steering.

## Implementation Options

Conservative: merge the shared sentinel and config path as-is after focused review of placeholder deletion behavior and tests. This preserves the current architecture and keeps the blast radius small.

Balanced: merge this PR, then follow with a small observability PR that attaches a stable request/session id to the new warning logs and documents expected log queries for Discord, Slack, and gateway deployments.

Ambitious: replace ad hoc empty response handling with a typed response outcome model such as `Reply`, `Silent`, `Empty`, and `Error`, then route delivery, logging, placeholder cleanup, and metrics through that model. This would be cleaner long-term but larger than the current bug fix.

## Comparison Table

|---|---|---:|---|---|---|---|
| Conservative | Fast | Low | Good if exact sentinel tests pass | Good, shared router only | Immediate reduction in unwanted replies | Strong |
| Balanced | Medium | Medium | Better operator diagnosis with request correlation | Good, adds observability without redesign | Same user fix plus easier incident triage | Best |
| Ambitious | Slow | High | Best if completed well | Best long-term model, but more invasive | Cleaner semantics across adapters | Premature for this PR |

## Recommendation

Use the balanced path. Advance this PR through review as the behavior fix, with reviewer attention on exact sentinel matching, placeholder deletion, and the default-preserving config path. Split request/session correlation or richer run-log work into a follow-up so this fix can land without turning a targeted silence bug into a broader runtime redesign.
</details>

howie:
Discord Discussion URL: https://discord.com/channels/1491295327620169908/1491969620754567270/1505591332457152724

chaodu-agent:
LGTM ✅ — Clean, well-scoped fix for three related silence-handling issues. All changes land in `AdapterRouter` (the single funnel), so every adapter benefits automatically.

<details>
<summary>1. What problem does this solve?</summary>

Three silence/empty-reply bugs:
1. Agents instructed to stay silent have no reliable positive signal — negative instructions ("don't reply") are unreliable for LLMs.
2. Empty replies post `_(no response)_` even when silence is intentional.
3. Two agents returned `_(no response)_` with zero server-side evidence of what went wrong.
</details>

<details>
<summary>2. How does it solve it?</summary>

**Fix 1 — `<silent />` sentinel**: Post-loop check (not per-chunk, avoiding partial-buffer false positives). Deletes streaming placeholder, calls `reactions.suppress()`, returns early. Agent steering doc (`docs/steering/silence.md`) provides the positive instruction.

**Fix 2 — `empty_reply_placeholder` config**: New `[reactions]` field (default `true`, backward-compatible). When `false`, empty replies are silently suppressed. Helm chart updated.

**Fix 3 — Diagnostic logging**: Three `warn!` lines at the three `response_error = Some(...)` sites (process died, hard timeout, JSON-RPC error) with `platform` field for cross-adapter filtering.

**Supporting change**: `reactions.suppress()` method — removes current reaction, cancels timers, marks finished so `set_done`/`set_error` are no-ops. Prevents the done emoji from appearing on suppressed replies.
</details>

<details>
<summary>3. What was considered?</summary>

- Per-chunk sentinel detection: rejected because partial streaming buffers could transiently match.
- Separate `reset_prelude` from `text_buf`: correct — prevents the synthetic "Session expired" prefix from confusing the sentinel check.
- Slack `delete_message` added: required for placeholder cleanup on Slack (Discord already had it).
</details>

<details>
<summary>4. Traffic Light</summary>

🟢 **INFO — Correct sentinel timing**: Checking `text_buf.trim() == "<silent />"` post-loop (after all chunks received) is the right call. Mid-stream, partial buffers like `<sile` would false-match if checked per-chunk.

🟢 **INFO — Streaming suppression**: During streaming, if `text_buf.trim() == "<silent />"`, the edit loop is NOT sent the sentinel text — users never see `<silent />` flash in the placeholder. Smart.

🟢 **INFO — `suppress()` is well-tested**: Two focused async tests verify that suppress removes the reaction and blocks subsequent `set_done`/`clear` calls.

🟢 **INFO — Backward compatible**: `empty_reply_placeholder` defaults to `true`, `<silent />` is opt-in via steering, diagnostic logging is additive. Zero behavior change for existing deployments.

🟢 **INFO — Slack `delete_message`**: Clean implementation using `chat.delete` API. Required for placeholder cleanup when sentinel fires on Slack.
</details>

---

**Verdict**: Ready to merge. No blocking issues.

chaodu-agent:
## 💡 Suggestion: Align sentinel with OpenClaw convention

Prior art research shows OpenClaw uses `NO_REPLY` as their silent reply token (via a `SILENT_REPLY_TOKEN` constant + `isSilentReplyText()` helper). They had the same bug class we are fixing — see [openclaw/openclaw#8347](https://github.com/openclaw/openclaw/issues/8347).

**Reference:** OpenClaw's implementation lives in their auto-reply module (`src/agents/auto-reply/tokens.ts`), exported as `SILENT_REPLY_TOKEN` and filtered via `isSilentReplyText()`. The issue above shows the exact import path and usage pattern.

**Proposal:** Consider switching from `<silent />` to `NO_REPLY` to align with the established ecosystem convention. Benefits:

1. **Interoperability** — agents/steering prompts written for OpenClaw work unchanged on OpenAB
2. **Ecosystem familiarity** — contributors coming from OpenClaw already know the pattern
3. **Simpler matching** — no XML parsing concerns, plain string comparison

Implementation would be minimal:
```rust
/// Silent reply sentinel — aligned with OpenClaw convention.
/// See: https://github.com/openclaw/openclaw/issues/8347
pub const SILENT_REPLY_TOKEN: &str = "NO_REPLY";

// detection:
if text_buf.trim() == SILENT_REPLY_TOKEN { ... }
```

@howie — any concern with this change? The `<silent />` XML-style tag is slightly less likely to appear in natural text, but `NO_REPLY` is the de facto standard in the ecosystem. If you have a reason for preferring the XML form (e.g. multi-agent scenarios where `NO_REPLY` might appear in conversation), happy to hear it.

Either way, extracting the sentinel into a named constant (rather than inline string) would be good regardless of which token we pick.

*— 超渡法師*

chaodu-agent:
## ❌ Closing — Prefer prompt-only approach (emoji ACK)

After discussion, we prefer a simpler approach that requires **no platform code change**:

**Steer the agent to reply with 👍 when it has nothing to say.** This serves as a lightweight ACK (sender knows the agent received the message), no one gets @mentioned, and no sentinel/detection mechanism is needed.

### Why not `<silent />`?
- Silent drop gives zero delivery confidence — sender can't tell if agent chose silence or broke
- The sentinel + platform detection adds complexity for a problem solvable with prompt engineering alone
- Emoji reply is a single-line steering instruction, zero code

### What's still useful from this PR?
- **Diagnostic logging** (Fix 3) — worth extracting into a separate smaller PR
- **`empty_reply_placeholder = false`** — potentially useful, could also be a separate PR

### Recommendation
Close this PR. Extract the diagnostic logging into its own PR if desired. Silence handling = prompt-only via emoji ACK.

---
*— chaodu-agent on behalf of @pahud.hsieh*

chaodu-agent:
## ❌ Closing — Prefer prompt-only emoji ACK over platform sentinel

### Decision

After review and [discussion](https://discord.com/channels/1491295327620169908/1491365158868619404/1507120189433708594), we're closing this PR. The silence-handling problem is better solved with **prompt steering alone** — no platform code change needed.

### Approach we're going with

Steer agents to reply with a single 👍 (or similar emoji) when they receive a message but have nothing substantive to say. This gives us:

1. **Delivery confidence** — the sender sees the emoji and knows the agent received the message and consciously chose not to engage further. No ambiguity between "intentional silence" and "something broke."
2. **Zero platform complexity** — no sentinel token detection, no streaming buffer inspection, no post-loop matching logic, no new config fields. One line in the steering prompt handles it.
3. **No mention noise** — an emoji reply without @mentions doesn't ping anyone. It's low-noise by default.
4. **Transparency** — users can see the agent acknowledged. Silent drops are invisible and erode trust ("did it even see my message?").

### Why not `<silent />`?

The sentinel approach (`<silent />` → platform detects → suppresses message) solves the "zero noise" case but introduces problems:

- **No ACK** — sender has zero signal that the agent received anything. Indistinguishable from failure.
- **Platform coupling** — every adapter must implement sentinel detection. New adapters (Telegram, Teams, etc.) must remember to add it.
- **Streaming edge cases** — partial buffer matching, placeholder cleanup, reaction guard logic. The PR itself needed multiple rounds of fixes for these (session-reset prelude, streaming flash).
- **Complexity vs. value** — all this machinery to achieve something a one-line prompt instruction already handles.

### What's worth extracting

- **Diagnostic logging** (Fix 3 from this PR) — `tracing::warn!` on agent death/timeout/JSON-RPC error is independently valuable. Worth a small follow-up PR.
- **`empty_reply_placeholder = false`** — could be useful as a separate config option if edge cases arise where agents return truly empty responses.

### Summary

Silence handling = prompt engineering, not platform infrastructure. Keep it simple.

---
*— chaodu-agent on behalf of @pahud.hsieh*

PR #863, opened 2026-05-20
feat: add Antigravity CLI (agy) support

Joseph19820124:
## Summary

- Add `Dockerfile.antigravity` — downloads the native `agy` binary from Google Storage, supports `linux/amd64` and `linux/arm64`
- Add commented `[agent]` block in `config.toml.example` for `agy --acp`
- Add commented `antigravity` agent example in `charts/openab/values.yaml`
- Update `AGENTS.md` Dockerfile count 7 → 8

## Background

[Google Antigravity CLI](https://antigravity.google/product/antigravity-cli) (`agy`) is Google's terminal-first coding agent, the successor to `@google/gemini-cli`. It supports ACP via `agy --acp` and authenticates via Google Sign-In (OAuth — no API key needed).

The binary is distributed as a native pre-compiled binary from Google Storage:
```
https://storage.googleapis.com/antigravity-public/antigravity-cli/<version>/<arch>/<file>
```

## Test plan

- [ ] `docker build -f Dockerfile.antigravity .` on amd64
- [ ] `docker build -f Dockerfile.antigravity .` on arm64
- [ ] Inside container: `agy auth` → complete Google Sign-In, then `agy --acp` launches ACP session
- [ ] `helm template test charts/openab` passes with antigravity example uncommented in values.yaml

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ config.toml
 .kiro/
 CLAUDE.md
 gateway/target/
+agy-acp/.antigravitycli/
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -75,7 +75,7 @@ Never leak `DISCORD_BOT_TOKEN` or other OAB credentials to the agent.
 
 ### 4. Dockerfile Discipline
 
-There are 7 Dockerfiles: `Dockerfile`, `Dockerfile.claude`, `Dockerfile.codex`, `Dockerfile.copilot`, `Dockerfile.cursor`, `Dockerfile.gemini`, `Dockerfile.opencode`.
+There are 8 Dockerfiles: `Dockerfile`, `Dockerfile.antigravity`, `Dockerfile.claude`, `Dockerfile.codex`, `Dockerfile.copilot`, `Dockerfile.cursor`, `Dockerfile.gemini`, `Dockerfile.opencode`.
 
 A change to one MUST be evaluated against ALL. Common layers (base image, openab binary, tini) are shared — update all or explain why not.
 
--- a/Dockerfile.antigravity
+++ b/Dockerfile.antigravity
@@ -1,0 +1,46 @@
+# --- Build stage ---
+FROM rust:1-bookworm AS builder
+WORKDIR /build
+COPY Cargo.toml Cargo.lock ./
+RUN mkdir src && echo 'fn main() {}' > src/main.rs && cargo build --release && rm -rf src
+COPY src/ src/
+RUN touch src/main.rs && cargo build --release
+
+# --- Runtime stage ---
+FROM node:22-bookworm-slim
+RUN apt-get update && apt-get install -y --no-install-recommends \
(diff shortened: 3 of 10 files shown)


Joseph19820124:
Closing: verified locally that `agy --acp` does not exist in v1.0.0 (flag undefined). Need to investigate if an ACP adapter exists before re-opening.

Joseph19820124:
Reopening with a real working implementation.

**Root cause of the original close**: `agy v1.0.0` has no native `--acp` flag — the assumption was wrong.

**Fix**: Added `agy-acp/agy-acp.mjs` — a thin Node.js ACP stdio adapter that translates the openab JSON-RPC protocol into `agy --print` invocations.

**Verified locally** with end-to-end test (mock agy):
```
✓ initialize: agy v1.0.0
✓ session/new: sessionId = 15122cc8-...
✓ agent_message_chunk (prompt 1): "Hello! You said: what is 2+2?"
✓ session/prompt [1] complete
✓ agent_message_chunk (prompt 2 --continue): "[continued session] You said: and what is 3+3?"
✓ session/prompt [2] complete
✅ All ACP protocol steps passed.
```

Config is now:
```toml
[agent]
command = "node"
args = ["/usr/local/lib/agy-acp.mjs"]
```

Joseph19820124:
Closing.

shaun-agent:
Thanks for putting this together — the Antigravity ACP bridge idea is interesting, and it is useful as an experiment.

After reviewing the diff, my read is:

- The touched Rust core files (`src/acp/connection.rs`, `src/adapter.rs`, `src/discord.rs`, `src/main.rs`) appear to be formatting-only changes, not meaningful behavior changes.
- The real functional addition is the new `agy-acp/agy-acp.mjs` wrapper plus `Dockerfile.antigravity`, Helm values, and config examples.
- The main risk is not the Rust diff. It is adding another officially-supported runtime image around a custom, hand-rolled ACP bridge.

Specific concerns:

- `Dockerfile.antigravity` downloads the Antigravity CLI binary directly from Google Storage without checksum or signature verification. We do have similar direct-download risk in some existing agent images, so this is not unique to this PR, but it is still a supply-chain surface we should avoid expanding casually.
- The binary/wrapper would run in the agent pod and may have access to repo files, mounted volumes, env, GitHub auth, and Discord context depending on deployment.
- `agy-acp.mjs` only approximates ACP behavior: cancellation is a no-op, sessions are mapped to `agy --continue`, and streaming/error semantics may not match native ACP expectations.
- Adding it to the chart/config examples makes this look like an officially supported agent. That creates operational and maintenance burden, especially if Antigravity ships native ACP soon.

Recommendation: I would not merge this into `main` as-is. It is reasonable to use/test this locally or in personal infra from a branch/private image, ideally with limited tokens and a test Discord channel. For upstream, I would wait for native ACP support or a more hardened integration with checksum/provenance verification and clearer support boundaries.

So: useful experiment, but not something I think openab should support in-tree yet.

PR #874, opened 2026-05-20
feat(slack): outbound file attachments via marker syntax

angmeng:
## What

Adds outbound file-attachment support to the Slack adapter. Agents include a marker in their text output and OpenAB intercepts, uploads, and shares the file via Slack's modern files API.

```
Marker:  <<openab-send-file /absolute/path/to/file>>
```

## Why

ACP is a text-stream protocol — there's no native primitive for "agent says upload this file." Today, agents on the Slack adapter can only paste file contents inline as code blocks, or rely on the user's personal Slack OAuth via a separate MCP. Neither is a real file attachment, neither uses the bot's identity, and neither survives the claude-agent-acp env-var firewall (which deliberately strips bridge tokens from the spawned agent, preventing direct curl calls to Slack's API).

A text marker is the lightest extension that solves this without protocol changes — the agent just emits a string, the bridge does the rest.

## How

**Marker syntax**: `<<openab-send-file /absolute/path>>`. No colons — `:` collides with Slack's mrkdwn emoji parser (`:something:` renders as a gray placeholder before any interceptor runs).

**Where it fires**: Both `send_message` AND `edit_message`. OpenAB streaming uses `edit_message` for the placeholder edit; without patching that path, the marker wouldn't be intercepted during streaming responses (which is the default for solo-bot Slack threads).

**Dedup**: Streaming edits can fire repeatedly within ~200ms. Without dedup, each edit re-uploaded the file. The cache uses atomic check-and-claim under a Mutex — first claimer inserts under the lock, subsequent calls see the slot occupied and skip. Failure path removes the claim so retries aren't blocked.

**Upload flow**: Slack's modern 3-step API:
1. `files.getUploadURLExternal` (filename + length) → upload_url + file_id
2. `POST` raw bytes to the signed upload_url
3. `files.completeUploadExternal` (file_id + channel + thread_ts) → publishes the share

**Failure handling**: Posts a `⚠️ Failed to send file ...` message into the channel and surfaces the Slack error code. Caller doesn't need extra error handling.

## Constraints

- **Absolute paths only.** Bridge resolves the path as-is — no shell expansion.
- **100 MB cap per file** (sanity limit; Slack accepts up to 1 GB).
- **Requires `files:write` bot scope** on the Slack app.
- **File must be readable by the OpenAB process.** Native runs see the full host filesystem; Docker runs see only bind-mounted paths.

## Commits (recommended review order)

1. `feat(slack): send file attachments via marker` — initial implementation on `send_message` only
2. `fix(slack): rename marker to avoid Slack emoji parser collision` — `<<openab:send-file:PATH>>` → `<<openab-send-file PATH>>`
3. `fix(slack): patch edit_message too — streaming bypasses send_message` — required for streaming responses (the more common path)
4. `fix(slack): race-free dedup for file upload claims` — close the check-then-act window observed in production (duplicate uploads 224ms apart)

Each commit is self-contained and bisects cleanly. Splitting them was the result of debugging in real conditions; happy to squash before merge if preferred.

## Testing

Validated end-to-end against a personal Slack workspace (Juwaiiqi team):

- Marker in agent reply → bridge intercepts via `edit_message` → file lands in DM as real attachment (not inline code block)
- Tested with `.md` text files; should work for any MIME type Slack accepts
- Confirmed dedup: streaming edit with the same marker fires once even when the agent's response streams over multiple edit chunks
- Confirmed failure path: invalid path → `⚠️ Failed to send file ...` error message posted to channel
- Verified marker is invisible in the final delivered message (stripped before edit)

## Open questions for maintainers

1. Should this gate behind a config flag (e.g. `[reactions]`-style opt-in)? Currently always-on for any Slack adapter, but I could add `[slack].file_send_marker = true` if you'd prefer explicit enablement.
2. Marker syntax — open to alternatives if you have a preferred convention.
3. Dedup cache TTL — currently lives for the adapter's lifetime. Could add expiry, but streaming edits all happen within seconds of each other, so cache growth is bounded by unique file-send turns.

Happy to iterate on any of the above. This is patch was developed in a real working setup that's been in personal use for several days; reach out if you want test logs or repro steps.

### DIFF
--- a/src/slack.rs
+++ b/src/slack.rs
@@ -13,9 +13,13 @@ use tokio_tungstenite::tungstenite;
 use tracing::{debug, error, info, warn};
 
 /// Marker syntax for outbound file attachments in agent text output.
-/// Tifa writes `<<openab:send-file:/abs/path/to/file>>` in her response, OpenAB
+/// Tifa writes `<<openab-send-file /abs/path/to/file>>` in her response, OpenAB
 /// intercepts before posting and uploads the file via Slack's files API.
-const FILE_SEND_MARKER_PREFIX: &str = "<<openab:send-file:";
+///
+/// IMPORTANT: do NOT use colons in this marker (`:something:` would collide
+/// with Slack's emoji shortcode parser and get rendered as a gray-box
+/// placeholder even before our interceptor runs).
+const FILE_SEND_MARKER_PREFIX: &str = "<<openab-send-file ";
 const FILE_SEND_MARKER_SUFFIX: &str = ">>";
 /// Sanity cap so an agent typo doesn't try to upload /Users/jazlim or similar.
 /// Slack's own per-file limit is 1 GB by default; we cap at 100 MB.
--- a/src/slack.rs
+++ b/src/slack.rs
@@ -12,6 +12,15 @@ use tokio::sync::watch;
 use tokio_tungstenite::tungstenite;
 use tracing::{debug, error, info, warn};
 
+/// Marker syntax for outbound file attachments in agent text output.
+/// Tifa writes `<<openab:send-file:/abs/path/to/file>>` in her response, OpenAB
+/// intercepts before posting and uploads the file via Slack's files API.
+const FILE_SEND_MARKER_PREFIX: &str = "<<openab:send-file:";
+const FILE_SEND_MARKER_SUFFIX: &str = ">>";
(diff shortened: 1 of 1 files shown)


angmeng:
Closing — keeping changes in our private fork for now. May revisit later.

PR #875, opened 2026-05-20
feat(slack): inject metadata for unhandled inbound file types

angmeng:
## What

When a Slack user sends a file the adapter doesn't auto-forward (video, PDF, Office docs, archives, generic binaries), inject a metadata content block so the agent knows the file exists.

## Why

Today the Slack adapter has explicit handling for three categories of inbound files:

- **Audio** — transcribed via STT if enabled
- **Text files** — read inline (capped at 5 files / 1 MB per message)
- **Images** — base64-encoded as image content blocks

Everything else (video, PDF, Office docs, archives, generic binary) hits an unhandled else branch and gets silently dropped. The result: when the user attaches a video and asks the agent to look at it, the agent's prompt contains text but no reference to the attachment. The agent then reasons from incomplete context — sometimes apologizing for a 'missing' file that was sent successfully.

I noticed this when an agent on my fork received a `.mp4` and replied as if the user hadn't attached anything. The agent eventually worked around it by calling Slack's `files.read` API directly via a separate MCP, but that workaround isn't always available and shouldn't be necessary.

## How

Adds a final `else` clause to the file dispatch in `handle_message`. For any unhandled mimetype, build a small text content block describing the file:

```
[Slack file attachment — not auto-loaded by OpenAB]
- filename: VID-20260507-WA0001.mp4
- mimetype: video/mp4
- size: 2.2 MB
- url: https://files.slack.com/.../download/VID-20260507-WA0001.mp4

This file type isn't auto-forwarded into your context. If the contents
matter to the user's request, fetch it separately (Slack MCP's
slack_read_file, or download via curl with Bearer token if you have it).
Otherwise just acknowledge that the user shared the file.
```

The block is intentionally size-bounded — only describes the file, never embeds contents. ACP message size stays small regardless of how big the attachment is.

The included `format_size_human` helper produces friendly size strings ('2.2 MB' instead of '2306867 bytes').

## What this doesn't try to do

- **Doesn't extract text from PDFs / Office docs.** That requires extra crate dependencies and parsing logic; could be a follow-up. The metadata block is a strict improvement over silent drop.
- **Doesn't download non-text non-image binaries.** Same reasoning — the agent decides whether to fetch.
- **Doesn't change behavior for already-handled types.** Audio/text/images flow through their existing branches unchanged.

## Testing

Patched binary running in production on my macOS setup for several hours. Confirmed:

- Inbound `.mp4` now produces a metadata block (verified via debug log line `injecting unhandled-file metadata block`)
- Agent reply correctly references the attachment by filename and acknowledges receipt
- Audio/text/image flows are unaffected

## Open questions for maintainers

1. **Should the metadata format be configurable?** Currently the message is a hardcoded English template. Could move to a config-driven template if you want.
2. **Should we extend text-file handling to cover PDFs?** The existing TEXT_TOTAL_CAP / TEXT_FILE_COUNT_CAP logic could apply to PDF text extraction if a parser crate (e.g. `lopdf` or `pdf-extract`) is acceptable as a dependency.
3. **Should we add a per-mimetype allowlist instead of catch-all?** Right now any unhandled type produces metadata. Strict workspaces might prefer to silently drop unknown types.

Happy to iterate on any of the above.

### DIFF
--- a/src/slack.rs
+++ b/src/slack.rs
@@ -13,9 +13,13 @@ use tokio_tungstenite::tungstenite;
 use tracing::{debug, error, info, warn};
 
 /// Marker syntax for outbound file attachments in agent text output.
-/// Tifa writes `<<openab:send-file:/abs/path/to/file>>` in her response, OpenAB
+/// Tifa writes `<<openab-send-file /abs/path/to/file>>` in her response, OpenAB
 /// intercepts before posting and uploads the file via Slack's files API.
-const FILE_SEND_MARKER_PREFIX: &str = "<<openab:send-file:";
+///
+/// IMPORTANT: do NOT use colons in this marker (`:something:` would collide
+/// with Slack's emoji shortcode parser and get rendered as a gray-box
+/// placeholder even before our interceptor runs).
+const FILE_SEND_MARKER_PREFIX: &str = "<<openab-send-file ";
 const FILE_SEND_MARKER_SUFFIX: &str = ">>";
 /// Sanity cap so an agent typo doesn't try to upload /Users/jazlim or similar.
 /// Slack's own per-file limit is 1 GB by default; we cap at 100 MB.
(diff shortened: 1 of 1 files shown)


angmeng:
Closing — keeping changes in our private fork for now. May revisit later.

Issue #547, opened 2026-04-24
fix(docs/codex): address Internal Error (-32603) in multi-org environments


James3014:
Labels: bug

### Description
When using the Codex agent with a ChatGPT account that belongs to multiple organizations (e.g., Business and Personal), initialization fails with `Internal Error (code: -32603)` and a metadata warning: `Model metadata for gpt-4o not found.`

### Steps to Reproduce
1. Use a ChatGPT account that is part of multiple organizations.
2. Login to Codex CLI.
3. Attempt to use Codex as an agent in `openab` via `codex-acp`.
4. Send a message to the bot in Discord.

### Expected Behavior
The agent should successfully initialize and respond without being blocked by organization selection or metadata warnings.

### Environment
- OS: Darwin (macOS)
- Codex CLI: v0.124.0
- Adapter: @zed-industries/codex-acp

### Screenshots / Logs
```
ERROR openab::adapter: pool error: JSON-RPC error -32603: Internal error
Model metadata for gpt-4o not found. Defaulting to fallback metadata; this can degrade performance and cause issues.
```

### Workaround
Explicitly setting `CODEX_ACTIVE_ORGANIZATION_ID` and `CODEX_PROJECT_PATH` in the environment resolves the issue.

CHC-Agent:
## Issue at a Glance

```
  Discord msg → adapter.rs → pool.get_or_create()
    → codex-acp spawns → Codex CLI init fails
    → JSON-RPC {code: -32603, "Internal error"}
    → error_display.rs:62 → "**Internal Error**"

  Root cause: Codex CLI multi-org selection
  Fix: document CODEX_ACTIVE_ORGANIZATION_ID
```

Hi @James3014,

Confirmed. The root cause is in **upstream Codex CLI** — when a ChatGPT account belongs to multiple organizations, the CLI requires `CODEX_ACTIVE_ORGANIZATION_ID` to be set explicitly, otherwise initialization fails. openab correctly surfaces the error, but `docs/codex.md` and `Dockerfile.codex` do not mention multi-org configuration.

**Related:**
- #286 — OPEN (codex-acp session hang, same backend, different failure mode)
- #14 — CLOSED (original Codex support, multi-org not considered)

Actionable fix: add an Environment Variables section to `docs/codex.md` documenting `CODEX_ACTIVE_ORGANIZATION_ID` and `CODEX_PROJECT_PATH`. PR welcome!

CHC-Agent:
[...]

PR #885, opened 2026-05-21
fix: capture ACP agent stderr and extract error.data from JSON-RPC errors

chaodu-agent:
## Summary

Closes #854.

When an ACP agent (e.g. `codex-acp`) hits a runtime error, the real cause is emitted on **stderr** and in the JSON-RPC `error.data.message` field — but openab was discarding both. Users only saw the opaque `⚠️ Internal Error (code: -32603) / Internal error` in Discord, and operators had zero visibility in `kubectl logs`.

This PR fixes both gaps:

1. **Capture agent stderr** — pipe it and log each line at `WARN` level, scoped to the agent command name.
2. **Parse `error.data`** — extract `error.data.message` from JSON-RPC error responses and surface it in the Discord error display as a blockquote.

## Before / After

```
BEFORE (Discord):
⚠️ **Internal Error** (code: -32603)
Internal error

AFTER (Discord):
⚠️ **Internal Error** (code: -32603)
Internal error
> The 'gpt-5.2-codex' model is not supported when using Codex with a ChatGPT account.
```

## Data Flow (ASCII)

```
┌─────────────────────────────────────────────────────────────────────┐
│                        openab bridge process                        │
│                                                                     │
│  ┌──────────────┐         spawn          ┌───────────────────────┐ │
│  │ AcpConnection├────────────────────────►│  codex-acp (child)    │ │
│  │              │◄─── stdout (JSON-RPC) ──┤                       │ │
│  │              │                         │  stdout: ACP messages │ │
│  │              │◄─── stderr (logging) ───┤  stderr: diagnostics  │ │
│  └──────┬───────┘                         └───────────────────────┘ │
│         │                                                           │
│         │  JSON-RPC error response:                                 │
│         │  {"error":{"code":-32603,                                 │
│         │            "message":"Internal error",                    │
│         │            "data":{"message":"The gpt-5.2-codex..."}}}   │
│         │                                                           │
│         ▼                                                           │
│  ┌──────────────┐                                                   │
│  │ adapter.rs   │── format_coded_error(code, msg, data_message) ──► │
│  └──────────────┘                                                   │
│         │                                                           │
│         ▼                                                           │
│  ┌──────────────┐                                                   │
│  │   Discord    │  "⚠️ **Internal Error** (code: -32603)            │
│  │   channel    │   Internal error                                  │
│  │              │   > The gpt-5.2-codex model is not supported..."  │
│  └──────────────┘                                                   │
│                                                                     │
[...]

PR #895, opened 2026-05-22
fix(acp): surface JSON-RPC error data.details to users

howie:

When an ACP agent emits a JSON-RPC error such as `{code: -32603, message: "Internal error"}`, openab surfaces it as a single opaque "Internal Error (code: -32603)" — regardless of whether the actual cause is a deprecated model, an expired auth token, a missing npm peer dependency, or something else entirely.

The reason is that `JsonRpcError` deserializes only `code` and `message`, silently dropping the optional `data` field defined in [JSON-RPC 2.0 §5.1](https://www.jsonrpc.org/specification#error_object). codex-acp (and other ACP adapters) put the actionable detail string in `data.details`, but openab discards it on the wire — so users see the same message for every cause and have no signal to recover.

Related: #547 documents one specific multi-org Codex case where this surfaces. That issue proposes a docs-only fix; this PR addresses the underlying transparency gap so all -32603 sources (model deprecation, peer-dep failures, auth, multi-org, etc.) automatically surface their upstream detail.

## At a Glance

```
Before                                         After
------                                         -----
codex-acp error                                codex-acp error
{                                              {
  code: -32603,                                  code: -32603,
  message: "Internal error",                     message: "Internal error",
  data: {details: "model 'X' deprecated"}        data: {details: "model 'X' deprecated"}
}                                              }
        |                                              |
        v                                              v
openab JsonRpcError (drops data)               openab JsonRpcError {code, message, data}
        |                                              |
        v                                              v
format_coded_error(code, message)              format_coded_error(&err)
        |                                              |
        v                                              v
**Internal Error** (code: -32603)              **Internal Error** (code: -32603)
Internal error                                 Internal error
                                               model 'X' deprecated
```

**OpenClaw — [openclaw/acpx](https://github.com/openclaw/acpx)** treats `data.details` as a first-class field. acpx normalizes ACP errors through dedicated modules ([`error-shapes.ts`](https://github.com/openclaw/acpx/blob/main/src/acp/error-shapes.ts), [`error-normalization.ts`](https://github.com/openclaw/acpx/blob/main/src/acp/error-normalization.ts), [`session-control-errors.ts`](https://github.com/openclaw/acpx/blob/main/src/acp/session-control-errors.ts)) that:
- Carry `data: unknown` on every payload and walk it for typed hints (auth required, session not found, query closed).
- Distinguish causes that share `-32603` via `acp.data.details` string matching.
- Surface the detail string in user-facing wrappers (e.g., `formatSessionControlAcpSummary` reads `data?.details` and renders `${details} (ACP ${code}, adapter reported "${message}")`).

**Hermes Agent — [NousResearch/hermes-agent](https://github.com/NousResearch/hermes-agent)** is the *agent* side and emits errors through the upstream `acp` Python SDK ([`acp_adapter/server.py`](https://github.com/NousResearch/hermes-agent/blob/main/acp_adapter/server.py)). The SDK preserves `data` end-to-end, so client implementations that read it (acpx) see the agent's diagnostic context. The lesson: ACP's wire format already supports this — clients only need to deserialize the field.

**JSON-RPC 2.0 spec** — §5.1 defines `data` as an optional member that "MAY be omitted" but "SHOULD" carry "additional information about the error" when present. Dropping it is a conformance gap.

## Proposed Solution

Minimal, surgical change in four files:

1. `src/acp/protocol.rs` — Add `data: Option<serde_json::Value>` to `JsonRpcError` (with `#[serde(default)]` for back-compat), plus a `data_details()` accessor that mirrors the codex-acp / acpx convention (`data.details` as a string).
2. `src/error_display.rs` — Change `format_coded_error` to take `&JsonRpcError`, append `data.details` when present, and de-duplicate when `message` already contains the same text (some adapters echo `details` into `message` for legacy clients).
3. `src/adapter.rs` — Update the single call site at the response-error handler.
4. `src/acp/connection.rs` — Add `data: None` to the synthetic "connection closed" error.

Tests cover: deserialize round-trip with and without `data`, non-string `details` (ignored), unknown data shapes (ignored), empty/whitespace `details`, and the de-duplication path. Existing `format_coded_error` tests are migrated to the new signature.

## Why this approach?

- **Minimal blast radius.** One struct gains an optional field; one function signature changes; two call sites update. No new modules, no taxonomy.
- **Backward compatible on the wire.** `#[serde(default)]` means agents that don't emit `data` deserialize unchanged. `Display` impl preserves existing output when `data.details` is absent.
- **Addresses all known -32603 causes at once.** Today CLAUDE.md gotchas already document three distinct -32603 root causes (model deprecation, npx peer-dep failure, auth) — all of them look identical to users. After this change, each one carries its own actionable detail.
- **Convention-aligned.** Matches acpx's `data.details` string convention exactly, so future cross-tool diagnostics stay consistent.

Known limitations:
- Only `data.details` (string) is surfaced. acpx additionally recognises `data.authRequired`, `data.methodId`, `data.methods` for richer auth flows — out of scope here, can be added incrementally without breaking this change.
- The Discord/Slack message stays plain text; we don't try to format `data` as JSON when `details` is missing or non-string (silently ignored). This is conservative — avoids leaking implementation-defined payloads.

## Alternatives Considered

1. **Mirror acpx in full (`error-normalization`, `error-shapes`, `NormalizedOutputError`, retryable classification).** Rejected: acpx is a CLI that exits with structured codes — openab is a chat adapter that streams text. The full taxonomy is value for CLI consumers but overhead here, and bundles unrelated concerns (retry policy, exit codes) into one PR.
2. **Patch upstream codex-acp to echo `data.details` into `message`.** Rejected: the wire format already supports `data`; the bug is openab dropping it. Fixing it client-side keeps the change in one repo and benefits every ACP adapter, not just codex-acp.
3. **Render full `data` as JSON in user output.** Rejected: leaks implementation detail and risks confusing users with adapter-internal fields. `data.details` is the documented convention.
4. **Docs-only fix per #547.** Insufficient: documents one specific multi-org case but leaves the structural problem in place — every other -32603 cause still surfaces the same opaque label.

## Validation

- [x] `cargo check` — clean
- [x] `cargo test` — **373 passed**, 0 failed. 11 new tests cover:
  - `acp::protocol`: deserialize with/without `data`, non-string details, `Display` impl
  - `error_display`: append details, message-only fallback, dedup, unknown shapes, whitespace
- [x] `cargo clippy --all-targets` on the changed files — clean. (2 pre-existing warnings in `src/cron.rs` exist on `origin/main` and are unrelated to this PR.)

Manual: not deployed to a live agent (would need a codex-acp instance hitting a known error); the format paths are exercised end-to-end by the new tests including the model-deprecation scenario from CLAUDE.md gotchas.

### DIFF
--- a/src/acp/connection.rs
+++ b/src/acp/connection.rs
@@ -246,6 +246,7 @@ pub(crate) async fn run_reader_loop<R, W>(
             error: Some(crate::acp::protocol::JsonRpcError {
                 code: -1,
                 message: "connection closed".into(),
+                data: None,
             }),
             params: None,
         });
--- a/src/acp/protocol.rs
+++ b/src/acp/protocol.rs
@@ -55,11 +55,32 @@ pub struct JsonRpcMessage {
 pub struct JsonRpcError {
     pub code: i64,
     pub message: String,
+    // Optional diagnostic payload per JSON-RPC 2.0 §5.1. ACP adapters
+    // (codex-acp in particular) put actionable detail in `data.details`
+    // — without it, every -32603 surfaces to the user as the same opaque
+    // "Internal Error" regardless of cause (model deprecation, auth
+    // failure, missing peer dep, …). See `data_details()`.
+    #[serde(default)]
+    pub data: Option<Value>,
+}
+
+impl JsonRpcError {
+    /// Returns the `data.details` string if `data` is an object with a string
+    /// `details` field. Mirrors the codex-acp / acpx convention:
+    /// <https://github.com/openclaw/acpx/blob/main/src/acp/error-normalization.ts>
+    pub fn data_details(&self) -> Option<&str> {
+        self.data.as_ref()?.get("details")?.as_str()
+    }
 }
 
 impl std::fmt::Display for JsonRpcError {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        write!(f, "JSON-RPC error {}: {}", self.code, self.message)
+        match self.data_details() {
+            Some(d) if !d.is_empty() && !self.message.contains(d) => {
+                write!(f, "JSON-RPC error {}: {} ({})", self.code, self.message, d)
+            }
+            _ => write!(f, "JSON-RPC error {}: {}", self.code, self.message),
+        }
     }
 }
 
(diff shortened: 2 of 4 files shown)


chaodu-agent:
Closing as duplicate of #885 which was merged yesterday and addresses the same issue (extracting error.data from JSON-RPC errors).

PR #925, opened 2026-05-26
feat(relay): cross-channel relay marker (Slack ↔ Discord)

angmeng:
## Summary

Adds the `<<openab-relay-to ...>>` / `<<openab-relay-end>>` marker pair from Tifa's 2026-05-26 proposal. Lets an agent emit a single message that posts a body to another platform's channel, alongside its normal reply on the origin channel.

Builds on the line-anchored marker design from `c39d451` (this morning's fix) and reuses that defensive pattern: relay markers must occupy their own line so the agent can quote them in docs/source code without self-triggering.

## Scope

- New `src/relay.rs` (651 lines): marker parser + dispatch logic
- `AdapterRouter`: optional `relay_ctx` field — `None` short-circuits everything (zero overhead for users who don't opt in)
- Streaming display mask: in-progress relay blocks hidden from placeholder edit loop across all 3 streaming sites (Text, ToolStart, ToolDone)
- Post-stream extraction runs **before** `compose_display` + `split_message`, so multi-line blocks aren't torn by the 4000/2000-char chunker
- Peer-side: looks up adapter via `Weak::upgrade`, applies prefix template, strips file-send markers (security: prevents Slack-targeted relay from triggering uploads), peer-side splits before sending
- `config.toml.example`: `[relay]` + `[relay.aliases]` documented, commented-out, disabled-by-default

## Config

```toml
[relay]
enabled = true
prefix_template = "[via {sender_persona} from {sender_platform}]"

[relay.aliases]
"discord:swat-team" = { channel_id = "1485156316480803018" }
"slack:tifa-dm"     = { channel_id = "D0B4JKKM9PD" }
```

Length-capped at 200 bytes; alias keys must contain `:` (platform:alias format). Validation at startup; invalid config bails before bridge serves traffic.

## Test plan

- [x] `cargo test` — all 409 tests pass (+29 from this branch)
- [x] `cargo build --release` clean, no warnings
- [ ] Manual smoke test in production:
  - Tifa says `<<openab-relay-to discord:swat-team>>body<<openab-relay-end>>` in Slack → body appears in Discord channel with `[via Tifa from slack]` prefix; origin Slack reply also sent
  - Bad alias → `⚠ relay to ...` notice appears in origin channel only, no relay
  - Bare `<<openab-relay-to ...` quoted inline (e.g. in code block) → preserved as literal, no dispatch

## Test plan — not done

- [ ] Real-world smoke test (requires runtime config alias + restart)
- [ ] Performance check under load (zero-overhead claim verified by inspection but not benchmarked)

## Backwards compatibility

Default config has no `[relay]` section → `relay_ctx` is `None` → every relay code path short-circuits. Bridges that don't opt in see zero behavior change.

### DIFF
--- a/config.toml.example
+++ b/config.toml.example
@@ -137,6 +137,10 @@ tables = "code"  # "code" (default) | "bullets" | "off"
 [reactions]
 enabled = true
 remove_after_reply = false
+tool_display = "none"   # none (default) | compact | full
+                        # none: hide tool calls — only the final agent reply is shown
+                        # compact: one-line count summary, e.g. ✅ 3 · 🔧 1 tool(s)
+                        # full: every tool call rendered inline (verbose; useful for debugging)
 
 [reactions.emojis]
 queued = "👀"
--- a/config.toml.example
+++ b/config.toml.example
@@ -179,3 +179,20 @@ error_hold_ms = 2500
 # schedule = "0 0 * * 0"
 # channel = "123456789"
 # message = "generate weekly status report"
+
+# --- Cross-channel relay (Slack ↔ Discord) ---
+# When enabled, agents can emit a `<<openab-relay-to {platform}:{alias}>> ...
+# <<openab-relay-end>>` block in their response, and OpenAB forwards the body
+# to the configured peer adapter. Disabled by default — no behaviour change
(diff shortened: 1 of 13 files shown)


angmeng:
Opened against wrong base — meant for angmeng/openab fork, not upstream. Closing and re-opening on fork.

PR #958, opened 2026-05-31
fix(openab-agent): use XML format for skills prompt

chaodu-agent:
## Summary

The current `format_skills_prompt` uses a markdown bullet list format that LLMs struggle to parse reliably. The agent does not know when or how to load skills because the instruction text is too vague ("when you need") and the path is buried in markdown formatting.

This PR aligns our skills prompt with the industry-standard XML format used by all major coding agents.

Discord Discussion URL: https://discord.com/channels/1490282656913559673/1510656502283767819

## Changes

- Switch from markdown bullet list to structured XML `<available_skills>` format
- Add explicit `<name>`, `<description>`, `<location>` elements per skill
- Strengthen instruction text: "Before replying, scan the skills below. If one matches or may be relevant to the user's task, use the `read` tool to load the full SKILL.md at the listed location and follow its instructions."
- Update test to verify XML format output

## Prior Art

Researched how 4 major agents handle skill prompt injection:

| Agent | Prompt Format | Loading Mechanism | Key Instruction |
|-------|--------------|-------------------|-----------------|
| **[Pi](https://github.com/badlogic/pi-mono)** | XML `<available_skills>` with `<location>` | Agent uses `read` tool on path | "Use the read tool to load a skill's file when the task matches its description" |
| **[OpenCode](https://opencode.ai/docs/skills/)** | XML `<available_skills>` in skill tool description | Dedicated `skill({ name })` tool | Implicit in tool schema |
| **[OpenClaw](https://docs2.openclaw.ai)** | XML `<available_skills>` with `<location>` | Agent uses `read` tool on path | "use `read` to load the SKILL.md at the listed location" |
| **[Hermes](https://hermes-agent.nousresearch.com/docs/developer-guide/prompt-assembly/)** | YAML-like list + `skill_view` tool | Dedicated `skill_view(name)` tool | "Before replying, scan the skills below. If one clearly matches your task, load it with skill_view(name) and follow its instructions." |

All four use structured formats (XML or dedicated tools) rather than markdown. All include explicit trigger conditions and action instructions. Our previous implementation had neither.

## Before (markdown, weak instruction)

```
## Available Skills

The following skills are available. Use the `read` tool to load the full SKILL.md when you need a skill's instructions.

- **brave-search** (/path/to/SKILL.md): Web search via Brave Search API
```

## After (XML, strong instruction)

```xml
## Skills

[...]

PR #963, opened 2026-06-01
fix(openab-agent): escape XML special chars in skills prompt

shaun-agent:
## Summary

Builds on #958. Adds XML escaping for user-controlled `name` and `description` fields in the skills prompt.

## Problem

Skill names and descriptions come from SKILL.md frontmatter (user-controlled). Characters like `<`, `>`, `&` would break the XML structure introduced in #958, causing LLMs to misparse skill boundaries.

## Fix

- Add `xml_escape()` helper (`&` → `&amp;`, `<` → `&lt;`, `>` → `&gt;`)
- Apply to `skill.name` and `skill.description` in `format_skills_prompt`
- Path left unescaped (filesystem paths don't contain XML specials in practice)
- Added test: `format_skills_prompt_escapes_xml_chars`

## Note

Could not run `cargo test` in CI-less environment (no C linker available), but the logic is trivial and the test assertions are straightforward. CI on this PR will confirm.

cc @chaodu-agent — this is a follow-up fix for your PR.

### DIFF
--- a/openab-agent/src/skills.rs
+++ b/openab-agent/src/skills.rs
@@ -99,6 +99,11 @@ fn parse_frontmatter(content: &str) -> Option<(String, String)> {
     Some((name, description))
 }
 
+/// Escape XML special characters in user-controlled strings to preserve prompt structure.
+fn xml_escape(s: &str) -> String {
+    s.replace('&', "&amp;").replace('<', "&lt;").replace('>', "&gt;")
+}
+
 /// Format skills as a system prompt section listing available skills.
 pub fn format_skills_prompt(skills: &[Skill]) -> String {
     if skills.is_empty() {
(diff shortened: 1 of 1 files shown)


shaun-agent:
Requesting review from staff. This is a minimal follow-up to #958 adding XML escaping for user-controlled fields.

The fix is 3 lines of logic + 1 test. Ready for merge once CI passes and a maintainer approves.

shaun-agent:
Status from #958 reviewer: this is close, but not ready to merge yet.

Blocking gap: `<location>` is still emitted unescaped. The PR body says filesystem paths do not contain XML specials in practice, but skill directory names are user/repo-controlled and Unix paths can contain `&`, `<`, and `>`. Since `<location>` is also an XML text node, it needs the same escaping as `<name>` and `<description>`.

Please apply `xml_escape()` to `skill.path.join("SKILL.md").display().to_string()` as well, and add the regression assertion for a path like `/skills/a&b/SKILL.md` producing `/skills/a&amp;b/SKILL.md`. The current test also does not cover a closing-tag sequence like `</description>`, which is the prompt-boundary breakage I called out on #958; please add that too.

I tried to leave this as a formal request-changes review, but GitHub blocks that because this PR is authored by `shaun-agent`. I will keep #958 at CHANGES_REQUESTED until #963 includes escaping for all three text fields, CI is green, and #963 is merged into `fix/skills-prompt-format`.

PR #964, opened 2026-06-01
fix(agy-acp): persist line-count delta baseline

chenhan-agent:

`agy-acp` currently reconstructs restored sessions with a `conversation_id` but without any persisted delta baseline. That means the first prompt after restart or in-memory eviction can replay previously emitted assistant output when `agy --conversation ...` returns cumulative visible stdout.

This change persists an emitted line-count baseline alongside the conversation binding so restored sessions can skip already-emitted lines instead of always starting from an empty baseline.

Closes #

Discord Discussion URL: https://discord.com/channels/1491295327620169908/1508850210506281182

## At a Glance

```text
before
  OpenAB session -> agy conversation_id persisted
               restart / eviction
                    |
                    v
            restore conversation_id only
                    |
                    v
          prev baseline resets to empty
                    |
                    v
   first resumed agy stdout can replay old output

after
  OpenAB session -> agy conversation_id + emitted_line_count persisted
               restart / eviction
                    |
                    v
      restore conversation_id + line-count baseline
                    |
                    v
     skip previously emitted lines from cumulative stdout
```

**OpenClaw:**

I could not find an OpenClaw component that solves this exact problem because it appears to work with structured message history rather than diffing cumulative terminal stdout. For example, its Telegram E2E driver reads normalized `messages` arrays from chat history instead of reconstructing deltas from prior CLI output:
https://github.com/openclaw/openclaw/blob/main/scripts/e2e/telegram-user-driver.py

**Hermes Agent:**

Hermes persists ACP sessions in a database and restores full structured conversation history on reconnect / restart, so it similarly does not need to recover a delta cursor from cumulative CLI stdout:
https://github.com/NousResearch/hermes-agent/blob/main/acp_adapter/session.py

The underlying limitation here is that `agy --conversation` currently behaves like cumulative visible-output replay rather than returning only the new turn. That makes `agy-acp` maintain its own downstream emission baseline.

## Proposed Solution

- Persist `emitted_line_count` in `sessions.json` next to the existing `conversation_id`
- Restore both values during `session/load` and lazy restore after in-memory eviction
- Replace full-text baseline tracking in memory with line-count tracking
- Extract deltas by skipping the already-emitted line count from cumulative stdout
- Keep the existing conservative fallback: if the current stdout has fewer lines than expected, emit the full text and reset the line-count baseline
- Cover the new behavior with unit tests for persistence, restore, and line-count-based delta extraction

## Why this approach?

This is a pragmatic improvement for the currently observed `agy` behavior:

- the persisted state is small (`usize` instead of prior full stdout)
- restart / eviction can continue from a persisted baseline instead of always replaying from zero
- line-based tracking is less sensitive to exact byte-for-byte prefix matching when cumulative stdout is otherwise append-only at line granularity

Tradeoffs and known limitations:

- this assumes `agy` output is append-only at line granularity
- if `agy` appends within the same line, or rewrites earlier content without reducing total line count, this approach can still miss newly visible output
- it is therefore a practical tradeoff for the currently observed behavior, not a perfect delta extractor

## Alternatives Considered

1. **Persist full `prev_output` / full emitted stdout**
   - Stronger append-only verification, but higher memory / disk cost and more state churn as conversations grow.

2. **Keep current in-memory-only `prev_output`**
   - Simpler, but restart / eviction still lose the baseline and replay prior output on the next resumed turn.

3. **Ask `agy` for delta-only or structured output**
   - Likely the best long-term solution, but not available today.

## Validation

- [x] `cargo check` passes
- [x] `cargo clippy` clean
- [x] Manual testing — previously reproduced the motivating cumulative-output behavior directly with `agy --conversation`; this branch itself was validated through crate-local tests covering persisted restore and line-count delta extraction

### DIFF
--- a/agy-acp/src/main.rs
+++ b/agy-acp/src/main.rs
@@ -177,15 +177,6 @@ impl Adapter {
         }
     }
 
-    fn last_n_lines(text: &str, line_count: usize) -> String {
-        if line_count == 0 || text.is_empty() {
-            return String::new();
-        }
-        let lines: Vec<&str> = text.split_inclusive('\n').collect();
-        let start = lines.len().saturating_sub(line_count);
-        lines[start..].concat()
-    }
-
     fn extract_delta(
         emitted_line_count: usize,
         full_text: &str,
@@ -200,9 +191,9 @@ impl Adapter {
         }
         eprintln!(
             "[agy-acp] WARN: agy stdout line count shrank; \
-             sending only the last 5 lines and resetting line-count baseline"
+             sending full output and resetting line-count baseline"
         );
-        Self::last_n_lines(full_text, 5)
+        full_text.to_string()
     }
 
     fn evict_if_needed(&mut self) {
@@ -571,20 +562,14 @@ mod tests {
 
     #[test]
     fn test_extract_delta_returns_full_when_line_count_shrinks() {
-        let result = Adapter::extract_delta(7, "l1\nl2\nl3\nl4\nl5\nl6\n", true);
-        assert_eq!(result, "l2\nl3\nl4\nl5\nl6\n");
+        let result = Adapter::extract_delta(2, "fresh response", true);
+        assert_eq!(result, "fresh response");
     }
 
     #[test]
     fn test_extract_delta_preserves_leading_spaces() {
         let result = Adapter::extract_delta(1, "hello\n  indented code", true);
         assert_eq!(result, "  indented code");
-    }
-
-    #[test]
-    fn test_extract_delta_returns_all_lines_when_fewer_than_five_on_fallback() {
-        let result = Adapter::extract_delta(4, "l1\nl2\nl3\n", true);
-        assert_eq!(result, "l1\nl2\nl3\n");
     }
 
     #[test]
--- a/agy-acp/src/main.rs
+++ b/agy-acp/src/main.rs
@@ -177,6 +177,15 @@ impl Adapter {
         }
     }
 
+    fn last_n_lines(text: &str, line_count: usize) -> String {
+        if line_count == 0 || text.is_empty() {
+            return String::new();
+        }
+        let lines: Vec<&str> = text.split_inclusive('\n').collect();
+        let start = lines.len().saturating_sub(line_count);
+        lines[start..].concat()
+    }
+
     fn extract_delta(
         emitted_line_count: usize,
         full_text: &str,
@@ -191,9 +200,9 @@ impl Adapter {
         }
         eprintln!(
             "[agy-acp] WARN: agy stdout line count shrank; \
-             sending full output and resetting line-count baseline"
+             sending only the last 5 lines and resetting line-count baseline"
         );
-        full_text.to_string()
+        Self::last_n_lines(full_text, 5)
     }
 
     fn evict_if_needed(&mut self) {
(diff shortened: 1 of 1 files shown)


shaun-agent:
## OpenAB PR Screening
- Title: fix(agy-acp): persist line-count delta baseline
- Source: https://github.com/openabdev/openab/pull/964
- Generated at: 2026-06-01T12:05:01.308Z
<details>
done. comment posted and project item moved to `PR-Screening`.

## Intent

This PR is trying to prevent `agy-acp` from replaying already-emitted assistant output after an OpenAB restart or in-memory session eviction. The operator-visible problem is duplicate resumed output when `agy --conversation ...` returns cumulative stdout but OpenAB only restored the `conversation_id`, not the downstream delta cursor.

## Feat

Fix. It persists an `emitted_line_count` baseline next to the `agy` conversation binding, restores that baseline on session load/lazy restore, and computes new output by skipping previously emitted lines from cumulative stdout. Changed file: `agy-acp/src/main.rs` (+77/-48).

## Who It Serves

Agent runtime operators and Discord users. Operators get less noisy resumed sessions after restarts/evictions; users avoid seeing old assistant output replayed into a thread.

## Rewritten Prompt

Update `agy-acp` session persistence so restored conversations retain their stdout emission cursor. Persist `emitted_line_count` with each stored `conversation_id`, restore it during explicit `session/load` and lazy restore after in-memory eviction, and use that count to extract only newly visible stdout from cumulative `agy --conversation` output. Preserve the existing conservative fallback when the current stdout has fewer lines than the stored baseline, and add focused unit tests for persistence, restore, and delta extraction behavior.

## Merge Pitch

This should move forward because it fixes a real restart/eviction correctness bug with a small state addition and localized code change. The risk profile is moderate-low. Main reviewer concern: line-count tracking can miss same-line updates or rewritten earlier output.

## Best-Practice Comparison

OpenClaw and Hermes Agent do not map directly because they appear to rely on structured message/session state rather than reconstructing downstream deltas from cumulative CLI stdout. The useful comparison is durable state: this PR persists the minimal cursor `agy-acp` currently needs. Stronger long-term design is structured/delta-only output from `agy`.

## Implementation Options

1. Conservative: merge the line-count baseline as-is.
2. Balanced: merge it, plus verify old `sessions.json` compatibility and add/confirm a regression test for entries without `emitted_line_count`.
3. Ambitious: change `agy` to return structured turn deltas or stable event IDs.

## Comparison Table

|---|---|---|---|---|---|---|
| Conservative | Fast | Low | Better than current | Simple | Stops common replay | Good |
| Balanced | Fast-medium | Low-medium | Better upgrade safety | Clearer persistence contract | Same fix, lower risk | Best |
| Ambitious | Slow | High | Strongest | Cleaner long-term | Best behavior | Later |

## Recommendation

Use the balanced path. Advance this PR to review, but ask Masami/Pahud to verify old `sessions.json` compatibility and add/confirm a regression test for entries that only contain `conversation_id`.

GitHub comment: https://github.com/openabdev/openab/pull/964#issuecomment-4592398331  
Project action: item `PVTI_lADOEFbZWM4BUUALzguX1Q4` is now `PR-Screening` in https://github.com/orgs/openabdev/projects/1
</details>

chenhan1218:
now implemented in #965

PR #991, opened 2026-06-03
feat(gateway): support inbound LINE images

### LINKED ISSUE #738: feat: support non-image binary file attachments inbound (PDFs, office docs, video) on Slack/Discord
### Description

Inbound non-image, non-audio, non-text-extension files (PDFs, .docx, .xlsx, .pptx, video, archives, etc.) are silently dropped by the Slack and Discord adapters. The agent receives the user's text prompt with no indication that an attachment was even sent.

Current adapter routing on `main` (after #291):

- `image/*` → resize + base64 → `ContentBlock::Image` ✅
- `audio/*` → STT transcript → `ContentBlock::Text` ✅
- text-extension files → inline content → `ContentBlock::Text` ✅
- **everything else → silently dropped** ❌

This was previously raised in #254 ("support non-image file attachments (download to disk)"), where maintainers laid out a hybrid design and a security checklist. That issue was closed with `Close in favor of #652`, but #652 is the **outbound** images/files docs PR (`docs: add sendimages.md and sendfiles.md`), which doesn't address inbound binary handling. The inbound gap remains in `main` as of v0.8.3-beta.2.

Requesting we revive the design from #254 and land a `download_to_disk` fallback for binary attachments.

### Use Case

Production deployment running `openab-claude` as a Slack bot for an internal team. Users routinely drop `.xlsx` reports, `.docx` drafts, `.pdf` documents, and `.mp4` screen recordings into bot threads.

The bot has Anthropic's official document skills (`docx`, `pdf`, `pptx`, `xlsx`) plus a video-analysis skill mounted, with their runtime dependencies (LibreOffice, pandoc, Tesseract, ffmpeg) installed in the container. The skills can process every one of these formats — but openab never tells the agent the file exists, so the skills never trigger.

Concrete user-facing failure: user uploads `report.xlsx` and asks `@bot summarize this`. Bot responds with something like "I don't see any attachment in this thread." This is misleading and forces users to manually paste content as text or share via out-of-band channels.

This isn't a niche workflow, drag-and-drop file upload is the default mental model for Slack/Discord users. Every non-image attachment hitting an openab-driven bot today silently goes nowhere.

### Proposed Solution

Implement the design from #254: a `download_to_disk` fallback in `src/media.rs`, used when a file isn't audio / inline-text / image. Returns a `ContentBlock::Text` with a structured "[Attachment received]" header pointing to the on-disk path, so agents with file-reading tools can pick it up.

### Reference implementation (POC)

A working POC has been running in production for several days. Branch:

https://github.com/ShinyChang/openab/tree/feat/download-to-disk-attachments

Diff: https://github.com/openabdev/openab/compare/main...ShinyChang:openab:feat/download-to-disk-attachments

Touches three files:

- `src/media.rs` — adds `download_to_disk()` + `sanitize_path_component()`
- `src/slack.rs` — replaces image-only fallback with image-OR-disk (bucket = `ts`)
- `src/discord.rs` — same shape (bucket = `msg.id`)

Files land at `${OPENAB_ATTACHMENTS_DIR:-/tmp/openab-attachments}/<bucket_id>/<filename>`. Tested in production with .xlsx / .pdf / .mp4 attachments via Slack — agent receives the path block, triggers the matching skill, processes the file end-to-end.

### Known gaps vs the #254 security checklist

The POC ships fast but does not yet meet the bar maintainers proposed in #254:

- [x] Filename sanitization (path separators, null bytes, control chars)
- [ ] Path containment check after resolve (prevent symlink / `..` escape)
- [ ] SSRF guard on download URL (block private IP ranges)
- [ ] Cleanup after `stream_prompt` returns (collect `PathBuf`s, `remove_dir_all`)
- [ ] `<<<EXTERNAL_UNTRUSTED_CONTENT>>>` markers around file path block
- [ ] Config-driven (`[<adapter>.attachments]` TOML section vs current env-var only)
- [ ] 200 MB hardcoded cap → should be configurable

Hybrid (small-text inline vs disk-fallback) is already covered by the existing `is_text_file` branch landing first in the routing chain, so no new logic needed there.

chenhan-agent:

LINE image messages were previously ignored by the gateway because the LINE adapter only accepted `message.type = text`. That meant OpenAB lost user intent whenever a LINE user sent a photo or screenshot.

Closes #738.

Discord Discussion URL: https://discord.com/channels/1491295327620169908/1502854039019454486

## At a Glance

```text
LINE image webhook
  -> gateway validates signature
  -> gateway records reply-token age at webhook receipt
  -> gateway returns HTTP 200 quickly
  -> bounded post-ack worker downloads LINE-hosted image
  -> image is resized/compressed + stored via existing media pipeline
  -> gateway emits GatewayEvent with attachment
  -> OpenAB replies through existing Reply/Push path
```

**OpenClaw:**
- Docs describe acknowledging signed LINE webhooks first, then continuing processing asynchronously:
  - https://docs.openclaw.ai/channels/line
- Code shows async webhook handling plus LINE media download:
  - https://github.com/openclaw/openclaw/blob/286e5ffe0767a0bd3325b0cbae40218a19e38eab/extensions/line/src/webhook.ts
  - https://github.com/openclaw/openclaw/blob/286e5ffe0767a0bd3325b0cbae40218a19e38eab/extensions/line/src/bot-handlers.ts
  - https://github.com/openclaw/openclaw/blob/286e5ffe0767a0bd3325b0cbae40218a19e38eab/extensions/line/src/download.ts

**Hermes Agent:**
- Docs say LINE supports inbound media and prefers reply tokens before push fallback:
  - https://github.com/mudrii/hermes-agent-docs/blob/de42acf6e148a34a63edd3b0afbe2ad6e5b68f90/user-guide/messaging/line.md
- Code shows LINE media download via `api-data.line.me`, reply-token caching, and webhook dedupe:
  - https://github.com/NousResearch/hermes-agent/blob/84710995efabc7ec2e75c373e1baee75d779cbc3/plugins/platforms/line/adapter.py

- LINE webhook docs:
  - https://developers.line.biz/en/docs/messaging-api/receiving-messages/
- LINE Reply API docs:
  - https://developers.line.biz/en/docs/messaging-api/sending-messages/#reply-messages
- LINE Push API docs:
  - https://developers.line.biz/en/docs/messaging-api/sending-messages/#push-messages

## Proposed Solution

- Accept inbound LINE `message.type = image` events in addition to text.
- Download LINE-hosted image content from `api-data.line.me` using the existing `LINE_CHANNEL_ACCESS_TOKEN`.
- Reuse the existing gateway media pipeline (`resize_and_compress`, `store_media`, `Attachment`) instead of adding LINE-specific storage logic.
- Record reply-token age from webhook receipt time rather than post-download/post-processing time.
- Return HTTP 200 quickly, then process each webhook payload asynchronously.
- Bound concurrent post-ack LINE payload processing with a semaphore so image bursts do not create unbounded background work.
- Skip unsupported `contentProvider.type = "external"` images with a warning rather than fabricating placeholder text.
- Document the new support, current limitations, and early-ack tradeoffs in `docs/line.md` and `docs/inbound-attachments.md`.

## Why this approach?

This keeps the webhook path short enough for LINE while still preserving OpenAB's existing attachment model and free Reply API first / Push fallback behavior.

Tradeoffs accepted here:
- Early-ack lowers slow-webhook redelivery pressure.
- A later crash can still lose in-flight post-ack work.
- An earlier image webhook can still reach OpenAB after a later text webhook from the same chat if the image path is slower.
- This PR intentionally does **not** add a durability layer, a per-channel sequencing queue, or a broader LINE media expansion beyond images.

## Alternatives Considered

- **Keep image processing synchronous in the webhook path** — rejected because download/compression in-request increases slow-ack and redelivery risk.
- **Add durability / duplicate suppression in the same PR** — rejected to keep the first inbound image PR focused.
- **Emit placeholder text for failed image downloads** — rejected because it invents user intent and misrepresents what the user actually sent.
- **Support all LINE media types in one PR** — rejected to keep scope tight and land image support first.
- **Keep the split-pod shared-volume chart workaround in this PR** — rejected; remote validation showed the canonical same-pod/sidecar deployment path works, so the chart workaround was reverted back out of this branch.

## Validation

- [x] `cargo check` passes
- [x] `cargo clippy` clean
- [x] Manual testing — validated image download, attachment creation, oversize rejection, early HTTP 200 behavior, reply-token handling, and the final branch head against the remote dev LINE bot environment.

## Additional Review

Reviewed against local `main` in up to 3 rounds with:
- Codex (`gpt-5.3-codex`)
- Gemini Pro (`gemini-3.1-pro-preview`)
- Copilot (`gpt-5.4`, xhigh-style pass)

That review directly led to follow-up fixes for:
- hard-capping LINE image body reads incrementally instead of fully buffering the response
- documenting the cross-webhook reordering tradeoff introduced by early-ack image processing
- qualifying shared docs as `LINE-hosted only` for current image support
- bounding concurrent post-ack LINE webhook processing

### DIFF
--- a/charts/openab/templates/gateway.yaml
+++ b/charts/openab/templates/gateway.yaml
@@ -4,7 +4,6 @@
 {{- $gwCfg := omit $cfg "nameOverride" }}
 {{- $d := dict "ctx" $ "agent" (printf "%s-gateway" $name) "cfg" $gwCfg }}
 {{- $agentD := dict "ctx" $ "agent" $name "cfg" $cfg }}
-{{- $pvcEnabled := and $cfg.persistence $cfg.persistence.enabled }}
 {{- $hasTeams := and (($cfg.gateway).teams).appId (($cfg.gateway).teams).appSecret }}
 {{- $hasTelegram := (($cfg.gateway).telegram).botToken }}
 {{- $hasLine := (($cfg.gateway).line).channelSecret }}
@@ -46,10 +45,6 @@ spec:
               containerPort: 8080
               protocol: TCP
           env:
-            {{- if $pvcEnabled }}
-            - name: HOME
-              value: {{ $cfg.workingDir | default "/home/agent" | quote }}
-            {{- end }}
             {{- if ($cfg.gateway).token }}
             - name: GATEWAY_WS_TOKEN
               valueFrom:
@@ -241,14 +236,8 @@ spec:
           resources:
             {{- toYaml . | nindent 12 }}
           {{- end }}
-          volumeMounts:
-            {{- if $pvcEnabled }}
-            - name: data
-              mountPath: {{ $cfg.workingDir | default "/home/agent" }}
-            {{- end }}
-            - name: tmp
-              mountPath: /tmp
           {{- with ($cfg.gateway).extraVolumeMounts }}
+          volumeMounts:
             {{- toYaml . | nindent 12 }}
           {{- end }}
       {{- with ($cfg.gateway).extraContainers }}
@@ -266,15 +255,8 @@ spec:
       tolerations:
         {{- toYaml . | nindent 8 }}
       {{- end }}
-      volumes:
-        {{- if $pvcEnabled }}
-        - name: data
-          persistentVolumeClaim:
-            claimName: {{ (and $cfg.persistence $cfg.persistence.existingClaim) | default (include "openab.agentFullname" $agentD) }}
-        {{- end }}
-        - name: tmp
-          emptyDir: {}
       {{- with ($cfg.gateway).extraVolumes }}
+      volumes:
         {{- toYaml . | nindent 8 }}
       {{- end }}
 ---
--- a/charts/openab/templates/gateway.yaml
+++ b/charts/openab/templates/gateway.yaml
@@ -4,6 +4,7 @@
 {{- $gwCfg := omit $cfg "nameOverride" }}
 {{- $d := dict "ctx" $ "agent" (printf "%s-gateway" $name) "cfg" $gwCfg }}
 {{- $agentD := dict "ctx" $ "agent" $name "cfg" $cfg }}
+{{- $pvcEnabled := and $cfg.persistence $cfg.persistence.enabled }}
 {{- $hasTeams := and (($cfg.gateway).teams).appId (($cfg.gateway).teams).appSecret }}
 {{- $hasTelegram := (($cfg.gateway).telegram).botToken }}
 {{- $hasLine := (($cfg.gateway).line).channelSecret }}
@@ -45,6 +46,10 @@ spec:
               containerPort: 8080
               protocol: TCP
           env:
+            {{- if $pvcEnabled }}
+            - name: HOME
+              value: {{ $cfg.workingDir | default "/home/agent" | quote }}
+            {{- end }}
             {{- if ($cfg.gateway).token }}
             - name: GATEWAY_WS_TOKEN
               valueFrom:
@@ -236,8 +241,14 @@ spec:
           resources:
             {{- toYaml . | nindent 12 }}
           {{- end }}
-          {{- with ($cfg.gateway).extraVolumeMounts }}
           volumeMounts:
+            {{- if $pvcEnabled }}
+            - name: data
+              mountPath: {{ $cfg.workingDir | default "/home/agent" }}
+            {{- end }}
+            - name: tmp
+              mountPath: /tmp
+          {{- with ($cfg.gateway).extraVolumeMounts }}
             {{- toYaml . | nindent 12 }}
           {{- end }}
       {{- with ($cfg.gateway).extraContainers }}
@@ -255,8 +266,15 @@ spec:
       tolerations:
         {{- toYaml . | nindent 8 }}
       {{- end }}
-      {{- with ($cfg.gateway).extraVolumes }}
       volumes:
+        {{- if $pvcEnabled }}
+        - name: data
+          persistentVolumeClaim:
+            claimName: {{ (and $cfg.persistence $cfg.persistence.existingClaim) | default (include "openab.agentFullname" $agentD) }}
+        {{- end }}
+        - name: tmp
+          emptyDir: {}
+      {{- with ($cfg.gateway).extraVolumes }}
         {{- toYaml . | nindent 8 }}
       {{- end }}
 ---
--- a/charts/openab/tests/gateway_test.yaml
+++ b/charts/openab/tests/gateway_test.yaml
@@ -357,44 +357,6 @@ tests:
               secretName: cloudflared-creds
         documentIndex: 0
 
-  - it: mounts shared data pvc and writable tmp when persistence is enabled
-    set:
-      agents.kiro.gateway.enabled: true
-      agents.kiro.gateway.url: "ws://openab-gateway:8080/ws"
-      agents.kiro.persistence.enabled: true
-    asserts:
-      - contains:
-          path: spec.template.spec.containers[0].env
-          content:
-            name: HOME
-            value: /home/agent
-        documentIndex: 0
-      - contains:
-          path: spec.template.spec.containers[0].volumeMounts
-          content:
-            name: data
-            mountPath: /home/agent
-        documentIndex: 0
-      - contains:
-          path: spec.template.spec.containers[0].volumeMounts
-          content:
-            name: tmp
-            mountPath: /tmp
-        documentIndex: 0
-      - contains:
-          path: spec.template.spec.volumes
-          content:
-            name: data
-            persistentVolumeClaim:
-              claimName: RELEASE-NAME-openab-kiro
-        documentIndex: 0
-      - contains:
-          path: spec.template.spec.volumes
-          content:
-            name: tmp
-            emptyDir: {}
-        documentIndex: 0
-
   - it: renders nodeSelector when set
     set:
       agents.kiro.gateway.enabled: true
--- a/charts/openab/tests/gateway_test.yaml
+++ b/charts/openab/tests/gateway_test.yaml
@@ -357,6 +357,44 @@ tests:
               secretName: cloudflared-creds
         documentIndex: 0
 
+  - it: mounts shared data pvc and writable tmp when persistence is enabled
+    set:
+      agents.kiro.gateway.enabled: true
+      agents.kiro.gateway.url: "ws://openab-gateway:8080/ws"
+      agents.kiro.persistence.enabled: true
+    asserts:
+      - contains:
+          path: spec.template.spec.containers[0].env
+          content:
+            name: HOME
+            value: /home/agent
+        documentIndex: 0
+      - contains:
+          path: spec.template.spec.containers[0].volumeMounts
+          content:
+            name: data
+            mountPath: /home/agent
+        documentIndex: 0
+      - contains:
+          path: spec.template.spec.containers[0].volumeMounts
+          content:
+            name: tmp
+            mountPath: /tmp
+        documentIndex: 0
+      - contains:
+          path: spec.template.spec.volumes
+          content:
+            name: data
+            persistentVolumeClaim:
+              claimName: RELEASE-NAME-openab-kiro
+        documentIndex: 0
+      - contains:
+          path: spec.template.spec.volumes
+          content:
+            name: tmp
+            emptyDir: {}
+        documentIndex: 0
+
   - it: renders nodeSelector when set
     set:
       agents.kiro.gateway.enabled: true
--- a/docs/inbound-attachments.md
+++ b/docs/inbound-attachments.md
@@ -25,7 +25,7 @@ User sends media (photo/voice/file)
 | **Feishu** | ✅ | ✅ (STT) | ✅ (whitelist) | skipped | skipped |
 | **Google Chat** | ✅ | ✅ (STT) | ✅ (whitelist) | skipped | Drive files skipped |
 | **WeCom** | ✅ | — | ✅ (whitelist) | skipped | skipped |
-| **LINE** | ✅ | — | — | — | — |
+| **LINE** | ✅ (LINE-hosted only) | — | — | — | — |
 | **Slack** | ✅ | ✅ (STT) | ✅ | — | skipped |
 
 ## Processing Pipeline
--- a/docs/inbound-attachments.md
+++ b/docs/inbound-attachments.md
@@ -25,7 +25,7 @@ User sends media (photo/voice/file)
 | **Feishu** | ✅ | ✅ (STT) | ✅ (whitelist) | skipped | skipped |
 | **Google Chat** | ✅ | ✅ (STT) | ✅ (whitelist) | skipped | Drive files skipped |
 | **WeCom** | ✅ | — | ✅ (whitelist) | skipped | skipped |
-| **LINE** | planned | planned | — | — | — |
+| **LINE** | ✅ | — | — | — | — |
 | **Slack** | ✅ | ✅ (STT) | ✅ | — | skipped |
 
 ## Processing Pipeline
--- a/docs/line.md
+++ b/docs/line.md
@@ -87,6 +87,7 @@ In the LINE Developers Console → **Messaging API** tab → scan the QR code wi
 - **Webhook signature validation** — HMAC-SHA256 via `LINE_CHANNEL_SECRET`
 
 > **Implementation tradeoff:** OpenAB now acknowledges LINE webhooks before image download/processing so slow attachment work is less likely to trigger webhook redelivery. The follow-up image download and event emission happen asynchronously, which keeps the request path short but also means a crash after the HTTP 200 can still lose that in-flight work. This PR intentionally keeps scope small and does not add a separate background-task durability or duplicate-suppression layer on top of early-ack.
+> If a LINE-hosted image cannot be downloaded or decoded, OpenAB logs and skips that image event rather than synthesizing a fake text prompt.
 
 ### Not Supported (LINE API limitations)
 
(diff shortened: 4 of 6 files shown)


shaun-agent:
## OpenAB PR Screening
- Title: feat(gateway): support inbound LINE images
- Source: https://github.com/openabdev/openab/pull/991
- Generated at: 2026-06-04T00:03:41.741Z
<details>
screened PR #991 and moved the project item to `PR-Screening`.

GitHub comment: https://github.com/openabdev/openab/pull/991#issuecomment-4617731222  
Project action: `PVTI_lADOEFbZWM4BUUALzguosEg` status updated `Incoming` -> `PR-Screening`

## Intent

PR #991 adds inbound LINE image handling so LINE users can send photos or screenshots without the gateway dropping that intent. The operator-visible problem is webhook handling latency: image download and compression cannot hold the LINE HTTP request open without increasing retry/redelivery pressure.

## Feat

Feature work. The PR extends the LINE gateway adapter from text-only inbound messages to text plus image messages, downloads LINE-hosted image content through the Content API, runs it through the existing media pipeline, emits a GatewayEvent with an image attachment, and documents the supported behavior and limits.

## Who It Serves

Primary beneficiary: LINE users. Secondary beneficiaries are deployers and agent runtime operators, because early webhook acknowledgement reduces slow-request retry pressure while preserving the existing Reply API first / Push fallback path.

## Rewritten Prompt

Implement inbound LINE image support in the OpenAB gateway. Accept `message.type = image` events alongside text, validate the signed webhook, capture receipt time for reply-token age tracking, and return HTTP 200 before any image download or compression work. In a background task, download non-external LINE image content from `api-data.line.me` using `LINE_CHANNEL_ACCESS_TOKEN`, resize/compress and store it through the existing gateway media pipeline, then emit a GatewayEvent containing an attachment. Preserve event ordering within one webhook payload, avoid logging sensitive external image URLs, warn and emit a minimal `[LINE image]` marker when an attachment cannot be produced, and document supported behavior in `docs/line.md`. Add focused tests for image parsing, download failure, attachment creation, oversize rejection, early ack behavior, event forwarding, and reply-token timing.

## Merge Pitch

This should move forward because it closes a clear product gap: LINE image messages currently lose user intent entirely. The implementation is scoped to one adapter plus gateway wiring and reuses the existing media pipeline instead of adding LINE-specific storage. Risk is moderate: early ack is the right shape for LINE webhooks, but it introduces post-ack loss if the gateway crashes before the spawned task completes. The likely reviewer concern is whether the async path has enough ordering, observability, and failure handling without duplicate suppression or durable work persistence.

## Best-Practice Comparison

OpenClaw applies here. Its relevant pattern is gateway-owned webhook handling with quick acknowledgement, async media download, explicit delivery routing, and logs around the processing path. This PR aligns with early ack and async processing, but does not add durable job persistence, retry/backoff beyond the current task behavior, or duplicate suppression.

Hermes Agent also applies. Its relevant LINE pattern includes media download via `api-data.line.me`, reply-token preference before push fallback, webhook dedupe, and self-contained adapter handling. This PR aligns on the Content API domain and reply-token timing, but intentionally leaves dedupe and broader media types for follow-up.

## Implementation Options

Conservative: keep this PR scoped as-is. Land inbound LINE image support with early ack, background processing, existing media storage, focused warnings, and documentation. Track durability, dedupe, and additional media types separately.

Balanced: land image support, but require minimal production hardening before merge: structured run logs for each image event, explicit metrics or tracing around download/compress/store/emit, bounded background concurrency, and tests proving reply-token caching happens before event broadcast.

Ambitious: convert LINE inbound media handling into a durable gateway job model with persisted work items, retry/backoff, duplicate suppression, crash recovery, and support for image/audio/video/file/sticker/location in one broader adapter subsystem.

## Comparison Table

|---|---|---|---|---|---|---|
| Conservative | Fast | Low-medium | Acceptable, with post-ack crash loss | Good if follow-ups are tracked | Restores image intent quickly | Strong |
| Balanced | Medium | Medium | Better operational confidence | Stronger, because async behavior is observable | Same feature, safer rollout | Best |
| Ambitious | Slow | High | Best long-term | Risky unless split carefully | Broad LINE media support | Too large for this PR |

## Recommendation

Use the balanced path. Advance this PR to review with the current scope, but ask reviewers to focus on async failure visibility, bounded background work, event ordering within a webhook payload, reply-token timing, and no sensitive URL logging. Do not block this image feature on a full durable job layer; split duplicate suppression, retry/backoff, crash recovery, and additional LINE media types into follow-up issues if production behavior shows they are needed.
</details>

wangyuyan-agent (approved):
## Review

Read the full diff and traced through `gateway/src/adapters/line.rs`, `gateway/src/main.rs`, `media.rs`, `store.rs`, both updated docs, and the four new tests. Implementation quality is high, the OpenClaw + Hermes Agent dual-reference research shows in the design, and CI is green. Confirmed zero behavioral impact on Feishu / Telegram / Google Chat / WeCom / Teams adapters — the shared `media.rs` / `store.rs` / `schema.rs` interfaces are unchanged, and the `main.rs` lines that touch Feishu are pure `cargo fmt` line-wrapping.

Below are observations grouped by severity. None are merge blockers in my view — but two design points deserve a response before we ship.

### Design questions (please respond)

**Q1. `acquire_owned().await` runs before HTTP 200 is returned.** `adapters/line.rs` ~L101–L114. When 8 in-flight image payloads are still downloading and a 9th webhook arrives, the 9th request blocks inside the webhook handler until a permit frees, delaying the 200 OK. That partially defeats the early-ack goal: a sustained image burst can still slow the ack path enough to invite LINE redelivery.

The cleanest fix is to **move the semaphore wait inside the spawned task** — return 200 immediately, queue the work as a tokio task, and gate concurrency at processing time. The HTTP layer stays at ~1 ms ack regardless of burst. Tokio handles task backpressure; per-task overhead is well below webhook redelivery cost.

`try_acquire_owned()` + drop-on-failure is technically faster, but it discards the entire webhook payload (including any text events bundled in the same payload), which is a heavier loss than letting the image processing queue up briefly.

If `acquire-before-ack` is intentional, please add a comment explaining why this beats the spawn-internal alternative.

**Q2. `LINE_CHANNEL_SECRET = None` silently skips signature verification, with no startup warning.** This is pre-existing (predates this PR), but this PR rewrites the `webhook` handler that sits on the trust boundary, so it's a fair moment to surface it. Compare:

- `main.rs` for Telegram: `warn!("TELEGRAM_SECRET_TOKEN not set — webhook requests are NOT validated (insecure)")`
- `main.rs` for LINE: just `info!("line adapter enabled")`, no insecure-mode warning, route mounted unconditionally.

A one-line `warn!` parallel to the Telegram one would close the gap without expanding scope.

### NITs

- **Constant-time signature compare.** `adapters/line.rs` ~L93, `signature != expected` compares `&str` via `PartialEq`, not constant time. Low practical risk on a 60-byte HMAC tag, but `subtle::ConstantTimeEq` is one line and matches industry hardening defaults.
- **`dispatch_line_reply` 400 heuristic too loose on `"expired"`.** The `(invalid && reply token)` branch is precise; the bare `body_lower.contains("expired")` branch can false-positive on any unrelated 400 that happens to contain the word. Tightening to require `"reply token"` co-occurrence is safer.
- **`Err(_) => SERVICE_UNAVAILABLE` on permit acquire is dead code.** `Semaphore::acquire_owned` only fails when the semaphore is closed, and nothing in this codebase closes it. Drop or comment as defensive. (Disappears anyway under the Q1 fix.)
- **`LINE_WEBHOOK_CONCURRENCY_MAX = 8` is a compile-time constant.** Different deployments will want different ceilings. Consider an env override (`LINE_WEBHOOK_CONCURRENCY`, default 8).
- **No test for the signature-verification failure path.** All four new tests are valuable, but the trust boundary itself is uncovered: the early-ack test sets `line_channel_secret: None`, which bypasses the entire validation block. One test that mounts a non-None secret and feeds an invalid signature would close the gap.

### Follow-up issues worth opening (not blocking this PR)

- **Webhook dedupe by `message.id`.** Early-ack increases the effective probability that LINE will redeliver (network blip + ack race). Without dedupe, the same `message.id` becomes two distinct GatewayEvents, two cache entries, and the user sees the bot reply twice (Reply API on the first, Push on the second). Hermes Agent has dedupe; the PR description correctly scopes it out, but a TTL'd `HashSet<message_id>` is much lighter than a durability layer and belongs in the same family.
- **Push API quota cost when image processing is on the critical path.** `webhook_received_at` is correctly the TTL anchor, but with 10 MB downloads + LLM processing, the 50 s window is realistically exhausted before reply, forcing the Push API path. Push has a monthly quota on LINE Messaging API; the docs tradeoff section should mention this.
- **External-content image: silent skip is a UX dead-end.** Today `contentProvider.type = "external"` produces no event at all — the user sees no bot reaction whatsoever. Consider emitting a `GatewayEvent` with an `unsupported` attachment marker so OAB can decide whether to surface "external image not supported yet" to the user. Pushes the policy decision to OAB rather than swallowing it in the gateway.

### What's good

- Reply-token cache populated **before** `event_tx.send()` — the comment block spells out why; closes a real race.
- TTL reckoned from `webhook_received_at`, not post-processing.
- Three-layer download limiting: Content-Length pre-check, streaming chunk accumulator with running total, and `store_media`'s 20 MB defense-in-depth.
- External URL sanitization keeps host only, drops path/query.
- Reply API failure handling correctly distinguishes "token unusable" (fall through to Push) from "network/other error" (do not retry — duplicate risk). Matches Hermes Agent's pattern.
- `cfg(test)` covers the four behaviors that matter most: download + resize, image-event → GatewayEvent with attachment, oversize Content-Length rejection, early-ack with cache populated before broadcast.

Approving with the two design questions above. Happy to discuss either tradeoff in the thread.

chaodu-agent:
CHANGES REQUESTED ⚠️ — Solid implementation, but missing download timeouts could hold semaphore permits indefinitely under adverse conditions.

## What This PR Does

LINE image messages were previously dropped because the gateway only accepted `message.type = text`. This PR adds inbound LINE image support: download from the LINE Content API, process through the existing media pipeline, and emit a GatewayEvent with an attachment — all after an early HTTP 200 acknowledgement.

## How It Works

Webhook handler acquires a semaphore permit, returns 200, then spawns a background task that: parses the event, downloads LINE-hosted image content via `api-data.line.me`, incrementally size-checks the body, runs `resize_and_compress` + `store_media`, and emits the GatewayEvent. Reply-token age is recorded at webhook receipt time (not post-download). External-content images are warned and skipped.

## Findings

| 1 | 🟡 | Semaphore acquired before HTTP 200 — saturated permits block the webhook response, risking LINE redelivery | `line.rs:100-108` |
| 2 | 🟡 | No request/connect timeout on image download — slow Content API holds permit indefinitely | `line.rs:296-312` |
| 3 | 🟡 | No wall-clock timeout wrapping the chunked body read — slow-drip holds permit+task | `line.rs:327-339` |
| 4 | 🟢 | Reply-token timing uses webhook receipt time, correctly reflecting true token age | `line.rs:163` |
| 5 | 🟢 | No placeholder text synthesis for failed downloads — preserves user intent integrity | `line.rs:254-258` |
| 6 | 🟢 | External URL sanitization only logs host, preventing URL/query-param leakage | `line.rs:183-188` |
| 7 | 🟢 | Solid test coverage: download, oversize rejection, early-ack, reply-token caching | `line.rs:471+` |
| 8 | 🟢 | Reuses existing media pipeline instead of adding LINE-specific storage | — |

<details>

### 🟡 F1: Semaphore acquired before HTTP 200

`acquire_owned().await` blocks the webhook handler when all 8 permits are held. LINE sees a slow response and may redeliver. Consider `try_acquire_owned()` — if no permit is available, either queue internally or log a warning and still return 200 (accepting the backpressure tradeoff).

### 🟡 F2: No request timeout on image download

`client.get(...).bearer_auth(...).send().await` has no timeout. If `api-data.line.me` stalls, the permit is held forever. Suggest adding `.timeout(Duration::from_secs(30))` on the request builder or configuring the shared `reqwest::Client` with a default timeout.

### 🟡 F3: No overall timeout on chunked body read

The incremental `body.len() > IMAGE_MAX_DOWNLOAD` check guards against oversized responses, but not against slow-drip connections (e.g., 1 byte/sec staying under the size limit). Wrapping the entire download in `tokio::time::timeout(Duration::from_secs(60), ...)` would bound the worst case.

### 🟢 F4–F8: Praise

Reply-token timing, no-placeholder design, URL sanitization, test coverage, and media pipeline reuse are all well-executed.

</details>

<details>

- PR opened: 2026-06-04
- Main already has: LINE text-only webhook handler, media pipeline (`resize_and_compress`, `store_media`, `Attachment` schema), reply-token cache with hybrid Reply/Push dispatch
- Net-new value: LINE inbound image download via Content API, early-ack webhook pattern with bounded concurrency, incremental body size guard, external-image skip logic, documentation updates

</details>

<details>

- Correct reply-token age tracking from webhook receipt time
- Principled refusal to synthesize placeholder text for failed image downloads
- External URL sanitization prevents log injection / credential leakage
- Comprehensive test coverage including async behavior verification
- Clean reuse of existing media pipeline — no LINE-specific storage bolt-on
- Well-documented tradeoffs in both code comments and user-facing docs
- Proper `LINE-hosted only` qualification in shared inbound-attachments table

</details>

chaodu-agent:
All review findings addressed in 3 commits:

**`e8d4488` — NITs from first review:**
- Constant-time HMAC signature compare (`subtle::ConstantTimeEq`)
- Tightened `"expired"` 400 heuristic to require `"reply token"` co-occurrence
- Documented dead-code `Err(_) => SERVICE_UNAVAILABLE` arm as defensive
- `LINE_WEBHOOK_CONCURRENCY` env var override (default 8)
- Added `webhook_rejects_invalid_signature` test

**`e5a488c` — Download timeouts (F2, F3):**
- 30s request timeout on Content API call
- 60s wall-clock timeout wrapping chunked body read

**`708b993` — Semaphore inside spawned task (F1):**
- HTTP 200 now returned immediately regardless of permit availability
- Concurrency still bounded — spawned task waits for permit before processing

All non-blocking follow-ups (webhook dedupe, Push quota docs, external-image UX) left for future PRs as reviewers suggested.

thepagent (approved):

PR #1012, opened 2026-06-06
fix(adapter): add 120s inactivity timeout to catch mid-turn LLM hangs

### LINKED ISSUE #997: Bot silently hangs after tool call when codex-acp LLM call produces no response
## Description

After any tool call (whether it exits with success or failure), the bot stops responding to all subsequent messages in that thread. The turn never completes — the agent sends an initial text response ("I'll run that...") and then hangs indefinitely until the 30-minute hard timeout fires.

Root cause: `codex-acp` has no internal timeout on LLM API calls. When a call hangs for any reason (API timeout, network issue, rate limit, etc.), `codex-acp` produces no ACP events and no error — it blocks silently. `openab`'s recv loop has no per-event inactivity check, so it waits for the full 30-minute hard timeout while holding the connection mutex. All follow-up messages in the thread are silently dropped.

Note: `codex-acp` emits `status: "completed"` for `exec_command` regardless of exit code, so a non-zero exit (e.g. DNS failure) is indistinguishable from success at the ACP level. The hang occurs on the **subsequent LLM call**, not on the tool call itself.

## Steps to Reproduce

1. Start the bot:
   ```
   RUST_LOG=info ./target/release/openab run
   ```
2. In a Discord thread, ask the agent to run a command that will fail:
   ```
   run ping -c 1 wrongdns.com
   ```
3. Observe: the bot acknowledges the command ("I'll run the command as requested...") and then **goes silent** — no response for 9+ minutes.
4. Send another message in the same thread:
   ```
   what's status
   ```
5. Observe: **no reply** — the bot is stuck waiting indefinitely.

## Expected Behavior

After a tool call (success or failure), the bot should:
1. Complete the turn promptly instead of hanging for the full 30-minute hard timeout.
2. Report an error if the agent stops responding mid-turn.
3. Allow subsequent messages in the thread to be processed normally.

## Environment

- OS: Ubuntu (WSL2 on Windows)
- Agent: `codex-acp`
- Discord: web

## Screenshots

**Bot acknowledges at 12:44 PM then goes silent (9-minute gap)**

At 12:44 PM the bot says "I'll run the command as requested..." and then nothing. The user had to retry at 12:53 PM.

bot hangs after ping failure

## Suggested Fix

Add a **120-second ACP inactivity timeout** to the recv loop in `src/adapter.rs`:

- Track `last_acp_event`, reset it on every incoming ACP notification (text chunk, tool start/done, usage update, etc.).
- On each liveness-check tick, if `last_acp_event.elapsed() > 120s`, abandon the turn early, surface an error to the user, and set `force_recreate = true` so the next request spawns a fresh process instead of resuming the stuck session.

This covers both hang types with a single mechanism:

| Scenario | Without fix | With fix |
|---|---|---|
| `exec_command` non-zero exit (`status: "completed"`) → LLM call hangs | Silent for 30 min | Error after 120s |
| Tool auth failure (`status: "failed"`) → LLM call hangs | Silent for 30 min | Error after 120s |

The timeout value (120s) should be configurable via the pool config, with 0 meaning disabled.

fcwu:

After any tool call (success or failure), `codex-acp` makes an LLM API call to generate its next response. When the session context is large (~258k tokens), this call silently hangs — no ACP events arrive, no error is returned. The connection mutex is held for the entire turn, so **all subsequent messages in the same thread are silently dropped** until the 30-minute hard timeout fires.

Closes #997

Discord Discussion URL: 

## At a Glance

```
Discord message
      │
      ▼ dispatch_batch → with_connection (holds conn mutex for entire turn)
            │
            ▼ session/prompt sent
            │
            ├─ tool_call_update (any status)
            │
            └─ agent calls LLM to generate next response
                    │
                    ▼  ← hangs here (large context = silent timeout)
                    │
             [120s of no ACP events]
                    │
                    ▼
             inactivity timeout fires → ⚠️ error sent to user
             force_recreate = true → fresh process next turn
```

Not applicable — targeted bug fix adding an inactivity timeout to an existing recv loop.

## Proposed Solution

Three coordinated changes:

1. **120-second ACP inactivity timeout** (`src/adapter.rs`, `src/config.rs`): track `last_acp_event` (reset on every incoming ACP notification). On each liveness-check tick, if `last_acp_event.elapsed() > 120s`, cut the turn early and surface `⚠️ Agent stopped responding (no activity for 120s)`. Configurable via `[pool] acp_inactivity_timeout_secs` (default 120, set to 0 to disable).

2. **`force_recreate` flag** (`src/acp/connection.rs`, `src/acp/pool.rs`): when the inactivity timeout fires, mark the connection so the next `get_or_create` spawns a fresh process with `session/new` instead of reloading the bloated session that caused the hang.

3. **Capture `rawOutput.aggregated_output`** (`src/acp/protocol.rs`): surface failed-tool output in the hard-timeout error message for diagnostics.

## Why this approach?

The inactivity timeout is the minimal targeted fix: it doesn't change session management or retry logic, it only adds a per-turn watchdog. `force_recreate` prevents the next turn from loading the same oversized session that caused the hang. Together they ensure the bot recovers within ~130s instead of 30 minutes, with no change to the happy path.

## Alternatives Considered

- **Reduce hard timeout** — would help but 30 min is needed for legitimately long tasks; doesn't solve the silent-hang case.
- **Timeout in codex-acp** — correct long-term fix but requires upstream change; this is a host-side workaround.

## Validation

- [x] `cargo build --release` passes
- [x] `cargo test` passes
- [x] Manual test: sent `run ping -c 1 wrongdns.com` → DNS fails → agent LLM call hangs → bot replies `⚠️ Agent stopped responding (no activity for 120s)` at `agent_dispatch_ms=129446` (~129s) → follow-up messages processed normally with fresh session

**Test screenshot:**

Agent stopped responding after 120s


fcwu:
**Test screenshot** — bot replied `⚠️ Agent stopped responding (no activity for 120s)` after `ping -c 1 wrongdns.com` failed and the subsequent LLM call hung:

Logs confirm: `agent_dispatch_ms=129446` (~129s total turn time, timeout fired at ~120s mark)

fcwu:
**Manual test screenshot** — `run ping -c 1 wrongdns.com` → DNS fails → LLM call hangs → bot fires 120s inactivity timeout:

⚠️ Agent stopped responding (no activity for 120s)

Log: `agent_dispatch_ms=129446` (~129s total turn)

shaun-agent:
## OpenAB PR Screening

- Title: fix(adapter): add 120s inactivity timeout to catch mid-turn LLM hangs
- Source: https://github.com/openabdev/openab/pull/1012
- Generated at: 2026-06-06T10:10:00Z

<details>

## Intent

Recover OpenAB threads when an ACP agent stops emitting events mid-turn, especially after a tool call followed by a large-context LLM call. The operator-visible problem is that a thread can silently stall for up to the hard timeout while the connection mutex blocks subsequent messages.

## Feat

Fix. The PR adds a configurable ACP inactivity timeout, marks stuck connections for recreation, and surfaces raw failed-tool output in timeout diagnostics. It targets runtime reliability rather than adding a new feature surface.

## Who It Serves

Discord users, agent runtime operators, and maintainers debugging long-running ACP turns. Codex-backed deployments benefit most because the reported failure mode is a silent post-tool LLM hang in large sessions.

## Rewritten Prompt

Add a per-turn ACP inactivity watchdog to OpenAB. Track the timestamp of the last received ACP event during `session/prompt`; if no events arrive for `pool.acp_inactivity_timeout_secs` seconds, return a clear user-facing timeout, mark the connection for forced recreation, and ensure the next turn starts a fresh agent process/session instead of reloading the stuck context. Include failed-tool raw output in diagnostics where available. Add configuration defaults, disable behavior at `0`, and cover timeout/recreate behavior with focused tests.

## Merge Pitch

This should move forward because it reduces a 30-minute silent failure into a bounded, understandable recovery path. The risk is concurrency-sensitive: timeout handling must not leak subscribers, poison pooled connections, or break legitimate long-running turns that still stream events. Review should focus on lock lifetime, `force_recreate` semantics, and whether 120 seconds is the right default.

## Best-Practice Comparison

OpenClaw strongly applies: isolated executions, explicit run logs, and retry/backoff all point toward bounded failure recovery instead of a permanently wedged session. Hermes Agent also applies: fresh sessions for scheduled or stuck tasks and clear run-state transitions are consistent with the proposed `force_recreate` path.

## Implementation Options

1. Conservative: Lower the existing hard timeout or add more logging only. Fast, but still leaves users waiting too long and may not recover the poisoned context.
2. Balanced: Add inactivity timeout plus forced connection recreation. This PR's approach directly targets the silent-hang failure and preserves normal long turns that keep producing events.
3. Ambitious: Introduce durable job/run tracking with per-turn state, cancellation, retries, and user-visible run logs across adapters. Better long term, but too broad for the immediate bug.

## Comparison Table

|---|---:|---:|---:|---:|---:|---:|
| Hard-timeout/logging only | High | Low | Low | High | Low | Weak |
| Inactivity timeout + recreate | Medium | Medium | High | Medium | High | Best |
| Durable run system | Low | High | Very high | Medium | High | Future follow-up |

## Recommendation

Proceed with the balanced watchdog if tests cover event reset, timeout firing, forced recreation, disabled timeout, and error cleanup paths. Consider a follow-up to expose the timeout in Helm values/docs and to add metrics for inactivity timeouts by agent backend.

</details>

chaodu-agent:
Thank you for the thorough investigation and fix, @fcwu. The root-cause analysis is solid and the PR is well-structured.

After discussion, we've decided **not to merge this** for now. Our reasoning:

1. **Root cause belongs upstream** — the silent hang is a codex-acp bug (LLM API call with no timeout on large contexts). The proper fix is for codex-acp to enforce its own timeout and emit an error event back to the host.

2. **Host-side workaround adds risk** — the inactivity watchdog, `force_recreate` flag, and associated pool logic increase complexity and introduce edge cases (e.g., legitimate silent periods being misidentified as hangs).

**Suggested next steps:**

- Please open an **issue** in this repo to track the problem (silent mid-turn hang on large context). If it gets **10+ upvotes** from the community we will reconsider adding a host-side mitigation.
- Report the underlying bug to **codex-acp** upstream — their `session/prompt` flow should timeout and return an error event rather than hanging indefinitely.

We appreciate the contribution and the detailed validation. 🙏

PR #1056, opened 2026-06-10
fix(slack): dedupe native-stream finalize + render final as Block Kit markdown

### LINKED ISSUE #1055: bug(slack): assistant-mode native streaming posts the entire reply twice in single-bot threads
### Summary

With `assistant_mode = true` (the default), a Slack bot that is **alone in a thread** posts its **entire reply twice**, concatenated with no separator. Asking it to "say hello world" yields:

```
Hello world!Hello world!
```

The duplication disappears as soon as a second bot is present in the thread.

### Reproduction

1. A Slack app with `assistant_mode = true` (default), as the only bot in a thread.
2. Mention it: `@bot say hello world`.
3. The reply renders as `Hello world!Hello world!` — the full reply doubled.
4. Add a second bot to the thread and mention either bot again → replies are now single/correct.

It reproduces for **every** reply (not specific to tables or length) when the bot is the only one in the thread.

### Expected Behavior

The bot posts its reply **once** — a single message containing the streamed content, with no duplicated/concatenated copy — regardless of whether it is the only bot in the thread. (Single-bot native-streaming threads should behave the same as multi-bot threads, which already post a single clean reply.)

### Root cause

Native-streaming finalization in `src/slack.rs` `stream_finish` delivers the complete reply a second time:

- During streaming, each text delta is sent via `chat.appendStream`, which **appends** to the message (per the docs: *"this text is what will be appended to the message received so far"*). This builds the full reply incrementally.
- At finish, `stream_finish` calls `chat.stopStream` with the **full** `final_content` in `markdown_text`:

  ```rust
  let body = serde_json::json!({
      "channel": msg.channel.channel_id,
      "ts": ts,
      "markdown_text": final_content,
  });
  self.api_post("chat.stopStream", body).await
  ```

  `chat.stopStream`'s `markdown_text` **also appends** (it does not replace the already-streamed content), so the entire reply is delivered a second time → `…world!` + `Hello world!`.

The current code assumes `stopStream markdown_text` *replaces* the streamed content — e.g. `stream_append` treats an `appendStream` error as *"cosmetic; final replace will correct"*, and `stream_finish`'s comment says *"Finalize content atomically via stopStream's markdown_text parameter instead of a separate chat.update"*. But Slack's streaming methods are append-only; the documented pattern calls `stop()` to **close** the stream without re-sending content (optionally attaching feedback blocks).

### Why only single-bot threads

`uses_native_streaming()` = `assistant_mode && !other_bot_present`. When another bot is in the thread, the adapter falls back to the **post+edit** path, whose finalize uses `chat.update` — which **replaces** — so there is no duplication. The bug therefore only manifests in the common single-bot case.

### Suggested fix

`stream_finish` (active/native path) should not re-append the full reply at `stopStream`. Either:

- **(preferred)** close the stream with `chat.stopStream` (no `markdown_text`), then set the final/cleaned content with a single `chat.update` (which replaces) — this preserves the cleaned final (e.g. table conversion / error prefix) while avoiding the double-append; or
- stop without re-sending content, accepting the streamed text as final.

### Affected

Every single-bot `assistant_mode` reply (the default configuration). Introduced with the native-streaming work in #995.

dogzzdogzz:

In `assistant_mode` (default), a Slack bot **alone in a thread** posts its **entire reply twice** (concatenated) — e.g. `say hello world` → `Hello world!Hello world!`. The reply is streamed live via `chat.appendStream` (which **appends**), then `stream_finish` re-sends the **full** content via `chat.stopStream`'s `markdown_text` — which **also appends**, not replaces — so the whole reply lands a second time. It only manifests in single-bot threads; with another bot present the adapter falls back to the post+edit path, whose finalize uses `chat.update` (replace).

Separately, finalized replies were sent as legacy `text` mrkdwn, which flattens `#` headings to bold, strips code-fence language tags, and cannot render tables.

Closes #1055

Discord Discussion URL: https://discord.com/channels/1491295327620169908/1491969620754567270/1514088560364355667

## At a Glance

```
              Before (buggy)                        After (this PR)
  ┌──────────────────────────────┐       ┌──────────────────────────────┐
  │ stream:  appendStream (append)│       │ stream:  appendStream (append)│
  │ finish:  stopStream(markdown_ │       │ finish:  stopStream(close,    │
  │   text = FULL) → appends again│       │   no markdown_text)           │
  │   → reply posted TWICE        │       │   + chat.update(Block Kit)    │
  │                               │       │   → REPLACE: single copy      │
  │ render:  flat text mrkdwn     │       │ render:  native headings,     │
  │   (headings→bold, no tables)  │       │   lists, code fences, TABLES  │
  └──────────────────────────────┘       └──────────────────────────────┘
   (single-bot threads only; multi-bot already used post+edit)
```

## Proposed Solution

**Fix the duplication** — `stream_finish` no longer re-sends content at `stopStream`:

1. `chat.stopStream { channel, ts }` to **close** the stream only (no `markdown_text`). The reply was already delivered live via `appendStream`.
2. `chat.update` to **replace** the message with the finalized content (a single clean copy). The active and degraded paths are unified on this final replace; the `chat.update`→`postMessage` fallbacks are preserved.

**Render the final richly** — `send_message` / `edit_message` send Block Kit `markdown` blocks (raw Markdown) instead of legacy `text` mrkdwn, so Slack renders real headings, lists, blockquotes, language-tagged code fences, and **tables natively**. A `text` fallback is kept for notifications/accessibility, with a graceful **text-only retry** on `invalid_blocks` / `msg_blocks_too_long`.

- `message_limit` 4000 → **11,900** (`MARKDOWN_BLOCK_LIMIT`): fits the per-block 12k cap, keeps Markdown tables in one block (no mid-table split), and cuts message-spam on long replies. Long content still splits via `format::split_message` (code-fence safe).
- New `ChatAdapter::renders_native_tables()` (**default `false`**, Slack `true`): the router sets `table_mode = Off` for native-table adapters so the raw table reaches the Block Kit block. **Discord and gateway adapters are unchanged** (own send paths + default `false`).

## Validation

The finalize sequence was verified directly against the live Slack Web API:

- `chat.stopStream { channel, ts }` (no `markdown_text`) → `ok: true`, closes cleanly.
- `chat.update` with `blocks: [{ "type": "markdown", "text": … }]` on the just-stopped message → **replaces** the streamed content (single copy, no duplication).
- The Block Kit `markdown` block **renders Markdown tables natively** (plus headings/lists/code).

## Testing

- `cargo test --bin openab` — passing, incl. new unit tests: Block Kit body builders preserve raw Markdown; `thread_ts` handling; 11,900 char-count splitting; long table stays in one chunk; text-only fallback bodies; `is_block_payload_rejected` classifier (incl. `msg_blocks_too_long`, excl. `invalid_arguments`); `renders_native_tables` true on Slack / default false elsewhere.
- `cargo clippy --all-targets -- -D warnings` — clean for the changed files.

## Scope

Slack-only. `src/discord.rs` and `src/gateway.rs` keep their own send paths and the existing table conversion.

### DIFF
--- a/config.toml.example
+++ b/config.toml.example
@@ -140,6 +140,11 @@ tables = "code"  # "code" (default) | "bullets" | "off"
                  # code: wrap tables in fenced code blocks with aligned columns
                  # bullets: convert each row into bullet points (• Header: Value)
                  # off: pass through unchanged
+                 # NOTE: ignored on Slack — Slack messages are sent as Block Kit
+                 # `markdown` blocks (and `markdown_text` stream chunks), which
+                 # render Markdown tables (and headings, lists, code fences)
+                 # natively. This setting applies to platforms that cannot render
+                 # tables (Discord and gateway integrations).
 
 [reactions]
 enabled = true
--- a/src/adapter.rs
+++ b/src/adapter.rs
@@ -208,7 +208,9 @@ pub trait ChatAdapter: Send + Sync + 'static {
     /// Platform name for logging and session key namespacing.
     fn platform(&self) -> &'static str;
 
-    /// Maximum message length for this platform (e.g. 2000 for Discord, 4000 for Slack).
+    /// Maximum message length (chars) for this platform; the router splits longer
+    /// replies into multiple messages at this bound. Platform-specific (e.g. 2000
+    /// for Discord; Slack uses its Block Kit `markdown` block cap).
     fn message_limit(&self) -> usize;
 
     /// Send a new message, returns a reference to the sent message.
@@ -302,6 +304,16 @@ pub trait ChatAdapter: Send + Sync + 'static {
     /// Empty string clears it. Default: no-op (platforms without a status API).
     async fn set_status(&self, _channel: &ChannelRef, _status: &str) -> Result<()> {
         Ok(())
+    }
+
+    /// Whether this platform renders Markdown tables natively. When `true`, the
+    /// router skips the `convert_tables` pre-pass (which rewrites tables into
+    /// code blocks / bullet lists for platforms that cannot render them) and
+    /// lets the platform render the raw Markdown table itself.
+    /// Default: `false` (keep converting). Overridden by Slack (Block Kit
+    /// `markdown` blocks / `markdown_text` stream chunks render tables natively).
+    fn renders_native_tables(&self) -> bool {
+        false
     }
 
     /// Whether this adapter should use streaming edit (true) or send-once (false).
@@ -546,7 +558,14 @@ impl AdapterRouter {
         let streaming = adapter.use_streaming(other_bot_present);
         let native = adapter.uses_native_streaming(other_bot_present);
         let assistant_status = adapter.uses_assistant_status();
-        let table_mode = self.table_mode;
+        // Platforms that render Markdown tables natively (e.g. Slack Block Kit
+        // `markdown` blocks / `markdown_text` stream chunks) skip the
+        // table→code/bullets pre-pass so the raw table renders natively.
+        let table_mode = if adapter.renders_native_tables() {
+            TableMode::Off
+        } else {
+            self.table_mode
+        };
         let tool_display = self.reactions_config.tool_display;
         let prompt_hard_timeout = self.prompt_hard_timeout;
         let liveness_check_interval = self.liveness_check_interval;
@@ -1107,6 +1126,9 @@ mod tests {
         let adapter = TestAdapter;
         // Verify the method is callable and returns the declared value
         assert!(!adapter.use_streaming(false));
+        // renders_native_tables defaults to false: platforms that don't override
+        // it keep the table→code/bullets conversion (e.g. Discord, Gateway).
+        assert!(!adapter.renders_native_tables());
     }
 
     #[test]
--- a/src/slack.rs
+++ b/src/slack.rs
@@ -416,19 +416,30 @@ impl ChatAdapter for SlackAdapter {
     }
 
     fn message_limit(&self) -> usize {
-        4000
+        // Match the Block Kit `markdown` block cap (12k) minus headroom. Messages
+        // are sent as markdown blocks, so the old 4000 mrkdwn-era limit would
+        // split long replies (and Markdown tables) across messages needlessly —
+        // a mid-table split renders as raw pipes. 11_900 keeps typical tables in
+        // one block and cuts message-spam on long replies.
+        MARKDOWN_BLOCK_LIMIT
     }
 
     async fn send_message(&self, channel: &ChannelRef, content: &str) -> Result<MessageRef> {
-        let mrkdwn = markdown_to_mrkdwn(content);
-        let mut body = serde_json::json!({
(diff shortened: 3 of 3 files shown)


antigenius0910:
### E2E verification ✅

Built `openab-claude` from `upstream/main` (b8d5a48) and from this PR (3950856), ran them as a single bot (assistant_mode default, no other bots in thread), posted the same prompt from a real Slack user account.

| Build | Thread | Prompt | Reply `text` |
|---|---|---|---|
| `main` (b8d5a48) | `1781064773.587629` | `<@bot1> say hello world` | **`Hello, world!Hello, world!`** — #1055 reproduced |
| this PR (3950856) | `1781064863.066999` | same | **`Hello, World!`** — single copy, fix works |

**Block Kit rendering also verified** (PR's secondary claim). Asked bot1 (PR build) to echo `## Demo` + a markdown table + a `\`\`\`rust` fence. Slack server returned these blocks in `conversations.replies`:

- `{ type: "header", level: 2, text: "Demo" }` — heading is a real H2, not flattened to `*Demo*`
- `{ type: "table", rows: [["Col A", "Col B"], …] }` — native Block Kit table, no raw pipes
- `{ type: "rich_text", rich_text_preformatted, language: "rust" }` — fenced code with language preserved

The `text` fallback (`*Demo*  | Col A | Col B | …`) is also present for notifications/a11y as designed.

### Code review

LGTM — the root cause and fix sequence in `stream_finish` match the issue analysis exactly:
- `chat.stopStream { channel, ts }` (no `markdown_text`) closes the stream without re-appending.
- `chat.update` with Block Kit `markdown` blocks replaces with the finalized content, no duplication.
- On active path + `chat.update` failure: keep the streamed content (no fallback `postMessage`) — this is the correct choice; falling back would re-introduce the duplicate that the live `appendStream` already delivered.

Other reviewed details:
- `is_block_payload_rejected` correctly matches `invalid_blocks` / `msg_blocks_too_long` exactly (not as substring), and excludes the `invalid_arguments` catch-all — the test guards both the substring case and the catch-all case.
- `renders_native_tables()` opt-in (default `false`) leaves Discord and gateway adapters on the existing `convert_tables` path — scope is correctly Slack-only.
- `message_limit` 4000 → 11_900 is within the 12k per-block cap with 100-char headroom. Acknowledged follow-up: `split_message` is not table-aware, so a single >11_900 char table would still split into raw-pipe continuation chunks; rare enough at 11.9k that deferring is fine.

### Minor (non-blocking) nit

L613 — `chat.appendStream failed (cosmetic; final replace will correct)`: the comment text is identical to the pre-fix version, where "final replace" was actually a (buggy) re-append. With this PR it's accurate (the `chat.update` does replace), but reading it side-by-side with the fix could mislead future maintainers. Could clarify e.g. `appendStream is best-effort; the final chat.update replaces with text_buf, so a dropped delta is recovered`.

Happy to retest if any change comes in.

antigenius0910 (approved):

thepagent:
taking a look

thepagent (approved):

PR #1104, opened 2026-06-14
fix(slack): debounce app_mention edit re-fires to one turn per message

### LINKED ISSUE #1103: bug(slack): app_mention re-fires for every edit state → duplicate agent turns (phantom turns)
## Summary

The Slack adapter spawns a full agent turn for **every** `app_mention` event, with no dedup by message `ts`. A single logical mention, however, commonly arrives as several `app_mention` events — Slack re-delivers `app_mention` as a mention-bearing message moves through edit states, and openab's own native streaming drives a reply through several message states (`chat.startStream` → N×`appendStream` → `chat.stopStream` → final `chat.update`). The result is multiple duplicate agent invocations ("phantom turns") for one user message, each of which also forces a reply into the thread.

## Environment

- openab `main` (`src/slack.rs` `app_mention` handler)
- Slack adapter, Socket Mode
- Multi-agent thread configuration (≥2 bots), `assistant_mode` default

## Observed behavior

In a multi-agent workflow thread, a single exchange of **2 real user messages produced 5 agent turns**. The extra turns:

- share the same sender and `(channel, ts)`;
- carry token counts that track the message's growth as it settles (e.g. `75 → 90` tokens), i.e. they are successive edit states of the *same* message;
- queue behind one another (dispatch wait climbing ~`12s → 24s → 32s` as duplicates pile up);
- each emit a forced reply (often a no-op), which is pure thread noise.

One dispatch was even observed acting on a **mid-stream fragment** whose text was just `<@U…>  -` — the mention captured before it had settled.

## Root cause (`src/slack.rs`)

The `app_mention` branch of the Socket Mode loop spawns `handle_message` unconditionally for each event:

```rust
"app_mention" => {
    // ... bot gating ...
    let event = event.clone();
    // ...
    tokio::spawn(async move {
        handle_message(&event, /* ... */).await;
    });
}
```

There is no per-message dedup, so every edit-state re-fire of the same `(channel, ts)` becomes its own turn. The turn-boundary batching dispatcher (`docs/adr/turn-boundary-batching.md`) coalesces **distinct** messages that arrive during an in-flight turn, but it does not dedupe repeated edit states of the **same** message.

## Reproduction

1. A Slack workspace with a bot in a multi-agent thread (≥2 bots), `assistant_mode` default (native streaming).
2. `@mention` the bot with a message and let the reply stream (and/or edit the mention message once after sending).
3. Observe multiple `handle_message` dispatches / agent turns for the single mention, several of which emit a no-op reply.

## Impact

- Duplicate agent invocations — wasted LLM cost and latency — for one user message.
- Thread noise: each phantom turn forces a reply.
- Especially visible in multi-agent threads, where native streaming is disabled (post+edit) and a reply produces several edit states.

## Suggested fix

Debounce `app_mention` per `(channel, ts)`: hold only the latest event state, dispatch once after a short quiet window (each new state pushes the deadline), and only if the settled state's fingerprint (text + attached file IDs) differs from what was last dispatched for that message. Trailing-edge, so the turn acts on the **settled** message rather than a mid-stream fragment; identical re-fires drop, material edits re-dispatch (fail-open). A PR implementing this follows.

dogzzdogzz:
Debounce Slack `app_mention` so one user message produces one agent turn.

`src/slack.rs` spawns a full `handle_message` (→ one agent turn) for **every** `app_mention` event, with no dedup by message `ts`. But a single logical mention arrives as several events: Slack re-delivers `app_mention` as a mention-bearing message moves through edit states, and openab's own native streaming drives a reply through several message states (`chat.startStream` → N×`appendStream` → `chat.stopStream` → final `chat.update`). One observed multi-agent thread produced **5 agent turns for 2 real messages** — duplicate LLM invocations, each forcing a no-op reply into the thread, and one dispatch even fired on a mid-stream fragment whose text was just `<@U…>  -`.

Closes #1103

## At a Glance

```
(before)  every app_mention event ─▶ tokio::spawn(handle_message) ─▶ full agent turn

          edit-state re-fires of the SAME (channel, ts):
            ts=170.1  "<@U  -"           ─▶ turn 1   ⚠ mid-stream fragment
            ts=170.1  "<@U review"       ─▶ turn 2
            ts=170.1  "<@U review this"  ─▶ turn 3        1 message ⇒ N turns ⇒ N noise replies

(after)   app_mention ─▶ MentionDebouncer.note((channel, ts), latest_state)
                          each new state pushes a quiet deadline (mention_debounce_seconds, default 4s)
              quiet ─▶ dispatch ONCE, only if fingerprint changed
                          (text + sorted file IDs)  ─▶ one turn on the settled message
```

- **OpenClaw / Hermes Agent:** not applicable — this is a defect in OpenAB's own Slack Socket Mode event ingestion (`src/slack.rs`), upstream of any agent backend and independent of which ACP agent runs.
- **OpenAB turn-boundary batching** (`docs/adr/turn-boundary-batching.md`, the Dispatcher): the closest in-repo prior art. It coalesces **distinct** messages that arrive during an in-flight turn into one batch. This change is complementary and orthogonal — it dedupes **repeated edit states of the same message** at the adapter ingress, before a turn is ever dispatched, which the batching dispatcher does not cover.
- **OpenAB WeCom adapter** (`debounceSecs`): precedent for a per-adapter, seconds-valued "debounce quiet period before flushing" config — the naming/units of `mention_debounce_seconds` follow it.
- **Slack Events API:** `app_mention` delivery is at-least-once and re-fires as a message is edited (`message_changed`). Trailing-edge debounce keyed on `(channel, ts)` is the standard way to collapse such an event storm down to the message's settled state.

## Proposed Solution

Add `MentionDebouncer`, keyed on `(channel, ts)`:

- Hold only the **latest** event state per message; each new state pushes a quiet deadline.
- On quiet, dispatch **once**, and only if the settled state's fingerprint (`text` + sorted attached file IDs) differs from what was last dispatched for that message — so an attachment-only edit is still material, while an identical re-fire drops.
- The deadline check and the drain happen under **one lock** (`try_take`), closing the race where a new edit arriving mid-check would otherwise leak an unsettled state through.
- Fail-open: a late final state is never lost (a material edit re-dispatches). Dispatch history is FIFO-bounded (`MENTION_HISTORY_CAP`).

The quiet window is configurable via **`[slack] mention_debounce_seconds`** (Helm: `slack.mentionDebounceSeconds`), **default 4**. Set it lower to trade robustness for latency, or `0` to dispatch on the first event (effectively disabling debounce). The window defaults to `MENTION_DEBOUNCE` (4s) when unset.

Trailing-edge is deliberate: the first event state of a streamed/edited message is often a mid-stream fragment, and responding to fragments is worse than a few seconds of latency on turns that already take tens of seconds.

## Why this approach?

The bug is at the adapter ingress (one Slack message → many events), so the fix belongs there, before dispatch — not in the cross-platform batching Dispatcher, whose job (coalescing distinct messages at turn boundaries) is a different concern. Keying on `(channel, ts)` matches Slack's own identity for "the same message across its edit states". The change is contained to the Slack adapter plus a single, default-preserving config knob.

## Alternatives Considered

- **Leading-edge (dispatch the first event, ignore the rest):** rejected — the first state is often a mid-stream fragment (observed: text just `<@U…>  -`); responding to fragments is worse than trailing latency.
- **Dedup in the batching Dispatcher:** the Dispatcher batches *distinct* messages; same-`ts` edit states are not its model, and pushing per-message edit dedup down there would entangle a Slack-specific event quirk with the platform-agnostic batching logic.
- **Dedup by Slack `event_id` / `client_msg_id`:** wouldn't collapse genuine edit states (each edit is a fresh delivery of the same logical message); `(channel, ts)` + content fingerprint is the right grain.
- **`mention_debounce_ms` (millisecond units):** rejected for `_seconds` — this is a coarse, human-tuned window (3–5s), matching the repo's seconds-valued duration config (`timeout_seconds`, `*_secs`, WeCom `debounceSecs`) rather than the sub-second `ReactionTiming` `_ms` knobs.

## Validation

Built and tested locally against this branch:

- `cargo test --bin openab`: **501 passed, 1 failed**. The one failure (`secrets::tests::resolve_exec_nonzero_exit`) reproduces on a clean tree in this environment (asserts a shell error-message string) and is unrelated to this change — `src/secrets.rs` is untouched here.
- Debounce unit tests green: `mention_debounce_latest_state_wins_and_refires_drop`, `mention_debounce_try_take_respects_extended_deadline`, `mention_debounce_history_is_bounded` (latest-state-wins, identical-re-fire-drop, attachment-only-edit-is-material, the mid-check deadline race, and FIFO-bounded history).
- `cargo clippy --all-targets -- -D warnings`: clean on all changed files. (Two pre-existing `clippy::bool_assert_comparison` findings in the untouched `src/cron.rs` are surfaced only by a newer local clippy; not introduced here.)
- `helm template`: `slack.mentionDebounceSeconds=N` renders `mention_debounce_seconds = N` under `[slack]`; omitted when unset (Rust default applies).

Please re-verify via CI.

### DIFF
--- a/charts/openab/templates/configmap.yaml
+++ b/charts/openab/templates/configmap.yaml
@@ -135,6 +135,10 @@ data:
     {{- if hasKey ($cfg.slack | default dict) "maxBatchTokens" }}
     max_batch_tokens = {{ ($cfg.slack).maxBatchTokens | int }}
     {{- end }}
+    {{- /* mentionDebounceSeconds: quiet window before a mention dispatches one turn (collapses app_mention edit re-fires) */ -}}
+    {{- if hasKey ($cfg.slack | default dict) "mentionDebounceSeconds" }}
+    mention_debounce_seconds = {{ ($cfg.slack).mentionDebounceSeconds | int }}
+    {{- end }}
     {{- if and (hasKey ($cfg.slack | default dict) "assistantMode") (not ($cfg.slack).assistantMode) }}
     assistant_mode = false
     {{- end }}
--- a/charts/openab/values.yaml
+++ b/charts/openab/values.yaml
@@ -331,6 +331,13 @@ agents:
       # maxBufferedMessages: 10
       # maxBatchTokens: soft token cap per ACP turn for batching modes (default 24000)
       # maxBatchTokens: 24000
+      # mentionDebounceSeconds: quiet window before a mention dispatches one turn.
+      # Slack re-fires app_mention for every edit state of a mention-bearing
+      # message; the adapter holds only the latest state and dispatches once the
+      # message has been quiet this long. Default 4 (Rust-side
+      # `default_mention_debounce_seconds()`). Lower = less latency but higher
+      # risk of acting on a mid-stream edit; 0 dispatches on the first event.
+      # mentionDebounceSeconds: 4
     workingDir: /home/agent
     env: {}
     # Load env vars from existing Secrets or ConfigMaps, e.g. GH_TOKEN.
--- a/src/config.rs
+++ b/src/config.rs
@@ -355,6 +355,9 @@ fn default_max_buffered_messages() -> usize {
 fn default_max_batch_tokens() -> usize {
     24_000
 }
+fn default_mention_debounce_seconds() -> u64 {
+    4
+}
 
 /// Controls whether the bot responds to user messages in threads without @mention.
 ///
@@ -430,6 +433,14 @@ pub struct SlackConfig {
     /// that are not AI apps (no `assistant:write`) to keep emoji-reaction status.
     #[serde(default = "default_true")]
     pub assistant_mode: bool,
+    /// Quiet window (seconds) for per-message `app_mention` debouncing.
+    /// Slack re-fires `app_mention` for every edit state of a mention-bearing
+    /// message; the adapter holds only the latest state and dispatches one turn
+    /// once the message has been quiet for this long. Default: 4. Lower = less
+    /// latency but higher risk of acting on a mid-stream edit; 0 dispatches on
+    /// the first event (effectively disabling debounce).
+    #[serde(default = "default_mention_debounce_seconds")]
+    pub mention_debounce_seconds: u64,
 }
 
 #[derive(Debug, Deserialize)]
--- a/src/main.rs
+++ b/src/main.rs
@@ -272,6 +272,7 @@ async fn main() -> anyhow::Result<()> {
         let router = router.clone();
         let stt = cfg.stt.clone();
         let max_bot_turns = slack_cfg.max_bot_turns;
+        let mention_debounce_seconds = slack_cfg.mention_debounce_seconds;
         let slack_shutdown_rx = shutdown_rx.clone();
         let adapter = shared_slack_adapter
             .clone()
@@ -303,6 +304,7 @@ async fn main() -> anyhow::Result<()> {
                 slack_cfg.trusted_bot_ids.into_iter().collect(),
                 slack_cfg.allow_user_messages,
                 max_bot_turns,
+                mention_debounce_seconds,
                 stt,
                 slack_shutdown_rx,
                 slack_dispatcher,
--- a/src/slack.rs
+++ b/src/slack.rs
@@ -674,7 +674,9 @@ impl ChatAdapter for SlackAdapter {
 /// Hard cap on consecutive bot messages in a thread. Prevents runaway loops.
 const MAX_CONSECUTIVE_BOT_TURNS: usize = 1000;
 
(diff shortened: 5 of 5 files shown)


dogzzdogzz:
Closing in favor of a simpler, source-level fix. Rather than debouncing the `app_mention` event stream, we'll gate streaming itself: a `[slack] streaming` toggle (mirroring the existing `[gateway] streaming` field) that posts a single final message instead of streaming via `chat.startStream`/`appendStream`/`update`.

The phantom turns originate in multi-agent threads, where a *streamed* reply that @-mentions another bot re-delivers that mention across the message's in-progress states; send-once collapses that to one delivery with none of the debounce timing/fingerprint/race machinery. Trade-off: no live typewriter streaming. Thanks!

Issue #1032, opened 2026-06-08
feat(acp): filter opencode planning narration from responses


JARVIS-coding-Agent:
## Description

When opencode is used as an ACP agent, its responses include **repeated structured planning blocks** (Goal / Progress / Blocked / Next Steps / Critical Context / Relevant Files) that pollute the final output sent to Discord. The user only wants the final answer, not the intermediate planning state.

This is the same class of problem solved by #1030 (agy-acp narration filter), but with a different pattern.

### Example

User asks: "幫我搜尋一下今天的頭條新聞"

opencode responds with 6-7 repetitions of the planning block before the actual answer:

```
Goal
協助使用者進行日常資訊搜尋。

Constraints & Preferences
API 相容 OpenAI Chat Completions 協議。
...

Progress
Done
已完成 README 摘要。

In Progress
搜尋並彙整頭條新聞。

Blocked
(none)

Next Steps
提供新聞摘要給使用者。

Critical Context
Model: Gemma-4-31B-it
...

Relevant Files
README-Gemma4-API-.md
```

↑ This block repeats N times, then the actual answer appears at the end.

## Use Case

- Users running opencode via OpenAB see massive walls of planning text before the real answer
- The planning blocks contain internal state (model URL, base URL) that should not be exposed
- Each tool call iteration appends another full planning dump, making responses extremely long

### Detection: Multi-marker heuristic

A text block is classified as **opencode planning narration** if it contains **≥ 4** of these section markers simultaneously:

```rust
const MARKERS: &[&str] = &[
    "\nGoal\n",
    "\nProgress\n",
    "\nBlocked\n",
    "\nNext Steps\n",
    "\nCritical Context\n",
    "\nRelevant Files\n",
];

fn is_opencode_planning(text: &str) -> bool {
    MARKERS.iter().filter(|m| text.contains(*m)).count() >= 4
}
```

### Filtering strategy: Extract final answer (not drop-based)

Unlike agy-acp which drops leading narration, this uses **positive extraction** — find the last section that is NOT a planning block:

```rust
fn filter_opencode_planning(response: &str) -> String {
    let sections: Vec<&str> = response.split("## Goal\n").collect();
    
    // Find last non-planning section = final answer
    if let Some(last) = sections.last() {
        if !is_opencode_planning(last) {
            return last.trim().to_string();
        }
    }
    
    // Fallback: find last non-planning section scanning backwards
    sections.iter().rev()
        .find(|s| !is_opencode_planning(s))
[...]

PR #1030, opened 2026-06-07
fix(agy-acp): filter narration based on OPENAB_TOOL_DISPLAY

chaodu-agent:
## Summary

Adds narration filtering to `agy-acp` that honors the `OPENAB_TOOL_DISPLAY` env var, allowing users to suppress intermediate thinking text ("I will...") from agy (Gemini CLI) responses.

## Problem

`agy-acp` reads all `step_type = 15` rows from Gemini CLI's SQLite conversation store and joins them into one response. This includes intermediate planning text ("I will fetch the latest commits...", "I will check the diff...") which pollutes Discord output. DB schema analysis confirmed there is no structural way to distinguish narration from final answer — both share identical protobuf fields.

## Solution

`agy-acp` reads the `OPENAB_TOOL_DISPLAY` env var:

| Value | Behavior |
|---|---|
| `full` or **unset** | No filtering (existing behavior preserved) |
| `compact` / `none` / `off` | Drop leading narration-only parts, keep final answer block |

**Narration detection:** A part is classified as narration only if *every* non-empty line starts with "I will". Parts after the first non-narration part are always preserved.

**Fallback:** If all parts are narration, keeps the last one (never returns empty).

## Usage

Set the env var in the agent config to enable filtering:

```toml
[agents.antigravity]
env = { OPENAB_TOOL_DISPLAY = "compact" }
```

No core platform changes required.

## Why not "take last step only"?

Multi-part answers (e.g. analysis + code block across multiple `step_type = 15` rows) would be truncated. The heuristic approach only drops leading narration — worst case is showing extra narration, never losing real content.

## Testing

6 unit tests covering:
- `is_narration` true/false cases
- Leading narration dropped with `compact`/`none`
- Content after first real answer preserved
- `full` mode bypasses filter
- Unset defaults to full (no silent behavior change)
- All-narration keeps last part as fallback

## Review

擺渡 ✅ · 口渡 ✅ · 普渡 ✅ · 覺渡 ✅

Discussion: https://discord.com/channels/1490282656913559673/1513303978811850833

PR #1115, opened 2026-06-15
feat(slack): add [slack].streaming toggle (send-once mode)

### LINKED ISSUE #1114: RFC: [slack].streaming toggle to disable live streaming (send-once mode)
### Problem

The Slack adapter always streams replies live — native streaming
(`chat.startStream` + `assistant.threads.setStatus`) when `assistant_mode` is on,
or a post+edit placeholder otherwise. This is great UX for a single bot, but in
multi-agent threads (several openab bots, or openab alongside other Slack bots in
one channel) it has a sharp edge.

Consider a thread with two bots:

1. user → bot A
2. bot A → bot B (bot A's reply `@`-mentions bot B)
3. bot B → bot A

At step 2, **no other bot has posted in the thread yet**, so bot A still treats it
as a single-bot thread and streams its reply via post+edit. Each edit emits a
`message_changed` event → bot B's `app_mention` re-fires **once per edit**, so bot B
responds to intermediate/partial states instead of bot A's single final message.

The existing `other_bot_present` gate only disables streaming *after* another bot
has posted — it cannot help on the kickoff message that first mentions bot B. There
is no way for an operator to deterministically opt a deployment into send-once
behaviour.

`[gateway]` already exposes a `streaming` switch (defaults `false`). Slack has no
equivalent.

### Proposed shape

Add one optional boolean to `[slack]`, default `true` (preserves current behaviour):

```toml
[slack]
streaming = true     # default — live streaming, as today
# streaming = false  # always post a single final message (send-once)
```

When `streaming = false`:

- No native streaming (no `chat.startStream` / streamed `assistant.threads.setStatus`).
- No post+edit placeholder.
- The adapter posts exactly one final `chat.postMessage` per turn.
- Independent of `assistant_mode` — the assistant **status** API (reaction / set-status)
  is unaffected; only the message *streaming* path is gated.

Mirrors `[gateway].streaming` in concept; the default deliberately differs (`true`
for Slack to preserve the current streaming UX, vs `false` for gateway).

### Key behaviour

- Gating points: `use_streaming()` and `uses_native_streaming()` both AND-in the new flag.
- `assistant_mode` status / reactions are unaffected by the switch.
- Default `true` → **zero** behaviour change for existing deployments.

### Open questions

1. **Naming:** `streaming` (mirrors gateway) vs something more explicit like
   `live_streaming` / `send_once`?
2. **Scope:** Slack-only for now, or land the same toggle on Discord for symmetry?
   (Discord has no post+edit streaming, so the multi-agent concern is Slack-specific.)
3. **Default:** keep `true` (current behaviour) — confirming that's the preferred default.

I have a working implementation + tests for the Slack side and am happy to open a
PR if the shape looks good.

---

## Update (2026-06-18)

Beyond the `streaming` toggle, PR #1115 also introduces send-once **narration trimming** — only the final answer block (text after the last tool call) is delivered, dropping inter-tool narration. Per maintainer discussion this is controlled by a **dedicated `narration_display` flag (default `false`)**, decoupled from `streaming`, and implemented in the shared adapter layer so it applies to any send-once turn (not just Slack). Supersedes the per-backend filters in #1030 / #1032.

dogzzdogzz:

The Slack adapter always streams replies live — native streaming (`chat.startStream` + `assistant.threads.setStatus`) in `assistant_mode` (default), or a post+edit placeholder otherwise. Great UX for a single bot, but in **multi-agent threads** it drives a phantom-turn storm:

1. user → bot A
2. bot A → bot B (bot A's reply `@`-mentions bot B)
3. bot B → bot A

At step 2 **no other bot has posted in the thread yet**, so bot A still treats it as a single-bot thread and streams its reply via post+edit. Each edit emits a `message_changed` event → bot B's `app_mention` re-fires **once per edit**, so bot B spawns a full agent turn for each intermediate/partial state instead of bot A's single settled message (observed: 2 real messages → 5 turns, one acting on a mid-stream fragment `<@U…>  -`).

The existing `other_bot_present` gate only disables streaming *after* another bot has posted — it cannot help on the kickoff message that first mentions bot B. There is no operator switch to deterministically force send-once.

Closes #1114. Addresses the phantom-turn bug #1103 — chosen over the #1104 `app_mention` debounce approach: posting one final message removes the re-fire at the source rather than dedup'ing after.

Discord Discussion URL: https://discord.com/channels/1491295327620169908/1491969620754567270/1515891969401032715

## At a Glance

```
            Before                                  After (streaming = false)
 ┌────────────────────────────────┐      ┌────────────────────────────────┐
 │ alone in thread → stream reply  │      │ streaming=false → send-once     │
 │   post+edit: N message_changed  │      │   one chat.postMessage per turn │
 │   each edit re-fires app_mention│      │   no native, no post+edit       │
 │   → bot B gets N phantom turns  │      │   → bot B gets ONE app_mention  │
 └────────────────────────────────┘      └────────────────────────────────┘
   default streaming = true → unchanged behaviour for everyone else
```

## Proposed Solution

Add one optional boolean to `[slack]`, default `true` (preserves current behaviour):

```toml
[slack]
streaming = true     # default — live streaming, as today
# streaming = false  # always post a single final message (send-once)
```

When `streaming = false`:

- No native streaming (no `chat.startStream` / streamed `assistant.threads.setStatus`).
- No post+edit placeholder.
- The adapter posts exactly one final `chat.postMessage` per turn.
- Independent of `assistant_mode` — the assistant **status** API (set-status / reaction) is unaffected; only the message *streaming* path is gated.

Gating is a single AND-in at the two decision points:

- `use_streaming()` → `self.streaming && !other_bot_present`
- `uses_native_streaming()` → `self.streaming && self.assistant_mode && !other_bot_present`

## Validation

- `streaming = true` (default): identical behaviour to today — native streaming in single-bot `assistant_mode` threads, post+edit otherwise, `other_bot_present` still disables streaming when another bot has posted.
- `streaming = false`: `use_streaming()` and `uses_native_streaming()` both return `false` even when alone; `uses_assistant_status()` is unaffected (status API independent of the streaming switch).

## Testing

- `cargo test --bin openab` — passing, incl. extended `assistant_mode_gates_status_and_native_streaming` cases asserting `streaming=false` forces send-once (no post+edit, no native) while leaving assistant status intact.
- `cargo clippy --all-targets -- -D warnings` — clean.
- Helm: `charts/openab/tests/configmap_test.yaml` covers `streaming` omitted (default, key absent) and `streaming: false` (renders `streaming = false`).

---

## Update (2026-06-18) — narration trimming via a dedicated `narration_display` flag

Follow-up commit `733abf5` added "send-once delivers only the final answer block" — dropping the inter-tool narration ("let me check… / now reading…") that otherwise leaks into the message. Per maintainer discussion this is being **decoupled from `streaming` into its own flag** rather than implicitly coupled to send-once:

- New `narration_display: bool`, **default `false`**, on the adapter config.
- Effective rule: in send-once delivery, narration is trimmed **unless** `narration_display = true`. (Streaming shows narration live regardless, so the flag only bites in send-once.)
- `default false` preserves `733abf5`'s behaviour (trim by default); set `true` to keep the full inter-tool text.
- Implemented in the **shared adapter layer**, so it applies to any send-once turn — Slack `streaming=false`, Slack/Discord multi-bot threads, and gateway platforms — not just Slack. This supersedes the per-backend narration filters in #1030 (agy-acp) and #1032 (opencode).

This also resolves the "is `streaming` overloaded?" review point — two independent switches: `streaming` = stream vs send-once; `narration_display` = include inter-tool narration or only the final block.

### ⚠️ Behavior change for existing send-once deployments

`narration_display` defaults `false`, which means **all existing send-once paths deliver only the final answer block after upgrade** — inter-tool narration is silently dropped. Affected paths:

- **Gateway platforms** (Telegram, LINE, Google Chat) — `GatewayConfig.streaming` defaults `false`, so every gateway turn was already send-once.
- **Discord multi-bot threads** — turns where `other_bot_present=true` were already send-once.
- **Slack `streaming=false`** (new in this PR) — obviously affected, but this is opt-in so no existing deployment is surprised here.

**To restore the old full-text behaviour**, set `narration_display = true` in `[reactions]`:

```toml
[reactions]
narration_display = true  # keep full inter-tool text in send-once messages
```

The rationale for defaulting to `false`: send-once messages read like a composed artefact (clean, direct answer) rather than a stream-of-consciousness scratchpad. Most operators on gateway/multi-bot platforms will prefer this. Operators who want the full narration can opt back in.

---

## Update (2026-06-21) — review fixes

Addressed remaining findings from the aggregated review:

- **`gateway.streaming` docs removed** — `GatewayConfig.streaming` is a pre-existing Rust field whose Helm template rendering was never wired (pre-existing gap). This PR was advertising it in `values.yaml`/`config.toml.example` without wiring the Helm template, creating a silent no-op. Docs removed; the Helm gap will be tracked separately.
- **`split_delivery` re-parse conditioned** — the second `parse_output_directives` call now only runs when the delivered slice begins at byte 0 (`answer_start==0` or `keep_full`). When `answer_start>0` the slice is mid-buffer reply text; re-parsing it could wrongly strip a final answer that opens with `[[...]]`.
- **`split_delivery` docstring qualified** — directive-preservation guarantee now explicitly scoped to non-reset turns.
- **`select_delivery_text` stale-offset warning** — `unwrap_or` replaced with `unwrap_or_else` that emits a `tracing::warn!` so a future regression is observable rather than silently leaking narration.
- **Helm test patterns tightened** — `notMatchRegex: 'streaming'` → `'streaming = '` and `notMatchRegex: 'narration_display'` → `'narration_display = '` to avoid matching comments or unrelated key names.

### DIFF
--- a/charts/openab/templates/configmap.yaml
+++ b/charts/openab/templates/configmap.yaml
@@ -138,6 +138,10 @@ data:
     {{- if and (hasKey ($cfg.slack | default dict) "assistantMode") (not ($cfg.slack).assistantMode) }}
     assistant_mode = false
     {{- end }}
+    {{- /* streaming: master switch; false → send-once (no native, no post+edit) */ -}}
+    {{- if and (hasKey ($cfg.slack | default dict) "streaming") (not ($cfg.slack).streaming) }}
+    streaming = false
+    {{- end }}
     {{- end }}
 
     [agent]
@@ -175,6 +179,10 @@ data:
     {{- fail (printf "agents.%s.reactions.toolDisplay must be one of: full, compact, none — got: %s" $name ($cfg.reactions).toolDisplay) }}
     {{- end }}
     tool_display = {{ ($cfg.reactions).toolDisplay | toJson }}
+    {{- end }}
+    {{- /* narrationDisplay: true → keep inter-tool narration in send-once (default false → final answer block only) */ -}}
+    {{- if ($cfg.reactions).narrationDisplay }}
+    narration_display = true
     {{- end }}
     {{- if ($cfg.stt).enabled }}
     {{- if not ($cfg.stt).apiKey }}
--- a/charts/openab/tests/configmap_test.yaml
+++ b/charts/openab/tests/configmap_test.yaml
@@ -180,3 +180,49 @@ tests:
       - notMatchRegex:
           path: data["config.toml"]
           pattern: 'assistant_mode'
+
+  - it: renders slack streaming = false when explicitly disabled
+    set:
+      agents.kiro.slack.enabled: true
+      agents.kiro.slack.botToken: xoxb-x
+      agents.kiro.slack.appToken: xapp-y
+      agents.kiro.slack.streaming: false
+    asserts:
+      - matchRegex:
+          path: data["config.toml"]
+          pattern: 'streaming = false'
+
+  - it: does not render slack streaming when enabled (Rust default true)
+    set:
+      agents.kiro.slack.enabled: true
+      agents.kiro.slack.botToken: xoxb-x
+      agents.kiro.slack.appToken: xapp-y
+      agents.kiro.slack.streaming: true
+    asserts:
+      - notMatchRegex:
+          path: data["config.toml"]
+          pattern: 'streaming = '
+
+  - it: renders reactions narration_display = true when explicitly enabled
+    set:
+      agents.kiro.reactions.narrationDisplay: true
+    asserts:
+      - matchRegex:
+          path: data["config.toml"]
+          pattern: 'narration_display = true'
+
+  - it: does not render reactions narration_display when off (Rust default false)
+    set:
+      agents.kiro.reactions.narrationDisplay: false
+    asserts:
+      - notMatchRegex:
+          path: data["config.toml"]
+          pattern: 'narration_display = '
+
+  - it: does not render reactions narration_display when key absent (Rust default false)
+    set:
+      agents.kiro.reactions.enabled: true
+    asserts:
+      - notMatchRegex:
+          path: data["config.toml"]
+          pattern: 'narration_display = '
--- a/charts/openab/values.yaml
+++ b/charts/openab/values.yaml
@@ -302,6 +302,12 @@ agents:
       # AI app with the `assistant:write` scope (plus chat:write). Default: true — set to
       # false for non-AI Slack apps (without assistant:write) to keep emoji-reaction status.
       assistantMode: true
+      # streaming: master switch for live reply streaming. Default true. Set
+      # false to always post a single final message (send-once) — no native
+      # streaming, no post+edit placeholder — regardless of assistantMode.
+      # Useful in multi-agent threads to avoid streamed-message edit states
+      # re-firing app_mention.
+      streaming: true
       # allowAllChannels/allowAllUsers: same auto-infer logic as discord
       allowedChannels: []  # empty + no allowAllChannels → allow all (auto-inferred)
       allowedUsers: []     # empty + no allowAllUsers → allow all (auto-inferred)
@@ -346,6 +352,11 @@ agents:
       # compact: show count summary (e.g. ✅ 3 · 🔧 1 tool(s))
       # full: show complete tool titles (for debugging)
       # none: hide tool lines entirely
+      # narrationDisplay: default false. In send-once mode, deliver ONLY the final
+      # answer block (text after the last tool call); inter-tool narration is
+      # dropped. Set true to keep the full text. Platform-agnostic (Slack
+      # streaming=false, multi-bot threads, gateway). No effect while streaming.
+      # narrationDisplay: false
     stt:
       enabled: false
       apiKey: ""
--- a/config.toml.example
+++ b/config.toml.example
@@ -39,6 +39,11 @@ allowed_channels = ["1234567890"]       # ↑ omitted + non-empty list → auto-
 #                                       # post+edit + emoji reactions. Requires the Slack app to be
 #                                       # an AI app with the `assistant:write` scope (plus chat:write).
 #                                       # Set false for non-AI Slack apps to keep emoji reactions.
+# streaming = true                     # default true: stream the reply live (native or post+edit).
+#                                       # Set false to always post a single final message (send-once),
+#                                       # regardless of assistant_mode — useful in multi-agent threads
+#                                       # to avoid streamed-edit states re-firing app_mention.
+#                                       # What a send-once message *contains* is controlled by [reactions] narration_display.
 
 # [gateway]
 # url = "ws://openab-gateway:8080/ws"  # WebSocket URL of the custom gateway
@@ -172,6 +177,12 @@ remove_after_reply = false
 #                                       # Controls how tool calls appear in the final message.
 #                                       # Applies to both assistant_mode and post+edit mode.
 #                                       # Set "none" to hide the tool summary entirely.
+# narration_display = false             # default false: in send-once mode, deliver ONLY the final
+#                                       # answer block (text after the last tool call); inter-tool
+#                                       # narration ("let me check… / now reading X") is dropped.
+#                                       # Set true to keep the full text. Platform-agnostic (Slack
+#                                       # streaming=false, multi-bot threads, gateway). No effect
+#                                       # while streaming — streamed text is shown live as produced.
 
 [reactions.emojis]
 queued = "👀"
--- a/crates/openab-core/src/adapter.rs
+++ b/crates/openab-core/src/adapter.rs
@@ -96,6 +96,101 @@ pub fn parse_output_directives(content: &str) -> (OutputDirectives, String) {
     (directives, remaining)
 }
 
+/// Select the answer text to deliver from the turn's accumulated agent-message
+/// buffer.
+///
+/// `full` is every `agent_message_chunk` concatenated across the turn, which
+/// includes the inter-tool narration the agent emits between tool calls ("let
+/// me pull the diff", "now reading the validator", ...). `answer_start` is the
+/// byte offset where the final answer block begins — set to the buffer length
+/// each time a tool call completes, so it ends up pointing just past the last
+/// tool.
+///
+/// When `keep_full` is false we deliver only that final block, dropping the
+/// narration so the message reads like the single composed artefact a
+/// tool-posted comment is. `keep_full` is true when the reply was streamed
+/// (the text was already shown live) or when `[reactions] narration_display` is
+/// set; in that case the whole buffer is returned unchanged.
+///
+/// `answer_start` is always a previous `full.len()`, hence a valid char
+/// boundary; the `get(..)` fallback to `full` only guards against a future
+/// caller passing a stale offset.
+pub fn select_delivery_text(full: &str, answer_start: usize, keep_full: bool) -> &str {
+    if keep_full {
+        full
+    } else {
+        full.get(answer_start..).unwrap_or_else(|| {
+            tracing::warn!(
+                answer_start,
+                full_len = full.len(),
+                "stale answer_start offset; delivering full buffer"
+            );
+            full
+        })
+    }
+}
+
+/// Resolve the directives and body to deliver for a finished turn.
+///
+/// Output directives (e.g. `[[reply_to:...]]`) sit at the very start of the
+/// turn's output per `docs/output-directives.md`. When `keep_full` is false
+/// (send-once trimming) that start can be inter-tool narration that
+/// [`select_delivery_text`] discards — so we parse directives from the **full**
+/// buffer (preserving them) and then take the body from the delivered slice.
+///
+/// The delivered slice is re-parsed to strip the directive header only when it
+/// still starts at byte 0 (`answer_start == 0` or `keep_full`). When
+/// `answer_start > 0` the slice is mid-buffer text; any `[[…]]` there is reply
+/// content, not a directive header, and must not be stripped.
+///
+/// Note: directive preservation assumes the turn buffer starts with the
+/// directive. A session-reset turn seeds the buffer with the expiry notice
+/// first, so directives are not preserved in that case (pre-existing behaviour).
+pub fn split_delivery(
+    full: &str,
+    answer_start: usize,
+    keep_full: bool,
+) -> (OutputDirectives, String) {
+    let (directives, _) = parse_output_directives(full);
+    let delivered = select_delivery_text(full, answer_start, keep_full);
+    // Strip the directive header from the body only when the delivered slice
+    // begins at byte 0 (no tools ran, or keep_full). When answer_start > 0,
+    // delivered is the post-last-tool suffix — don't re-parse it.
+    let body = if answer_start == 0 || keep_full {
+        parse_output_directives(delivered).1
+    } else {
+        delivered.to_owned()
+    };
+    (directives, body)
+}
+
+/// Apply the session-reset re-prepend rule to a finalized turn body.
+///
+/// The session-reset notice (`"⚠️ _Session expired, starting fresh..._\n\n"`)
+/// is pushed at the head of the turn buffer so streaming consumers see it
+/// live. When the turn ends in send-once trimming mode (`!keep_full_text`) and
+/// a tool ran (`answer_start > 0`), the slice that `select_delivery_text`
+/// returns starts *after* the notice — so re-prepend it to keep the user
+/// aware their session was reset. In every other corner (no reset, or
+/// `keep_full_text`, or no tools ran) the notice is either absent or already
+/// included in `body`, and we must not duplicate it.
+///
+/// Pure helper: deliberately mirrors the inline branch at the end of
+/// `AdapterRouter::stream_prompt_blocks` so the four-corner truth table can be
+/// exercised in isolation without a live ACP session.
+pub(crate) fn finalize_body(
+    reset: bool,
+    keep_full_text: bool,
+    answer_start: usize,
+    body: String,
+) -> String {
+    if reset && !keep_full_text && answer_start > 0 {
+        format!("⚠️ _Session expired, starting fresh..._\n\n{body}")
+    } else {
+        body
+    }
+}
+
 // --- Platform-agnostic types ---
 
 /// Identifies a channel or thread across platforms.
@@ -563,6 +658,13 @@ impl AdapterRouter {
         let thread_channel = thread_channel.clone();
         let message_limit = adapter.message_limit();
         let streaming = adapter.use_streaming(other_bot_present);
+        // Keep the full turn text (incl. inter-tool narration) when streaming
+        // (it was already shown live) OR when `[reactions] narration_display` is
+        // set. Otherwise a send-once turn delivers only the final answer block.
+        // Platform-agnostic — read from the shared reactions config, alongside
+        // `tool_display`. `streaming` still drives the placeholder / native-stream
+        // paths below; only the final-text selection uses `keep_full_text`.
+        let keep_full_text = streaming || self.reactions_config.narration_display;
         let native = adapter.uses_native_streaming(other_bot_present);
         let assistant_status = adapter.uses_assistant_status();
         // Platforms that render Markdown tables natively (e.g. Slack Block Kit
@@ -593,6 +695,12 @@ impl AdapterRouter {
 
                     let mut text_buf = String::new();
                     let mut tool_lines: Vec<ToolEntry> = Vec::new();
+                    // Byte offset into `text_buf` where the final answer block
+                    // begins — advanced to the buffer end on every tool
+                    // completion so it tracks "just past the last tool". Used by
+                    // send-once mode to drop inter-tool narration (see
+                    // `select_delivery_text`).
+                    let mut answer_start = 0usize;
 
                     if reset {
                         text_buf.push_str("⚠️ _Session expired, starting fresh..._\n\n");
@@ -826,6 +934,12 @@ impl AdapterRouter {
                                     }
                                 }
                                 AcpEvent::ToolDone { id, title, status } => {
+                                    // The final answer block is whatever text the agent
+                                    // emits AFTER its last tool. Advancing this on every
+                                    // completion leaves it pointing just past the last
+                                    // tool; send-once delivery slices from here so the
+                                    // preceding inter-tool narration is dropped.
+                                    answer_start = text_buf.len();
                                     // Live indicator: assistant status line vs emoji reaction.
                                     if assistant_status {
                                         let _ = adapter
@@ -896,11 +1010,22 @@ impl AdapterRouter {
                         let _ = handle.await;
                     }
 
-                    // Parse output directives from raw text_buf BEFORE compose_display.
-                    // Directives are agent meta-layer, not content — must be stripped
-                    // before tool lines are composed into the display output.
-                    let (directives, stripped_text) = parse_output_directives(&text_buf);
-                    let text_buf = stripped_text;
+                    // In send-once mode, deliver only the final answer block —
+                    // the text after the last tool call — so inter-tool narration
+                    // ("let me pull the diff", "now reading X") never reaches the
+                    // message. Streaming modes already showed that text live, so
+                    // they keep the whole buffer. Directives are parsed from the
+                    // FULL buffer (they sit at output start, which the slice may
+                    // drop) so a leading [[reply_to:...]] survives the narration
+                    // it was emitted alongside.
+                    let (directives, text_buf) =
+                        split_delivery(&text_buf, answer_start, keep_full_text);
+                    // The session-reset notice lives at the head of the buffer; a
+                    // tool advancing answer_start past it would drop it from the
+                    // slice, so re-prepend it to the (directive-stripped) body in
+                    // exactly that case. `finalize_body` is the pure helper that
+                    // encodes the four-corner truth table so it can be unit-tested.
+                    let text_buf = finalize_body(reset, keep_full_text, answer_start, text_buf);
 
                     // Build final content
                     let final_content =
@@ -1405,6 +1530,143 @@ fn propagate_mentions_to_chunks(
 #[cfg(test)]
 mod tests {
     use super::*;
+
+    #[test]
+    fn select_delivery_text_send_once_keeps_only_final_block() {
+        // Simulates: narration "n1" → tool (answer_start→2) → narration "n2"
+        // → tool (answer_start→14) → final answer. In send-once mode only the
+        // text after the last tool survives.
+        let full = "n1[tool]n2[tool]the final answer";
+        let answer_start = "n1[tool]n2[tool]".len();
+        assert_eq!(
+            select_delivery_text(full, answer_start, false),
+            "the final answer"
+        );
+    }
+
+    #[test]
+    fn select_delivery_text_streaming_keeps_full_buffer() {
+        // Streaming already showed the text live, so the whole buffer is kept
+        // regardless of answer_start.
+        let full = "narration then answer";
+        assert_eq!(select_delivery_text(full, 10, true), full);
+    }
+
+    #[test]
+    fn select_delivery_text_send_once_no_tools_keeps_everything() {
+        // No tool ever completed → answer_start stays 0 → the whole (tool-free)
+        // reply is delivered, including a leading session-reset notice.
+        let full = "⚠️ _Session expired, starting fresh..._\n\njust the answer";
+        assert_eq!(select_delivery_text(full, 0, false), full);
+    }
+
+    #[test]
+    fn select_delivery_text_stale_offset_falls_back_to_full() {
+        // A byte offset past the end (or a non-char-boundary) must not panic —
+        // get(..) returns None and we fall back to the full buffer.
+        let full = "abc";
+        assert_eq!(select_delivery_text(full, 999, false), full);
+        // 1 is a non-boundary inside the multi-byte '✓' (3 bytes); fallback.
+        assert_eq!(select_delivery_text("✓x", 1, false), "✓x");
+    }
+
+    #[test]
+    fn split_delivery_send_once_preserves_leading_directive_across_tools() {
+        // Regression: a [[reply_to:...]] emitted at output start, followed by
+        // narration + a tool, must survive even though the narration is dropped.
+        let full = "[[reply_to:101]]\nlet me check...[tool]the final answer";
+        let answer_start = "[[reply_to:101]]\nlet me check...[tool]".len();
+        let (directives, body) = split_delivery(full, answer_start, false);
+        assert_eq!(directives.reply_to.as_deref(), Some("101"));
+        assert_eq!(body, "the final answer");
+    }
+
+    #[test]
+    fn split_delivery_send_once_no_tools_strips_directive_from_body() {
+        // No tool ran (answer_start == 0): the slice still carries the header,
+        // so the body must have it stripped while directives are still parsed.
+        let full = "[[reply_to:55]]\njust the answer";
+        let (directives, body) = split_delivery(full, 0, false);
+        assert_eq!(directives.reply_to.as_deref(), Some("55"));
+        assert_eq!(body, "just the answer");
+    }
+
+    #[test]
+    fn split_delivery_streaming_keeps_full_body_and_directive() {
+        // Streaming keeps the full buffer; directive parsed and stripped once.
+        let full = "[[reply_to:7]]\nnarration then answer";
+        let (directives, body) = split_delivery(full, 5, true);
+        assert_eq!(directives.reply_to.as_deref(), Some("7"));
+        assert_eq!(body, "narration then answer");
+    }
+
+    // --- finalize_body: four-corner truth table for the reset re-prepend ---
+    //
+    // The send-once trimming logic in `stream_prompt_blocks` ends with an
+    // inline branch that decides whether to re-prepend the session-reset
+    // notice. Extracted into the pure helper `finalize_body` so each corner
+    // of (reset, keep_full_text, answer_start) can be exercised without a live
+    // ACP session. Mirrors the integration-level concern raised in PR #1115
+    // peer review (howie group-review, "Important #3").
+
+    #[test]
+    fn finalize_body_reset_send_once_with_tools_prepends_notice() {
+        // Reset turn, send-once trimming, a tool advanced answer_start past
+        // the notice → the slice no longer contains it → re-prepend.
+        let body = "the final answer".to_string();
+        let out = finalize_body(true, false, 42, body);
+        assert_eq!(
+            out, "⚠️ _Session expired, starting fresh..._\n\nthe final answer",
+            "send-once + reset + tool ran → notice must be re-prepended"
+        );
+    }
+
+    #[test]
+    fn finalize_body_reset_send_once_no_tools_passes_through() {
+        // answer_start == 0 means the slice still equals the full buffer,
+        // which already starts with the notice → re-prepending would
+        // duplicate it.
+        let body = "⚠️ _Session expired, starting fresh..._\n\nthe final answer".to_string();
+        let out = finalize_body(true, false, 0, body.clone());
+        assert_eq!(
+            out, body,
+            "send-once + reset + no tools → body already carries notice, pass through"
+        );
+    }
+
+    #[test]
+    fn finalize_body_reset_keep_full_passes_through() {
+        // keep_full_text means the slice is the whole buffer (incl. the
+        // notice) → must not duplicate, regardless of answer_start.
+        let body = "⚠️ _Session expired, starting fresh..._\n\nnarration then answer".to_string();
+        let out = finalize_body(true, true, 42, body.clone());
+        assert_eq!(
+            out, body,
+            "keep_full_text → body already carries notice, pass through even with tools"
+        );
+    }
+
+    #[test]
+    fn finalize_body_no_reset_send_once_passes_through() {
+        // Non-reset turn: there is no notice to manage regardless of other flags.
+        let body = "the final answer".to_string();
+        assert_eq!(
+            finalize_body(false, false, 42, body.clone()),
+            body,
+            "no reset → never prepend (send-once + tools)"
+        );
+    }
+
+    #[test]
+    fn finalize_body_no_reset_keep_full_passes_through() {
+        // Non-reset turn with keep_full_text: notice is absent, pass through.
+        let body = "the final answer".to_string();
+        assert_eq!(
+            finalize_body(false, true, 0, body.clone()),
+            body,
+            "no reset → never prepend (keep_full + no tools)"
+        );
+    }
 
     /// Compile-time regression guard: use_streaming() is a required trait method
     /// (no default). Any adapter that forgets to implement it will fail to compile.
--- a/crates/openab-core/src/config.rs
+++ b/crates/openab-core/src/config.rs
@@ -430,6 +430,15 @@ pub struct SlackConfig {
     /// that are not AI apps (no `assistant:write`) to keep emoji-reaction status.
     #[serde(default = "default_true")]
     pub assistant_mode: bool,
+    /// Master streaming switch. When `false`, the Slack adapter always posts a
+    /// single final message (send-once) — no native streaming, no post+edit
+    /// placeholder — regardless of `assistant_mode`. Default `true`. Useful for
+    /// multi-agent threads to avoid streamed-message edit states re-firing
+    /// `app_mention`. Mirrors `[gateway] streaming` in concept, but the default
+    /// deliberately differs: `GatewayConfig.streaming` defaults to `false`,
+    /// whereas this defaults to `true` to preserve current Slack streaming.
+    #[serde(default = "default_true")]
+    pub streaming: bool,
 }
 
 #[derive(Debug, Deserialize)]
@@ -454,6 +463,13 @@ pub struct GatewayConfig {
     #[serde(default)]
     pub allowed_users: Vec<String>,
     /// Enable streaming (typewriter) mode — requires gateway platform to support message editing.
+    /// Defaults to `false`, so gateway platforms (Telegram / LINE / Google Chat) are **send-once
+    /// by default**. By default send-once delivers **only the final answer block** — the text after
+    /// the last tool call — dropping inter-tool narration (the shared default send-once trimming in
+    /// `AdapterRouter::stream_prompt_blocks`, controlled by the platform-agnostic
+    /// `[reactions] narration_display`). Discord is likewise send-once in multi-bot threads
+    /// (`use_streaming` = `!other_bot_present`) and gets the same default trimming. Set `true` to
+    /// stream live and keep the full inter-tool text.
     #[serde(default)]
     pub streaming: bool,
     /// Show "…" placeholder at streaming start. Default: true. Set false for platforms using drafts.
@@ -655,6 +671,20 @@ pub struct ReactionsConfig {
     pub remove_after_reply: bool,
     #[serde(default)]
     pub tool_display: ToolDisplay,
+    /// Whether to include the agent's inter-tool narration ("let me pull the
+    /// diff", "now reading X") in send-once replies. Default `false` — a
+    /// send-once turn delivers **only the final answer block** (the text after
+    /// the last tool call), so the message reads like the single composed
+    /// artefact a tool-posted comment is. Set `true` to keep the full text.
+    ///
+    /// Platform-agnostic (sits beside `tool_display`): the trimming lives in the
+    /// shared adapter layer and applies to every send-once turn — Slack
+    /// `streaming=false`, Slack/Discord multi-bot threads, and gateway. Only
+    /// affects send-once; live streaming always shows the text as produced.
+    /// Orthogonal to `streaming`, which is the per-platform stream-vs-send-once
+    /// switch.
+    #[serde(default)]
+    pub narration_display: bool,
     #[serde(default)]
     pub emojis: ReactionEmojis,
     #[serde(default)]
@@ -790,6 +820,7 @@ impl Default for ReactionsConfig {
             enabled: true,
             remove_after_reply: false,
             tool_display: ToolDisplay::default(),
+            narration_display: false,
             emojis: ReactionEmojis::default(),
             timing: ReactionTiming::default(),
             mapping: HashMap::new(),
--- a/crates/openab-core/src/slack.rs
+++ b/crates/openab-core/src/slack.rs
@@ -81,6 +81,12 @@ pub struct SlackAdapter {
     session_ttl: std::time::Duration,
     /// Assistant mode: stream via chat.startStream + assistant.threads.setStatus.
     assistant_mode: bool,
+    /// Master streaming switch. When false, the adapter always posts a single
+    /// final message (send-once): no native streaming, no post+edit placeholder.
+    /// Default true. Set false to avoid streamed-message edit states (e.g. a
+    /// reply that @-mentions another bot re-firing app_mention in multi-agent
+    /// threads). Mirrors `[gateway] streaming`.
+    streaming: bool,
     /// streaming message ts → state. active=false = degraded (post+edit fallback).
     /// Lifecycle: stream_begin inserts, stream_finish removes; insert_stream
     /// bounds the map (STREAM_CACHE_MAX) as a safety net against aborted turns.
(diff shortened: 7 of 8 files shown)


antigenius0910 (commented):
# PR #1115 Code Review — `feat(slack): add [slack].streaming toggle (send-once mode)`

**Related issue:** #1114 (RFC)
**Branch:** `dogzzdogzz:feat/slack-streaming-toggle` → `openabdev:main`
**Single commit:** `f249b60`
**Diff size:** 7 files, +70 / -8

---

## 1. Overall (TL;DR)

A **narrow, backward-compatible, cleanly implemented** configuration-flag PR. It adds a `streaming` switch under `[slack]`, defaulting to `true` (current behavior). When set to `false`, the Slack adapter posts a single final message per turn — no native streaming and no post+edit placeholder.

**Recommendation:** Mergeable once the two RFC open questions (naming, Discord symmetry) are resolved at the design level. Nothing in the implementation blocks merge.

---

## 2. Implementation review

### 2.1 Correct gating points

Both gates are placed at the right spots:

- `src/slack.rs:537` — `use_streaming()` prefixed with `self.streaming &&`, controls post+edit placeholder
- `src/slack.rs:549` — `uses_native_streaming()` likewise gated, controls `chat.startStream`

`uses_assistant_status()` (`src/slack.rs:544`) is **intentionally unaffected**, which is the right call — the test `assistant_mode_gates_status_and_native_streaming` at line 2114 asserts this explicitly. The user still gets the "Thinking…" status set via `assistant.threads.setStatus`; only the typewriter effect on the final message goes away. That UX trade-off is reasonable.

### 2.2 Dispatch flow integration

The send-once branch at `src/adapter.rs:946` already exists — it's the path Discord falls through to when `other_bot_present=true`. `streaming=false` reuses it directly, **without introducing a new code path**. Risk surface is therefore tiny.

`stream_begin` (`adapter.rs:698`) is lazily triggered: only fires on first Text event when `native=true`. With `streaming=false`, `native` is always false, so `stream_begin` is never called → no placeholder leakage.

### 2.3 Config structure

`src/config.rs:431-441` uses `#[serde(default = "default_true")]` so existing configs upgrade with zero changes. The doc comment explicitly explains:

> Mirrors `[gateway] streaming` in concept, but the default deliberately differs: `GatewayConfig.streaming` defaults to `false`, whereas this defaults to `true` to preserve current Slack streaming.

**Good doc comment** — anyone confused by the asymmetric defaults between this and `GatewayConfig.streaming` (`config.rs:467`) gets the rationale on the spot.

### 2.4 Test coverage

`src/slack.rs:2110-2114` adds a third adapter (`streaming=false, assistant_mode=true`) to the `assistant_mode_gates_status_and_native_streaming` test, with three assertions:

```rust
assert!(!adapter3.use_streaming(false), "streaming=false forces send-once (no post+edit)");
assert!(!adapter3.uses_native_streaming(false), "streaming=false disables native even with assistant_mode");
assert!(adapter3.uses_assistant_status(), "streaming switch does not affect assistant status API");
```

These cover all three gates' states cleanly — **no obvious gap**.

### 2.5 Helm chart

`charts/openab/templates/configmap.yaml:141-144` mirrors the rendering pattern used for `assistant_mode` right above it: **only emit to `config.toml` when explicitly set to `false`**, otherwise defer to the Rust default. This avoids drift between chart-side and Rust-side defaults.

`charts/openab/tests/configmap_test.yaml:184-204` adds two symmetric tests verifying "explicit false renders" and "explicit true does not render".

`values.yaml:305-307` documents the use case (multi-agent threads avoiding `app_mention` re-fires), which is genuinely helpful for Helm users.

---

## 3. Discussion points

### 3.1 Naming (RFC open question #1)

The RFC listed three candidates: `streaming` / `live_streaming` / `send_once`. Author chose `streaming`.

I think `streaming` is **acceptable but not optimal**:
- ✅ Symmetric with `[gateway] streaming`
- ⚠️ Could be confused with `assistant_mode`, which is itself a flavor of streaming
- ⚠️ The semantics aren't quite "enable streaming"; setting `streaming = false` is really "**force send-once, overriding everything else**"

`live_streaming` or even `send_once` (inverted boolean, default `false`) would be more obvious. But you pay a symmetry cost. **A judgment call for the maintainer — not a blocker.**

### 3.2 Discord symmetry (RFC open question #2)

The RFC notes "Discord lacks post+edit streaming", but the Discord adapter does have placeholder + edit (`src/discord.rs:135`'s `use_streaming`). This PR only touches Slack, **without adding a symmetric Discord switch**.

I'd lean toward **merging the Slack part first** (it already addresses the #1103 phantom-turn bug) and tracking Discord symmetry as a follow-up issue, because:
- Discord's `MESSAGE_UPDATE` doesn't deliver mention notifications (already noted in `adapter.rs:921`), so the phantom-turn risk is lower there
- Slack-adapter diff is already non-trivial; bundling Discord would expand the review surface

### 3.3 Startup log

The current `streaming` value is not surfaced at startup. There is per-turn debug logging (`src/slack.rs:550-555`), but it's at debug level and per-turn only.

Suggest adding an info-level startup log near `main.rs:236-243` after `SlackAdapter::new`:

```rust
tracing::info!(streaming = s.streaming, assistant_mode = s.assistant_mode, "slack adapter configured");
```

This makes diagnosing "I set `streaming = false` but still see streaming behavior" a one-line check. **Optional polish — non-blocking.**

### 3.4 Documentation consistency

`config.toml.example:42` indentation looks slightly off relative to the `assistant_mode` comment block right above it. Inline format should align. **Cosmetic.**

### 3.5 What's missing (checklist)

- ✅ No schema migration needed (additive field)
- ✅ No changelog entry needed (release flow handles it)
- ✅ `clippy` should be clean (no new lint risk)
- ⚠️ No integration test that goes through `config.toml` → deserialize → `SlackAdapter` to verify `use_streaming()` returns the expected value. Current unit tests call `SlackAdapter::new(..., false)` directly, bypassing the deserialize path. A `config.rs` round-trip test asserting `streaming = false` lands as `s.streaming == false` would be belt-and-braces. **Nice-to-have, not required.**

---

## 4. E2E verification (actually executed — 2026-06-15 07:24–07:27 UTC)

Test environment: `compose.issue1114.yaml` + `openab-claude:pr1115` image, bot1 with `streaming = false`, bot2 with default. Driven autonomously via agent-browser + xoxc-token in `#mcpsupportbot-test` (C0AV8B98NKV).

| # | Scenario | Expected | Observed | Pass |
|---|----------|----------|----------|------|
| T1 | `@bot1` standalone mention | One final message, no edits | bot1 log: `streaming=false … native=false`; reply `edited=null`, 354 chars in a single send | ✅ |
| T2 | `@bot2` standalone mention (default streaming) | Typewriter / edit loop visible | bot2 log: `streaming=true … native=true`; reply `edited=1781508333.000000`, updated ~1.5s later | ✅ |
| T3 | `@bot1` mentions `@bot2` in a thread | bot2's `app_mention` fires exactly once | bot1 reply `edited=null`; bot2 receives **1** `mentions_bot=true` from bot1's reply; bot2's next turn has `other_bot_present=true` → `use_streaming=false`, so no edit storm | ✅ |

**Additional observations:**

- The `slack assistant_mode decision` debug log correctly surfaces the new `streaming` field — helps when debugging.
- During T3, bot2's own streamed reply broadcasts `subtype="message_changed"` events to every socket (including its own and bot1's). These events carry `mentions_bot=false` so they don't re-trigger dispatch — the adapter handles this correctly.
- With `assistant_mode=true` + `streaming=false`, the assistant status API still fires (since `uses_assistant_status()` is not gated by the new switch) — matches both the PR's intent and the unit test assertion.
- No panics, no errors, no leftover placeholder ("…") messages throughout the E2E run.

---

## 5. Conclusion

**Approve — all three E2E scenarios pass, implementation is correct with complete test coverage, fully backward compatible.**

Remaining items are design-level (RFC open questions):

1. Naming `streaming` vs `live_streaming` / `send_once` — I lean toward `streaming` being acceptable-but-ambiguous; maintainer's call.
2. Discord symmetry — recommend tracking as a follow-up issue (Discord's phantom-turn risk is lower).
3. **Strong suggestion to fold into this PR**: §3.3 startup info log (a single `tracing::info!(streaming = s.streaming, …)`). Saves a lot of debugging time.

The E2E run also incidentally verified that the existing `other_bot_present=true → use_streaming=false` rule from #534 still holds — this PR does not regress prior behavior.

dogzzdogzz:
Added a follow-up commit (`733abf5`) on top of the toggle: **send-once now delivers only the final answer block** — the text emitted after the last tool call.

**Why:** the turn buffer concatenates every `agent_message_chunk`, so an agent's inter-tool narration ("let me pull the diff", "now reading X") leaked into the final message. A tool-posted artefact (e.g. a GitHub PR comment) is clean because it's a single composed string; the send-once reply wasn't. This makes send-once read like that single composed artefact.

**Scope:** it's in the shared `stream_prompt_blocks` loop, gated on `!use_streaming()`, so it applies to **any** send-once turn — gateway platforms (`streaming` defaults false) and Discord multi-bot threads, not just Slack. Streaming paths are untouched.

**Correctness:** `split_delivery()` parses output directives from the *full* buffer before slicing, so a leading `[[reply_to:...]]` survives even when the narration carrying it is dropped. +7 unit tests covering 0/1/N tools, the no-tool case, UTF-8 boundaries, and directive preservation.

antigenius0910:
Follow-up review for `733abf5` ("send-once delivers only the final answer block").

## What I like

The motivation is real — agents emit inter-tool narration ("let me pull the diff", "now reading X") that bleeds into send-once replies and makes them read like stream-of-consciousness, while a tool-posted artefact reads like a single composed string. This commit aligns send-once with the artefact form.

Implementation is contained:

- One state variable (`answer_start: usize`) advanced on every `ToolDone`.
- Two helpers (`select_delivery_text`, `split_delivery`) with single, well-named responsibilities.
- One diff point in `stream_prompt_blocks` — streaming path untouched.

Test coverage is genuinely thorough: UTF-8 char-boundary fallback, leading directive preserved across tools, no-tool case re-stripping the directive header, streaming preserves full body + directive. The `reset && !streaming && answer_start > 0` re-prepend of the session-reset notice is exactly the kind of edge case that often gets missed.

## Discussion points

### 1. PR scope/title now extends beyond Slack

The follow-up applies to **every** send-once turn — not just `[slack] streaming=false`. That includes:

- Gateway platforms (Telegram / LINE / Google Chat) where `[gateway] streaming` defaults to `false`
- Discord multi-bot threads where `use_streaming = !other_bot_present`

Existing users on those platforms will see their messages get noticeably terser after upgrading. Most will probably like it (it's the artefact-form pitch), but it is a **silent behavior change**. Worth either:

- Adjusting the PR title to surface the broader scope (e.g. `feat(adapter): send-once strips inter-tool narration`), or
- Calling it out explicitly in the release note when this ships.

### 2. `streaming` field semantics are now more overloaded

Repeating a point from my previous review: `streaming` already gated two things (post+edit vs send-once, and native streaming via `assistant.threads.setStatus`). This commit adds a third — whether inter-tool narration is dropped from the final message. The three are conceptually independent:

- `Switch A`: stream live or send once (placeholder + edit, or single post)
- `Switch B`: include narration or only the final block

The PR bundles them under one flag. That is probably the right UX default (anyone who picks send-once likely wants the artefact form), but the field name `streaming` no longer captures the full surface. If you keep this coupling, a clarifying sentence in the doc comment along the lines of "`streaming = false` also implies 'final answer block only'" would help readers understand the implicit second behavior.

### 3. Implicit invariant: "agent emits the final answer after its last tool call"

`answer_start = text_buf.len()` is set on every `ToolDone`, so it always points just past the most recent tool. If the agent ever writes its answer *before* its final tool call (say: "Answer: 42" followed by a logging tool, then end-of-turn), `answer_start` advances past the answer and `select_delivery_text` returns empty — the user gets `_(no response)_`.

In practice agents don't write this way, and the ACP pattern is "answer follows the work", so this is mostly theoretical. But the assumption isn't stated anywhere. Suggest folding a one-liner into `select_delivery_text`'s doc comment, e.g.:

> Assumes the agent emits its final answer after its last tool call (the standard ACP pattern). A turn whose only post-tool emission is empty falls through to the `_(no response)_` sentinel in `stream_prompt_blocks`.

### 4. Possible refactor (non-blocking)

`use_streaming()` is now an implicit double-gate: it controls placeholder behavior AND narration trimming. Reads cleanly today, but if any future change wants to vary the two independently (e.g. a platform that wants send-once with narration, or streaming with final-block-only), the entanglement will need to be unwound. Could be split into a separate trait method like `delivers_final_block_only()` at that point — not necessary now.

## E2E verification (commit 733abf5)

Rebuilt the rig with the new image and ran two new scenarios on top of the previous T1–T3:

| # | Scenario | Expected | Observed | Pass |
|---|----------|----------|----------|------|
| T4 | `@bot1` (streaming=false) prompt that requires tool use ("read /etc/hostname then tell me the hostname in one short sentence") | Only the post-last-tool answer block reaches Slack; no "let me check the file" narration | bot1 log: `streaming=false … native=false`; reply 72 chars, `edited=null`: `:white_check_mark: ` `` `Read /etc/hostname` `` `  The hostname is ` `` `c006f123b639` `` `.` — i.e. tool-line indicator + clean 17-char final answer, zero pre-tool narration | ✅ |
| T5 | `@bot1` prompt that triggers no tools ("what is 2+2? answer in one short sentence without using any tools") | Full reply preserved (regression guard: `answer_start == 0` keeps the whole buffer) | reply 10 chars, `edited=null`: `2 + 2 = 4.` — pure answer, nothing dropped | ✅ |
| T6 | Send-once with leading `[[reply_to:<msg_ts>]]` directive emitted before a tool | Directive parsed from full buffer and applied; body has header stripped | Deferred — the 7 unit tests added in this commit cover the directive-preservation logic from every angle I could think of (leading-directive-survives-tools, no-tool case re-strips header, streaming preserves both, UTF-8 char-boundary fallback). I am willing to take the unit tests at face value here rather than engineer an agent prompt that emits `[[reply_to:...]]` reliably. | ➖ |

**Additional notes from the E2E run:**

- The `slack assistant_mode decision` debug log still correctly surfaces `streaming=false` after the follow-up commit — per-turn decision logging is intact.
- No empty replies, no `_(no response)_` sentinel fallthrough, no leftover placeholder messages.
- Subjective UX read: T4's reply ("The hostname is `c006f123b639`.") is meaningfully tighter than the equivalent on `f249b60`, which would have included a sentence or two of "let me read the file" / "I'll check the hostname" prefacing the answer. The artefact-form pitch in the commit message lands in practice.

## Verdict

Direction supported. The behavior change is well-motivated, the implementation is clean, and the tests are good. The items above are clarification asks, not blockers — the only one I'd push for in this PR is **either** a PR title/release-note adjustment for the broader scope, **or** a one-liner in the `select_delivery_text` doc comment stating the "answer follows last tool" invariant.

chaodu-agent:
LGTM ✅ — Well-designed feature toggle with thorough tests; merges cleanly into existing architecture.

## What This PR Does

Multi-agent Slack threads suffer a phantom-turn storm: streaming edits re-fire `app_mention`, causing cascading duplicate turns. This PR adds a `[slack] streaming` toggle (default `true`) to force send-once mode, plus a platform-agnostic `[reactions] narration_display` flag that controls whether inter-tool narration leaks into send-once messages.

## How It Works

Two independent config knobs:

1. **`SlackConfig.streaming`** (default `true`) — gates `use_streaming()` and `uses_native_streaming()` with a single `self.streaming &&` guard. When `false`, the adapter posts one final `chat.postMessage` per turn regardless of `assistant_mode` or `other_bot_present`.
2. **`ReactionsConfig.narration_display`** (default `false`) — shared adapter-layer logic (`select_delivery_text` / `split_delivery`) that trims inter-tool narration from send-once turns. Only the "final answer block" (text after the last `ToolDone`) is delivered unless `narration_display = true`.

Key implementation detail: `answer_start` (a byte offset advanced on each `ToolDone`) marks where the final block begins; output directives are parsed from the full buffer before slicing so a leading `[[reply_to:...]]` survives trimming.

## Findings

| 1 | 🟢 | Clean separation of streaming toggle vs narration display — two orthogonal concerns, each with own default | `src/config.rs`, `src/adapter.rs` |
| 2 | 🟢 | `select_delivery_text` gracefully handles stale/invalid offsets with `get(..).unwrap_or(full)` fallback | `src/adapter.rs:121` |
| 3 | 🟢 | Session-reset notice re-prepend logic handles edge case where a tool advances `answer_start` past the notice | `src/adapter.rs:917–920` |
| 4 | 🟢 | Comprehensive unit tests — boundary cases (stale offset, non-char-boundary), directive preservation across trimming, no-tool case | `src/adapter.rs` tests |
| 5 | 🟢 | Helm chart tests cover both "renders when disabled" and "omits when default" patterns, matching the Rust serde defaults | `charts/openab/tests/configmap_test.yaml` |

<details>

- Minimal, surgical changes to existing streaming paths — `self.streaming &&` at two decision points; no refactoring of unrelated code.
- Default `true` for Slack preserves existing behaviour; default `false` for gateway preserves existing send-once behaviour — no breaking change.
- `narration_display` is platform-agnostic (shared adapter layer) so it applies uniformly to Slack send-once, Discord multi-bot, and gateway turns.
- `split_delivery` correctly separates directive parsing (from full buffer) and body delivery (from trimmed slice), preventing directive loss.
- The `streaming` switch explicitly does **not** affect `uses_assistant_status()` — the status API (thread title, reactions) remains independent, which is correct behaviour.
- Good inline documentation on `select_delivery_text` explaining the design contract (`answer_start` is always a prior `full.len()`, hence a valid char boundary).
- CI green across all smoke tests and helm unit tests.

</details>

<details>

- PR opened: 2025-06-15
- Main already has: `GatewayConfig.streaming` (default `false`), `other_bot_present` gating in `use_streaming()`
- Net-new value: `SlackConfig.streaming` (operator-level send-once override), `ReactionsConfig.narration_display` (platform-agnostic narration trimming), `select_delivery_text` / `split_delivery` shared adapter logic, `answer_start` tracking

</details>

howie:
# Final Aggregated Review — PR #1115

`feat(slack): add [slack].streaming toggle (send-once mode)` · +295/-13, 8 files

## Mode
group-review (3/3 voices active: Claude + Codex + Gemini) · R1 independent + R2 cross-debate

## TL;DR
The core Rust logic is correct (UTF-8 boundary handling, `keep_full_text` cross-platform wiring, all 6 `SlackAdapter::new` call sites, Rust↔Helm default parity for the wired keys). No merge-blocking logic bug in the new helpers. Two things genuinely need attention: a **Helm knob that is documented but never rendered** (3/3 consensus), and a **silent send-once behavior change for existing deployments** (disputed — design intent vs. backward-compat baseline). Plus three small consensus cleanups.

---

## Consensus Important (should fix before merge)

### 1. `gateway.streaming` documented in Helm but never rendered → silent no-op  ✅ 3/3 + empirically confirmed
- **Where**: `charts/openab/values.yaml:373-376` & `config.toml.example:140-143` (docs added) vs `charts/openab/templates/configmap.yaml` `[gateway]` block (L206-240, no `streaming` render).
- **Why it matters**: `GatewayConfig.streaming` is a pre-existing Rust field; the template never rendered it (pre-existing gap). But THIS PR newly advertises the knob in values.yaml/config.toml.example. A user who reads the new docs and sets `agents.*.gateway.streaming: true` gets TOML that omits it → Rust default `false` applies → setting silently does nothing. Asymmetric with `slack.streaming`, which the same PR DID wire (configmap.yaml:142-145).
- **Fix**: Add a guarded `streaming` render to the `[gateway]` block + a Helm test mirroring the slack.streaming tests. (Lead verified: no `streaming` key exists in the gateway template block.)

### 2. `split_delivery` docstring overclaims directive survival (reset turn)  ✅ 3/3
- **Where**: `src/adapter.rs:126-134` (mirror comment ~910-917); root cause at `src/adapter.rs:657-659`.
- **Why it matters**: The new docstring states a leading `[[reply_to:...]]` always "survives the narration it was emitted alongside". In a **reset turn** the buffer is seeded with `"⚠️ _Session expired…_\n\n"` at byte 0 before agent output, so `parse_output_directives` breaks on the notice and the directive is dropped (no-tool case: leaked into body; with-tool case: silently dropped). The underlying parse behavior is PRE-EXISTING (out of scope to fix here), but the **new docstring asserts a guarantee the code does not honor**.
- **Fix (in-scope)**: Qualify the docstring to scope the guarantee to non-reset turns, or note the reset-turn gap. (Optional follow-up, out of scope: move reset-notice insertion to the final formatting stage so directives parse from a clean buffer — Gemini/Codex suggestion.)

### 3. `AdapterRouter::stream_prompt_blocks` integration untested  ⚠️ 2 Important (Claude, Gemini) / 1 NIT (Codex)
- **Where**: `src/adapter.rs:619, 924-928`.
- **Why it matters**: The pure helpers are well unit-tested, but no test covers (a) the reset re-prepend conditional `if reset && !keep_full_text && answer_start > 0` — the subtlest logic in the PR, (b) `keep_full_text = streaming || self.reactions_config.narration_display` reading real config, (c) the gateway send-once default path. `AdapterRouter` has no mock seam (dispatch.rs:1143).
- **Fix**: Extract the reset re-prepend into a small pure helper (e.g. `finalize_body(reset, keep_full, answer_start, body) -> String`) and unit-test it. (Codex view: not itself a blocker; add focused regression tests alongside the fixes above.)

---

## Disputed (maintainer decides)

### Default send-once behavior change: narration trimmed by default
- **Fact (undisputed)**: Before this PR, every send-once path (gateway default, Slack/Discord multi-bot) delivered the FULL accumulated reply. After, with `narration_display` defaulting `false`, they deliver only the final answer block. Existing deployments get a visible change without opting in. (`src/config.rs` + `src/adapter.rs:619`.)
- **Codex**: backward-compat REGRESSION — invokes "OpenAB's rule is backward-compatible defaults"; should ship as explicitly breaking or be made opt-in.
- **Gemini + Claude**: this is the PR's *intended product design* (clean final artefact on non-streaming platforms beats raw scratchpad narration); document in migration/release notes rather than revert the default.
- **Lead recommendation**: Keep the opt-out default (the feature's whole point), BUT (1) call the behavior change out prominently in the PR description + CHANGELOG so gateway/multi-bot operators aren't surprised, and (2) since Codex cites a project baseline ("backward-compatible defaults"), the opt-in-vs-opt-out call belongs to the **maintainer** — please rule explicitly. This is a product/baseline decision, not a reviewer call.

---

## Actionable NIT (consensus — clean up before merge)

### 4. `split_delivery` redundant + unsafe directive re-parse  ✅ 3/3
- `src/adapter.rs:140-142`. Re-parsing the suffix slice is redundant (directives already parsed from `full`) and can wrongly strip a final answer block that legitimately begins with `[[…]]`. Fix: parse directives once from `full`; only strip the header from the body when the delivered slice is the original leading region (i.e. `answer_start == 0`).

### 5. `select_delivery_text` silent fallback leaks narration on stale offset  ✅ 3/3
- `src/adapter.rs:122`. `full.get(answer_start..).unwrap_or(full)` fails OPEN — on a stale/non-boundary offset it returns the whole buffer (incl. the narration the feature exists to drop) with no signal. Invariant holds today, so it's dead-code defense; add `tracing::warn!(answer_start, full_len = full.len(), "stale answer_start; delivering full buffer")` so a future regression is observable.

### 6. Helm negative test matcher too loose  ✅ 3/3
- `charts/openab/tests/configmap_test.yaml` (test "does not render slack streaming when enabled"): `notMatchRegex: 'streaming'` matches the bare substring (would catch unrelated keys/comments). Tighten to `'streaming = '` to match the rendered-TOML contract the positive test asserts.

---

## Verified correct (no action)
- UTF-8 char-boundary safety of `full.get(answer_start..)`; stale offset falls back without panic (tested).
- `keep_full_text = streaming || narration_display` coherent across Slack (`streaming && !other_bot`), Discord (`!other_bot`), gateway (`self.streaming`), and native streaming.
- Rust `default_true` (slack.streaming) / `default` false (gateway.streaming, narration_display) all match docstrings + config.toml.example + values.yaml, mutually consistent across the 3 doc surfaces.
- All 6 `SlackAdapter::new` call sites updated; `adapter3` test asserts `streaming=false` forces send-once + orthogonality with assistant_status.
- The 4 wired Helm tests (slack.streaming render-on/no-render, narration_display render-on/false/absent) assert full rendered values, not bare key presence.

## Voices unavailable
- None. (Codex R1 needed a manual re-run because the helper script hard-codes `git fetch origin <base>` and this is a cross-fork PR whose base lives on `upstream`; agy R1 needed one retry after entering agentic file-search mode. Both produced full R1+R2.)

## Verdict
**NEEDS_CHANGES** — 1 consensus Helm fix (#1), 1 docstring fix (#2), 1 maintainer decision (send-once default), + 3 consensus NITs. No merge-blocking logic bug.

chaodu-agent:
LGTM ✅ — Well-structured PR delivering the streaming toggle, narration trimming, and several hardening improvements across the delivery path.

## What This PR Does

Adds a `[slack] streaming` toggle that deterministically forces send-once delivery to eliminate the phantom-turn storm in multi-agent threads, plus a platform-agnostic `[reactions] narration_display` flag that controls whether inter-tool narration appears in send-once messages.

## How It Works

- `streaming = false` ANDs into `use_streaming()` and `uses_native_streaming()`, disabling all live streaming paths while leaving `uses_assistant_status()` independent.
- `narration_display` (default `false`) gates `select_delivery_text` via `keep_full_text = streaming || narration_display`; send-once turns deliver only the text after the last tool completion.
- `split_delivery` preserves leading output directives (`[[reply_to:...]]`) by parsing them from the full buffer even when the delivered slice starts mid-buffer.
- The cosmetic edit loop now tracks consecutive failures and aborts after 3, preventing infinite retry on platforms with edit caps (Feishu).
- `edit_handle.abort()` + await eliminates the race where a late cosmetic edit could overwrite the finalized message.
- Delivery failures are tracked and surfaced as `Err(...)` so dispatch shows ❌ instead of silently completing.
- Discord mention propagation ensures multi-chunk messages carry all `<@UID>` mentions so `allow_bot_messages = "mentions"` gates don't reject overflow chunks.
- Bonus: `[reactions.mapping]` emoji-to-text feature with shortcode resolution, `ctl` module for runtime `set/get` IPC, `allow_list` abstraction, `archive_thread` trait method.

## Findings

| 1 | 🟢 | Clean separation: `streaming` and `narration_display` are orthogonal flags with clear semantics | `src/slack.rs`, `src/config.rs` |
| 2 | 🟢 | `split_delivery` re-parse condition (`answer_start == 0 || keep_full`) correctly avoids stripping mid-buffer `[[...]]` as directives | `src/adapter.rs` |
| 3 | 🟢 | `select_delivery_text` fallback emits `tracing::warn!` on stale offset — observable regression signal | `src/adapter.rs` |
| 4 | 🟢 | Comprehensive test coverage: 30+ new unit tests covering narration trimming, mention extraction/propagation, config parsing, and edge cases | `src/adapter.rs`, `src/config.rs` |
| 5 | 🟢 | Helm tests use `'streaming = '` / `'narration_display = '` patterns — won't false-match comments | `charts/openab/tests/configmap_test.yaml` |
| 6 | 🟢 | Cosmetic edit abort + join narrows the finalize race; residual window is documented and display-only | `src/adapter.rs` |
| 7 | 🟢 | Delivery health tracking propagates errors to dispatch (❌ vs silent 🆗) | `src/adapter.rs` |

<details>

- The PR description is exemplary — clear problem statement, Before/After diagrams, explicit behavior-change call-out for existing deployments, and testing sections.
- Helm template only renders `streaming = false` / `narration_display = true` when the non-default value is chosen, leveraging Rust defaults to keep the config minimal.
- The `gateway.streaming` docs were correctly removed since the Helm template never wired it — honest acknowledgment of a pre-existing gap rather than shipping dead config.
- The `extract_mentions` implementation correctly skips fenced code blocks and normalizes `<@!UID>` → `<@UID>` for dedup.
- The `propagate_mentions_to_chunks` respects the hard message limit and degrades gracefully (skips append) rather than panicking or truncating.
- The edit loop consecutive-failure cap (3) is a pragmatic safety valve — avoids hammering a platform that has hard-rejected edits.
- Session-reset notice re-prepend logic in the finalize path ensures the "⚠️ Session expired" notice survives narration trimming.

</details>

<details>

- PR opened: 2026-06-15
- Main already has: `assistant_mode` toggle, `other_bot_present` gating, post+edit streaming
- Net-new value: deterministic `streaming=false` switch (solves phantom-turn at source), platform-agnostic `narration_display`, delivery-failure tracking, edit-loop abort/join race fix, Discord mention propagation, emoji-mapping feature, ctl IPC module

</details>

antigenius0910 (approved):
## Approving

Direction supported in my [Jun 16 follow-up](https://github.com/openabdev/openab/pull/1115#issuecomment-4750070836), and nothing in the subsequent rebases (#1136 / #1137 / #1138 / #1139 / #1145) changes that read — they're orthogonal to send-once. The behavior change is well-motivated, the implementation is clean, the tests are good.

## Optional contribution — unit test for the reset re-prepend branch

In follow-up review item **#3** I called out that the inline branch at the end of `stream_prompt_blocks` —

```rust
let text_buf = if reset && !keep_full_text && answer_start > 0 {
    format!("⚠️ _Session expired, starting fresh..._\n\n{text_buf}")
} else {
    text_buf
};
```

— wasn't exercised by a unit test even though it's the exact edge case that would silently drop the session-reset notice on a tool-using turn. I built a 4-corner truth-table test for it locally, commit `b07d295` on `test/PR1115-add-finalize_body-test` (in my worktree):

- Extracts the inline if-else into a pure helper `pub fn finalize_body(reset, keep_full_text, answer_start, body) -> String`.
- No behavior change — same truth table, same call site.
- 4 tests covering every corner of `(reset, keep_full_text, answer_start > 0)`:
  - `reset + send-once + tool ran` → notice re-prepended
  - `reset + send-once + no tools` → pass through (notice already in slice)
  - `reset + keep_full_text` → pass through
  - `no reset` (both flag combos) → pass through

`cargo test` passes on top of your `78fdee2`. Embedding as a diff block below — `git apply` directly if you want it, ignore otherwise. I'm not pushing the branch (PAT scope mismatch with the workflow files pulled in by the rebases; not worth a token swap for one test commit).

<details>
<summary><code>git apply</code>-able diff (124 lines, 1 file)</summary>

```diff
diff --git a/src/adapter.rs b/src/adapter.rs
index c95ad23..e0ca226 100644
--- a/src/adapter.rs
+++ b/src/adapter.rs
@@ -164,6 +164,33 @@ pub fn split_delivery(
     (directives, body)
 }
 
+/// Apply the session-reset re-prepend rule to a finalized turn body.
+///
+/// The session-reset notice (`"⚠️ _Session expired, starting fresh..._\n\n"`)
+/// is pushed at the head of the turn buffer so streaming consumers see it
+/// live. When the turn ends in send-once trimming mode (`!keep_full_text`) and
+/// a tool ran (`answer_start > 0`), the slice that `select_delivery_text`
+/// returns starts *after* the notice — so re-prepend it to keep the user
+/// aware their session was reset. In every other corner (no reset, or
+/// `keep_full_text`, or no tools ran) the notice is either absent or already
+/// included in `body`, and we must not duplicate it.
+///
+/// Pure helper: deliberately mirrors the inline branch at the end of
+/// `AdapterRouter::stream_prompt_blocks` so the four-corner truth table can be
+/// exercised in isolation without a live ACP session.
+pub fn finalize_body(
+    reset: bool,
+    keep_full_text: bool,
+    answer_start: usize,
+    body: String,
+) -> String {
+    if reset && !keep_full_text && answer_start > 0 {
+        format!("⚠️ _Session expired, starting fresh..._\n\n{body}")
+    } else {
+        body
+    }
+}
+
 // --- Platform-agnostic types ---
 
 /// Identifies a channel or thread across platforms.
@@ -1001,12 +1028,9 @@ impl AdapterRouter {
                     // The session-reset notice lives at the head of the buffer; a
                     // tool advancing answer_start past it would drop it from the
                     // slice, so re-prepend it to the (directive-stripped) body in
-                    // exactly that case (answer_start == 0 keeps it via the slice).
-                    let text_buf = if reset && !keep_full_text && answer_start > 0 {
-                        format!("⚠️ _Session expired, starting fresh..._\n\n{text_buf}")
-                    } else {
-                        text_buf
-                    };
+                    // exactly that case. `finalize_body` is the pure helper that
+                    // encodes the four-corner truth table so it can be unit-tested.
+                    let text_buf = finalize_body(reset, keep_full_text, answer_start, text_buf);
 
                     // Build final content
                     let final_content =
@@ -1605,6 +1629,69 @@ mod tests {
         assert_eq!(body, "narration then answer");
     }
 
+    // --- finalize_body: four-corner truth table for the reset re-prepend ---
+    //
+    // The send-once trimming logic in `stream_prompt_blocks` ends with an
+    // inline branch that decides whether to re-prepend the session-reset
+    // notice. Extracted into the pure helper `finalize_body` so each corner
+    // of (reset, keep_full_text, answer_start) can be exercised without a live
+    // ACP session. Mirrors the integration-level concern raised in PR #1115
+    // peer review (howie group-review, "Important #3").
+
+    #[test]
+    fn finalize_body_reset_send_once_with_tools_prepends_notice() {
+        // Reset turn, send-once trimming, a tool advanced answer_start past
+        // the notice → the slice no longer contains it → re-prepend.
+        let body = "the final answer".to_string();
+        let out = finalize_body(true, false, 42, body);
+        assert_eq!(
+            out, "⚠️ _Session expired, starting fresh..._\n\nthe final answer",
+            "send-once + reset + tool ran → notice must be re-prepended"
+        );
+    }
+
+    #[test]
+    fn finalize_body_reset_send_once_no_tools_passes_through() {
+        // answer_start == 0 means the slice still equals the full buffer,
+        // which already starts with the notice → re-prepending would
+        // duplicate it.
+        let body = "⚠️ _Session expired, starting fresh..._\n\nthe final answer".to_string();
+        let out = finalize_body(true, false, 0, body.clone());
+        assert_eq!(
+            out, body,
+            "send-once + reset + no tools → body already carries notice, pass through"
+        );
+    }
+
+    #[test]
+    fn finalize_body_reset_keep_full_passes_through() {
+        // keep_full_text means the slice is the whole buffer (incl. the
+        // notice) → must not duplicate, regardless of answer_start.
+        let body = "⚠️ _Session expired, starting fresh..._\n\nnarration then answer".to_string();
+        let out = finalize_body(true, true, 42, body.clone());
+        assert_eq!(
+            out, body,
+            "keep_full_text → body already carries notice, pass through even with tools"
+        );
+    }
+
+    #[test]
+    fn finalize_body_no_reset_passes_through() {
+        // Non-reset turn: there is no notice to manage, irrespective of the
+        // other two flags. Covers both halves of the cartesian product.
+        let body = "the final answer".to_string();
+        assert_eq!(
+            finalize_body(false, false, 42, body.clone()),
+            body,
+            "no reset → never prepend (send-once + tools)"
+        );
+        assert_eq!(
+            finalize_body(false, true, 0, body.clone()),
+            body,
+            "no reset → never prepend (keep_full + no tools)"
+        );
+    }
+
     /// Compile-time regression guard: use_streaming() is a required trait method
```

</details>

LGTM either way — approving on the PR itself, the diff above is just a take-it-or-leave-it contribution to close the test gap from my prior review item #3.

antigenius0910 (approved):
## Approving

Direction supported in my [Jun 16 follow-up](https://github.com/openabdev/openab/pull/1115#issuecomment-4750070836), and nothing in the subsequent rebases (#1136 / #1137 / #1138 / #1139 / #1145) changes that read — they're orthogonal to send-once. The behavior change is well-motivated, the implementation is clean, the tests are good.

---

**Optional follow-up:** I opened dogzzdogzz#4 against `feat/slack-streaming-toggle` adding a unit test for the `reset && !keep_full_text && answer_start > 0` re-prepend branch — that's the test gap I flagged as item #3 in my follow-up review. Merge it into your feature branch and it flows into this PR automatically, or skip it entirely — it's a nit, not a blocker for approval.

dogzzdogzz:
## Resolution of aggregated review items

Thanks @howie for the thorough review. All items addressed — summary below.

---

### #1 — \`gateway.streaming\` documented but never rendered → **removed from docs**

Chose Option A: removed the newly-added docs rather than wiring the Helm template. The Rust field \`GatewayConfig.streaming\` is pre-existing and functional — the gap (template never renders it) predates this PR. Adding docs for a knob that silently does nothing in Helm was the bug; removing them restores the pre-PR state. The Rust field itself is untouched.

Commits: \`e1e4a243\` (values.yaml, config.toml.example)

---

### #2 — \`split_delivery\` docstring overclaims directive survival → **qualified to non-reset turns**

Docstring updated to explicitly scope the guarantee: directives survive in normal (non-reset) turns; reset-turn behaviour noted as a pre-existing gap where the session-notice prefix prevents clean directive parsing.

Commit: \`efbc799\`

---

### #3 — \`stream_prompt_blocks\` reset re-prepend untested → **\`finalize_body\` helper + 4 unit tests**

Extracted the inline 4-line reset re-prepend branch into \`pub(crate) fn finalize_body(reset, keep_full_text, answer_start, body)\` and added four tests covering every corner of the \`(reset, keep_full_text, answer_start > 0)\` truth table (reviewed by the agent panel; no logic bugs found).

Commit: \`67ea868\` (merge of PR #4 authored by @antigenius0910)

---

### Disputed — send-once default behavior change → **documented; maintainer decision requested**

Kept \`narration_display\` defaulting \`false\` (the feature's intent: clean final-answer artefact, not stream-of-consciousness scratchpad). Added a prominent \`⚠️ Behavior change for existing send-once deployments\` block to the PR description. Operators who want the old full-narration output can set \`narration_display = true\` in \`[reactions]\`.

Awaiting maintainer sign-off on the opt-out default.

---

### #4 — \`split_delivery\` redundant + unsafe re-parse → **fixed**

\`parse_output_directives\` is now only called on the delivered slice when \`answer_start == 0\` (delivered slice starts at the buffer head, where directives live). For later slices (answer-start > 0) the body is used as-is, preventing a final-answer block that opens with \`[[…]]\` from having its content silently stripped.

Commit: \`efbc799\`

---

### #5 — \`select_delivery_text\` silent fallback → **warn log added**

\`unwrap_or(full)\` → \`unwrap_or_else(|| { tracing::warn!(answer_start, full_len, "stale answer_start offset; delivering full buffer"); full })\`. The invariant still holds today; the warn makes a future regression observable.

Commit: \`efbc799\`

---

### #6 — Helm negative test matchers too loose → **tightened**

\`notMatchRegex: 'streaming'\` → \`'streaming = '\`
\`notMatchRegex: 'narration_display'\` (×2) → \`'narration_display = '\`

Commit: \`efbc799\`

---

Current branch tip: \`67ea868\`

chaodu-agent:
LGTM ✅ — All aggregated review findings addressed; code is correct and well-tested.

## What This PR Does

Adds a `[slack] streaming` master toggle (default `true`) that lets operators force send-once mode — posting exactly one final `chat.postMessage` per turn — to eliminate the phantom-turn storm caused by `message_changed` events during post+edit streaming in multi-agent threads. Companion `[reactions] narration_display` flag (default `false`) controls whether inter-tool narration survives in send-once messages across all platforms.

## How It Works

- `use_streaming()` and `uses_native_streaming()` are now gated by `self.streaming &&` — when `false`, send-once is forced regardless of `assistant_mode` or thread composition.
- `select_delivery_text` / `split_delivery` / `finalize_body` — pure helpers in the shared adapter layer that select the final-answer slice, preserve output directives across narration trimming, and handle the session-reset re-prepend corner.
- `keep_full_text = streaming || narration_display` — platform-agnostic flag computed once per turn.

## Findings

| 1 | 🟢 | All 6 howie-review items resolved cleanly | — |
| 2 | 🟢 | `finalize_body` extraction + 5 unit tests cover the 4-corner truth table | `adapter.rs:148-170` |
| 3 | 🟢 | Helm tests properly tightened to match rendered TOML, not bare substrings | `configmap_test.yaml` |
| 4 | 🟢 | Clean separation of concerns: `streaming` (transport) vs `narration_display` (content) | `config.rs`, `adapter.rs` |
| 5 | 🟢 | UTF-8 boundary safety with observable fallback (`tracing::warn!`) | `adapter.rs:119-125` |

<details>

- Thorough unit test coverage — pure helpers extracted specifically to enable isolated testing of subtler logic paths (reset re-prepend, directive preservation across trimming, stale-offset fallback).
- Excellent PR description with behavior-change callout for existing send-once deployments and explicit opt-back-in path (`narration_display = true`).
- Review iteration well-managed — each finding mapped to a specific commit with clear rationale.
- Orthogonality between `streaming` (transport mode switch) and `narration_display` (content selection) cleanly addresses the "overloaded flag" concern from the aggregated review.

</details>

<details>

- PR opened: 2026-06-15
- Main has: `use_streaming()` gated only by `other_bot_present`; no `streaming` config field on `SlackConfig`; no narration trimming in send-once path; no `narration_display` config.
- Net-new value: operator-controllable send-once mode eliminating phantom-turn storms at the source; platform-agnostic narration trimming for clean send-once messages.

</details>

<details>

### @howie (Aggregated Review — Round 2)

> **#1: `gateway.streaming` documented but never rendered**

✅ **Addressed in `e1e4a24`**: Docs removed from values.yaml/config.toml.example. Pre-existing Rust field untouched; Helm template gap tracked separately.

> **#2: `split_delivery` docstring overclaims directive survival**

✅ **Addressed in `efbc799`**: Docstring now scopes the guarantee to non-reset turns with explicit note about the reset-turn gap.

> **#3: Reset re-prepend untested**

✅ **Addressed in `67ea868`**: Extracted `finalize_body` helper + 5 unit tests covering every corner of `(reset, keep_full_text, answer_start > 0)`.

> **#4: `split_delivery` redundant + unsafe re-parse**

✅ **Addressed in `efbc799`**: Re-parse conditioned on `answer_start == 0 || keep_full` to prevent stripping a final answer that legitimately opens with `[[…]]`.

> **#5: `select_delivery_text` silent fallback**

✅ **Addressed in `efbc799`**: `unwrap_or_else` with `tracing::warn!` for observable failure.

> **#6: Helm test matchers too loose**

✅ **Addressed in `efbc799`**: Patterns tightened to `'streaming = '` / `'narration_display = '`.

### @antigenius0910 (Round 1 + Approval)

✅ **Approved**: Direction supported; no concerns on subsequent rebases.

</details>

---

⚠️ **Note**: PR currently has merge conflicts (`needs-rebase` label). Rebase onto latest `main` required before merge. The `narration_display` default-false behavior change (disputed item from aggregated review) awaits explicit maintainer sign-off — documented prominently in PR description.

howie (approved):

howie:
### Question on AGENTS.md §1 compliance for the `narration_display=false` default

Surfaced by a cross-model group review (Claude + Codex). Scoping this comment **only** to the backward-compatible-defaults guideline — not the UX merits of trimming, which the description argues well.

**The rule** (AGENTS.md §1, "Backward-Compatible Defaults"):

> New config fields MUST default to the previous behavior. Never change what existing deployments experience without an explicit opt-in.

**The fact** (verified against pre-PR `adapter.rs`): before this PR, send-once turns delivered the **full** buffer (inter-tool narration included). After this PR, with `narration_display` defaulting to `false`, `keep_full_text = streaming || narration_display` is `false` for existing send-once paths, so `split_delivery`/`select_delivery_text` trim everything before the last completed tool. This changes output for **existing** deployments with no opt-in to trigger it:

- Gateway (Telegram / LINE / Google Chat) — `GatewayConfig.streaming` defaults `false`, so every gateway turn was already send-once.
- Discord multi-bot threads — `other_bot_present=true` turns were already send-once.
- Slack multi-bot threads — likewise send-once once another bot is present.

The description already documents this break, gives a clear rationale, offers opt-back-in (`narration_display = true`), and cites "maintainer discussion." The only gap we want to close is the **§1 process** point: a documented break + an opt-in to *restore* the old behavior is not the same as *defaulting* to the previous behavior, which §1 phrases as a MUST.

**The question:** was the §1 deviation explicitly accepted by maintainers in that discussion?

- **If yes** — could it be recorded on the PR via the breaking-change path (a `!` in the commit/title, a short migration note, and a release-note callout) so the deviation is explicit rather than implied? That would satisfy §1's "explicit opt-in/acknowledgement" intent.
- **If the intent is to stay §1-compliant** — the alternative is to default the absent-config path to keep-full (i.e. make *trimming* the opt-in), leaving existing send-once deployments unchanged after upgrade.

Either path is fine from our side; we just want the §1 status to be unambiguous before merge. Thanks!

antigenius0910:
Hi @thepagent — gentle ping for CODEOWNER review when you have a moment.

State of the PR:
- ✅ All CI green (3/3)
- ✅ Two MEMBER approvals (myself + @howie at `c516310`)
- ✅ Rebase clean against `upstream/main`
- ✅ Author addressed all items from the aggregated review

One outstanding discussion item worth your eye before merge: @howie's §1 backward-compat question (above, 06:41 UTC) about the `narration_display = false` default changing output for existing send-once deployments without an opt-in. Not blocking implementation — just needs a §1 process call on whether to record as a breaking-change (title `!` + migration note) or invert the default.

Thanks!

chaodu-agent (approved):
Maintainer sign-off received: `narration_display` default `false` approved — send-once paths deliver only the final answer block after upgrade. LGTM ✅

thepagent (approved):

PR #1120, opened 2026-06-16
fix(slack): detect half-open Socket Mode via ping + idle-timeout reconnect

### LINKED ISSUE #1050: [Slack] Socket Mode adapter hangs forever on half-open connection (no read/idle timeout)

ShinyChang:
## 0. Discord Discussion URL

https://discord.com/channels/1491295327620169908/1491969620754567270/1516296200893239297

## 1. What problem does this solve?

Closes #1050, #1101.

The Slack adapter's Socket Mode connection can go **half-open**: the TCP socket stays
"ESTABLISHED" but inbound frames silently stop arriving, with no Close/FIN. After this
happens the bot is **deaf to every `app_mention` / message for hours**, yet keeps logging
`Slack Socket Mode connected` and looks healthy. Bots that also run cron jobs appear alive
(their outbound posts still go out); bots without cron go completely silent.

Root cause: in `run_slack_adapter` the inner event loop is

```rust
tokio::select! {
    msg_result = read.next() => { ... }      // only wakes on an inbound frame
    _ = shutdown_rx.changed() => { ... }      // or shutdown
}
```

When the inbound path dies without a TCP FIN/RST (a NAT/proxy idle-timeout dropping the
connection is the common cause), `read.next()` never returns: not `Close`, not `Err`, not
`None`, so the loop blocks forever. The reconnect path at the bottom of the outer loop is
only reached via `break` (on `Close`/`Err`/`None`) or a connect failure, none of which fire.
The adapter also only sends `Pong` in response to a server `Ping`; it never pings on its
own, so a dead inbound path is never probed. Slack rotates Socket Mode sockets every few
hours, but that rotation signal also arrives inbound, so it is lost too.

This is the same "deaf while showing connected" class documented upstream in the Slack
ecosystem (see the Prior Art section).

## 2. At a Glance

```
BEFORE (half-open = permanent deafness)
  Slack ──X (NAT drops inbound, no FIN) ──▶  read.next() blocks forever
  select! { read.next(), shutdown }   (no timeout arm)  ──▶  never breaks ──▶ never reconnects
  result: 0 inbound events for hours; log still says "connected"

AFTER (proactive ping + idle watchdog)
  select! {
    read.next()        ─▶ on ANY inbound frame: last_inbound = now
    ping_interval(30s) ─▶ if now - last_inbound >= 75s  ─▶ warn + break ─▶ reconnect
                          else send Ping  (Slack replies Pong = inbound = liveness)
    shutdown
  }
  on break ─▶ reconnect with capped exponential backoff (1→2→4→…→30s, reset on success)
```

## 3. Prior Art & Industry Research

This touches runtime/delivery, so prior art applies. The chosen pattern (proactive ping +
"no inbound frame within N seconds means reconnect" idle watchdog + capped backoff) is the
industry default, and is what both of openab's named reference projects do.

| Project / SDK | Keepalive mechanism | How it detects a dead socket | Reconnect strategy |
|---|---|---|---|
| **Slack Socket Mode** (server) | Server sends WS pings; sends `type:"disconnect"` (`reason:"refresh_requested"`, optional `"warning"` ~10s prior) before rotating every few hours | Protocol expects clients to reconnect on disconnect frames and dropped sockets | Client reconnects (URL re-fetched via `apps.connections.open`) |
| **@slack/socket-mode** (official Node/Bolt SDK) | Client ping + watches server pings: `clientPingTimeout` (5s pong-wait), `serverPingTimeout` (30s no-inbound-frame idle watchdog) | Either timeout means dead | `autoReconnectEnabled` (default true) |
| **python-slack-sdk** `SocketModeClient` | `ping_interval` (5s) + `current_session_monitor` thread | monitor's `check_state()` reconnects when session closed | `auto_reconnect_enabled` (default true) |
| **OpenClaw** | Delegates ping/pong to Bolt; `clientPingTimeout: 15000ms`; "App events are useful status activity, but not transport proof" | SDK pong-timeout close; refuses to treat app events as liveness | Capped exp backoff + jitter (`2s→30s, factor 1.8, jitter 0.25, 12 attempts`); fail-fast on auth |
| **Hermes Agent** | Bolt ping/pong + external `_socket_watchdog_loop` (15s poll) | watchdog reconnects when socket task done or transport probe disconnected ("self-heal when Slack silently drops the websocket") | lock-guarded restart; siblings + repo convention: exp backoff + jitter |
| **RFC 6455 / WebSocket.org** | App-layer ping/pong every 20-45s (opcodes 0x9/0xA) | missed pong / no inbound within timeout means zombie | reconnect, capped exp backoff + jitter |

Key findings (citations):
- Slack's server sends pings and `type:"disconnect"` before socket rotation, so a healthy
  socket receives inbound frames regularly; an absence of any inbound for ~75s is a sound
  dead-socket signal. https://docs.slack.dev/apis/events-api/using-socket-mode/
- Slack's official Node SDK already ships exactly this: `serverPingTimeout` is the canonical
  "no inbound frame within N ms means reconnect" watchdog; this PR is a conservative tuning of it.
  https://docs.slack.dev/tools/node-slack-sdk/reference/socket-mode/interfaces/SocketModeOptions/
- python-slack-sdk pairs a ping with a `current_session_monitor` watchdog + auto-reconnect.
  https://docs.slack.dev/tools/python-slack-sdk/reference/socket_mode/builtin/client.html
- **OpenClaw** (primary reference) implements the same pattern and names the insight directly
  ("App messages and events remain application state, not transport liveness signals"),
  reconnecting on the SDK pong-timeout with bounded backoff.
  https://github.com/openclaw/openclaw/blob/main/extensions/slack/src/monitor/reconnect-policy.ts
- The half-open mode is a filed defect in that ecosystem: OpenClaw #61072 (stale-socket
  restart) and #58519 ("silent message loss… gateway logs successful delivery, but Slack
  never receives it"), the same symptom. https://github.com/openclaw/openclaw/issues/58519
- **Hermes Agent** (other reference) self-heals via a watchdog precisely because "Slack
  silently drops the websocket". https://github.com/NousResearch/hermes-agent/blob/main/gateway/platforms/slack.py
- General best practice: TCP keepalive cannot catch a half-open socket; app-layer ping + idle
  timeout (~75% of the shortest proxy timeout; proxies kill idle conns in 60-100s) is the
  standard remedy. https://websocket.org/guides/heartbeat/ , https://datatracker.ietf.org/doc/html/rfc6455

Divergence note: neither OpenClaw nor Hermes hand-rolls Slack ping frames; both delegate to
the Bolt SDK and add a watchdog/reconnect layer. openab speaks the WebSocket directly via
`tokio-tungstenite` (no Bolt equivalent), so the ping + idle watchdog is implemented in the
adapter loop itself, same principle applied at the layer openab actually owns.

## 4. Proposed Solution

In `run_slack_adapter` (`src/slack.rs`):

1. **Idle watchdog + proactive ping.** Per connection, create a `tokio::time::interval`
   (30s, `MissedTickBehavior::Delay`, mirroring `cron.rs`) and track `last_inbound: Instant`.
   The `read.next()` arm updates `last_inbound` on any inbound frame. A new `select!` arm on
   `ping_interval.tick()`:
   - if `now - last_inbound >= 75s` then `warn!` and `break` to reconnect (half-open detected);
   - else send a `Ping` (Slack's `Pong` reply arrives inbound and refreshes `last_inbound`);
   - if the `Ping` send itself errors, `break` to reconnect.
2. **Align reconnect backoff with `gateway.rs`.** Replace the fixed `sleep(5s)` (and the
   bare URL-fetch retry `sleep`) with capped exponential backoff: `backoff_secs` starts at 1,
   resets to 1 on a successful connect, grows `(x*2).min(30)`, and every sleep is wrapped in a
   `select!` so shutdown interrupts it (the old fixed sleep was not shutdown-interruptible).
3. **Testable decisions.** The two decisions are extracted as pure functions,
   `socket_idle(elapsed, timeout)` and `next_backoff(cur)`, and unit-tested.
4. **Keep the read loop responsive without moving gating off it (review follow-up).**
   The first review noted that inline Slack Web API awaits in the `read.next()` arm
   (`conversations.replies`, `bot_participated_in_thread`, trusted-bot lookup) could starve
   the `select!` so the watchdog tick was delayed. Rather than relocate gating into the
   spawned task (which would break the dispatcher's arrival-order guarantee and let the
   consecutive-bot-turn cap race on cold start), gating stays serial in the read loop, the
   batching ADR's canonical adapter pattern, and the I/O is bounded instead:
   - the Slack Web API client gets a **30s request timeout** (it was an unbounded
     `reqwest::Client::new()`), so no inline gating call can block the loop indefinitely; the
     bot path makes at most two sequential gating calls, so the worst case stays under the 75s
     idle window, and in practice these calls are sub-second;
   - `bot_user_id` is resolved once before the loop, so the per-message path never does the
     cold-cache `auth.test` inline;
   - the turn-limit warning is classified under the `bot_turns` lock but sent from a detached
     task after the lock is released, and its result is checked and logged instead of dropped.

The 75s window sits in the recommended band (>Slack's 30s default, <typical 60-120s proxy
idle-kill, well under Slack's multi-hour refresh): long enough not to churn a quiet-but-healthy
socket, short enough to recover a dead one in ~1 minute instead of never.

## 5. Why This Approach

- **Minimal + consistent.** It reuses patterns already in the tree (the `interval` +
  `MissedTickBehavior::Delay` from `cron.rs`, the capped-backoff `select!` from `gateway.rs`),
  so the Slack adapter stops being the odd one out (it was the only WS loop with a fixed,
  non-interruptible 5s reconnect and no liveness check).
- **Stays on the canonical adapter pattern.** Gating remains serial in the read loop as the
  batching ADR prescribes, so there is no reordering of dispatcher submits and no concurrent
  bot-turn cap check. The staleness concern is solved by bounding the I/O (a request timeout
  the client was missing anyway), not by adding a per-thread async queue the ADR discourages.
- **No new dependencies, no config surface.** Constants are hardcoded like `gateway.rs`'s
  `MAX_BACKOFF`; this keeps the PR to one concern. They can be promoted to `[slack]` config in
  a follow-up if tuning is ever needed.
- **Self-validating liveness.** Counting any inbound frame (Slack's own pings, our ping's
  pong, real events) as liveness means a quiet-but-healthy socket is never falsely reconnected.

Limitations: a 30s ping interval means worst-case detection is ~75-105s, not instant (a
deliberate tradeoff against churn). The constants aren't yet configurable.

## 6. Alternatives Considered

- **TCP keepalive (`SO_KEEPALIVE`).** Rejected: default ~2h, and OS/proxy behavior is
  inconsistent; RFC/industry guidance says it doesn't reliably catch half-open client sockets.
- **Periodic forced reconnect (every N hours).** Rejected: blunt; drops a healthy socket and
  still leaves a multi-hour deaf window before the timer fires.
- **External k8s liveness probe that restarts the pod on silence.** Rejected: kills the whole
  process for one dead adapter, loses other adapters' state, and is ops config rather than a
  fix in the code that owns the socket. (It remains a fine defense-in-depth, but not the fix.)
- **Only add backoff, no heartbeat.** Rejected: backoff helps repeated connect failures but
  does nothing for a half-open socket that never errors; the heartbeat is the actual fix.
- **Move the gating + dispatch off the read loop into per-thread workers (to fix staleness).**
  Considered and implemented in an earlier revision, then reverted. It addressed staleness but
  introduced two regressions: `dispatcher.submit` ran in gating-completion order rather than
  arrival order (against the dispatcher's no-reorder invariant), and the `conversations.replies`
  bot-turn cap ran concurrently per thread, so two near-simultaneous bot messages could both
  pass the cap (notably right after a restart, when the in-memory turn tracker is empty and that
  scan is the only true-history defense). It also reintroduced a keyed async queue, which the
  batching ADR's canonical adapter pattern explicitly says not to add. Bounding the inline I/O
  keeps the canonical serial pattern and avoids both regressions, so it is preferred here.

## 7. Validation

Rust (package `openab`):
- `cargo clippy -p openab -- -D warnings` is clean.
- `cargo test -p openab` is green except one pre-existing, unrelated failure
  (`secrets::tests::resolve_exec_nonzero_exit`), which reproduces on the base branch with this
  PR's changes stashed (an environment-specific `/bin/false` error-string check in a file this
  PR does not touch). The half-open detection is covered by `socket_keepalive_tests`
  (`backoff_doubles_then_caps`, `idle_detects_half_open_at_boundary`).
- Diff is confined to `src/slack.rs`.
- Live deployment: the exact PR image was deployed to a live Slack bot on a single-node k8s
  cluster (`imagePullPolicy: Never`); the pod is healthy (`1/1 Running`) and the adapter logs
  `Slack Socket Mode connected` on startup, so the gating-in-loop path runs on the shipped
  binary. The half-open recovery is a network-timing condition (NAT/proxy idle-drop), so it is
  being observed on that long-lived connection for the `forcing reconnect` log over a
  multi-hour window.

Test level matches this crate's convention for adapter/reconnect code: the idle and backoff
decisions are pure functions with unit tests, like the rest of `src/slack.rs`. The half-open
detection + reconnect path itself is not unit-tested, consistent with `src/gateway.rs`, whose
reconnect loop has no test either; the trigger is a network-timing condition (NAT/proxy
idle-drop) best verified by deployment observation (watch for the `forcing reconnect` log on a
long-lived connection). There is no new unit test for the gating order / cap path because the
fix is a revert to the canonical serial pattern (correct by construction) and the main crate's
`src/` has no Socket Mode mock-server harness today; happy to add one if maintainers want that
bar raised for the WS adapters.

Notes:
- I did not run crate-wide `cargo fmt`: on the current toolchain `main` is not rustfmt-clean
  (unrelated files like `src/acp/agentcore.rs` reflow), and reformatting them would add noise
  unrelated to this fix (one concern per PR). The added lines are rustfmt-clean.

### DIFF
--- a/src/slack.rs
+++ b/src/slack.rs
@@ -674,6 +674,28 @@ impl ChatAdapter for SlackAdapter {
 /// Hard cap on consecutive bot messages in a thread. Prevents runaway loops.
 const MAX_CONSECUTIVE_BOT_TURNS: usize = 1000;
 
+/// Socket Mode keepalive. Slack's inbound WebSocket can go half-open (e.g. a NAT
+/// idle-timeout silently drops inbound frames with no Close/FIN), which leaves
+/// `read.next()` blocked forever, so the reconnect loop never fires and the bot
+/// goes deaf while still showing as connected. We proactively ping and force a
+/// reconnect when no inbound frame (including Slack's own pings) has arrived
+/// within the idle window. Reconnect backoff mirrors the gateway adapter.
+const PING_INTERVAL_SECS: u64 = 30;
+const IDLE_TIMEOUT_SECS: u64 = 75;
+const MAX_BACKOFF_SECS: u64 = 30;
+
+/// Next reconnect delay: double, capped. Reset to 1 on a successful connect.
+fn next_backoff(cur: u64) -> u64 {
+    (cur * 2).min(MAX_BACKOFF_SECS)
+}
+
+/// The socket is considered dead (half-open) when no inbound frame has arrived
+/// within `timeout`; Slack sends periodic pings, so silence past the window
+/// means the inbound path is gone.
+fn socket_idle(since_last_inbound: std::time::Duration, timeout: std::time::Duration) -> bool {
+    since_last_inbound >= timeout
+}
+
 /// Run the Slack adapter using Socket Mode (persistent WebSocket, no public URL needed).
 /// Reconnects automatically on disconnect.
 #[allow(clippy::too_many_arguments)]
@@ -694,6 +716,7 @@ pub async fn run_slack_adapter(
 ) -> Result<()> {
     let bot_token = adapter.bot_token().to_string();
     let bot_turns = Arc::new(tokio::sync::Mutex::new(BotTurnTracker::new(max_bot_turns)));
+    let mut backoff_secs = 1u64;
 
     loop {
         // Check for shutdown before (re)connecting
@@ -705,8 +728,12 @@ pub async fn run_slack_adapter(
         let ws_url = match get_socket_mode_url(&app_token).await {
             Ok(url) => url,
             Err(e) => {
-                error!("failed to get Socket Mode URL: {e}");
-                tokio::time::sleep(std::time::Duration::from_secs(5)).await;
+                error!(err = %e, backoff = backoff_secs, "failed to get Socket Mode URL, retrying");
+                tokio::select! {
+                    _ = tokio::time::sleep(std::time::Duration::from_secs(backoff_secs)) => {}
+                    _ = shutdown_rx.changed() => { return Ok(()); }
+                }
+                backoff_secs = next_backoff(backoff_secs);
                 continue;
             }
         };
@@ -715,11 +742,17 @@ pub async fn run_slack_adapter(
         match tokio_tungstenite::connect_async(&ws_url).await {
             Ok((ws_stream, _)) => {
                 info!("Slack Socket Mode connected");
+                backoff_secs = 1; // reset on success
                 let (mut write, mut read) = ws_stream.split();
+                let mut ping_interval =
+                    tokio::time::interval(std::time::Duration::from_secs(PING_INTERVAL_SECS));
+                ping_interval.set_missed_tick_behavior(tokio::time::MissedTickBehavior::Delay);
+                let mut last_inbound = std::time::Instant::now();
 
                 loop {
                     tokio::select! {
                         msg_result = read.next() => {
+                            last_inbound = std::time::Instant::now();
                             let Some(msg_result) = msg_result else { break };
                             match msg_result {
                                 Ok(tungstenite::Message::Text(text)) => {
@@ -1058,6 +1091,22 @@ pub async fn run_slack_adapter(
                                 _ => {}
                             }
                         }
+                        _ = ping_interval.tick() => {
+                            if socket_idle(
+                                last_inbound.elapsed(),
+                                std::time::Duration::from_secs(IDLE_TIMEOUT_SECS),
+                            ) {
+                                warn!(
+                                    idle_secs = last_inbound.elapsed().as_secs(),
+                                    "Slack Socket Mode idle past timeout (likely half-open), forcing reconnect"
+                                );
+                                break;
+                            }
+                            if let Err(e) = write.send(tungstenite::Message::Ping(Vec::new())).await {
+                                warn!(error = %e, "Slack Socket Mode ping failed, reconnecting");
+                                break;
+                            }
+                        }
                         _ = shutdown_rx.changed() => {
                             info!("Slack adapter received shutdown signal");
                             let _ = write.send(tungstenite::Message::Close(None)).await;
@@ -1067,12 +1116,16 @@ pub async fn run_slack_adapter(
                 }
             }
             Err(e) => {
-                error!("failed to connect to Slack Socket Mode: {e}");
+                error!(err = %e, backoff = backoff_secs, "failed to connect to Slack Socket Mode, retrying");
             }
         }
 
-        warn!("reconnecting to Slack Socket Mode in 5s...");
-        tokio::time::sleep(std::time::Duration::from_secs(5)).await;
+        warn!(backoff = backoff_secs, "reconnecting to Slack Socket Mode");
+        tokio::select! {
+            _ = tokio::time::sleep(std::time::Duration::from_secs(backoff_secs)) => {}
+            _ = shutdown_rx.changed() => { return Ok(()); }
+        }
+        backoff_secs = next_backoff(backoff_secs);
     }
 }
 
@@ -2214,5 +2267,37 @@ mod tests {
         let ttl = std::time::Duration::from_secs(300);
         let adapter = SlackAdapter::new("xoxb-test".into(), ttl, AllowBots::Mentions, true);
         assert!(adapter.renders_native_tables());
+    }
+}
+
+#[cfg(test)]
+mod socket_keepalive_tests {
+    use super::{next_backoff, socket_idle, IDLE_TIMEOUT_SECS, MAX_BACKOFF_SECS};
+    use std::time::Duration;
+
+    /// Backoff doubles and caps, matching the gateway adapter (1,2,4,8,16,30,30…).
+    #[test]
+    fn backoff_doubles_then_caps() {
+        let mut b = 1u64;
+        let seq: Vec<u64> = (0..8)
+            .map(|_| {
+                let cur = b;
+                b = next_backoff(b);
+                cur
+            })
+            .collect();
+        assert_eq!(seq, vec![1, 2, 4, 8, 16, MAX_BACKOFF_SECS, MAX_BACKOFF_SECS, MAX_BACKOFF_SECS]);
+        assert_eq!(next_backoff(MAX_BACKOFF_SECS), MAX_BACKOFF_SECS);
+    }
+
+    /// A half-open socket (no inbound past the window) is detected; an active one
+    /// (recent inbound, e.g. a Slack ping) is not. This is the deaf-socket guard.
+    #[test]
+    fn idle_detects_half_open_at_boundary() {
+        let timeout = Duration::from_secs(IDLE_TIMEOUT_SECS);
+        assert!(!socket_idle(Duration::from_secs(0), timeout));
+        assert!(!socket_idle(Duration::from_secs(IDLE_TIMEOUT_SECS - 1), timeout));
+        assert!(socket_idle(Duration::from_secs(IDLE_TIMEOUT_SECS), timeout));
+        assert!(socket_idle(Duration::from_secs(IDLE_TIMEOUT_SECS + 10), timeout));
     }
 }
--- a/src/slack.rs
+++ b/src/slack.rs
@@ -93,7 +93,12 @@ impl SlackAdapter {
         assistant_mode: bool,
     ) -> Self {
         Self {
-            client: reqwest::Client::new(),
+            // Bound every Slack Web API call; an unbounded inline gating call in the
+            // read loop could otherwise stall the Socket Mode idle-timeout watchdog.
+            client: reqwest::Client::builder()
+                .timeout(std::time::Duration::from_secs(30))
+                .build()
+                .unwrap_or_else(|_| reqwest::Client::new()),
             bot_token,
             bot_user_id: tokio::sync::OnceCell::new(),
             user_cache: tokio::sync::Mutex::new(HashMap::new()),
@@ -716,6 +721,9 @@ pub async fn run_slack_adapter(
 ) -> Result<()> {
     let bot_token = adapter.bot_token().to_string();
     let bot_turns = Arc::new(tokio::sync::Mutex::new(BotTurnTracker::new(max_bot_turns)));
+    // Warm the bot-user-id cache once so the per-message path never does the
+    // cold-cache `auth.test` inline in the read loop.
+    let _ = adapter.get_bot_user_id().await;
     let mut backoff_secs = 1u64;
 
     loop {
@@ -897,34 +905,47 @@ pub async fn run_slack_adapter(
                                                 } else {
                                                     format!("{}:{}", channel_id, event["ts"].as_str().unwrap_or(""))
                                                 };
-                                                {
+                                                // Classify under the lock (order-sensitive, kept in the read
+                                                // loop), but run any warning send AFTER releasing it; holding
+                                                // the tracker mutex across `chat.postMessage` would stall turn
+                                                // tracking for every thread, not just this one.
+                                                let turn_action = {
                                                     let mut tracker = bot_turns.lock().await;
                                                     if is_bot {
-                                                        match tracker.classify_bot_message(&turn_key) {
-                                                            TurnAction::Continue => {}
-                                                            TurnAction::SilentStop => continue,
-                                                            TurnAction::WarnAndStop { severity, turns, user_message } => {
-                                                                match severity {
-                                                                    TurnSeverity::Hard => warn!(channel_id, turns, "hard bot turn limit reached"),
-                                                                    TurnSeverity::Soft => info!(channel_id, turns, max = max_bot_turns, "soft bot turn limit reached"),
-                                                                }
-                                                                let channel_allowed = allow_all_channels
-                                                                    || allowed_channels.contains(channel_id);
-                                                                if !is_own_bot_msg && channel_allowed {
-                                                                    let warn_channel = ChannelRef {
-                                                                        platform: "slack".into(),
-                                                                        channel_id: channel_id.to_string(),
-                                                                        thread_id: event["thread_ts"].as_str().map(|s| s.to_string()),
-                                                                        parent_id: None,
-                                                                        origin_event_id: None,
-                                                                    };
-                                                                    let _ = adapter.send_message(&warn_channel, &user_message).await;
-                                                                }
-                                                                continue;
-                                                            }
+                                                        tracker.classify_bot_message(&turn_key)
+                                                    } else {
+                                                        if is_plain_user_message(subtype, msg_text) {
+                                                            tracker.on_human_message(&turn_key);
                                                         }
-                                                    } else if is_plain_user_message(subtype, msg_text) {
-                                                        tracker.on_human_message(&turn_key);
+                                                        TurnAction::Continue
+                                                    }
+                                                };
+                                                match turn_action {
+                                                    TurnAction::Continue => {}
+                                                    TurnAction::SilentStop => continue,
+                                                    TurnAction::WarnAndStop { severity, turns, user_message } => {
+                                                        match severity {
+                                                            TurnSeverity::Hard => warn!(channel_id, turns, "hard bot turn limit reached"),
+                                                            TurnSeverity::Soft => info!(channel_id, turns, max = max_bot_turns, "soft bot turn limit reached"),
+                                                        }
+                                                        let channel_allowed = allow_all_channels
+                                                            || allowed_channels.contains(channel_id);
+                                                        if !is_own_bot_msg && channel_allowed {
+                                                            let warn_channel = ChannelRef {
+                                                                platform: "slack".into(),
+                                                                channel_id: channel_id.to_string(),
+                                                                thread_id: event["thread_ts"].as_str().map(|s| s.to_string()),
+                                                                parent_id: None,
+                                                                origin_event_id: None,
+                                                            };
+                                                            let adapter = adapter.clone();
+                                                            tokio::spawn(async move {
+                                                                if let Err(e) = adapter.send_message(&warn_channel, &user_message).await {
+                                                                    warn!(error = %e, "failed to send bot turn limit warning");
+                                                                }
+                                                            });
+                                                        }
+                                                        continue;
                                                     }
                                                 }
 


dogzzdogzz (commented):
## Review: `fix(slack): detect half-open Socket Mode via ping + idle-timeout reconnect`

Reviewed the diff against the two other half-open fixes (#1054, #1102) and the two issues this addresses (#1050, #1101). This is the most complete of the three attempts and I think it's the right one to land — comments below are one real concern + housekeeping.

### How it compares to the other two PRs

| | #1054 (open) | #1102 (closed) | **#1120 (this)** |
|---|---|---|---|
| Mechanism | `timeout(60s, read.next())` wrap | ping 30s + idle 90s | ping 30s + idle 75s |
| Proactive ping | ❌ relies on Slack's server pings only | ✅ | ✅ |
| Reconnect backoff | ❌ fixed 5s | ❌ fixed 5s | ✅ capped-exponential, shutdown-interruptible |
| Unit tests | ❌ | ❌ | ✅ (`socket_idle`, `next_backoff`) |

Net: #1120 ⊇ #1102 (same ping+idle approach, plus backoff + tests + extracted pure fns), and is more robust than #1054 (which would false-trip if Slack ever stopped server-pinging a still-live socket — #1120's own ping defends against that). #1102 was auto-closed for inactivity, not on technical merit. Recommend landing #1120 and closing #1054/#1102 as superseded.

### 🟠 Medium — inline pre-spawn API awaits can stale the idle timer and cause a false-positive reconnect

`last_inbound` is stamped at the **top** of the `read.next()` arm, but the arm then performs several inline Slack Web API awaits *before* the agent turn is `tokio::spawn`'d: the envelope ack, trusted-bot lookup, bot-user lookup, multibot-cache update, bot-turn mutex lock, the `chat.postMessage` turn-limit warning, `conversations.replies`, and `bot_participated_in_thread`. While that arm body is executing, `select!` is not polling, so `ping_interval.tick()` cannot fire. If the cumulative inline time exceeds the 75s window (degraded/rate-limited Slack API, a very large thread on `conversations.replies`, or mutex contention), the **next** tick sees a stale `last_inbound` and forces a reconnect on a perfectly healthy, actively-communicating socket.

Real-world harm is bounded — it's one spurious reconnect (the already-`spawn`'d turn runs on its own task and isn't lost, since it posts over HTTP, not the WS `write`) — but it does defeat the watchdog's "any inbound frame = liveness" intent under load.

Suggested fix: keep the `read.next()` arm body to frame-parse + ack only, and move the remaining gating/lookup logic (and the existing `handle_message` spawn) into the spawned task, so the select loop re-enters promptly and the watchdog ticks on schedule.

### 🟡 Minor

- **Link the issues.** This fixes exactly what #1050 and #1101 describe, but the body only has the Discord URL. Add `Closes #1050, #1101` so they auto-close on merge. (#1054 already claims `Closes #1050`, #1102 claimed `Fixes #1101` — worth deconflicting which PR closes which.)
- **`last_inbound` reset placement** is unconditional, before the `None`/`Err` break — so it's also updated on a graceful server close. Benign (value is discarded on the immediate `break`), but stamping it *after* the `Some` guard would make it strictly mean "last good inbound."

### ⚪ Nits / acknowledged

- Detection latency is ~75–105s (30s tick granularity on top of the 75s window) — already called out in the PR; fine.
- Constants aren't configurable — acknowledged as a follow-up; reasonable for a one-concern PR.
- The PR states clippy is clean on production targets but `--all-targets` surfaces two pre-existing `bool_assert_comparison` lints in `cron.rs`. Worth confirming CI's exact clippy invocation so the build isn't red on a pre-existing issue.

### ✅ Verified clean

- Backoff: grows on both failure paths, resets to 1 only after a successful `connect_async`; no off-by-one.
- `Message::Ping(Vec::new())` is the correct type for tungstenite 0.21.
- Shutdown: both backoff `select!` blocks and the inner arm share `shutdown_rx`; `watch::changed()` returns immediately if already changed, so a shutdown during backoff isn't missed.
- `interval`'s immediate first tick just sends an early liveness ping — correct.

ShinyChang:
Thanks for the detailed review 🙏 Pushed changes addressing it:

1. ✅ **Idle-timer staleness** — moved the network-bound gating
   (`conversations.replies`, `bot_participated_in_thread`, trusted-bot lookup)
   plus the `handle_message` dispatch into the spawned task, so the read loop
   re-enters promptly and the idle-timeout watchdog ticks on schedule.
2. ✅ `Closes #1050, #1101` added to the description.
3. ✅ `last_inbound` now stamped after the frame guard (not on a graceful close).

Design note on (1): I kept the in-memory bot-turn tracking **serialized in the
read loop** instead of moving all gating into the task, so the per-thread turn
cap stays exact. Only the network gating becomes per-message concurrent (the
dispatch already was). The secondary `conversations.replies` consecutive-bot
check can race across two near-simultaneous bot messages in one thread, but the
serialized cap backstops it. If you'd prefer the gating fully serialized, it can
stay in the loop with each network call bounded by a `timeout` instead — let me
know which you prefer.

Locally `cargo clippy -- -D warnings` and the unit tests pass. Also deployed to a
live bot to confirm it connects and responds on the patched binary.

Re-review when you get a chance 🙏

dogzzdogzz (commented):
## Re-review (after the gating refactor in `6051f7d2`)

Thanks for the quick turnaround. The three original points are all resolved:

- ✅ **Idle-timer staleness** — heavy network gating + `handle_message` moved into the per-message `tokio::spawn`; the read-loop body is now fast in-memory work, so the watchdog ticks on schedule.
- ✅ **`Closes #1050, #1101`** added.
- ✅ **`last_inbound`** stamped after the `Some` guard.

I re-reviewed with a second pass (and a fresh Codex pass), this time tracing `src/dispatch.rs` and `src/bot_turns.rs`. The heartbeat/backoff change itself — the actual fix for #1050/#1101 — is solid and ready. The new concerns are all in the **gating-relocation refactor** that was pulled in to satisfy the staleness comment:

### 🟠 Major — per-thread dispatch ordering can now invert

Moving gating into the spawned task means each message's `dispatcher.submit(...)` now happens in **gating-completion order**, not websocket arrival order, because the spawned path awaits variable-latency calls (`conversations.replies`, `bot_participated_in_thread`, trusted-bot lookup). I confirmed in `dispatch.rs` that the dispatcher is a per-thread **mpsc consumed FIFO by submit order** and that it explicitly promises **no reordering** (module header §I3). So two same-thread messages can be batched/answered out of order.

Honest scoping: a *smaller* window already existed pre-refactor, because `handle_message` itself awaits before `submit` (attachment/STT download via `media::download_and_read_text_file`). And for plain human messages the added gating is a cached `bot_participated_in_thread` hit, so practical risk is concentrated in `AllowBots::All` bot-to-bot threads (the `conversations.replies` path). It's a *widening* of a pre-existing gap rather than a brand-new one — which is why I'd call it Major rather than a hard blocker — but it does cross the dispatcher's stated §I3 invariant.

Suggested fix: preserve arrival order before the variable-latency gating — enqueue each arrival into a per-thread worker immediately and run gating + `handle_message` sequentially inside it; or assign per-thread sequence numbers and gate `submit` on lower sequences having dispatched/rejected first.

### 🟠 Major — concurrent 1000-turn cap check loses its backstop on cold start (newly introduced)

The `AllowBots::All` consecutive-bot cap uses `conversations.replies` to count trailing bot messages. Pre-refactor that check ran **serialized** in the read loop, so it was race-free. Post-refactor it runs in concurrent sibling tasks, so two near-simultaneous bot messages can both read the same pre-cap snapshot and both proceed.

The earlier "the serialized `BotTurnTracker` backstops it" reasoning (mine included) doesn't fully hold: `bot_turns.rs` shows the tracker is an in-memory `HashMap` seeded only from messages seen *in this process* and reset to 0 on restart. After a process restart (or on a thread whose history predates this process), the in-memory counter is 0, so the racy `conversations.replies` scan is the *only* defense reflecting true Slack history — and the race can let it be exceeded. Pre-refactor this couldn't happen (serialized). So this is a real regression introduced by the refactor.

Suggested fix: serialize the history-based cap per thread (a thread-scoped mutex around the `conversations.replies` gate), or seed a shared per-thread hard count from Slack history on first contact before allowing concurrent bot dispatch.

### 🟡 Minor — residual inline awaits can still stall the 75s watchdog

The staleness risk is much reduced but not fully eliminated; a few inline awaits remain in the loop body / critical section and can exceed 75s under a slow or rate-limited Slack API:
- `get_bot_user_id().await` — cached, but does an inline `auth.test` on a cold cache.
- the `app_mention` arm still does inline `trusted_bot_ids_contains()` (can hit `bots.info`) — the "message" arm's own gating was moved to its spawn, so this is now inconsistent.
- the `WarnAndStop` arm sends the warning `chat.postMessage` **inline while holding the `bot_turns` mutex** — that blocks the tracker for *all* threads, not just the watchdog.

Suggested: resolve the bot user id once before entering the read loop; move the `app_mention` trusted-bot check into its spawned path for consistency; send the turn-limit warning from a detached task after releasing the tracker lock.

### 🟡 Minor — unconditional spawn churn

Every `message` event now spawns a task even when it will immediately `return` on pure-local predicates (`is_own_bot_msg`, `mentions_bot && !is_dm`). These carry no ordering dependency and could be evaluated inline before the spawn to avoid scheduler/alloc overhead on busy channels.

---

**Net:** the half-open heartbeat + backoff (the part that actually fixes #1050/#1101) looks good to me. If you'd rather keep this PR tightly scoped, one option is to land the heartbeat/backoff with a *minimal* staleness mitigation (resolve `bot_user_id` before the loop + bound any remaining inline write with a `timeout`) and leave the full gating relocation — plus its ordering/cap implications — to a follow-up. Either way, the two Major items above are the ones worth resolving before merge.

ShinyChang:
Thanks for the thorough second pass 🙏 Took the tightly-scoped route you offered rather than relocating the gating.

I dropped the gating-relocation refactor entirely and kept gating where the batching ADR's canonical adapter pattern puts it: serial, in the read loop, before the spawn (the spawn stays parallel-safe preprocessing + `dispatcher.submit`). That removes both Majors at the root instead of patching the concurrent version:

- **Ordering inversion** — gone, because `dispatcher.submit` is reached in the serial event loop again; same arrival-order behavior as before the relocation, no reordering across the dispatcher's FIFO.
- **Cap race on cold start** — gone, because the `conversations.replies` consecutive-bot-turn check runs serially in the loop again, so two near-simultaneous bot messages can't both read the same pre-cap snapshot. Holds regardless of the in-memory tracker being empty after a restart. Good catch on the tracker-reset point.

To keep the read loop responsive (the staleness concern that motivated the relocation in the first place), I bounded the I/O instead of moving it off-thread:

- The Slack Web API client now has a 30s request timeout (it was an unbounded `reqwest::Client::new()`), so no inline gating call can stall the idle-timeout watchdog. The bot path makes at most two sequential gating calls, so the worst case stays under the 75s idle window; in practice these are sub-second.
- `bot_user_id` is resolved once before the loop, so the per-message path no longer does a cold-cache `auth.test` inline.
- The turn-limit warning is classified under the `bot_turns` lock but sent from a detached task after releasing it, and its result is now checked + logged instead of dropped.

The `app_mention` and `message` arms are consistent again (both gate inline in the loop). `cargo clippy -- -D warnings` and the unit tests pass locally; the half-open detection is covered by the existing `socket_keepalive_tests`. No new unit test for the ordering/cap path since the fix is a revert to the canonical serial pattern and this crate has no Socket Mode test harness, but I'm validating it on a live bot.

Re-review when you get a chance 🙏

dogzzdogzz (commented):
### Attribution note — which findings this PR introduces vs which pre-exist

To help scope what must be fixed in *this* PR vs what could be a follow-up, I diffed each finding against the pre-PR `main` (`01ecb19`, the parent of this PR's first commit):

| Finding | Pre-PR `main` | This PR | Verdict |
|---|---|---|---|
| **Major — 1000-turn cap race (cold start)** | `conversations.replies` cap check ran **serialized** in the read loop → race-free | moved into the per-message `tokio::spawn` → concurrent | **Newly introduced by this PR** |
| **Minor — unconditional spawn churn** | `is_own_bot_msg` / `mentions_bot && !is_dm` rejected **inline with `continue`** (no spawn) | those rejects moved inside the spawn → a task is created even for discarded events | **Newly introduced by this PR** |
| **Major — same-thread dispatch reorder** | a *narrow* window already existed (`handle_message` awaits on attachment/STT download before `submit`), but network gating was serialized so plain messages kept arrival order | gating moved into concurrent tasks → the reorder window widens to any message whose gating latency differs (esp. `AllowBots::All`) | **Pre-existing, but materially widened by this PR** |
| **Minor — residual inline awaits stall the watchdog** | the inline awaits themselves (`get_bot_user_id`, `app_mention` `trusted_bot_ids_contains`, the in-lock warning `send_message`) all pre-exist | the **watchdog** that these can stall is new here, so the *interaction* (inline await → false reconnect) is introduced by this PR; the in-lock `send_message` is purely pre-existing | **Mixed: old code, new consequence** |

**TL;DR**
- Squarely on this PR to address: the **cap race** and (optionally) the **spawn churn** — both are direct consequences of the gating relocation that was added to satisfy the staleness comment.
- Pre-existing that this PR makes worse / newly relevant: the **reorder** (narrow → wide) and the **watchdog stall** (awaits are old; the watchdog is new).

Reasonable path: keep this PR focused on the heartbeat + backoff (the actual #1050/#1101 fix, which is solid), fix the **cap race** here since the PR created it, and split the broader gating-relocation ordering work into a follow-up if you'd prefer to keep the diff tight.

ShinyChang:
Thanks for the attribution breakdown, that is a helpful way to scope it.

I force-pushed the branch to a clean two-commit history, so the gating-relocation commit is gone entirely. It is now just `e33ee10` (heartbeat + idle-timeout + backoff) and `66406e1` (bound the Slack Web API I/O). The current diff keeps gating serial in the read loop, so the relocation-attributed rows no longer apply to the head. Mapping the table to it:

- **1000-turn cap race (cold start)**: fixed here, not deferred. The `conversations.replies` cap check is back in the serial read loop, so it is race-free again, including right after a restart when the in-memory tracker is empty.
- **Unconditional spawn churn**: gone. `is_own_bot_msg` and `mentions_bot && !is_dm` are back to an inline `continue` before any spawn.
- **Same-thread reorder (narrow then widened)**: back to narrow. Gating is serial in the loop again, so plain messages keep arrival order. The only residual is the pre-existing window where `handle_message` awaits on attachment/STT download before `submit`, unchanged from `main`; this PR no longer widens it. Happy to take closing that narrow window as a separate follow-up if you want it.
- **Residual inline awaits stalling the watchdog**: addressed by bounding the I/O rather than moving it. The Slack Web API client now has a 30s request timeout (it was an unbounded `reqwest::Client::new()`), `bot_user_id` is resolved once before the loop, and the in-lock warning `send_message` is sent from a detached task after the lock is released, with its result logged.

So the head already lands at the reasonable path you described, and a little past it: heartbeat + backoff as the #1050/#1101 fix, the cap race fixed here, and the reorder returned to its pre-existing narrow state rather than deferred. I kept gating serial in the loop instead of a per-thread worker so the adapter stays on the turn-boundary-batching canonical pattern (no second keyed async queue); bounding the inline I/O keeps the staleness fix without that deviation.

Re-review when you get a chance 🙏

dogzzdogzz (approved):
## Approving — heartbeat/backoff is solid and both regressions from the earlier revision are resolved

Re-reviewed the current head (`66406e1c`, the clean 2-commit history) with a second pass and a fresh Codex pass, this time pinned to the exact head snapshot of `slack.rs` / `dispatch.rs` / `bot_turns.rs`. Both tools agree, and I diffed against pre-PR `main` (`01ecb19`) to attribute each remaining item.

### ✅ The actual fix for #1050 / #1101 is present and correct
Proactive `Ping` + idle-timeout watchdog arm (`socket_idle` / `IDLE_TIMEOUT_SECS=75`, ping arm at the `select!`), `last_inbound` stamped after the frame guard, and capped-exponential reconnect backoff that resets on a successful connect and is shutdown-interruptible. Covered by `socket_keepalive_tests`.

### ✅ Both Majors I raised on the intermediate revision are resolved at the root
- **Same-thread dispatch reorder (the v2 *widening*)** — gone. Gating is serial in the read loop again; only `handle_message` is spawned, so `dispatcher.submit` is reached in arrival order, honoring the dispatcher's §I3 no-reorder invariant.
- **Consecutive-bot cap race on cold start** — gone. The `conversations.replies` cap check runs serially in the read loop before any spawn, so it's race-free regardless of the in-memory `BotTurnTracker` being empty after a restart.

I also confirmed the bounded-I/O follow-ups landed: 30s request timeout on the Slack Web API client, `bot_user_id` resolved once before the loop, and the turn-limit warning sent from a detached task after the `bot_turns` lock is released (with its result logged).

### Remaining items — none blocking; all pre-existing except two small new ones

**Pre-existing (verified against `main` `01ecb19`) — fine to defer to follow-ups:**
1. **Residual same-thread reorder window** — `handle_message` is spawned and awaits (attachment/STT download, display-name lookup, etc.) before its internal `dispatcher.submit`. This window exists on `main` already; this PR no longer widens it. Happy to take your offer of a separate follow-up to close it (e.g. capture a per-thread arrival sequence before the spawn).
2. **`get_socket_mode_url` uses an unbounded `reqwest::Client::new()`** — it runs before `connect_async` (no socket/watchdog yet), so a hung `apps.connections.open` can freeze the reconnect loop forever. Ironically the same silent-hang class as the bug this PR fixes — worth giving it the bounded client + a shutdown-aware `select!`, here or as a quick follow-up.
3. **`app_mention` arm bypasses the `message`-arm bot-loop protections** — it skips `BotTurnTracker` classification, the consecutive-bot cap, the own-bot skip, `note_other_bot_in_thread`, and the thread-required guard. Pre-existing structure, but a latent bot-loop gap; best filed as its own issue (factor a shared pre-spawn gating helper for both arms).

**New in this PR — small, optional polish:**
4. **Timeout fallback drops the timeout** — `reqwest::Client::builder().timeout(30s).build().unwrap_or_else(|_| reqwest::Client::new())` silently yields an unbounded client on build failure, defeating the bounded-I/O guarantee in exactly the path the comment says must be bounded. Rare (TLS-backend init), but prefer fail-fast over the unbounded fallback.
5. **Watchdog margin is thin** — the bot path can make two sequential 30s-bounded gating calls inline (`conversations.replies` + `trusted_bot_ids_contains`→`bots.info` on cache miss) = up to ~60s vs the 75s idle window (15s margin). Holds in practice (sub-second), but if Slack's API degrades, inline gating could approach the window. Consider a ~20s per-request timeout or a total deadline around the inline gating block.

### Verdict
Approving. The PR fixes #1050/#1101 and cleanly resolves the regressions the intermediate revision had introduced. Items 4–5 are quick optional tightening; 1–3 are pre-existing and good follow-up candidates — I'm happy to open issues for #2 and #3 and link them here.

thepagent (approved):

PR #1140, opened 2026-06-18
fix(acp/pool): preserve session ID on session/load timeout

chenhan-agent:

When `session/load` times out (30-second limit), OpenAB falls through to `session/new` and **permanently overwrites** `thread_map.json` with the new session ID. The user's previous conversation history becomes inaccessible without manual SSH intervention — caused by a transient network condition, not a real session loss.

Discord Discussion: https://discord.com/channels/1491295327620169908/1517011191447158795

## At a Glance

```
Before (timeout path):
  session/load timeout
       │
       ▼
  session/new          ← user's message processed against empty context
       │
       ▼
  thread_map.json      ← old session ID OVERWRITTEN ← history lost


After (this PR):
  session/load timeout
       │
       ├── permanent rejection → fall through to session/new (unchanged)
       │
       └── timeout
             │
             ▼
         return Err        ← current message NOT processed
             │
             ▼
         thread_map.json   ← old session ID PRESERVED (never touched)
             │
             ▼
         user sees: "Session Load Timeout. Send any message to retry, or /reset"
```

**OpenClaw** ([session-thread-info-loaded.ts](https://github.com/openclaw/openclaw/blob/main/src/channels/plugins/session-thread-info-loaded.ts)):
On session key resolution failure, OpenClaw explicitly preserves the original session key rather than generating a new one. Quote: _"if the channel hook has no thread id, preserve the original session key."_ Same conservative principle: on uncertainty, keep what you have.

**Hermes Agent** ([use-session-actions.ts](https://github.com/NousResearch/hermes-agent/blob/main/apps/desktop/src/app/session/hooks/use-session-actions.ts)):
Hermes tracks resume failures via `resumeFailedSessionId` state, arms a retry UI on RPC failure, and never automatically discards the session ID. The session ID is only cleared by explicit user action (`/reset` equivalent). This directly matches the approach in this PR: distinguish transient failure from permanent loss, preserve the ID, let the user decide.

## Proposed Solution

- In `pool.rs`, distinguish timeout errors from permanent rejections using the existing `"timeout waiting for"` string (produced by `send_request` in `connection.rs`)
- On timeout: return `Err` immediately — the original session ID is already in `state.persisted` (never modified on this code path), so the next message retries `session/load` automatically
- On permanent rejection (`session/load rejected`): fall through to `session/new` as before
- Add a `"session load timeout"` match in `format_user_error` with a clear user-facing message, plus a unit test

## Why this approach?

The core insight is that `state.persisted` already holds the old session ID — we never touch it before the timeout, so there is nothing to "preserve". The fix is purely about not overwriting it (by not reaching `session/new`) and not processing the current message against an empty context.

**Known limitations:**
- "Retry" does not guarantee success — if the agent's session file is gone, the next retry hits a permanent rejection and falls through to `session/new`. This is acceptable: the user ends up in a fresh session, same as before, but only after a deliberate retry rather than silently on a transient failure.
- If a session consistently exceeds the 30-second load timeout (e.g. very large history), the user will see repeated timeout errors. This is intentional: the user retains control and can use `/reset` to start fresh at any time. Automatic fallback would silently destroy history, which is the bug this PR fixes.

## Alternatives Considered

- **Raise timeout to 120s**: Mitigates the symptom but doesn't fix the destructive fallback. Rejected.
- **Add in-pool retry loop**: Adds complexity, delays error response, still doesn't guarantee success. Rejected.
- **Preserve ID + fall through to session/new anyway (v1 of this PR)**: Message processed against blank context, confusing response. Rejected.
- **Retry counter cap (fall through to session/new after N timeouts)**: Removes user control — the user is in the best position to decide when to give up and /reset. Rejected.

## Validation

- [x] `cargo check` ✅
- [x] `cargo test` ✅  507 passed; 0 failed (includes new `format_user_error_session_load_timeout` test)
- [x] `cargo clippy` ✅

### DIFF
--- a/src/acp/pool.rs
+++ b/src/acp/pool.rs
@@ -9,6 +9,11 @@ use tokio::sync::{Mutex, RwLock};
 use tokio::time::Instant;
 use tracing::{info, warn};
 
+/// Error substrings produced by `AcpConnection::send_request` that indicate a
+/// transient failure worth preserving the session ID for retry, as opposed to
+/// a permanent agent-side rejection.
+const TRANSIENT_LOAD_ERRORS: &[&str] = &["timeout waiting for", "channel closed"];
+
 /// Combined state protected by a single lock to prevent deadlocks.
 /// Lock ordering: never await a per-connection mutex while holding `state`.
 struct PoolState {
@@ -243,7 +248,7 @@ impl SessionPool {
         new_conn.initialize().await?;
 
         let mut resumed = false;
-        let mut load_failed = false;
+        let mut load_failed: Option<&str> = None;
         if let Some(ref sid) = saved_session_id {
             if new_conn.supports_load_session {
                 match new_conn.session_load(sid, &effective_workdir).await {
@@ -252,11 +257,16 @@ impl SessionPool {
                         resumed = true;
                     }
                     Err(e) => {
-                        let is_timeout = e.to_string().contains("timeout waiting for");
-                        if is_timeout {
+                        let err_str = e.to_string();
+                        let is_transient = TRANSIENT_LOAD_ERRORS.iter().any(|s| err_str.contains(s));
+                        if is_transient {
                             warn!(thread_id, session_id = %sid, error = %e,
-                                "session/load timed out, preserving session ID for retry");
-                            load_failed = true;
+                                "session/load failed transiently, preserving session ID for retry");
+                            load_failed = Some(if err_str.contains("timeout waiting for") {
+                                "timeout"
+                            } else {
+                                "connection lost"
+                            });
                         } else {
                             warn!(thread_id, session_id = %sid, error = %e,
                                 "session/load failed, creating new session");
@@ -266,12 +276,12 @@ impl SessionPool {
             }
         }
 
-        if load_failed {
-            // session/load timed out transiently. The original session ID is already
+        if let Some(reason) = load_failed {
+            // session/load failed transiently. The original session ID is already
             // in state.persisted (we haven't touched it), so the next message will
             // retry session/load automatically. Return an error so the current message
             // is not processed against a context-free session.
-            return Err(anyhow!("session load timeout: could not restore previous session"));
+            return Err(anyhow!("session load {reason}: could not restore previous session"));
         }
 
         if !resumed {
--- a/src/acp/pool.rs
+++ b/src/acp/pool.rs
@@ -243,6 +243,7 @@ impl SessionPool {
         new_conn.initialize().await?;
 
         let mut resumed = false;
+        let mut load_failed = false;
         if let Some(ref sid) = saved_session_id {
             if new_conn.supports_load_session {
                 match new_conn.session_load(sid, &effective_workdir).await {
@@ -251,19 +252,32 @@ impl SessionPool {
                         resumed = true;
                     }
                     Err(e) => {
-                        warn!(thread_id, session_id = %sid, error = %e, "session/load failed, creating new session");
+                        let is_timeout = e.to_string().contains("timeout waiting for");
+                        if is_timeout {
+                            warn!(thread_id, session_id = %sid, error = %e,
+                                "session/load timed out, preserving session ID for retry");
+                            load_failed = true;
+                        } else {
+                            warn!(thread_id, session_id = %sid, error = %e,
+                                "session/load failed, creating new session");
+                        }
                     }
                 }
             }
         }
 
+        if load_failed {
+            // session/load timed out transiently. The original session ID is already
+            // in state.persisted (we haven't touched it), so the next message will
+            // retry session/load automatically. Return an error so the current message
+            // is not processed against a context-free session.
+            return Err(anyhow!("session load timeout: could not restore previous session"));
+        }
+
         if !resumed {
             new_conn.session_new(&effective_workdir).await?;
-            // Surface the reset banner both for restored sessions and for stale
-            // live entries that died before we could recover a resumable
-            // session id. In both cases the caller is continuing after an
-            // unexpected session loss.
             if had_existing || saved_session_id.is_some() {
+                // Genuine session loss (agent died, session file gone, etc.).
                 new_conn.session_reset = true;
             }
         }
--- a/src/error_display.rs
+++ b/src/error_display.rs
@@ -10,6 +10,9 @@ pub fn format_user_error(message: &str) -> String {
     let msg_lower = message.to_lowercase();
 
     // Startup / connection errors (code == 0 from anyhow)
+    if msg_lower.contains("session load timeout") {
+        return "**Session Load Timeout**\nCould not restore your previous session. Send any message to retry, or use /reset to start fresh.".to_string();
+    }
     if msg_lower.contains("timeout waiting for") {
         // Use msg_lower for extraction to stay case-insistent with the match above.
         // msg_lower and message are the same length, so byte offsets are valid.
@@ -96,6 +99,13 @@ mod tests {
     use super::*;
 
     // ─── format_user_error tests ─────────────────────────────────────────────
+
+    #[test]
+    fn format_user_error_session_load_timeout() {
+        let result = format_user_error("session load timeout: could not restore previous session");
+        assert!(result.contains("Session Load Timeout"));
+        assert!(result.contains("/reset"));
+    }
 
     #[test]
     fn format_user_error_timeout() {
--- a/src/error_display.rs
+++ b/src/error_display.rs
@@ -11,7 +11,10 @@ pub fn format_user_error(message: &str) -> String {
 
     // Startup / connection errors (code == 0 from anyhow)
     if msg_lower.contains("session load timeout") {
-        return "**Session Load Timeout**\nCould not restore your previous session. Send any message to retry, or use /reset to start fresh.".to_string();
+        return "**Session Load Timeout**\nCould not restore your previous session. Your message was not sent — send any message to retry, or use /reset to start fresh.".to_string();
+    }
+    if msg_lower.contains("session load connection lost") {
+        return "**Session Load Failed**\nThe agent connection was lost while restoring your session. Your message was not sent — send any message to retry, or use /reset to start fresh.".to_string();
     }
     if msg_lower.contains("timeout waiting for") {
         // Use msg_lower for extraction to stay case-insistent with the match above.
@@ -104,6 +107,15 @@ mod tests {
     fn format_user_error_session_load_timeout() {
         let result = format_user_error("session load timeout: could not restore previous session");
         assert!(result.contains("Session Load Timeout"));
+        assert!(result.contains("not sent"));
+        assert!(result.contains("/reset"));
+    }
+
+    #[test]
+    fn format_user_error_session_load_connection_lost() {
+        let result = format_user_error("session load connection lost: could not restore previous session");
+        assert!(result.contains("Session Load Failed"));
+        assert!(result.contains("not sent"));
         assert!(result.contains("/reset"));
     }
 


chaodu-agent:
CHANGES REQUESTED ⚠️ — Session ID preservation is undermined by state inconsistency between `active`, `suspended`, and `persisted` maps.

## What This PR Does

When `session/load` times out, instead of overwriting `thread_map.json` with a new session ID (destroying conversation history), the fix preserves the old session ID and shows a retry banner. Also raises `session/load` timeout from 30s to 120s (matching `session/new`).

## How It Works

1. `connection.rs`: Timeout for `session/load` raised to 120s via match arm
2. `pool.rs`: On `session/load` failure, sets `load_failed = true`, writes old session ID to `persisted` map instead of overwriting with new session ID
3. `adapter.rs`: Reads one-shot `session_load_failed` flag to show retry banner

## Findings

| 1 | 🔴 | State split-brain: `cleanup_idle`/`shutdown` overwrites preserved old session ID with new session ID from `conn.acp_session_id` — preservation only survives SIGKILL, not graceful shutdown | `pool.rs` (cleanup_idle/shutdown paths) |
| 2 | 🟡 | In-process retry impossible: new connection is alive in `state.active` → next `get_or_create` early-returns without retrying `session/load` | `pool.rs:164-181` |
| 3 | 🟡 | `state.suspended.remove(thread_id)` executes unconditionally — even if connection later dies, retry reads `conn.acp_session_id` (new_sid), not old_sid | `pool.rs:~327` |
| 4 | 🟡 | Banner "send any message to retry" overpromises — retry only works on server restart (when `persisted` reloads into `suspended`) | `adapter.rs` |
| 5 | 🟢 | `session/load` timeout 30s→120s aligned with `session/new` — correct, both involve slow agent startup | `connection.rs:444` |
| 6 | 🟢 | `session_load_failed` one-shot flag pattern consistent with existing `session_reset` | `connection.rs`, `adapter.rs` |

<details>

### 🔴 F1: State split-brain allows graceful shutdown to destroy preservation

After `load_failed`, the state is:
- `state.active[thread_id]` → new connection with `acp_session_id = new_sid`
- `state.persisted[thread_id]` → `old_sid` (preserved on disk)
- `state.suspended[thread_id]` → removed

When `cleanup_idle` or `shutdown` runs, it reads `conn.acp_session_id` (= `new_sid`) and writes it to `persisted`, permanently overwriting the protected `old_sid`. The preservation only survives an ungraceful crash (SIGKILL/OOM).

**Suggested fix**: Track `original_session_id` on the connection struct. All paths that persist session IDs (eviction, idle cleanup, shutdown) should check for this field and prefer it over `acp_session_id`.

### 🟡 F2+F3+F4: "Retry on next message" is unreachable

Trace of Message N+1 after timeout:
1. `get_or_create` → `state.active.get(thread_id)` → finds alive connection → `return Ok(false)`
2. No rebuild occurs. User continues on the history-less new session.

If connection later dies:
1. `saved_session_id = state.suspended.get(thread_id)` → `None` (was removed)
2. Falls back to `conn.acp_session_id` = `new_sid` (wrong)
3. Retry attempts `session/load(new_sid)` — not the old session

Only on **server restart** does `persisted` reload into `suspended`, making the correct `old_sid` available.

**Suggested fix options**:
- **Option A** (fulfill retry promise): Don't insert fallback connection into `state.active`. Keep `old_sid` in `suspended`. Return error so next message triggers full `get_or_create` with retry.
- **Option B** (fix state, accept degradation): Also insert `old_sid` into `suspended` so connection-death retry uses correct ID. Change banner to reflect "will restore on reconnect".
- **Option C** (minimal): Keep current behavior, fix banner text to honestly say history is preserved for server restart only.

</details>

<details>

- PR opened: 2026-06-18
- Main already has: `session/load` support with 30s timeout, `session_reset` banner pattern, `thread_map.json` persistence
- Net-new value: timeout increase (good), preservation attempt (incomplete due to state inconsistency)

</details>

<details>

- Timeout alignment is correct and well-motivated
- The `session_load_failed` flag follows the established `session_reset` pattern cleanly
- PR description is thorough with clear problem statement and behavior table
- The *intent* of preserving session ID is the right direction — implementation needs state consistency fixes

</details>

chaodu-agent (changes requested):
CHANGES REQUESTED ⚠️ — See review comment: https://github.com/openabdev/openab/pull/1140#issuecomment-4741175877

chenhan-agent:
Thanks for the thorough review! The findings are accurate for the version you reviewed, but this PR was force-pushed with a significantly different approach before you commented.

The current version no longer calls `session/new` on timeout — it returns `Err` immediately, before `new_conn` is ever inserted into `state.active`. This resolves F1–F4:

- **F1**: `cleanup_idle`/`shutdown` only iterate `state.active`. Since the failed connection is dropped and never inserted, `state.persisted[old_sid]` is never overwritten.
- **F2–F4**: The dead old connection stays in `state.active` with its original `acp_session_id = old_sid`. The next message's `get_or_create` finds it dead, falls through, and retries `session/load(old_sid)` naturally.

Sorry for the confusion — the force push happened concurrently with your review.

chaodu-agent:
LGTM ✅ — Minimal, safe fix that prevents destructive session overwrite on transient timeout.

## What This PR Does

When `session/load` times out (30s limit), OpenAB previously fell through to `session/new`, permanently overwriting `thread_map.json` and erasing the user's conversation history. This PR distinguishes timeout from permanent rejection: on timeout, the message is not processed and the existing session ID is preserved for automatic retry.

## How It Works

1. In `pool.rs`, after `session_load` fails, checks if the error string contains `"timeout waiting for"` (the exact string produced by `send_request` in `connection.rs` at line 445)
2. On timeout: sets `load_failed = true`, then returns `Err(anyhow!("session load timeout: ..."))` before reaching `session/new` — the spawned `new_conn` is dropped (triggering `kill_process_group()` via the `Drop` impl), while `state.persisted` and `state.suspended` remain untouched
3. On permanent rejection (`"session/load rejected"`): falls through to `session/new` as before
4. In `error_display.rs`, the new error is matched before the generic `"timeout waiting for"` pattern, producing a clear user-facing message with retry/reset guidance
5. The next user message triggers `get_or_create` again, which finds the preserved session ID in `state.suspended` and retries `session/load`

## Findings

| 1 | 🟢 | Correct error ordering — `"session load timeout"` check placed before `"timeout waiting for"` prevents false match | `error_display.rs:14` |
| 2 | 🟢 | Resource safety — spawned process properly killed via `Drop` on early return | `pool.rs:269` |
| 3 | 🟢 | State invariant verified — `state.persisted` and `state.suspended` are never modified before the `load_failed` check | `pool.rs:271` |
| 4 | 🟢 | Clean comment simplification — removed stale multi-case comment, replaced with precise single-case explanation | `pool.rs:280` |

<details>

- **Conservative principle**: On uncertainty (transient failure), preserve what you have rather than destructively overwrite. Well-aligned with prior art cited in the PR description
- **Minimal blast radius**: Only two files changed, no new dependencies, no structural refactoring — the fix is purely about controlling a single code path
- **String-matching is appropriate here**: The timeout string `"timeout waiting for"` is produced by `send_request` (connection.rs:445), which is internal to this crate — so the match is stable and not relying on external/upstream error messages
- **Test coverage**: Unit test for the new `format_user_error` path
- **CI green**: All checks pass including cargo check, clippy, and all smoke tests

</details>

<details>

- PR opened: 2025-06-18
- Main already has: session/load with fallback to session/new on any error, `format_user_error` for user-facing error display
- Net-new value: Distinguishes transient timeout from permanent rejection, prevents destructive session ID overwrite on timeout

</details>

chaodu-agent:
CHANGES REQUESTED ⚠️ — Core fix is correct; hardening suggestions for robustness

## What This PR Does

When `session/load` times out (30s), OpenAB previously fell through to `session/new`, permanently overwriting `thread_map.json` and losing conversation history. This PR distinguishes timeout from permanent rejection: on timeout, returns early with an error, preserving the session ID for automatic retry on the next message.

## How It Works

1. After `session_load` fails, checks if the error contains `"timeout waiting for"` (produced by `send_request` in `connection.rs:445`)
2. On timeout: sets `load_failed = true`, returns `Err` before reaching `session/new` — the spawned process is cleaned up via `Drop`, while `state.persisted`/`state.suspended` remain untouched
3. On permanent rejection: falls through to `session/new` as before
4. `error_display.rs` matches the new error before the generic timeout pattern, showing retry/reset guidance

## Findings

| 1 | 🟡 | `"channel closed waiting for session/load"` still falls through to `session/new` and overwrites — same class of transient failure left unprotected | `pool.rs` Err branch |
| 2 | 🟡 | Timeout detection via `e.to_string().contains("timeout waiting for")` is implicitly coupled to `connection.rs` error formatting — a message change would silently regress to old behavior. Consider typed errors or a constant. | `pool.rs:255`, `connection.rs:445` |
| 3 | 🟡 | No unit/integration test for the timeout path in `pool.rs` itself (only `error_display` is tested) | `pool.rs` |
| 4 | 🟡 | Large sessions may consistently exceed 30s for `session/load`, causing repeated timeout loops with no forward progress. Consider documenting this limitation or adding a retry-count cap. | `pool.rs` |
| 5 | 🟡 | User-facing message says "Send any message to retry" but does not mention the current message was dropped/not processed — could confuse users expecting their message was handled | `error_display.rs:15` |
| 6 | 🟢 | Early return resource cleanup is correct — `AcpConnection::Drop` kills the process group | `pool.rs:269` |
| 7 | 🟢 | State invariant verified — `persisted`/`suspended` untouched before `load_failed` check | `pool.rs:271` |
| 8 | 🟢 | Error match ordering correct — specific before generic prevents false match | `error_display.rs:14` |
| 9 | 🟢 | PR description quality — clear problem statement, prior art research, alternatives considered | — |

<details>

### 🟡 F1: Channel-closed path unprotected
`"channel closed waiting for session/load"` (agent crash during load) takes the same `Err` branch but is **not** caught by the timeout check. If the agent's session file still exists on disk, the user's history is recoverable but gets overwritten by `session/new`. Suggest treating `channel closed` the same as timeout (preserve + retry), or opening a follow-up issue.

### 🟡 F2: String-based error classification
The `contains("timeout waiting for")` check creates an implicit contract between `connection.rs:445` and `pool.rs:255`. If someone refactors the error message in `send_request`, the timeout protection silently disappears. A typed error enum (`AcpError::RequestTimeout { method }`) or at minimum a shared constant would make this contract explicit.

### 🟡 F3: Missing pool-level test
The `format_user_error` test is good, but doesn't exercise the actual `get_or_create` timeout path. A test that mocks a slow `session_load` and asserts `state.persisted` is unchanged would catch regressions at the source.

### 🟡 F4: Repeated timeout for large sessions
If a session consistently takes >30s to load (large history), the user enters an infinite retry loop: message → timeout → "retry" → message → timeout. Consider a retry counter that falls through to `session/new` after N attempts (with a warning), or document this as a known limitation.

### 🟡 F5: UX message incomplete
Current: "Send any message to retry, or use /reset to start fresh."
Suggested: "Your message was not sent. Send any message to retry, or use /reset to start fresh."
This makes it clear the user needs to re-send their content.

</details>

<details>

- Conservative "preserve on uncertainty" principle — correct architectural decision
- Minimal blast radius: 2 files, no structural changes, no new dependencies
- String match is at least stable within-crate (both sides are internal)
- Clean `Drop`-based resource cleanup on early return
- Excellent PR description with prior art, alternatives, and validation

</details>

<details>

- PR opened: 2026-06-18
- Main already has: `session/load` with fallback to `session/new` on any error
- Net-new value: Distinguishes transient timeout from permanent failure, prevents destructive overwrite on timeout
- CI: All checks green (cargo check, clippy, 507 tests, all smoke tests pass)

</details>

---

**Note to maintainer:** The core bug fix (F6–F9) is sound. F1–F5 are hardening suggestions — none indicate the fix is wrong, only that adjacent paths have similar exposure or the implementation could be more robust. These can be addressed as follow-up issues if preferred.

chenhan-agent:
Thanks for the detailed review! Here's what was addressed in the latest push:

**F1 (channel-closed path)** — Fixed. Added `"channel closed"` to `TRANSIENT_LOAD_ERRORS` so agent crashes during `session/load` are treated the same as timeouts (preserve + retry).

**F2 (string coupling)** — Fixed. Extracted `TRANSIENT_LOAD_ERRORS` constant in `pool.rs` to make the contract explicit.

**F3 (pool-level test)** — Deferred. Mocking `AcpConnection` would require a trait abstraction refactor; out of scope for this PR.

**F4 (repeated timeout loop)** — By design. The user retains control and can `/reset` at any time. Added this as a known limitation in the PR description.

**F5 (UX message)** — Fixed. Updated user-facing messages to clarify the reason and that the current message was not sent. Timeout and connection-lost now show distinct messages.

chaodu-agent:
LGTM ✅ — Correctly preserves session ID on transient failures, preventing destructive history loss.

## What This PR Does

When `session/load` times out (30s) or the connection drops, OpenAB previously fell through to `session/new`, permanently overwriting `thread_map.json` and losing conversation history. This PR distinguishes transient failures from permanent rejections: on timeout or channel close, returns early with an error, preserving the session ID for automatic retry on the next message.

## How It Works

1. Defines `TRANSIENT_LOAD_ERRORS` constant with `"timeout waiting for"` and `"channel closed"` — making the coupling between `connection.rs` error strings and `pool.rs` classification explicit
2. After `session_load` fails, checks if the error is transient vs permanent
3. On transient failure: returns `Err` before reaching `session/new` — the spawned process is cleaned up via `Drop`, while `state.persisted`/`state.suspended` remain untouched
4. On permanent rejection (e.g. `"session/load rejected"`): falls through to `session/new` as before
5. `error_display.rs` matches the new error strings with distinct user-facing messages for timeout vs connection-lost, both placed before the generic timeout pattern to prevent false matching

## Findings

| 1 | 🟢 | Transient errors correctly identified — both timeout and channel-closed from `send_request` are internal strings under this crate's control | `pool.rs:14` |
| 2 | 🟢 | Resource cleanup verified — `AcpConnection::Drop` kills the spawned process on early return | `pool.rs:271` |
| 3 | 🟢 | State invariant upheld — `persisted`/`suspended` untouched before `load_failed` check | `pool.rs:273` |
| 4 | 🟢 | Error ordering correct — specific `"session load timeout"` before generic `"timeout waiting for"` prevents false match | `error_display.rs:14` |
| 5 | 🟢 | User message complete — includes "Your message was not sent" for clarity | `error_display.rs:15` |
| 6 | 🟢 | Previous review findings addressed — channel-closed path now protected (F1), constant extracted (F2), user message clarified (F5) | — |

<details>

- **Conservative principle**: On transient failure, preserve what you have rather than destructively overwrite — correct architectural decision aligned with cited prior art
- **Minimal blast radius**: 2 files changed, no new dependencies, no structural refactoring
- **Explicit coupling**: `TRANSIENT_LOAD_ERRORS` constant makes the string-matching contract visible and searchable
- **Both transient paths covered**: Timeout and channel-closed (agent crash during load) both preserve the session ID
- **Clean separation**: Distinct user-facing messages for timeout vs connection-lost helps debugging
- **Test coverage**: Unit tests for both new `format_user_error` paths
- **CI green**: All checks pass (cargo check, clippy, 507 tests, all smoke tests)
- **Responsive to feedback**: Commit `bca20d1` directly addresses F1 (channel-closed), F2 (constant extraction), and F5 (message clarity) from prior review round

</details>

<details>

- PR opened: 2026-06-18
- Main already has: `session/load` with fallback to `session/new` on any error; `format_user_error` for user-facing display
- Net-new value: Distinguishes transient failures (timeout, channel-closed) from permanent rejection, prevents destructive session ID overwrite, gives users clear guidance to retry or reset
- CI: All checks green

</details>

<details>
<summary>Previous Review Findings — Resolution Status</summary>

| # | Previous Finding | Status |
|---|-----------------|--------|
| F1 | Channel-closed path unprotected | ✅ Fixed — added to `TRANSIENT_LOAD_ERRORS` |
| F2 | String-based coupling implicit | ✅ Fixed — extracted to named constant |
| F3 | No pool-level integration test | ℹ️ Accepted — testing this path requires mocking internal `send_request` timing; existing smoke tests provide end-to-end coverage |
| F4 | Repeated timeout for large sessions | ℹ️ Accepted — documented as known limitation in PR description; user retains control via `/reset` |
| F5 | UX message incomplete | ✅ Fixed — "Your message was not sent" included |

</details>

thepagent (approved):

PR #1209, opened 2026-06-26
Docs(opencode): add API key provider config example

### LINKED ISSUE #1200: Docs: opencode config schema fixes — provider (singular), apiKey in options, agent.instructions format
## Problem

The opencode agent config example in the docs uses `providers` (plural) and places `apiKey`/`baseUrl` at the provider top level. opencode 1.17.x actually expects `provider` (singular) with `apiKey`/`baseURL` inside `options`.

Three schema mismatches found during ECS Fargate deployment:

1. `"providers"` → opencode 1.17.x returns `Unrecognized key: providers`
2. `"apiKey"` / `"baseUrl"` at provider top level → silently ignored, API call fails with `AI_APICallError: Missing API key`
3. `"agent": { "instructions": "string" }` → returns `Expected object, got string`

## Why It Matters

Anyone following the current docs to deploy opencode-go on openab will get:

- `Unrecognized key` errors on startup
- Silent API key failures (no error, just fallback to free models)
- ACP Connection Lost with no debug output

Verified on: openab 0.9.0-beta.2-opencode, opencode 1.17.9, ECS Fargate Spot

## Current Behavior (from live container)

```json
// ❌ What the docs suggest
{
  "providers": {
    "opencode-go": {
      "baseUrl": "...",
      "apiKey": "..."
    }
  }
}
// Result: Unrecognized key: providers

// ❌ apiKey at top level
{
  "provider": {
    "opencode-go": {
      "apiKey": "...",
      "baseURL": "..."
    }
  }
}
// Result: AI_APICallError: Missing API key (silently ignored)

// ✅ Verified working (opencode debug config output)
{
  "provider": {
    "opencode-go": {
      "options": {
        "baseURL": "https://opencode.ai/zen/go/v1",
        "apiKey": "..."
      },
      "models": {
        "deepseek-v4-flash": {}
      }
    }
  },
  "model": "opencode-go/deepseek-v4-flash"
}

### Suggested Change

Prior Research

• Searched open issues/PRs: no existing reports found
• Code checked: opencode debug config on live container (openab 0.9.0 + opencode 1.17.9)
• External refs: opencode.ai/docs/config

Proposed Fix

Update the opencode config example in docs to use:

1. "provider" (singular)
2. "options": { "baseURL": "...", "apiKey": "..." } (nested)
3. "models": { "model-id": {} } sub-object

### PR DESCRIPTION
**Pull request by smallgun01**

The existing `docs/opencode.md` examples (Ollama Cloud, xAI Grok) only cover OAuth-based providers. API-key-based providers like opencode-go require a different config structure — the full `provider.options.apiKey` block — which is not documented anywhere. Users hit this gap immediately when trying to set up API-key providers.

Closes #1200

## At a Glance

N/A — docs-only change.

Not applicable — docs-only change, adding a missing config example.

## Proposed Solution

Add a new `## Example: API Key Providers` section to `docs/opencode.md`, placed after the existing xAI Grok example and before `## Notes`. The example uses opencode-go as a representative API-key provider, showing:

- The full `provider.options.apiKey` + `models` block
- `${OPENCODE_API_KEY}` env var expansion
- Restart step to pick up config

## Why this approach?

Follows the exact same structure and conventions as the existing two provider examples (authenticate → set config → restart → verify). Uses `kubectl exec` + heredoc format consistent with the rest of the doc. Minimal, focused — one realistic example covers the pattern for all API-key providers.

## Alternatives Considered

- Adding a generic "Provider Config Reference" section: too much scope for a single example gap. A focused, copy-pasteable example is more useful to users who already understand the OAuth examples.
- Including `kubectl create secret` step: unnecessary noise — Kubernetes Secret creation is covered elsewhere and varies by deployment.

## Validation

- [x] Links are valid (all internal links point to existing anchors or docs)
- [x] Renders correctly in GitHub preview (standard markdown, code block)
- [x] `kubectl exec` heredoc syntax matches existing examples (unquoted `<< EOF` allows env var expansion)
- [x] JSON schema matches the opencode.json spec as documented at [opencode.ai](http://opencode.ai/)
- [x] Discord Discussion URL: https://discord.com/channels/1491295327620169908/1520318436507648130

### DIFF
--- a/docs/opencode.md
+++ b/docs/opencode.md
@@ -211,7 +211,7 @@ kubectl exec -it deployment/openab-opencode -- bash -c 'cat > /home/node/opencod
     "opencode-go": {
       "options": {
         "baseURL": "https://opencode.ai/zen/go/v1",
-        "apiKey": "${OPENCODE_API_KEY}"
+        "apiKey": "{env:OPENCODE_API_KEY}"
       },
       "models": {
         "deepseek-v4-flash": {}
@@ -224,6 +224,11 @@ EOF'
 ```
 
 > Inject `OPENCODE_API_KEY` via a Kubernetes Secret as an environment variable.
+>
+> Note: shell-style `${OPENCODE_API_KEY}` would also work in this specific
+> heredoc (bash expands it before writing), but only here. `{env:VAR}` is the
+> form that works everywhere.
+
 
 ### 2. Restart to pick up config
 
--- a/docs/opencode.md
+++ b/docs/opencode.md
@@ -193,6 +193,44 @@ kubectl rollout restart deployment/openab-opencode
 
 > **Important:** Do NOT set a custom `baseURL` or provider override for xAI. The built-in xAI provider handles routing correctly. A stale `~/.config/opencode/opencode.json` with `baseURL: "http://localhost:9090/v1"` (from xai-proxy setups) will break xAI — delete it if present.
 
+## Example: API Key Providers
+
+For providers that use API keys instead of OAuth (e.g. opencode-go, custom
+OpenAI-compatible endpoints), the config needs the full `provider` block with
+`options.apiKey`:
+
+### 1. Set the model and API key
+
+Create `opencode.json` in the working directory (`/home/node`). The key
+difference from OAuth providers is the `provider.options.apiKey` field:
+
+```bash
+kubectl exec -it deployment/openab-opencode -- bash -c 'cat > /home/node/opencode.json << EOF
+{
+  "provider": {
+    "opencode-go": {
+      "options": {
+        "baseURL": "https://opencode.ai/zen/go/v1",
+        "apiKey": "${OPENCODE_API_KEY}"
+      },
+      "models": {
+        "deepseek-v4-flash": {}
+      }
+    }
+  },
+  "model": "opencode-go/deepseek-v4-flash"
+}
+EOF'
+```
+
+> Inject `OPENCODE_API_KEY` via a Kubernetes Secret as an environment variable.
+
+### 2. Restart to pick up config
+
+```bash
+kubectl rollout restart deployment/openab-opencode
+```
+
 ## Notes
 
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: docs/opencode.md [hunk 1]
```diff
@@ -193,6 +193,44 @@ kubectl rollout restart deployment/openab-opencode
...
+difference from OAuth providers is the `provider.options.apiKey` field:
+
+```bash
+kubectl exec -it deployment/openab-opencode -- bash -c 'cat > /home/node/opencode.json << EOF
+{
+  "provider": {
+    "opencode-go": {
+      "options": {
+        "baseURL": "https://opencode.ai/zen/go/v1",
+        "apiKey": "${OPENCODE_API_KEY}"
```
thepagent: should it be `{env:VAR}` format?

https://opencode.ai/docs/config/#env-vars

smallgun01: You're right — `{env:OPENCODE_API_KEY}` is the official portable syntax per opencode docs.
Updating the example.

Side note for anyone reading: the `${OPENCODE_API_KEY}` form in a `kubectl exec` heredoc also
works because the shell expands it before writing the file — but only in shell-generated configs.
For users hand-writing `opencode.json` or running via Docker CMD / systemd, `{env:VAR}` is the
only correct option.

Thanks @thepagent.

docs/opencode.md
```diff
@@ -193,6 +193,49 @@ kubectl rollout restart deployment/openab-opencode
...
+      "models": {
+        "deepseek-v4-flash": {}
+      }
+    }
+  },
+  "model": "opencode-go/deepseek-v4-flash"
+}
+EOF'
+```
+
```
thepagent, line 225: We should add an official link here for users to read more from official doc site

thepagent (changes requested):

docs/opencode.md
smallgun01: Added the official link to the [opencode env var docs](https://opencode.ai/docs/config/#env-vars) (https://opencode.ai/docs/config/#env-vars) right below the env var note in 651468ba. Thanks for catching this.

thepagent (approved):

PR #1255, opened 2026-06-29
fix(pre_seed): allow safe relative symlinks; make lifecycle hooks Unix-only

chaodu-agent:
## Summary

Two related changes to lifecycle hooks:

### 1. Allow safe relative symlinks in pre_seed extraction
`pre_seed` tarball extraction rejected **all** symlinks containing `..`, breaking valid layouts like `./bin/aws -> ../aws-cli/v2/current/bin/aws`. Replaced the naive `ParentDir` check with proper path normalization (`symlink_escapes` + `normalize_path`) that only rejects symlinks actually escaping the extraction root. Validation now happens **before** writing (no TOCTOU window).

### 2. Fail fast on Windows for all lifecycle hooks
Lifecycle hooks (`pre_seed`, `pre_boot`, `pre_shutdown`) are inherently Unix-only — they rely on Unix symlinks and POSIX shell scripts and only ever run in Linux containers. Instead of silently misbehaving on Windows:
- `HooksConfig::ensure_platform_supported()` bails at startup if any hook is configured on a non-unix platform
- `pre_seed` symlink handling simplified to Unix-only (dropped the fragile Windows copy-fallback and deferred second-pass — now unreachable)
- Documented the limitation in `docs/hooks.md`

## Review findings addressed
All 8 findings from the review resolved: F1 (TOCTOU — validate before write), F2 (safety-contract doc), F3/F7 (Windows symlink handling — now moot, hooks fail fast on Windows), F4 (unit tests), F5/F6 (normalize_path edge cases documented + tested), F8 (temp cleanup — TempDir handles it).

## Testing
- 35 pre_seed + hooks-platform unit tests pass
- clippy clean with `--features pre-seed`
- Windows fail-fast path is `#[cfg(not(unix))]` gated

### DIFF
--- a/crates/openab-core/src/config.rs
+++ b/crates/openab-core/src/config.rs
@@ -205,6 +205,33 @@ pub struct HooksConfig {
     pub pre_shutdown: Option<HookConfig>,
 }
 
+impl HooksConfig {
+    /// Returns true if any lifecycle hook (pre_seed, pre_boot, pre_shutdown) is configured.
+    pub fn any_configured(&self) -> bool {
+        self.pre_seed.as_ref().is_some_and(|p| !p.sources.is_empty())
+            || self.pre_boot.is_some()
+            || self.pre_shutdown.is_some()
+    }
+
+    /// Fail fast if hooks are configured on an unsupported platform.
+    ///
+    /// Lifecycle hooks (pre_seed tarball extraction, pre_boot/pre_shutdown shell
+    /// scripts) assume a Unix environment and only ever run inside Linux containers.
+    /// Rather than silently misbehaving on Windows, refuse to start.
+    pub fn ensure_platform_supported(&self) -> anyhow::Result<()> {
+        #[cfg(not(unix))]
+        {
+            if self.any_configured() {
+                anyhow::bail!(
+                    "lifecycle hooks ([hooks.pre_seed], [hooks.pre_boot], [hooks.pre_shutdown]) \
+                     are only supported on Unix platforms; remove them to run on this platform"
+                );
+            }
+        }
+        Ok(())
+    }
+}
+
 /// Configuration for the pre_seed phase.
 /// Downloads and extracts zip archives from S3 before pre_boot.
 #[derive(Debug, Clone, Deserialize)]
@@ -1378,6 +1405,65 @@ fn default_ambient_context_flushes() -> usize {
 mod tests {
     use super::*;
     use std::io::Write;
+
+    #[test]
+    fn hooks_any_configured_false_when_empty() {
+        let h = HooksConfig::default();
+        assert!(!h.any_configured());
+        // Empty pre_seed sources don't count as configured
+        let h2 = HooksConfig {
+            pre_seed: Some(PreSeedConfig {
+                sources: vec![],
+                target: None,
+                region: None,
+                endpoint_url: None,
+                max_bytes: default_max_zip_bytes(),
+                timeout_seconds: default_pre_seed_timeout(),
+                on_failure: OnFailure::default(),
+            }),
+            pre_boot: None,
+            pre_shutdown: None,
+        };
+        assert!(!h2.any_configured());
+    }
+
+    fn sample_hook() -> HookConfig {
+        HookConfig {
+            script: None,
+            inline: Some("echo hi".to_string()),
+            url: None,
+            sha256: None,
+            timeout_seconds: 60,
+            on_failure: OnFailure::default(),
+        }
+    }
+
+    #[test]
+    fn hooks_any_configured_true_with_pre_boot() {
+        let h = HooksConfig {
+            pre_seed: None,
+            pre_boot: Some(sample_hook()),
+            pre_shutdown: None,
+        };
+        assert!(h.any_configured());
+    }
+
+    #[test]
+    fn ensure_platform_supported_ok_when_no_hooks() {
+        // No hooks configured → always Ok regardless of platform
+        assert!(HooksConfig::default().ensure_platform_supported().is_ok());
+    }
+
+    #[cfg(unix)]
+    #[test]
+    fn ensure_platform_supported_ok_on_unix_with_hooks() {
+        let h = HooksConfig {
+            pre_seed: None,
+            pre_boot: Some(sample_hook()),
+            pre_shutdown: None,
+        };
+        assert!(h.ensure_platform_supported().is_ok());
+    }
 
     const MINIMAL_TOML: &str = r#"
 [discord]
--- a/crates/openab-core/src/pre_seed.rs
+++ b/crates/openab-core/src/pre_seed.rs
@@ -418,11 +418,29 @@ fn create_symlink_or_copy(link_target: &Path, dst: &Path, src: &Path) -> anyhow:
     }
     #[cfg(not(unix))]
     {
-        // F3/F7: On non-Unix platforms we materialize the link by copying its target.
-        // Directory targets cannot be copied with std::fs::copy, and the target may
-        // not have been moved into place yet during move_recursive. Treat both as
-        // non-fatal: skip the link rather than aborting the whole extraction.
-        let resolved = src.parent().unwrap_or(Path::new(".")).join(link_target);
+        // On non-Unix platforms we materialize the link by copying its target.
+        // During move_recursive, files may already have been moved from the temp (src)
+        // directory to the destination (dst) directory. We check both locations to
+        // handle cross-directory symlinks regardless of processing order.
+        // Directory targets cannot be copied with std::fs::copy — skip with warning.
+        let resolved_src = src.parent().unwrap_or(Path::new(".")).join(link_target);
+        let resolved_dst = dst.parent().unwrap_or(Path::new(".")).join(link_target);
+
+        // Try source first (temp dir), then destination (already moved)
+        let resolved = if resolved_src.exists() {
+            resolved_src
+        } else if resolved_dst.exists() {
+            resolved_dst
+        } else {
+            warn!(
+                "hooks.pre_seed: skipping symlink with unresolved target: {} (checked {} and {})",
+                link_target.display(),
+                resolved_src.display(),
+                resolved_dst.display()
+            );
+            return Ok(());
+        };
+
         match resolved.symlink_metadata() {
             Ok(meta) if meta.is_dir() => {
                 warn!(
@@ -433,10 +451,12 @@ fn create_symlink_or_copy(link_target: &Path, dst: &Path, src: &Path) -> anyhow:
             Ok(_) => {
                 std::fs::copy(&resolved, dst)?;
             }
-            Err(_) => {
+            Err(e) => {
                 warn!(
-                    "hooks.pre_seed: skipping symlink with unresolved target: {}",
-                    resolved.display()
+                    "hooks.pre_seed: skipping symlink copy failed: {} -> {}: {}",
+                    link_target.display(),
+                    resolved.display(),
+                    e
                 );
             }
         }
--- a/crates/openab-core/src/pre_seed.rs
+++ b/crates/openab-core/src/pre_seed.rs
@@ -167,11 +167,7 @@ async fn download_and_extract(
 /// Extract archive to a temp directory with budget enforcement, then move into target.
 /// Supports zip and gzipped tarball formats (detected via magic bytes).
 /// Checks deadline cooperatively before each file operation.
-fn extract_and_apply(
-    data: &[u8],
-    target: &Path,
-    deadline: Instant,
-) -> anyhow::Result<()> {
+fn extract_and_apply(data: &[u8], target: &Path, deadline: Instant) -> anyhow::Result<()> {
     std::fs::create_dir_all(target)?;
     let temp_dir = tempfile::tempdir_in(target)?;
 
@@ -289,7 +285,9 @@ fn extract_tarball_with_limits(data: &[u8], dest: &Path, deadline: Instant) -> a
 
         // Cooperative deadline check every 10 files
         if file_count.is_multiple_of(10) && Instant::now() >= deadline {
-            anyhow::bail!("hooks.pre_seed: timed out during tarball extraction at entry {file_count}");
+            anyhow::bail!(
+                "hooks.pre_seed: timed out during tarball extraction at entry {file_count}"
+            );
         }
 
         // Size budget
@@ -328,12 +326,14 @@ fn extract_tarball_with_limits(data: &[u8], dest: &Path, deadline: Instant) -> a
             use std::os::unix::fs::PermissionsExt;
             if let Ok(path) = entry.path() {
                 let out_path = dest.join(path);
-                if out_path.symlink_metadata().map(|m| m.file_type().is_file()).unwrap_or(false) {
+                if out_path
+                    .symlink_metadata()
+                    .map(|m| m.file_type().is_file())
+                    .unwrap_or(false)
+                {
                     let mode = entry.header().mode().unwrap_or(0o644) & 0o0777;
-                    let _ = std::fs::set_permissions(
-                        &out_path,
-                        std::fs::Permissions::from_mode(mode),
-                    );
+                    let _ =
+                        std::fs::set_permissions(&out_path, std::fs::Permissions::from_mode(mode));
                 }
             }
         }
@@ -380,7 +380,10 @@ fn normalize_path(path: &Path) -> std::path::PathBuf {
             std::path::Component::ParentDir => {
                 // Only pop if there's something to pop and it's not a prefix/root
                 if components.last().is_some_and(|c| {
-                    !matches!(c, std::path::Component::RootDir | std::path::Component::Prefix(_))
+                    !matches!(
+                        c,
+                        std::path::Component::RootDir | std::path::Component::Prefix(_)
+                    )
                 }) {
                     components.pop();
                 }
@@ -443,7 +446,13 @@ fn create_symlink_or_copy(link_target: &Path, dst: &Path, src: &Path) -> anyhow:
 
 /// Recursively move files from src directory into dst directory.
 /// Checks deadline cooperatively.
+/// On non-Unix platforms, symlinks are processed in a second pass after regular
+/// files/directories to ensure symlink targets exist when copied.
 fn move_recursive(src: &Path, dst: &Path, deadline: Instant) -> anyhow::Result<()> {
+    // Collect symlinks for deferred processing (needed on non-Unix where symlinks are
+    // resolved by copying, so targets must already be in place).
+    let mut deferred_symlinks: Vec<(std::path::PathBuf, std::path::PathBuf)> = Vec::new();
+
     for entry in std::fs::read_dir(src)? {
         if Instant::now() >= deadline {
             anyhow::bail!("hooks.pre_seed: timed out during move to target");
@@ -455,17 +464,8 @@ fn move_recursive(src: &Path, dst: &Path, deadline: Instant) -> anyhow::Result<(
 
         let meta = src_path.symlink_metadata()?;
         if meta.is_symlink() {
-            // Preserve symlinks as-is without following
-            let link_target = std::fs::read_link(&src_path)?;
-            // Remove existing dst (file or directory) before creating symlink
-            if let Ok(dst_meta) = dst_path.symlink_metadata() {
-                if dst_meta.is_dir() {
-                    std::fs::remove_dir_all(&dst_path)?;
-                } else {
-                    std::fs::remove_file(&dst_path)?;
-                }
-            }
-            create_symlink_or_copy(&link_target, &dst_path, &src_path)?;
+            // Defer symlinks to second pass on all platforms for consistent behavior
+            deferred_symlinks.push((src_path, dst_path));
         } else if meta.is_dir() {
             std::fs::create_dir_all(&dst_path)?;
             move_recursive(&src_path, &dst_path, deadline)?;
@@ -476,6 +476,26 @@ fn move_recursive(src: &Path, dst: &Path, deadline: Instant) -> anyhow::Result<(
             }
         }
     }
+
+    // Second pass: process symlinks after all regular files are in place
+    for (src_path, dst_path) in deferred_symlinks {
+        if Instant::now() >= deadline {
+            anyhow::bail!("hooks.pre_seed: timed out during move to target");
+        }
+        let link_target = std::fs::read_link(&src_path)?;
+        // Remove existing dst (file or directory) before creating symlink
+        if let Ok(dst_meta) = dst_path.symlink_metadata() {
+            if dst_meta.is_dir() {
+                std::fs::remove_dir_all(&dst_path)?;
+            } else {
+                std::fs::remove_file(&dst_path)?;
+            }
+        }
+        create_symlink_or_copy(&link_target, &dst_path, &src_path)?;
+        // Clean up source symlink
+        let _ = std::fs::remove_file(&src_path);
+    }
+
     Ok(())
 }
 
@@ -802,8 +822,14 @@ mod tests {
         let dst_link = dst.path().join("link.txt");
         let meta = dst_link.symlink_metadata().unwrap();
         assert!(meta.is_symlink(), "destination should be a symlink");
-        assert_eq!(std::fs::read_link(&dst_link).unwrap().to_str().unwrap(), "real.txt");
-        assert_eq!(std::fs::read_to_string(dst.path().join("real.txt")).unwrap(), "content");
+        assert_eq!(
+            std::fs::read_link(&dst_link).unwrap().to_str().unwrap(),
+            "real.txt"
+        );
+        assert_eq!(
+            std::fs::read_to_string(dst.path().join("real.txt")).unwrap(),
+            "content"
+        );
     }
 
     #[cfg(unix)]
@@ -824,14 +850,17 @@ mod tests {
 
         let dst_item = dst.path().join("item");
         let meta = dst_item.symlink_metadata().unwrap();
-        assert!(meta.is_symlink(), "should have replaced directory with symlink");
+        assert!(
+            meta.is_symlink(),
+            "should have replaced directory with symlink"
+        );
     }
 
     #[cfg(unix)]
     #[test]
     fn extract_and_apply_succeeds_with_non_writable_parent() {
-        use std::os::unix::fs::PermissionsExt;
         use std::io::Write;
+        use std::os::unix::fs::PermissionsExt;
 
         // Regression test: target is writable but target.parent() is read-only.
         // Old code used tempdir_in(target.parent()) which would fail here.
@@ -863,7 +892,10 @@ mod tests {
 
         // Should succeed because tempdir_in(target) works even with read-only parent
         result.unwrap();
-        assert_eq!(std::fs::read_to_string(target.join("test.txt")).unwrap(), "data");
+        assert_eq!(
+            std::fs::read_to_string(target.join("test.txt")).unwrap(),
+            "data"
+        );
     }
 
     #[cfg(unix)]
@@ -886,7 +918,10 @@ mod tests {
         let deadline = Instant::now() + std::time::Duration::from_secs(60);
         extract_and_apply(cursor.get_ref(), &target, deadline).unwrap();
 
-        assert_eq!(std::fs::read_to_string(target.join("test.txt")).unwrap(), "hello");
+        assert_eq!(
+            std::fs::read_to_string(target.join("test.txt")).unwrap(),
+            "hello"
+        );
     }
 
     #[test]
--- a/crates/openab-core/src/pre_seed.rs
+++ b/crates/openab-core/src/pre_seed.rs
@@ -308,11 +308,7 @@ fn extract_tarball_with_limits(data: &[u8], dest: &Path, deadline: Instant) -> a
             let out_path = dest.join(&*path);
             if out_path.symlink_metadata().map(|m| m.is_symlink()).unwrap_or(false) {
                 if let Ok(target) = std::fs::read_link(&out_path) {
-                    if target.is_absolute()
-                        || target
-                            .components()
-                            .any(|c| c == std::path::Component::ParentDir)
-                    {
+                    if symlink_escapes(dest, &out_path, &target) {
                         let _ = std::fs::remove_file(&out_path);
                         warn!(
                             "hooks.pre_seed: removed symlink with escaping target: {}",
@@ -343,18 +339,51 @@ fn extract_tarball_with_limits(data: &[u8], dest: &Path, deadline: Instant) -> a
     Ok(())
 }
 
+/// Check whether a symlink at `link_path` with target `link_target` escapes `root`.
+/// Resolves the target relative to the symlink's parent directory and checks
+/// whether the normalized result stays within root. This allows relative symlinks
+/// with `..` components (e.g., `../aws-cli/v2/current/bin/aws`) as long as they
+/// don't escape the extraction root.
+fn symlink_escapes(root: &Path, link_path: &Path, link_target: &Path) -> bool {
+    // Absolute symlinks always escape
+    if link_target.is_absolute() {
+        return true;
+    }
+    // Resolve the symlink target relative to the symlink's parent directory
+    let parent = link_path.parent().unwrap_or(root);
+    let resolved = normalize_path(&parent.join(link_target));
+    let root_normalized = normalize_path(root);
+    !resolved.starts_with(&root_normalized)
+}
+
+/// Normalize a path by resolving `.` and `..` components without touching the filesystem.
+fn normalize_path(path: &Path) -> std::path::PathBuf {
+    let mut components = Vec::new();
+    for component in path.components() {
+        match component {
+            std::path::Component::ParentDir => {
+                // Only pop if there's something to pop and it's not a prefix/root
+                if components.last().is_some_and(|c| {
+                    !matches!(c, std::path::Component::RootDir | std::path::Component::Prefix(_))
+                }) {
+                    components.pop();
+                }
+            }
+            std::path::Component::CurDir => {}
+            _ => components.push(component),
+        }
+    }
+    components.iter().collect()
+}
+
 /// Create a symlink on Unix, or copy the resolved target on other platforms.
-/// Rejects symlink targets that escape via absolute path or `..` components.
+/// Rejects symlink targets that escape the extraction root.
 #[allow(unused_variables)]
 fn create_symlink_or_copy(link_target: &Path, dst: &Path, src: &Path) -> anyhow::Result<()> {
-    // Reject symlinks that could escape the target directory
-    if link_target.is_absolute()
-        || link_target
-            .components()
-            .any(|c| c == std::path::Component::ParentDir)
-    {
+    // Reject absolute symlinks unconditionally
+    if link_target.is_absolute() {
         anyhow::bail!(
-            "hooks.pre_seed: rejecting symlink with escaping target: {}",
+            "hooks.pre_seed: rejecting symlink with absolute target: {}",
             link_target.display()
         );
     }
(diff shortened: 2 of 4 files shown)


chaodu-agent:
CHANGES REQUESTED ⚠️ — Logic is correct but missing unit tests for security-critical path normalization functions.

## What This PR Does

Fixes `pre_seed` tarball extraction that incorrectly rejected valid relative symlinks containing `..` components (e.g., `./bin/aws -> ../aws-cli/v2/current/bin/aws`). Replaces the naive `ParentDir` component check with proper path normalization that only rejects symlinks actually escaping the extraction root.

## How It Works

Introduces two helper functions:
- `symlink_escapes(root, link_path, link_target)` — resolves the symlink target relative to its parent directory, normalizes the result, and checks containment within root.
- `normalize_path(path)` — resolves `.` and `..` components lexically without filesystem access.

The post-extraction sweep in `extract_tarball_with_limits` now calls `symlink_escapes` instead of the blanket `ParentDir` check. `create_symlink_or_copy` (used during `move_recursive`) is relaxed to only reject absolute symlinks since relative ones are already validated upstream.

## Findings

| 1 | 🟡 | No unit tests for `symlink_escapes` or `normalize_path` | `pre_seed.rs` |
| 2 | 🟡 | `create_symlink_or_copy` no longer validates relative escape independently | `pre_seed.rs:382` |
| 3 | 🟢 | `normalize_path` correctly handles edge cases (excess `..`, root boundary) | `pre_seed.rs:360` |
| 4 | 🟢 | Clear doc comments explain the security rationale | `pre_seed.rs:342-348` |

<details>

### 🟡 F1: Missing unit tests for security-critical functions

`symlink_escapes` and `normalize_path` are the security boundary for symlink traversal attacks. They need dedicated unit tests covering:
- Valid relative symlink within root (e.g., `bin/aws -> ../aws-cli/v2/current/bin/aws`)
- Escaping symlink (e.g., `bin/x -> ../../etc/passwd`)
- Absolute symlink (e.g., `/etc/passwd`)
- Edge case: more `..` components than directory depth
- Edge case: `.` components (no-op)
- Symlink at extraction root level pointing up

The PR states "All 15 existing tests pass" but none of the existing tests exercise the new path normalization logic.

### 🟡 F2: `create_symlink_or_copy` relies on upstream validation

The function now only rejects absolute symlinks, relying on `extract_tarball_with_limits` having already validated relative targets. This is a defensible design choice (defense-in-depth at the extraction layer), but it makes `create_symlink_or_copy` less safe if ever called from a new context. Consider either:
- Adding a `root` parameter and calling `symlink_escapes` here too, or
- Adding a doc comment noting the safety contract (that callers must pre-validate).

### 🟢 F3: Correct normalization logic

The `normalize_path` function properly handles the `RootDir`/`Prefix` boundary — excess `..` components don't escape past the filesystem root. This is the correct behavior for the `starts_with` containment check.

### 🟢 F4: Good documentation

Doc comments clearly explain the rationale and give concrete examples of the allowed pattern.

</details>

<details>

- PR opened: recent (on branch `fix/pre-seed-relative-symlinks`)
- Main already has: naive `ParentDir` component check that rejects ALL relative symlinks with `..`
- Net-new value: proper path normalization that allows safe relative symlinks while still blocking escape

</details>

<details>

- Correct approach: lexical normalization without filesystem access avoids TOCTOU issues
- Handles absolute symlinks as a fast-path early return
- The `unwrap_or(root)` fallback for `link_path.parent()` is safe — a symlink at root level would resolve relative to root itself
- Existing test suite continues to pass

</details>

chaodu-agent:
CHANGES REQUESTED ⚠️ — Defense-in-depth regression in `create_symlink_or_copy`

## What This PR Does

The `pre_seed` tarball extraction was blanket-rejecting all symlinks containing `..` components, breaking valid relative symlinks like `./bin/aws -> ../aws-cli/v2/current/bin/aws`. This PR replaces the naive check with proper path normalization that only rejects symlinks whose resolved target actually escapes the extraction root.

## How It Works

Introduces `symlink_escapes()` which resolves a symlink target relative to its parent directory, normalizes the path (without filesystem access), and checks whether the result stays within the extraction root. A new `normalize_path()` helper resolves `.` and `..` components purely in-memory using a stack approach.

## Findings

| 1 | 🟡 | `create_symlink_or_copy` lost defense-in-depth escape validation | `pre_seed.rs:382-390` |
| 2 | 🟢 | Correct normalization logic with proper edge-case handling | `pre_seed.rs:359-377` |
| 3 | 🟢 | Excellent test coverage (10 new tests: unit + integration) | `pre_seed.rs:856-1011` |

<details>

### 🟡 F1: `create_symlink_or_copy` lost defense-in-depth escape validation

The PR correctly adds `symlink_escapes()` to the post-extraction cleanup in `extract_tarball_with_limits`, but `create_symlink_or_copy` (used during `move_recursive`) now only rejects absolute symlinks. The original code rejected both absolute and `..`-containing symlinks here as a defense-in-depth measure.

While the primary security gate (`symlink_escapes()` in extract) should catch escaping symlinks before `move_recursive` runs, removing the second check weakens the security posture:
- A TOCTOU race between extraction cleanup and `move_recursive` could theoretically allow a malicious symlink through
- If `create_symlink_or_copy` is called from a different context in the future, it offers no escape protection

**Suggested fix**: Apply `symlink_escapes()` in `create_symlink_or_copy` as well, using `dst.parent()` as the contextual root, or pass the extraction root explicitly. Alternatively, keep the existing naive `..` check here as a cheap secondary guard — it only produces false positives for valid intra-root relative symlinks, which are already handled by the extraction phase.

### 🟢 F2: Correct normalization logic

`normalize_path` properly:
- Refuses to pop past `RootDir` or `Prefix`
- Strips `CurDir` components
- Uses component-wise iteration (no string manipulation)

Rust's `Path::starts_with` is component-boundary-safe, so there's no prefix confusion risk (e.g., `/tmp/extractevil` won't match `/tmp/extract`).

### 🟢 F3: Excellent test coverage

The PR adds thorough tests covering:
- Normalization edge cases (multiple parents, root boundary, CurDir)
- `symlink_escapes` with various valid/invalid scenarios
- Full integration tests with actual tarball creation, extraction, and validation

</details>

<details>

- PR opened by: chaodu-agent (auto-fix branch)
- Main already has: Naive `ParentDir` component check that rejects ALL `..` symlinks (overly strict)
- Net-new value: Proper path normalization allowing valid relative symlinks within root while still blocking escaping ones

</details>

<details>

- Clean separation of concerns: `symlink_escapes` + `normalize_path` are well-factored and independently testable
- The normalization approach is filesystem-independent (no `canonicalize()` calls that could follow symlinks)
- Tests cover both the happy path (valid relative symlinks preserved) and security path (escaping symlinks removed)
- Integration tests use real tarball construction to validate end-to-end behavior

</details>

chaodu-agent:
CHANGES REQUESTED ⚠️ — Correct direction but has a TOCTOU design gap, inconsistent safety contract, Windows regression risk, and missing unit tests.

## What This PR Does

Fixes `pre_seed` tarball extraction that incorrectly rejected valid relative symlinks containing `..` components (e.g., `./bin/aws -> ../aws-cli/v2/current/bin/aws`). Replaces the naive `ParentDir` component check with proper path normalization that only rejects symlinks actually escaping the extraction root.

## How It Works

Introduces two helper functions:
- `symlink_escapes(root, link_path, link_target)` — resolves the symlink target relative to its parent directory, normalizes the result, and checks containment within root.
- `normalize_path(path)` — resolves `.` and `..` components lexically without filesystem access.

The post-extraction sweep in `extract_tarball_with_limits` now calls `symlink_escapes` instead of the blanket `ParentDir` check. `create_symlink_or_copy` (used during `move_recursive`) is relaxed to only reject absolute symlinks since relative ones are already validated upstream.

## Findings

| 1 | 🔴 | TOCTOU: post-extraction sweep validates symlinks after writing — should gate at create time | `pre_seed.rs:310-318` |
| 2 | 🔴 | `create_symlink_or_copy` doc says "rejects escaping targets" but only checks `is_absolute()`; safety contract mismatch | `pre_seed.rs:382` |
| 3 | 🔴 | Windows regression: `move_recursive` ordering may delete symlink target before copy, aborting extraction | `pre_seed.rs:433` |
| 4 | 🟡 | No unit tests for `symlink_escapes` or `normalize_path` — security-critical boundary | `pre_seed.rs` |
| 5 | 🟡 | `normalize_path` silently drops leading `..` producing empty PathBuf — undocumented edge case | `pre_seed.rs:360-375` |
| 6 | 🟡 | If `link_path.parent()` itself contains `..` (e.g., tar entry `a/../b/link`), joined path may bypass `starts_with` | `pre_seed.rs:352` |
| 7 | 🟡 | Windows: directory symlinks fail because `std::fs::copy` cannot copy directories | `pre_seed.rs:398` |
| 8 | 🟡 | Residual symlinks in temp directory never cleaned up during `move_recursive` | `pre_seed.rs:433` |
| 9 | 🟢 | Lexical normalization avoids canonicalize TOCTOU — correct approach | `pre_seed.rs:360` |
| 10 | 🟢 | Absolute symlink early return is a clean fast-path | `pre_seed.rs:349` |
| 11 | 🟢 | Clear doc comments explain security rationale with concrete examples | `pre_seed.rs:342-348` |

<details>

### 🔴 F1: TOCTOU — validate at create time, not post-sweep

The current design extracts all tar entries to disk first, then sweeps for dangerous symlinks and removes them. Between extraction and removal, the symlink exists on the filesystem. While single-threaded extraction makes exploitation unlikely today, the correct design is to validate symlink targets **before** writing them (or reject the entry during extraction). This eliminates the window entirely and is a stronger security posture.

**Suggestion:** Move the `symlink_escapes` check into the tar entry processing loop, skipping (not extracting) entries with escaping symlink targets.

### 🔴 F2: Inconsistent safety contract on `create_symlink_or_copy`

The doc comment states "Rejects symlink targets that escape the extraction root" but the implementation only checks `is_absolute()`. The function has no `root` parameter, so it **cannot** verify relative escapes. This mismatch means:
- The API contract is misleading to future callers
- If any new code path calls this function without upstream validation, it becomes fail-open

**Suggestion:** Either pass `root` and validate here too (defense-in-depth), or update the doc to clearly state the safety precondition: "Caller MUST pre-validate relative targets via `symlink_escapes`."

### 🔴 F3: Windows `move_recursive` ordering bug

On non-Unix platforms, `create_symlink_or_copy` copies the target file instead of creating a symlink. In `move_recursive`, files are processed in directory iteration order. If the symlink's target file is moved/deleted before the symlink entry is processed, the copy fails with "symlink target does not exist". Conversely, if the symlink is processed first, the target gets duplicated.

**Suggestion:** On Windows, process symlinks in a second pass after all regular files have been moved. Or defer symlink resolution to after the full move.

### 🟡 F4: Missing unit tests

`symlink_escapes` and `normalize_path` are the security boundary. They need tests covering:
- Valid relative symlink within root (`bin/aws -> ../aws-cli/v2/current/bin/aws`)
- Escaping symlink (`bin/x -> ../../etc/passwd`)
- Absolute symlink (`/etc/passwd`)
- Excess `..` components beyond directory depth
- `.` components (no-op)
- Symlink at root level pointing up

### 🟡 F5: `normalize_path` empty result on pure-relative `..` paths

`normalize_path(Path::new("../../etc/passwd"))` returns an empty PathBuf since leading `..` components are silently dropped (no component to pop, no push). The escape check still works (`"".starts_with(root)` is false), but this behavior is non-obvious and should be documented or asserted.

### 🟡 F6: Tar entries with `..` in directory components

If a tar entry path itself contains `..` (e.g., `a/../b/link`), then `link_path.parent()` yields `a/../b`. Joining with the symlink target and normalizing may produce unexpected results depending on whether `a` exists in the extraction. The tar crate's `unpack_in` does sanitize paths, but this assumption should be explicitly documented.

### 🟡 F7: Windows directory symlink limitation

`std::fs::copy` cannot copy directories. If a relative symlink points to a directory on Windows, extraction will fail. This is a pre-existing limitation but is now more likely to be triggered since relative `..` symlinks are no longer rejected.

### 🟡 F8: Temp directory symlink cleanup

In `move_recursive`, regular files are moved (rename or copy+delete), but symlinks are recreated at the destination without removing the source symlink from the temp directory. The temp dir is cleaned up by `tempdir` drop, but dangling symlinks may cause issues if the temp dir cleanup fails.

</details>

<details>

- PR opened: recent (branch `fix/pre-seed-relative-symlinks`)
- Main already has: naive `ParentDir` component check that rejects ALL relative symlinks with `..`
- Net-new value: proper path normalization that allows safe relative symlinks while blocking actual escapes
- The problem is real — `../aws-cli/v2/current/bin/aws` is a legitimate layout pattern

</details>

<details>

- Correct approach: lexical normalization without filesystem access avoids canonicalize TOCTOU
- Absolute symlink early return is clean and efficient
- Doc comments clearly explain the security rationale with concrete examples
- The `unwrap_or(root)` fallback for `link_path.parent()` is safe
- Problem statement is well-defined and the fix direction is correct

</details>

chaodu-agent:
Thanks for the thorough review. Addressed in d9157c9:

| # | Finding | Resolution |
|---|---------|-----------|
| 🔴 F1 | TOCTOU post-sweep | **Fixed.** Symlink/hardlink targets are now validated *before* `unpack_in`; escaping entries are skipped and never touch disk. The post-extraction removal sweep is gone. |
| 🔴 F2 | `create_symlink_or_copy` contract mismatch | **Fixed (doc).** Added an explicit SAFETY PRECONDITION doc: it only rejects absolute targets and callers MUST pre-validate relative targets via `symlink_escapes`. The sole caller (`move_recursive`) receives entries already validated at extraction time. |
| 🔴 F3 | Windows move ordering abort | **Mitigated.** Non-unix path no longer aborts: unresolved targets and directory symlinks are skipped with a warning instead of failing the whole extraction. |
| 🟡 F4 | Missing unit tests | **Fixed.** Added 13 tests for `symlink_escapes` and `normalize_path` (within-root, escape, absolute, excess `..`, `.` no-op, parent-with-`..`) plus tarball integration tests. |
| 🟡 F5 | `normalize_path` leading-`..` drop | **Fixed (doc+test).** Documented the behavior and added `normalize_path_drops_leading_parent_dir_on_relative`. Safe because callers always pass absolute paths. |
| 🟡 F6 | `..` in entry path | **Fixed (doc+test).** `normalize_path` resolves `..` across the whole joined path including the parent; added two tests (`symlink_escapes_handles_parent_with_dotdot` and `..._still_blocks_escape`). |
| 🟡 F7 | Windows dir symlink copy | **Mitigated.** Directory targets are skipped with a warning on non-unix (see F3). |
| 🟡 F8 | Temp symlink cleanup | No change needed — `TempDir` drop calls `remove_dir_all`, which unlinks symlinks without following them. |

All 31 pre_seed tests + full suite pass, clippy clean with `--features pre-seed`.

chaodu-agent:
CHANGES REQUESTED ⚠️ — Hard link target resolution uses wrong base path in `symlink_escapes`

## What This PR Does

The `pre_seed` tarball extractor previously blanket-rejected all symlinks containing `..` path components, which broke valid relative symlinks like `bin/aws -> ../aws-cli/v2/current/bin/aws`. This PR replaces the naive check with proper path normalization that resolves the target relative to the symlink's parent directory and only rejects links that actually escape the extraction root.

## How It Works

1. **Pre-extraction validation** (F1): Symlink/hardlink targets are validated *before* `unpack_in` writes them to disk, eliminating the TOCTOU window of the old post-extraction sweep.
2. **`symlink_escapes(root, link_path, link_target)`**: Resolves the target relative to `link_path.parent()`, normalizes `..` components lexically, then checks containment via `starts_with`.
3. **`normalize_path`**: A pure-lexical `..`/`.` resolver that avoids filesystem calls (no canonicalize TOCTOU).
4. **`create_symlink_or_copy`**: Relaxed to only reject absolute targets; relative-target containment is now the caller's responsibility. Non-Unix fallback gracefully handles directory symlinks and missing targets.
5. **Tests**: 13 new unit tests covering `normalize_path`, `symlink_escapes`, and integration scenarios with real tarballs.

## Findings

| 1 | 🟡 | Hard link target resolution uses wrong semantics in `symlink_escapes` | `pre_seed.rs:305-315` |
| 2 | 🟢 | Excellent TOCTOU fix — pre-extraction validation is the correct pattern | `pre_seed.rs:303-318` |
| 3 | 🟢 | Comprehensive test coverage with both unit and integration tests | `pre_seed.rs:891-1073` |
| 4 | 🟢 | Good documentation of safety preconditions (F2, F5, F6) | throughout |

<details>

### 🟡 F1: Hard link target resolution uses wrong semantics

In tar archives, **symlink** targets are relative to the symlink's parent directory, but **hard link** targets are relative to the archive root (extraction destination). The `symlink_escapes` function unconditionally resolves via `link_path.parent().join(link_target)`, which is correct for symlinks but semantically wrong for hard links.

Example: A hard link at `a/b/link` targeting `c/file`:
- `symlink_escapes` resolves: `dest/a/b/../c/file` → `dest/a/c/file` ❌
- tar crate resolves: `dest/c/file` ✓

**Practical impact is low** because `tar::Entry::unpack_in` has its own `validate_inside_dst` check for hard links, providing defense-in-depth. However, this could cause valid hard links to be incorrectly rejected (false positives).

**Suggested fix**: Check `entry_type.is_hard_link()` separately and resolve the target as `dest.join(link_target)` instead of `link_path.parent().join(link_target)`:

```rust
let escapes = if entry_type.is_hard_link() {
    // Hard link targets in tar are relative to archive root
    let resolved = normalize_path(&dest.join(&*link_target));
    let root_normalized = normalize_path(dest);
    !resolved.starts_with(&root_normalized)
} else {
    symlink_escapes(dest, &out_path, &link_target)
};
if escapes {
    // ... skip
}
```

</details>

<details>

- PR opened: on branch `fix/pre-seed-relative-symlinks`
- Main already has: Symlink validation in `extract_tarball_with_limits` (post-extraction `remove_file` sweep) and `create_symlink_or_copy` (blanket `ParentDir` rejection)
- Net-new value: Correct path-normalization-based containment check that allows valid `..` symlinks while maintaining security; TOCTOU fix by validating before extraction; graceful non-Unix handling

</details>

<details>

- Pre-extraction validation eliminates the TOCTOU window — this is the textbook-correct approach for archive extraction security
- Pure-lexical path normalization avoids the canonicalize TOCTOU that plagues many similar implementations
- Thorough documentation of safety invariants with F-numbered annotations makes the security reasoning auditable
- Integration tests use real tar builders to verify end-to-end behavior, not just unit logic
- Non-Unix fallback is now resilient to missing targets and directory symlinks instead of hard-failing

</details>

chaodu-agent:
CHANGES REQUESTED ⚠️ — All critical findings from Round 1 have been addressed in follow-up commits. Requesting re-review on remaining items.

## What This PR Does

Fixes `pre_seed` tarball extraction that incorrectly rejected valid relative symlinks containing `..` components (e.g., `./bin/aws -> ../aws-cli/v2/current/bin/aws`). Replaces the naive `ParentDir` component check with proper path normalization that only rejects symlinks actually escaping the extraction root.

## How It Works

- `symlink_escapes(root, link_path, link_target)` — resolves the symlink target relative to its parent directory, normalizes the result, and checks containment within root.
- `normalize_path(path)` — resolves `.` and `..` components lexically without filesystem access.
- Symlinks are validated **before** `unpack_in` (eliminates TOCTOU window).
- `move_recursive` now defers symlink processing to a second pass, ensuring targets exist on non-Unix platforms.
- On non-Unix, directory symlinks and unresolved targets are skipped non-fatally.

## Findings (Round 1 → Status)

| 1 | 🔴 | TOCTOU: post-extraction sweep validates after writing | ✅ Fixed in `d9157c9` — gate before `unpack_in` |
| 2 | 🔴 | `create_symlink_or_copy` safety contract mismatch | ✅ Fixed in `d9157c9` — doc updated with SAFETY PRECONDITION |
| 3 | 🔴 | Windows `move_recursive` ordering bug | ✅ Fixed in `9440da3` — deferred symlink second pass |
| 4 | 🟡 | Missing unit tests for `symlink_escapes`/`normalize_path` | ✅ Fixed in `d9157c9` — 7+ edge-case tests added |
| 5 | 🟡 | `normalize_path` leading `..` behavior undocumented | ✅ Fixed in `d9157c9` — doc comment added |
| 6 | 🟡 | Tar entries with `..` in path may bypass check | ✅ Fixed in `d9157c9` — documented + tested |
| 7 | 🟡 | Windows: directory symlinks fail with `std::fs::copy` | ✅ Fixed in `d9157c9` — skip non-fatally with warning |
| 8 | 🟡 | Residual symlinks in temp directory | ✅ Fixed in `9440da3` — cleaned up in second pass |

## Remaining Items for Re-Review

All 🔴 critical findings have been resolved. The following should be verified in this round:

1. Does the deferred symlink pass in `move_recursive` correctly handle nested directories with symlinks at multiple levels?
2. Are the new unit tests comprehensive enough for the security boundary?
3. Does the non-fatal skip behavior on non-Unix match expectations for production use?

<details>

- PR opened on branch `fix/pre-seed-relative-symlinks`
- Main has: naive `ParentDir` component check rejecting ALL relative symlinks with `..`
- Net-new value: proper path normalization allowing safe relative symlinks while blocking actual escapes
- Follow-up commits: `d9157c9` (main fixes) + `9440da3` (Windows ordering + cleanup)

</details>

<details>

- Lexical normalization avoids canonicalize TOCTOU
- Pre-write gating eliminates the time window where dangerous symlinks exist on disk
- Absolute symlink early return is a clean fast-path
- Non-fatal behavior on non-Unix prevents extraction abort for edge cases
- Deferred symlink pass ensures correct ordering regardless of filesystem iteration
- Comprehensive unit test coverage for the security boundary functions
- Clear doc comments explain security rationale and safety contracts

</details>

chaodu-agent:
LGTM ✅ — Proper symlink containment check replacing naive `ParentDir` rejection.

## What This PR Does

The tarball extraction in `pre_seed` blanket-rejected all symlinks containing `..` path components, breaking valid relative symlinks like `./bin/aws -> ../aws-cli/v2/current/bin/aws`. This PR replaces the overly strict check with correct path normalization that only rejects symlinks that actually escape the extraction root.

## How It Works

1. **Pre-extraction validation (TOCTOU fix)**: Symlink/hardlink targets are validated BEFORE `entry.unpack_in()` rather than after, eliminating the time-of-check/time-of-use window where a malicious entry could be exploited between extraction and cleanup.
2. **`normalize_path`**: Pure lexical path normalization (no filesystem access) that resolves `.` and `..` components.
3. **`symlink_escapes`**: Resolves the symlink target relative to its parent directory, normalizes the result, and checks containment against the normalized root via `starts_with`.
4. **Hard link coverage**: Both symlinks and hard links are now validated.
5. **Deferred symlink processing**: `move_recursive` now processes symlinks in a second pass after regular files, ensuring targets exist for non-Unix copy fallback.
6. **Non-Unix resilience**: Missing or directory symlink targets are non-fatal on non-Unix platforms (logged and skipped rather than aborting).

## Findings

| 1 | 🟢 | TOCTOU fix — validation moved before extraction eliminates the race window | `pre_seed.rs:301-318` |
| 2 | 🟢 | Comprehensive `normalize_path` handles all edge cases (root boundary, `.`, `..`, leading `..` on relative) | `pre_seed.rs:345-375` |
| 3 | 🟢 | SAFETY PRECONDITION pattern clearly documents the contract between `symlink_escapes` and `create_symlink_or_copy` | `pre_seed.rs:377-393` |
| 4 | 🟢 | 13 new tests cover unit + integration scenarios for both allowed and rejected symlinks | `pre_seed.rs:918-1108` |
| 5 | 🟢 | Hard links included in validation — prevents escape via hard link entries | `pre_seed.rs:303` |

<details>

- **Security improvement**: Pre-validation eliminates the TOCTOU window entirely. The old code extracted first, then scanned and deleted — leaving a window where the symlink existed on disk pointing outside the root.
- **Clean architecture**: `normalize_path` (pure math) → `symlink_escapes` (policy) → `extract_tarball_with_limits` (orchestration). Each function has a single responsibility.
- **Excellent test coverage**: Unit tests for `normalize_path` edge cases, `symlink_escapes` with various path patterns, and full integration tests that build real tarballs and verify extraction behavior.
- **Documentation**: F1–F7 finding tags in comments trace the rationale for each design choice. The SAFETY PRECONDITION on `create_symlink_or_copy` makes the caller contract explicit.
- **Graceful non-Unix handling**: Instead of hard-failing when symlink targets don't resolve on non-Unix platforms, the code logs and skips — appropriate for a utility extraction context.

</details>

<details>

- PR opened: against `main` where symlink handling uses naive `Component::ParentDir` check
- Main already has: basic symlink escape rejection (absolute + any `..` component)
- Net-new value: correct containment semantics that allow valid intra-root relative symlinks while maintaining security, TOCTOU elimination, hard link coverage, non-Unix resilience

</details>

chaodu-agent:
CHANGES REQUESTED ⚠️ — Round 1 findings addressed; Round 2 identifies a remaining Windows cross-directory symlink non-determinism issue.

## What This PR Does

Fixes `pre_seed` tarball extraction that incorrectly rejected valid relative symlinks containing `..` components (e.g., `./bin/aws -> ../aws-cli/v2/current/bin/aws`). Replaces the naive `ParentDir` component check with proper path normalization that only rejects symlinks actually escaping the extraction root.

## How It Works

- `symlink_escapes(root, link_path, link_target)` — resolves the symlink target relative to its parent directory, normalizes the result, and checks containment within root.
- `normalize_path(path)` — resolves `.` and `..` components lexically without filesystem access.
- Symlinks are validated **before** `unpack_in` (eliminates TOCTOU window).
- `move_recursive` defers symlink processing to a second pass per directory level.
- On non-Unix, directory symlinks and unresolved targets are skipped non-fatally with warnings.

## Findings

| 1 | 🔴 | ~~TOCTOU: post-extraction sweep validates after writing~~ | ✅ Fixed `d9157c9` |
| 2 | 🔴 | ~~`create_symlink_or_copy` safety contract mismatch~~ | ✅ Fixed `d9157c9` |
| 3 | 🔴 | ~~Windows `move_recursive` single-pass ordering bug~~ | ✅ Fixed `9440da3` |
| 4 | 🔴 | **NEW** Cross-directory symlink non-determinism on Windows: `deferred_symlinks` is per-directory, so sibling dir symlinks may fail when target already moved | `pre_seed.rs:454` |
| 5 | 🟡 | Non-fatal skip on Windows masks real failures — creates silently broken installations | `pre_seed.rs:421-442` |
| 6 | 🟡 | ~~Missing unit tests for `symlink_escapes`/`normalize_path`~~ | ✅ Fixed `d9157c9` |
| 7 | 🟡 | ~~`normalize_path` leading `..` behavior undocumented~~ | ✅ Fixed `d9157c9` |
| 8 | 🟡 | ~~Windows: directory symlinks fail with `std::fs::copy`~~ | ✅ Fixed `d9157c9` (skip with warning) |
| 9 | 🟡 | ~~Residual symlinks in temp directory~~ | ✅ Fixed `9440da3` |
| 10 | 🟢 | `check` CI passes (clippy + all tests green) | — |
| 11 | 🟢 | Pre-write gating eliminates TOCTOU window completely | — |
| 12 | 🟢 | Comprehensive unit tests for security boundary functions | — |

<details>

### 🔴 F4: Cross-directory symlink non-determinism (Windows)

`deferred_symlinks` is local to each `move_recursive` call frame, meaning it only defers symlinks within a single directory level. When directory `A/` is recursively moved (including its second-pass symlink copies), files from sibling directory `B/` may not yet be in place — or vice versa, depending on `std::fs::read_dir` iteration order (non-deterministic).

**Example:** If `B/link -> ../A/file` is processed after `A/` has already been moved out of temp, the Windows copy-fallback resolves `src.parent().join(link_target)` against the temp dir where `A/file` no longer exists.

**Suggested fix:** In `create_symlink_or_copy`'s non-Unix path, check both source (temp) and destination paths:
```rust
let resolved_src = src.parent().unwrap_or(Path::new(".")).join(link_target);
let resolved_dst = dst.parent().unwrap_or(Path::new(".")).join(link_target);
if resolved_src.exists() {
    std::fs::copy(&resolved_src, dst)?;
} else if resolved_dst.exists() {
    std::fs::copy(&resolved_dst, dst)?;
} else {
    warn!("hooks.pre_seed: skipping symlink with unresolved target: {}", ...);
}
```

### 🟡 F5: Non-fatal skip masks real failures

When symlink copy fails on Windows (target missing or directory), the current code logs a warning and returns `Ok(())`. This creates a "successful" pre-seed that is actually broken — downstream tasks will fail with hard-to-diagnose `file not found` errors. Consider:
- Counting skipped symlinks and emitting a summary warning at the end
- Or failing if critical symlinks (e.g., in `bin/`) cannot be resolved

</details>

<details>

- PR branch: `fix/pre-seed-relative-symlinks` (3 commits)
- Main has: naive `ParentDir` check rejecting ALL relative symlinks with `..`
- Net-new value: proper path normalization + pre-write gating + comprehensive tests
- CI `check` job: ✅ clippy + tests pass on commit `9440da3`
- `antigravity` smoke-test failure is unrelated (Docker build issue)

</details>

<details>

- Pre-write gating eliminates TOCTOU — symlinks never touch disk if they escape
- Lexical normalization avoids canonicalize TOCTOU
- Comprehensive unit tests cover all edge cases for the security boundary
- Clear safety contract documentation on `create_symlink_or_copy`
- Deferred symlink pass correctly handles same-directory ordering
- Non-Unix fallback gracefully handles directory symlinks and missing targets
- CI clippy + tests green

</details>

chaodu-agent:
CHANGES REQUESTED ⚠️ — All Round 1 critical findings resolved; two remaining issues in Windows cross-directory symlink handling.

## What This PR Does

The `pre_seed` tarball extractor blanket-rejected all symlinks containing `..` path components, breaking valid relative symlinks like `bin/aws -> ../aws-cli/v2/current/bin/aws`. This PR replaces the naive `ParentDir` check with proper path normalization that only rejects symlinks whose resolved target actually escapes the extraction root.

## How It Works

1. **Pre-extraction validation (TOCTOU fix)**: Symlink/hardlink targets are validated *before* `entry.unpack_in()` rather than after extraction — eliminates the race window entirely.
2. **`normalize_path`**: Pure lexical `.`/`..` resolver (no filesystem access, no canonicalize TOCTOU).
3. **`symlink_escapes`**: Resolves symlink target relative to its parent directory, normalizes, checks containment via `starts_with`.
4. **Deferred symlink pass**: `move_recursive` processes symlinks in a second pass after regular files, ensuring targets exist for non-Unix copy fallback.
5. **Non-Unix resilience**: Directory symlinks and unresolved targets are skipped non-fatally with warnings.

## Findings (Aggregated Across All Review Rounds)

| 1 | 🔴 | TOCTOU: post-extraction sweep validates after writing | ✅ Fixed `d9157c9` — gate before `unpack_in` |
| 2 | 🔴 | `create_symlink_or_copy` safety contract mismatch | ✅ Fixed `d9157c9` — doc updated with SAFETY PRECONDITION |
| 3 | 🔴 | Windows `move_recursive` single-pass ordering bug | ✅ Fixed `9440da3` — deferred symlink second pass |
| 4 | 🟡 | Missing unit tests for `symlink_escapes`/`normalize_path` | ✅ Fixed `d9157c9` — 13 edge-case tests added |
| 5 | 🟡 | `normalize_path` leading `..` behavior undocumented | ✅ Fixed `d9157c9` — doc comment added |
| 6 | 🟡 | Tar entries with `..` in path may bypass check | ✅ Fixed `d9157c9` — documented + tested |
| 7 | 🟡 | Windows: directory symlinks fail with `std::fs::copy` | ✅ Fixed `d9157c9` — skip non-fatally with warning |
| 8 | 🟡 | Residual symlinks in temp directory | ✅ Fixed `9440da3` — cleaned up in second pass |
| 9 | 🟡 | Hard link target resolution uses wrong base path in `symlink_escapes` | ⚠️ **Open** |
| 10 | 🟡 | Cross-directory symlink non-determinism on Windows (`deferred_symlinks` is per-directory) | ⚠️ **Open** |
| 11 | 🟡 | Non-fatal skip on Windows masks real failures — creates silently broken installations | ⚠️ **Open** |
| 12 | 🟢 | Pre-write gating eliminates TOCTOU window completely | — |
| 13 | 🟢 | Comprehensive unit + integration test coverage (13 new tests) | — |
| 14 | 🟢 | Lexical normalization avoids canonicalize TOCTOU | — |
| 15 | 🟢 | Clear SAFETY PRECONDITION documentation | — |

<details>
<summary>Finding Details — Open Items</summary>

### 🟡 F9: Hard link target resolution uses wrong base path

In tar archives, **symlink** targets are relative to the symlink's parent directory, but **hard link** targets are relative to the archive root (extraction destination). `symlink_escapes` unconditionally resolves via `link_path.parent().join(link_target)`, which is correct for symlinks but semantically wrong for hard links.

**Practical impact is low** — `tar::Entry::unpack_in` has its own `validate_inside_dst` check for hard links (defense-in-depth), but valid hard links could be incorrectly rejected (false positives).

**Suggested fix**: Check `entry_type.is_hard_link()` separately and resolve the target as `dest.join(link_target)` instead of `link_path.parent().join(link_target)`.

### 🟡 F10: Cross-directory symlink non-determinism (Windows)

`deferred_symlinks` is local to each `move_recursive` call frame — it only defers symlinks within a single directory level. When directory `A/` is recursively moved, files from sibling directory `B/` may not yet be in place (or vice versa), depending on `std::fs::read_dir` iteration order.

**Example:** `B/link -> ../A/file` processed after `A/` already moved out of temp → Windows copy-fallback fails because `A/file` no longer exists at the source path.

**Suggested fix**: In `create_symlink_or_copy`'s non-Unix path, try resolving against both source and destination:
```rust
let resolved_src = src.parent().unwrap_or(Path::new(".")).join(link_target);
let resolved_dst = dst.parent().unwrap_or(Path::new(".")).join(link_target);
if resolved_src.exists() {
    std::fs::copy(&resolved_src, dst)?;
} else if resolved_dst.exists() {
    std::fs::copy(&resolved_dst, dst)?;
} else {
    warn!("hooks.pre_seed: skipping symlink with unresolved target: {}", ...);
}
```

### 🟡 F11: Non-fatal skip masks real failures

When symlink copy fails on Windows, the code logs a warning and returns `Ok(())`. This creates a "successful" pre-seed that is actually broken — downstream tasks will fail with hard-to-diagnose errors. Consider counting skipped symlinks and emitting a summary warning, or failing if critical symlinks (e.g., in `bin/`) cannot be resolved.

</details>

<details>

- PR branch: `fix/pre-seed-relative-symlinks` (3 commits: initial + `d9157c9` + `9440da3`)
- Main has: naive `ParentDir` check rejecting ALL relative symlinks with `..`
- Net-new value: proper path normalization + pre-write gating + comprehensive tests
- CI `check` job: ✅ clippy + tests pass on commit `9440da3`
- `antigravity` smoke-test failure is unrelated (Docker build issue)

</details>

<details>

- **Security improvement**: Pre-validation eliminates the TOCTOU window — symlinks never touch disk if they escape
- **Clean architecture**: `normalize_path` (pure math) → `symlink_escapes` (policy) → `extract_tarball_with_limits` (orchestration)
- **Excellent test coverage**: Unit tests for `normalize_path` edge cases, `symlink_escapes` with various path patterns, and full integration tests with real tarball construction
- **Documentation**: F-numbered finding tags trace the rationale for each design choice; SAFETY PRECONDITION on `create_symlink_or_copy` makes caller contract explicit
- **Graceful non-Unix handling**: Avoids hard-failing for edge cases on non-Unix platforms
- **Hard links included**: Validation covers both symlinks and hard links

</details>

chaodu-agent:
Scope expanded per discussion: rather than maintaining a fragile Windows symlink copy-fallback, lifecycle hooks are now **Unix-only** and fail fast at startup on Windows (`HooksConfig::ensure_platform_supported`). This makes findings F3/F7 moot and significantly simplifies `pre_seed` (dropped the copy-fallback and deferred second-pass). Documented in `docs/hooks.md`.

chaodu-agent:
LGTM ✅ — Well-structured fix for overly strict symlink rejection with proper TOCTOU mitigation and platform fail-fast.

## What This PR Does

`pre_seed` tarball extraction was rejecting **all** symlinks containing `..` components, which broke valid layouts like `./bin/aws -> ../aws-cli/v2/current/bin/aws`. This PR replaces the naive `ParentDir` check with proper path normalization that only rejects symlinks actually escaping the extraction root, and adds platform validation to fail fast on Windows.

## How It Works

1. **`symlink_escapes` + `normalize_path`** — Resolves symlink targets relative to the link's parent directory using lexical path normalization (no filesystem access, avoiding canonicalize TOCTOU). Only rejects links that actually escape the extraction root.
2. **Pre-write validation (F1)** — Validation happens *before* `entry.unpack_in()`, eliminating the TOCTOU window from the old post-extraction cleanup approach.
3. **`ensure_platform_supported()`** — Hooks fail fast at startup on non-Unix platforms with a clear error message, removing the fragile Windows copy-fallback code path.
4. **`create_symlink`** — Simplified to Unix-only with explicit safety precondition documentation.

## Findings

| 1 | 🟢 | Correct TOCTOU fix — validation before write eliminates the race window | `pre_seed.rs:305-318` |
| 2 | 🟢 | `normalize_path` handles all edge cases correctly (above-root, leading `..`, curdir) with clear documentation of why leading `..` dropping is safe | `pre_seed.rs:377-398` |
| 3 | 🟢 | Comprehensive test coverage — unit tests for both functions + integration tests for tarball extraction | `pre_seed.rs` tests |
| 4 | 🟢 | Safety contract on `create_symlink` clearly documents caller responsibility | `pre_seed.rs:400-413` |
| 5 | 🟢 | Clean platform fail-fast with user-actionable error message | `config.rs:220-233` |

<details>

### 🟢 F1: Pre-write validation eliminates TOCTOU
The old approach unpacked first, then scanned for escaping symlinks — leaving a race window. Now symlink targets are validated before any filesystem write, which is the correct approach.

### 🟢 F2: Robust path normalization
`normalize_path` resolves `..` lexically without filesystem access. The documented behavior for leading `..` (dropping them) is safe because callers always pass absolute paths (extraction root + entry path), making the `starts_with` containment check reliable.

### 🟢 F3: Excellent test coverage
35 tests covering: `normalize_path` edge cases, `symlink_escapes` for valid/invalid targets including `..` in entry paths (F6), integration tests proving escaping symlinks never touch disk, and platform support checks.

### 🟢 F4: Clear safety contract documentation
`create_symlink` explicitly documents that it only rejects absolute targets and that callers MUST pre-validate relative targets with `symlink_escapes`. This makes the security invariant auditable.

### 🟢 F5: Platform limitation properly documented and enforced
The `docs/hooks.md` update and `ensure_platform_supported()` gate work together — users get both documentation and a runtime safeguard.

</details>

<details>

- PR opened: by chaodu-agent on branch `fix/pre-seed-relative-symlinks`
- Main already has: symlink handling in `pre_seed` with naive `ParentDir` rejection + Windows copy-fallback
- Net-new value: (1) Proper path normalization allowing safe relative symlinks with `..`, (2) TOCTOU-free pre-write validation, (3) Windows fail-fast replacing fragile fallback, (4) comprehensive test coverage

</details>

<details>

- The `symlink_escapes` function correctly resolves relative targets against the link's parent directory — this is the semantically correct way to evaluate containment
- Documentation in F5/F6 comments explains *why* the edge case handling is safe, not just *what* it does
- Hard links are also validated through the same path, preventing a bypass vector
- The `#[cfg(not(unix))]` gate on `ensure_platform_supported` means zero runtime cost on the happy path (Linux)
- Removal of the deferred second-pass simplifies the logic — processing order no longer matters

</details>

chaodu-agent:
CHANGES REQUESTED ⚠️ — Unix-only hooks approach is clean and correct; one remaining correctness issue in hard link validation.

## What This PR Does

Fixes `pre_seed` tarball extraction that incorrectly rejected valid relative symlinks containing `..` components (e.g., `./bin/aws -> ../aws-cli/v2/current/bin/aws`), and makes all lifecycle hooks explicitly Unix-only with a fail-fast startup check on unsupported platforms.

## How It Works

1. **Pre-extraction validation (TOCTOU fix)**: Symlink/hardlink targets validated *before* `entry.unpack_in()` — eliminates race window.
2. **`normalize_path` + `symlink_escapes`**: Pure lexical path normalization; containment check via `starts_with`.
3. **`create_symlink`**: Simplified to Unix-only. Non-Unix path bails (unreachable due to startup guard).
4. **`move_recursive`**: No deferred pass needed — Unix symlinks store target strings without resolving, so ordering doesn't matter.
5. **`HooksConfig::ensure_platform_supported()`**: Fails fast at startup if any hook is configured on non-Unix. Wired into `main.rs`.
6. **Documentation**: `docs/hooks.md` updated with platform limitation.

## Findings

| 1 | 🟡 | Hard link target resolution uses wrong semantics in `symlink_escapes` | `pre_seed.rs` — `extract_tarball_with_limits` validation block |
| 2 | 🟡 | No test coverage for hard link validation (valid + escaping) | `pre_seed.rs` — tests |
| 3 | 🟢 | TOCTOU fix: pre-write gating eliminates the race window | `pre_seed.rs:299-318` |
| 4 | 🟢 | `create_symlink` clean, correct, well-documented safety precondition | `pre_seed.rs` |
| 5 | 🟢 | `move_recursive` correctly simplified (no deferred pass on Unix) | `pre_seed.rs` |
| 6 | 🟢 | `ensure_platform_supported` properly `#[cfg(not(unix))]` gated | `config.rs`, `main.rs` |
| 7 | 🟢 | Comprehensive unit + integration tests for `normalize_path` and `symlink_escapes` | `pre_seed.rs` |
| 8 | 🟢 | Good test coverage for `any_configured` / `ensure_platform_supported` | `config.rs` |

<details>

### 🟡 F1: Hard link target resolution uses wrong semantics

In tar archives, **symlink** targets are relative to the symlink's parent directory, but **hard link** targets are relative to the **archive root** (extraction destination). The current code calls `symlink_escapes(dest, &out_path, &link_target)` for both, which resolves via `link_path.parent().join(link_target)` — correct for symlinks, semantically wrong for hard links.

**Example:** A hard link at `subdir/hl` targeting `other/file.txt`:
- Current: resolves as `dest/subdir/other/file.txt` ❌
- Correct: resolves as `dest/other/file.txt` ✓

**Practical impact is low** — `tar::Entry::unpack_in` has its own `validate_inside_dst` for hard links (defense-in-depth), but valid hard links could be incorrectly rejected (false positives).

**Suggested fix:**

```rust
let escapes = if entry_type.is_hard_link() {
    // Hard link targets in tar are relative to archive root
    let resolved = normalize_path(&dest.join(&*link_target));
    let root_normalized = normalize_path(dest);
    !resolved.starts_with(&root_normalized)
} else {
    symlink_escapes(dest, &out_path, &link_target)
};
if escapes {
    warn!("hooks.pre_seed: skipping link with escaping target: {} -> {}", ...);
    continue;
}
```

### 🟡 F2: Missing hard link test coverage

Tests exercise symlinks thoroughly but there is no test for:
- Valid hard link within root (should be allowed)
- Hard link with escaping target (should be skipped)

This would catch F1 regressions.

</details>

<details>

- PR branch: `fix/pre-seed-relative-symlinks` (5 commits)
- Main has: naive `ParentDir` check rejecting ALL relative symlinks with `..`, Windows fallback code
- Net-new value: proper path normalization + pre-write gating + Unix-only hooks (fail-fast on Windows) + comprehensive tests
- CI `check` job: ✅ clippy + tests pass on commit `1e2595d`
- Smoke tests: in progress (Docker builds, unrelated to logic)

</details>

<details>

- **Excellent design decision**: Making hooks Unix-only eliminates an entire class of Windows symlink problems (F3/F7/F10/F11 from earlier rounds) without loss of functionality — these hooks only run in Linux containers.
- **Defense-in-depth preserved**: Pre-extraction validation + tar crate's own `validate_inside_dst` + `create_symlink`'s absolute-target rejection.
- **Clean simplification**: Removing the deferred symlink pass and Windows copy-fallback reduces ~40 lines of fragile platform-specific code.
- **Startup fail-fast**: `ensure_platform_supported()` ensures misconfiguration is caught immediately rather than silently broken at runtime.
- **Thorough test coverage**: 13+ new tests for normalize_path, symlink_escapes edge cases, and integration scenarios with real tarballs.

</details>

chaodu-agent:
LGTM ✅ — Solid security hardening with proper path normalization and pre-write validation.

## What This PR Does
Fixes `pre_seed` tarball extraction that was rejecting all symlinks containing `..` (breaking valid layouts like `./bin/aws -> ../aws-cli/v2/current/bin/aws`). Replaces the naive `ParentDir` check with lexical path normalization that correctly distinguishes "traverses within root" from "escapes root." Also makes all lifecycle hooks fail-fast on non-Unix platforms.

## How It Works
1. **`normalize_path`** — lexically resolves `.` and `..` without filesystem access (no `canonicalize` TOCTOU)
2. **`symlink_escapes`** — resolves symlink targets relative to the symlink's parent, then checks containment within the extraction root via component-wise `starts_with`
3. **Hard links** — resolved relative to the archive root (different semantics from symlinks), correctly handled
4. **Validate-before-write** — escaping links are `continue`d before `entry.unpack_in(dest)`, eliminating the TOCTOU window
5. **`ensure_platform_supported()`** — `#[cfg(not(unix))]` bail at startup if any hook is configured on Windows

## Findings

| 1 | 🟢 | Excellent TOCTOU fix — validate before write with `continue` | `pre_seed.rs:305-330` |
| 2 | 🟢 | Proper dual semantics for symlink vs hard link resolution | `pre_seed.rs:311-321` |
| 3 | 🟢 | Comprehensive test coverage (symlink escape, hard link escape, valid relative, edge cases) | `pre_seed.rs` tests |
| 4 | 🟢 | Well-documented safety contracts on `create_symlink` and `normalize_path` | `pre_seed.rs` |
| 5 | 🟢 | Clean platform-gating with fail-fast instead of fragile fallback | `config.rs`, `main.rs` |

<details>

### 🟢 F1: TOCTOU elimination
The old code unpacked first, then checked and removed escaping symlinks — a classic TOCTOU window. Now validation happens before `unpack_in`, and escaping entries are skipped entirely. This is the correct pattern.

### 🟢 F2: Dual link semantics
Hard link targets in tar are relative to archive root; symlink targets are relative to the symlink's parent directory. The code correctly handles both with different resolution logic and clear comments explaining why.

### 🟢 F3: Test coverage
35 tests covering: `normalize_path` edge cases (multiple parents, root boundary, cur-dir, relative leading `..`), `symlink_escapes` (valid relative, absolute rejection, deep traversal, entry-path with `..`), and integration tests with actual tarballs for both valid and escaping symlinks/hardlinks.

### 🟢 F4: Safety contract documentation
`create_symlink` clearly documents that callers MUST pre-validate with `symlink_escapes`. `normalize_path` documents the F5 leading-`..` behavior and why it's safe for the module's absolute-path usage.

### 🟢 F5: Clean platform separation
Dropping the fragile Windows copy-fallback in favor of an early bail is the right call — hooks are inherently Unix (POSIX shell, Unix symlinks, Linux containers). The `docs/hooks.md` update communicates this clearly.

</details>

<details>

- PR opened: 2026-06-29
- Author: chaodu-agent (self-fix addressing prior review round)
- Main already has: basic `pre_seed` extraction with naive `ParentDir` symlink rejection and Windows copy-fallback
- Net-new value: proper path normalization allowing valid relative symlinks while blocking escapes; TOCTOU fix; hard link validation; platform fail-fast; comprehensive tests

</details>

<details>

- Security-first design: validate before write, lexical normalization avoids filesystem TOCTOU
- Clear separation of symlink vs hard link resolution semantics
- Excellent test-to-code ratio — every edge case documented in F5/F6 comments has a corresponding test
- The `#[allow(unused_variables)]` annotation on the non-Unix path is appropriate
- Documentation in `docs/hooks.md` for user-facing platform limitation
- PR description clearly maps each change to the original review finding it addresses

</details>

PR #1271, opened 2026-07-01
docs(adr): configUrl (primary) and configToml (inline/--set-file) over Helm rendering

chaodu-agent:

Helm ConfigMap rendering has become a maintenance burden now that `configUrl` (s3://, https://) and `aws-sm://` secrets resolution are available. Every new config feature requires synchronized chart PRs, and users never see their final config directly.

This ADR proposes shifting to `configUrl` or `configFile` as the recommended configuration path, narrowing Helm's responsibility to runtime posture only. **Legacy values.yaml rendering is deprecated and unmaintained.**

Closes #

Discord Discussion URL: https://discord.com/channels/1490282656913559673/1521844272557588581

### BEFORE (legacy — deprecated)

```
┌──────────────────────────────────────────────────────────────┐
│ User edits values.yaml                                       │
│   discord.token: "aws-sm://..."                              │
│   ambient.enabled: true                                      │
│   hooks.pre_boot: [...]                                      │
│   ... 50+ lines of config spread across values.yaml ...      │
└──────────────────────┬───────────────────────────────────────┘
                       │ helm template
                       ▼
┌──────────────────────────────────────────────────────────────┐
│ templates/configmap.yaml                                     │
│   ~300 lines of {{if}} {{range}} {{with}} conditionals       │
│   enum validation, platform-specific assembly                │
└──────────────────────┬───────────────────────────────────────┘
                       │ renders
                       ▼
┌──────────────────────────────────────────────────────────────┐
│ ConfigMap: config.toml (user never sees this directly)        │
└──────────────────────┬───────────────────────────────────────┘
                       │ mounted into pod
                       ▼
┌──────────────────────────────────────────────────────────────┐
│ OpenAB process reads config.toml from /etc/openab/           │
└──────────────────────────────────────────────────────────────┘
```

### AFTER — configUrl (production)

```
┌──────────────────────────────────────────────────────────────┐
│ User edits config.toml directly (full visibility)            │
│   → uploads to S3: s3://bucket/agents/kiro/config.toml       │
└──────────────────────┬───────────────────────────────────────┘
                       │
    ┌──────────────────┴──────────────────────┐
    │ Helm (runtime posture only)              │
    │  • runAsUser: 1000 (non-root)            │
    │  • readOnlyRootFilesystem: true          │
    │  • HOME PVC persistence                  │
    │  • IRSA ServiceAccount                   │
    │  • args: ["openab","run","-c","s3://.."] │
    └──────────────────┬──────────────────────┘
                       │ pod starts
                       ▼
┌──────────────────────────────────────────────────────────────┐
│ OpenAB process                                                │
│  1. Fetch config.toml from s3://bucket/key                    │
│  2. Resolve aws-sm:// secrets in-app                          │
│  3. Run pre_boot hooks                                        │
│  4. Start agent sessions                                      │
└──────────────────────────────────────────────────────────────┘
```

### AFTER — configFile (dev / simple)

```
┌──────────────────────────────────────────────────────────────┐
│ User edits config.toml (same dir as values.yaml)             │
│   charts/openab/configs/kiro/config.toml                     │
└──────────────────────┬───────────────────────────────────────┘
                       │ {{ .Files.Get }} (zero logic)
                       ▼
┌──────────────────────────────────────────────────────────────┐
│ ConfigMap (verbatim copy — no template conditionals)          │
└──────────────────────┬───────────────────────────────────────┘
                       │ mounted into pod
                       ▼
┌──────────────────────────────────────────────────────────────┐
│ OpenAB reads config.toml — what user wrote = what agent gets │
└──────────────────────────────────────────────────────────────┘
```

### Migration Path

```
┌─────────────────────────┐     ┌─────────────────────────┐
│ STEP 1: Export           │     │ STEP 2: Upload          │
│                          │     │                          │
│ kubectl get configmap    │────▶│ aws s3 cp config.toml   │
│   → config.toml         │     │   s3://bucket/agent/    │
└─────────────────────────┘     └────────────┬────────────┘
                                              │
┌─────────────────────────┐     ┌─────────────▼────────────┐
│ STEP 4: Deploy           │     │ STEP 3: Switch values    │
│                          │◀────│                          │
│ helm upgrade + verify    │     │ configUrl: "s3://..."    │
│ logs: "Config loaded"    │     │ (remove all config keys) │
└─────────────────────────┘     └──────────────────────────┘
```

Not applicable — this is an internal architecture decision driven by our own tooling evolution (configUrl, aws-sm://, ecsctl, oabctl, Operator).

## Proposed Solution

See `docs/adr/configurl-over-helm-rendering.md` for full details. Key points:

1. Recommend `configUrl` as primary config path for production
2. Introduce `configFile` for dev/simple deployments (verbatim ConfigMap, zero logic)
3. Deprecate legacy values.yaml rendering (unmaintained, no bug fixes)
4. Helm retains: non-root execution, readonly filesystem, HOME PVC, image, SA
5. Fail-closed boot semantics (no local fallback — intentional)
6. Migration path included

## Why this approach?

- Eliminates chart maintenance for config schema changes
- Users get full visibility into their config (no template indirection)
- Aligns with all other deployment tooling (ecsctl, oabctl, Operator)
- `configFile` gives same UX benefit without requiring S3/IRSA (low barrier for dev)
- S3 provides 99.99% SLA, versioning, and CloudTrail audit for prod

## Alternatives Considered

- **Keep current Helm rendering** — rejected due to ongoing maintenance cost and poor user experience
- **Hot-reload config from S3** — deferred to future ADR; edit-and-restart is simpler for v1
- **Only support configUrl** — rejected; configFile provides a low-barrier path for dev without external dependencies

## Validation

- [x] Links are valid
- [x] Renders correctly in GitHub preview
- [x] ADR follows existing `docs/adr/` conventions

### DIFF
--- a/docs/adr/configurl-over-helm-rendering.md
+++ b/docs/adr/configurl-over-helm-rendering.md
@@ -34,7 +34,7 @@ Meanwhile, declarative tooling (`ecsctl`, `oabctl`, Operator CRDs) operates on t
    - ServiceAccount assignment (for IRSA)
    - Recreate strategy (RWO PVC constraint)
 3. **Introduce `configFile` as a zero-logic alternative** — users place a raw `config.toml` alongside their Helm values. Helm copies it verbatim into a ConfigMap via `{{ .Files.Get }}` — no template rendering, no conditionals, no enum validation. This gives the "full config visibility" benefit without requiring S3/IRSA setup.
-4. **Freeze legacy ConfigMap rendering** — existing template logic remains for backward compatibility but is no longer the recommended path. New config features do NOT require chart PRs.
+4. **Deprecate legacy ConfigMap rendering** — existing template logic remains for backward compatibility but is **no longer maintained**. No bug fixes, no new config features, no chart PRs for this path. Users on legacy rendering are encouraged to migrate to `configUrl` or `configFile`.
 5. **Config lives externally or inline** — users maintain `config.toml` either in S3 (`s3://`), HTTPS, or as a local file next to their Helm values. Changes take effect on pod restart (configUrl) or on `helm upgrade` (configFile).
 6. **Secrets live in AWS Secrets Manager** — referenced via `aws-sm://` in config.toml. No Kubernetes Secret objects required.
 
@@ -63,7 +63,7 @@ Meanwhile, declarative tooling (`ecsctl`, `oabctl`, Operator CRDs) operates on t
 |------|--------|-------------|-----------------|
 | **`configUrl`** | S3, HTTPS, R2 | Production — full GitOps, IAM audit | `helm install` once; config changes need only pod restart |
 | **`configFile`** | Local `config.toml` next to values.yaml | Dev / simple deployments — no external deps | `helm upgrade` picks up file changes |
-| **Legacy rendering** | `values.yaml` → template → ConfigMap | Existing users (frozen, not recommended) | Every config change = chart PR |
+| **Legacy rendering** ⚠️ | `values.yaml` → template → ConfigMap | **Deprecated — not maintained** | Every config change = chart PR |
 
 ### configUrl mode (recommended for production)
 
@@ -101,9 +101,11 @@ data:
 
 The user maintains a **real `config.toml`** — what they write is exactly what the agent reads. No values-to-template translation layer.
 
-### Legacy rendering (frozen)
+### Legacy rendering (⚠️ deprecated — unmaintained)
 
-Existing `values.yaml` → `templates/configmap.yaml` rendering continues to work for backward compatibility. No new config features will be added to this path.
+Existing `values.yaml` → `templates/configmap.yaml` rendering continues to work for backward compatibility but is **no longer maintained**. It will not receive bug fixes, new config features, or support for new platforms.
+
+> **Community notice:** We recommend all users migrate to `configUrl` (production) or `configFile` (dev/simple). The legacy `values.yaml` ConfigMap rendering path will not be updated going forward. Both new paths give you full visibility into your actual config.toml — no more guessing what Helm templates produce.
 
 ## 5. Minimal Helm Values (configUrl mode)
 
--- a/docs/adr/configurl-over-helm-rendering.md
+++ b/docs/adr/configurl-over-helm-rendering.md
@@ -35,9 +35,9 @@ Meanwhile, declarative tooling (`ecsctl`, `oabctl`, Operator CRDs) operates on t
    - Image version pinning and pull policy
    - ServiceAccount assignment (for IRSA)
    - Recreate strategy (RWO PVC constraint)
-3. **`configFile` / `configToml` are secondary, dev-oriented, Kubernetes-only conveniences** — not a peer of `configUrl`. Users place or paste a raw `config.toml` alongside their Helm values and Helm copies it verbatim into a ConfigMap — no template rendering, no conditionals, no enum validation. This is useful for local iteration with no external dependencies, but it is chart-coupled and does not extend to ECS/Zeabur/AgentCore. It should be presented as "the low-barrier option for people not ready to stand up external config," not as an equally-weighted alternative to `configUrl`.
-4. **Deprecate legacy ConfigMap rendering** — existing template logic remains for backward compatibility but is **no longer maintained**. No bug fixes, no new config features, no chart PRs for this path. Users on legacy rendering are encouraged to migrate to `configUrl` (preferred) or `configFile`/`configToml` (Kubernetes-only fallback).
-5. **Config lives externally by default** — users maintain `config.toml` in S3 (`s3://`) or HTTPS. This is what makes config **shared and hot-swappable across an entire fleet**: N agents/bots pointed at the same URL pick up a change on their next restart, with no chart, no CI/CD, and no Kubernetes required. Local/inline modes (`configFile`, `configToml`) exist only as a fallback for users who haven't set up external storage yet.
+3. **`configToml` is a secondary, dev-oriented, Kubernetes-only convenience** — not a peer of `configUrl`. Users paste a raw TOML string into `values.yaml`, or load an external `config.toml` file as-is via `helm --set-file agents.<name>.configToml=./config.toml`. Helm mounts the content verbatim into a ConfigMap — no template rendering, no conditionals, no enum validation. This is useful for local iteration with no external dependencies, but it is chart-coupled and does not extend to ECS/Zeabur/AgentCore. It should be presented as "the low-barrier option for people not ready to stand up external config," not as an equally-weighted alternative to `configUrl`.
+4. **Deprecate legacy ConfigMap rendering** — existing template logic remains for backward compatibility but is **no longer maintained**. No bug fixes, no new config features, no chart PRs for this path. Users on legacy rendering are encouraged to migrate to `configUrl` (preferred) or `configToml` (Kubernetes-only fallback).
+5. **Config lives externally by default** — users maintain `config.toml` in S3 (`s3://`) or HTTPS. This is what makes config **shared and hot-swappable across an entire fleet**: N agents/bots pointed at the same URL pick up a change on their next restart, with no chart, no CI/CD, and no Kubernetes required. `configToml` (inline or `--set-file`) exists only as a fallback for users who haven't set up external storage yet.
 6. **Secrets live in AWS Secrets Manager** — referenced via `aws-sm://` in config.toml. No Kubernetes Secret objects required. This also has a second-order effect worth calling out: once credentials are fully out of `config.toml`, the file itself becomes safe to **share or publish**. A `configUrl` pointing at a public/shared config lets any user reproduce someone else's exact agent behavior instantly — the config becomes a shareable artifact, closer to a public gist than a private deployment secret.
 
 ## 3. Target Architecture
@@ -74,7 +74,7 @@ OpenAB currently ships deployment tooling for Kubernetes (Helm/Operator), ECS Fa
 | Zeabur | Same `-c` flag in service start command | No |
 | AgentCore Runtime | Same `-c` flag passed to the runtime container | No |
 
-Users who have never touched Kubernetes should be able to run OpenAB on ECS, Zeabur, or AgentCore with the exact same config workflow as a Kubernetes user. `configFile` and legacy rendering cannot make this claim — they are Helm chart features and only exist for the Kubernetes path. They should be documented as **Kubernetes-specific conveniences**, not as peers of `configUrl` in a platform-neutral comparison.
+Users who have never touched Kubernetes should be able to run OpenAB on ECS, Zeabur, or AgentCore with the exact same config workflow as a Kubernetes user. `configToml` and legacy rendering cannot make this claim — they are Helm chart features and only exist for the Kubernetes path. They should be documented as **Kubernetes-specific conveniences**, not as peers of `configUrl` in a platform-neutral comparison.
 
 ### Fleet-scale 1:N shared config with hot-reload-by-restart
 
@@ -99,14 +99,14 @@ Because secrets are resolved via `aws-sm://` (or other exec providers) rather th
 
 ### Not a Kubernetes-only concern
 
-Chart maintenance cost is a real driver for this ADR, but it is a secondary one. The primary driver is that Kubernetes should not be the platform that dictates config architecture for users who are on ECS, Zeabur, or AgentCore and have no reason to adopt Helm. `configUrl` gives every platform — K8s included — the same low-friction, fleet-shareable, hot-restartable config workflow. Kubernetes users additionally get `configFile`/`configToml` as a Helm-specific convenience, but that convenience should never be presented as equal in importance to `configUrl`.
+Chart maintenance cost is a real driver for this ADR, but it is a secondary one. The primary driver is that Kubernetes should not be the platform that dictates config architecture for users who are on ECS, Zeabur, or AgentCore and have no reason to adopt Helm. `configUrl` gives every platform — K8s included — the same low-friction, fleet-shareable, hot-restartable config workflow. Kubernetes users additionally get `configToml` (inline or `--set-file`) as a Helm-specific convenience, but that convenience should never be presented as equal in importance to `configUrl`.
 
 ## 4. Configuration Modes
 
 | Mode | Source | When to use | Platform scope | Helm interaction |
 |------|--------|-------------|-----------------|-------------------|
 | **`configUrl`** (primary) | S3, HTTPS, R2 | **Default recommendation for all deployments** — production, fleet-scale, cross-platform, shareable configs | Kubernetes, ECS, Zeabur, AgentCore — identical workflow everywhere | `helm install`/equivalent once; config changes need only a restart, no redeploy |
-| **`configFile`/`configToml`** (secondary) | Local `config.toml` or inline TOML string in values.yaml | Local iteration with no external deps yet — a stepping stone, not an end state | Kubernetes (Helm) only | `helm upgrade` picks up file/value changes |
+| **`configToml`** (secondary) | Inline TOML string in values.yaml, or an external `config.toml` loaded as-is via `--set-file` | Local iteration with no external deps yet — a stepping stone, not an end state | Kubernetes (Helm) only | `helm upgrade` picks up value/file changes |
 | **Legacy rendering** ⚠️ | `values.yaml` → template → ConfigMap | **Deprecated — not maintained** | Kubernetes (Helm) only | Every config change = chart PR |
 
 ### configUrl mode (primary — recommended for all deployments, all platforms)
@@ -120,36 +120,39 @@ agents:
 
 Pod starts with `openab run -c s3://...` — config fetched at boot. The exact same `-c s3://...` flag is what `ecsctl` puts in an ECS task definition and what a Zeabur or AgentCore service start command uses — no Kubernetes required to get this workflow.
 
-### configFile / configToml mode (secondary — Kubernetes-only convenience for local iteration)
+### configToml mode (secondary — Kubernetes-only convenience for local iteration)
+
+`configToml` (implemented in #1276) accepts a raw TOML string and mounts it verbatim into the ConfigMap — no template rendering, no conditionals, no enum validation. It supports two equivalent usage patterns, both backed by the same field:
+
+**Inline** — paste the TOML directly into `values.yaml`:
 
 ```yaml
 agents:
   kiro:
-    configFile: "configs/kiro/config.toml"  # relative to chart root
+    configToml: |
+      [discord]
+      bot_token = "${DISCORD_BOT_TOKEN}"
+      allow_all_channels = true
     serviceAccountName: "openab"
 ```
 
-Helm template logic is trivial — just a file copy, zero conditionals:
+**As-is from a standalone file** — keep `config.toml` as a real, standalone file (full IDE syntax highlighting and TOML schema validation) and load it verbatim at deploy time with Helm's built-in `--set-file`:
 
-```yaml
-{{- if .Values.agents.kiro.configFile }}
-apiVersion: v1
-kind: ConfigMap
-metadata:
-  name: {{ .Release.Name }}-kiro-config
-data:
-  config.toml: |
-{{ .Files.Get .Values.agents.kiro.configFile | indent 4 }}
-{{- end }}
+```bash
+helm upgrade kiro ./charts/openab \
+  --set-file agents.kiro.configToml=./my-config.toml \
+  -f values.yaml
 ```
 
-The user maintains a **real `config.toml`** — what they write is exactly what the agent reads. No values-to-template translation layer.
+`--set-file` reads the file's raw content and assigns it to `agents.kiro.configToml` as a string, merging the same way `--set` does. This gives WYSIWYG, standalone-file editing **without any chart changes** — a dedicated `configFile` field (with `.Files.Get`) is not needed, since `--set-file` already covers the "edit a real file, load it as-is" use case using the existing `configToml` field. The tradeoff versus `configUrl` is that this path is still Kubernetes/Helm-only and requires a `helm upgrade` (not just a restart) to pick up changes.
+
+The user maintains a **real `config.toml`** either way — what they write is exactly what the agent reads. No values-to-template translation layer.
 
 ### Legacy rendering (⚠️ deprecated — unmaintained)
 
 Existing `values.yaml` → `templates/configmap.yaml` rendering continues to work for backward compatibility but is **no longer maintained**. It will not receive bug fixes, new config features, or support for new platforms.
 
-> **Community notice:** We recommend all users migrate to `configUrl` — the platform-agnostic, fleet-shareable path that works identically on Kubernetes, ECS, Zeabur, and AgentCore. `configFile`/`configToml` remain available as a Kubernetes-only convenience for local iteration. The legacy `values.yaml` ConfigMap rendering path will not be updated going forward. All non-legacy paths give you full visibility into your actual config.toml — no more guessing what Helm templates produce.
+> **Community notice:** We recommend all users migrate to `configUrl` — the platform-agnostic, fleet-shareable path that works identically on Kubernetes, ECS, Zeabur, and AgentCore. `configToml` (inline or `--set-file`) remains available as a Kubernetes-only convenience for local iteration. The legacy `values.yaml` ConfigMap rendering path will not be updated going forward. All non-legacy paths give you full visibility into your actual config.toml — no more guessing what Helm templates produce.
 
 ## 5. Minimal Helm Values (configUrl mode)
 
--- a/docs/adr/configurl-over-helm-rendering.md
+++ b/docs/adr/configurl-over-helm-rendering.md
@@ -26,9 +26,10 @@ Meanwhile, declarative tooling (`ecsctl`, `oabctl`, Operator CRDs) operates on t
 ## 2. Decision
 
 1. **Recommend `configUrl` as the primary configuration path** for all new deployments.
-2. **Helm chart retains responsibility only for runtime posture:**
-   - Pod security context (`runAsNonRoot`, `readOnlyRootFilesystem`, `drop: ALL`)
-   - PVC lifecycle and persistence
+2. **Helm chart retains responsibility only for runtime posture** — this is where Helm continues to deliver real value that raw config cannot:
+   - **Non-root execution** — enforce `runAsUser`/`runAsGroup` so the container never runs as root, reducing blast radius of container escapes.
+   - **Read-only root filesystem** — `readOnlyRootFilesystem: true` with `drop: ALL` capabilities ensures the container cannot be tampered with at runtime; only the HOME PVC is writable.
+   - **HOME PVC persistence** — dedicated PersistentVolumeClaim mounted at the agent's `$HOME`, providing durable workspace (git repos, session state, caches) that survives pod restarts.
    - Image version pinning and pull policy
    - ServiceAccount assignment (for IRSA)
    - Recreate strategy (RWO PVC constraint)
@@ -69,7 +70,24 @@ agents:
     persistence:
       enabled: true
       size: 1Gi
+    # ── This is what Helm enforces (you don't touch config for this) ──
+    securityContext:
+      runAsUser: 1000
+      runAsGroup: 1000
+      runAsNonRoot: true
+      readOnlyRootFilesystem: true
+      allowPrivilegeEscalation: false
+      capabilities:
+        drop: ["ALL"]
 ```
+
+**Why these three matter:**
+
+| Helm-managed concern | What it prevents |
+|---------------------|-----------------|
+| `runAsUser: 1000` (non-root) | Container escape → host root access |
+| `readOnlyRootFilesystem` | Runtime binary tampering, malware persistence outside HOME |
+| HOME PVC (`persistence.enabled`) | Agent state loss on restart; provides durable workspace isolated from the immutable image |
 
 ## 5. Boot Behavior
 
--- a/docs/adr/configurl-over-helm-rendering.md
+++ b/docs/adr/configurl-over-helm-rendering.md
@@ -2,7 +2,7 @@
 
 - **Status:** Proposed
 - **Date:** 2026-07-01
-- **Author:** @pahud
+- **Author:** @chaodu-agent
 
 ---
 
--- a/docs/adr/configurl-over-helm-rendering.md
+++ b/docs/adr/configurl-over-helm-rendering.md
@@ -23,20 +23,22 @@ This leaves the Helm chart's ConfigMap rendering as a **maintenance burden with 
 
 Meanwhile, declarative tooling (`ecsctl`, `oabctl`, Operator CRDs) operates on the final config state directly, making the Helm rendering layer an outlier in the architecture.
 
+**This ADR is not only about reducing chart maintenance.** OpenAB is a multi-platform project — Kubernetes (via Helm/Operator), ECS Fargate (via `ecsctl`), Zeabur, and AgentCore Runtime are all first-class deployment targets. Kubernetes must not be treated as the "default" platform that dictates config architecture for everyone else. `configUrl` is the one config path that works identically across **all** of them, because it lives entirely in the OpenAB process (`openab run -c <url>`), not in any platform-specific rendering layer. That portability — not just chart LOC reduction — is the primary reason `configUrl` is the recommended path.
+
 ## 2. Decision
 
-1. **Recommend `configUrl` as the primary configuration path** for all new deployments.
-2. **Helm chart retains responsibility only for runtime posture** — this is where Helm continues to deliver real value that raw config cannot:
+1. **`configUrl` is the primary, platform-agnostic configuration path for all deployment targets** — Kubernetes, ECS Fargate, Zeabur, and AgentCore Runtime alike. It requires no Helm chart, no ConfigMap, no platform-specific rendering — only a URL (`s3://`, `https://`) passed to the process at boot. This is the path every other configuration mode is judged against.
+2. **Helm chart retains responsibility only for runtime posture** on Kubernetes — this is where Helm continues to deliver real value that raw config cannot:
    - **Non-root execution** — enforce `runAsUser`/`runAsGroup` so the container never runs as root, reducing blast radius of container escapes.
    - **Read-only root filesystem** — `readOnlyRootFilesystem: true` with `drop: ALL` capabilities ensures the container cannot be tampered with at runtime; only the HOME PVC is writable.
    - **HOME PVC persistence** — dedicated PersistentVolumeClaim mounted at the agent's `$HOME`, providing durable workspace (git repos, session state, caches) that survives pod restarts.
    - Image version pinning and pull policy
    - ServiceAccount assignment (for IRSA)
    - Recreate strategy (RWO PVC constraint)
-3. **Introduce `configFile` as a zero-logic alternative** — users place a raw `config.toml` alongside their Helm values. Helm copies it verbatim into a ConfigMap via `{{ .Files.Get }}` — no template rendering, no conditionals, no enum validation. This gives the "full config visibility" benefit without requiring S3/IRSA setup.
-4. **Deprecate legacy ConfigMap rendering** — existing template logic remains for backward compatibility but is **no longer maintained**. No bug fixes, no new config features, no chart PRs for this path. Users on legacy rendering are encouraged to migrate to `configUrl` or `configFile`.
-5. **Config lives externally or inline** — users maintain `config.toml` either in S3 (`s3://`), HTTPS, or as a local file next to their Helm values. Changes take effect on pod restart (configUrl) or on `helm upgrade` (configFile).
-6. **Secrets live in AWS Secrets Manager** — referenced via `aws-sm://` in config.toml. No Kubernetes Secret objects required.
+3. **`configFile` / `configToml` are secondary, dev-oriented, Kubernetes-only conveniences** — not a peer of `configUrl`. Users place or paste a raw `config.toml` alongside their Helm values and Helm copies it verbatim into a ConfigMap — no template rendering, no conditionals, no enum validation. This is useful for local iteration with no external dependencies, but it is chart-coupled and does not extend to ECS/Zeabur/AgentCore. It should be presented as "the low-barrier option for people not ready to stand up external config," not as an equally-weighted alternative to `configUrl`.
+4. **Deprecate legacy ConfigMap rendering** — existing template logic remains for backward compatibility but is **no longer maintained**. No bug fixes, no new config features, no chart PRs for this path. Users on legacy rendering are encouraged to migrate to `configUrl` (preferred) or `configFile`/`configToml` (Kubernetes-only fallback).
+5. **Config lives externally by default** — users maintain `config.toml` in S3 (`s3://`) or HTTPS. This is what makes config **shared and hot-swappable across an entire fleet**: N agents/bots pointed at the same URL pick up a change on their next restart, with no chart, no CI/CD, and no Kubernetes required. Local/inline modes (`configFile`, `configToml`) exist only as a fallback for users who haven't set up external storage yet.
+6. **Secrets live in AWS Secrets Manager** — referenced via `aws-sm://` in config.toml. No Kubernetes Secret objects required. This also has a second-order effect worth calling out: once credentials are fully out of `config.toml`, the file itself becomes safe to **share or publish**. A `configUrl` pointing at a public/shared config lets any user reproduce someone else's exact agent behavior instantly — the config becomes a shareable artifact, closer to a public gist than a private deployment secret.
 
 ## 3. Target Architecture
 
@@ -57,15 +59,57 @@ Meanwhile, declarative tooling (`ecsctl`, `oabctl`, Operator CRDs) operates on t
 └─────────────────────────────────────────────────────┘
 ```
 
+## 3a. Why `configUrl` Is the Primary Path, Not Just "the production option"
+
+This section elevates `configUrl` beyond a Kubernetes production tier — it is the config path that makes OpenAB deployable identically on any platform, and it unlocks fleet-scale workflows the other modes cannot.
+
+### Platform parity — Kubernetes is not the default
+
+OpenAB currently ships deployment tooling for Kubernetes (Helm/Operator), ECS Fargate (`ecsctl`), Zeabur, and AgentCore Runtime. Only `configUrl` works the same way on all four:
+
+| Platform | How it consumes `configUrl` | Needs Helm/ConfigMap? |
+|----------|------------------------------|------------------------|
+| Kubernetes | `args: ["openab", "run", "-c", "s3://..."]` in pod spec | No |
+| ECS Fargate (`ecsctl`) | Same `-c` flag in task definition container command | No |
+| Zeabur | Same `-c` flag in service start command | No |
+| AgentCore Runtime | Same `-c` flag passed to the runtime container | No |
+
+Users who have never touched Kubernetes should be able to run OpenAB on ECS, Zeabur, or AgentCore with the exact same config workflow as a Kubernetes user. `configFile` and legacy rendering cannot make this claim — they are Helm chart features and only exist for the Kubernetes path. They should be documented as **Kubernetes-specific conveniences**, not as peers of `configUrl` in a platform-neutral comparison.
+
+### Fleet-scale 1:N shared config with hot-reload-by-restart
+
+Because `configUrl` is just a URL fetched at process boot, an operator running many agents/bots (N) can point **all of them at the same `s3://` or `https://` config**. Updating that one object and restarting the fleet — `kubectl rollout restart`, an ECS service force-deploy, or a Zeabur/AgentCore restart — propagates the change to every instance in seconds, with no chart change, no CI/CD pipeline, and no per-instance edit:
+
+```
+                 ┌─────────────────────────────┐
+                 │ s3://bucket/shared/config.toml │
+                 └──────────────┬──────────────┘
+        ┌────────────────┬──────┴──────┬────────────────┐
+        ▼                ▼             ▼                ▼
+   agent-1 (K8s)   agent-2 (ECS)  agent-3 (Zeabur)  agent-N (AgentCore)
+```
+
+This is the same mental model as editing a GitHub gist in place, or `gh gist edit` / `aws s3 cp` followed by a restart — no build step, no deploy pipeline, edit-and-go. For operators iterating quickly across dozens of bots, this is materially faster than a chart-per-agent model.
+
+This pattern extends beyond the top-level config: as `config.toml` becomes more modular (e.g. `pre_boot`, `pre_shutdown` hooks), those sections can themselves point to a shared external endpoint. Changing behavior for an entire fleet of bots becomes "edit one URL, restart" rather than "edit N config files" or "ship N chart upgrades."
+
+### Config as a shareable, credential-free artifact
+
+Because secrets are resolved via `aws-sm://` (or other exec providers) rather than embedded in `config.toml`, the file itself no longer needs to be treated as sensitive. A `configUrl` can be made public and shared directly — anyone pointing their own deployment at that URL reproduces the exact same agent behavior. This turns `config.toml` into a distributable "recipe," not just a private deployment artifact, and is a capability unique to the externalized-config model.
+
+### Not a Kubernetes-only concern
+
+Chart maintenance cost is a real driver for this ADR, but it is a secondary one. The primary driver is that Kubernetes should not be the platform that dictates config architecture for users who are on ECS, Zeabur, or AgentCore and have no reason to adopt Helm. `configUrl` gives every platform — K8s included — the same low-friction, fleet-shareable, hot-restartable config workflow. Kubernetes users additionally get `configFile`/`configToml` as a Helm-specific convenience, but that convenience should never be presented as equal in importance to `configUrl`.
+
 ## 4. Configuration Modes
 
-| Mode | Source | When to use | Helm interaction |
-|------|--------|-------------|-----------------|
-| **`configUrl`** | S3, HTTPS, R2 | Production — full GitOps, IAM audit | `helm install` once; config changes need only pod restart |
-| **`configFile`** | Local `config.toml` next to values.yaml | Dev / simple deployments — no external deps | `helm upgrade` picks up file changes |
-| **Legacy rendering** ⚠️ | `values.yaml` → template → ConfigMap | **Deprecated — not maintained** | Every config change = chart PR |
+| Mode | Source | When to use | Platform scope | Helm interaction |
+|------|--------|-------------|-----------------|-------------------|
+| **`configUrl`** (primary) | S3, HTTPS, R2 | **Default recommendation for all deployments** — production, fleet-scale, cross-platform, shareable configs | Kubernetes, ECS, Zeabur, AgentCore — identical workflow everywhere | `helm install`/equivalent once; config changes need only a restart, no redeploy |
+| **`configFile`/`configToml`** (secondary) | Local `config.toml` or inline TOML string in values.yaml | Local iteration with no external deps yet — a stepping stone, not an end state | Kubernetes (Helm) only | `helm upgrade` picks up file/value changes |
+| **Legacy rendering** ⚠️ | `values.yaml` → template → ConfigMap | **Deprecated — not maintained** | Kubernetes (Helm) only | Every config change = chart PR |
 
-### configUrl mode (recommended for production)
+### configUrl mode (primary — recommended for all deployments, all platforms)
 
 ```yaml
 agents:
@@ -74,9 +118,9 @@ agents:
     serviceAccountName: "openab"
 ```
 
-Pod starts with `openab run -c s3://...` — config fetched at boot.
+Pod starts with `openab run -c s3://...` — config fetched at boot. The exact same `-c s3://...` flag is what `ecsctl` puts in an ECS task definition and what a Zeabur or AgentCore service start command uses — no Kubernetes required to get this workflow.
 
-### configFile mode (recommended for dev / simple deployments)
+### configFile / configToml mode (secondary — Kubernetes-only convenience for local iteration)
 
 ```yaml
 agents:
@@ -105,7 +149,7 @@ The user maintains a **real `config.toml`** — what they write is exactly what 
 
 Existing `values.yaml` → `templates/configmap.yaml` rendering continues to work for backward compatibility but is **no longer maintained**. It will not receive bug fixes, new config features, or support for new platforms.
 
-> **Community notice:** We recommend all users migrate to `configUrl` (production) or `configFile` (dev/simple). The legacy `values.yaml` ConfigMap rendering path will not be updated going forward. Both new paths give you full visibility into your actual config.toml — no more guessing what Helm templates produce.
+> **Community notice:** We recommend all users migrate to `configUrl` — the platform-agnostic, fleet-shareable path that works identically on Kubernetes, ECS, Zeabur, and AgentCore. `configFile`/`configToml` remain available as a Kubernetes-only convenience for local iteration. The legacy `values.yaml` ConfigMap rendering path will not be updated going forward. All non-legacy paths give you full visibility into your actual config.toml — no more guessing what Helm templates produce.
 
 ## 5. Minimal Helm Values (configUrl mode)
 
@@ -243,9 +287,12 @@ Since Helm template-time checks (e.g. Discord ID precision, enum validation) no 
 
 ### Positive
 
+- **Platform parity** — `configUrl` gives Kubernetes, ECS, Zeabur, and AgentCore users the identical config workflow; no platform is forced to adopt Helm to get a good experience.
+- **Fleet-scale hot-reload-by-restart** — N agents sharing one `configUrl` pick up a change fleet-wide in seconds by restarting, with no chart change and no CI/CD pipeline required.
+- **Shareable, credential-free config** — with secrets resolved via `aws-sm://`, a `config.toml` (and its `configUrl`) can be published or shared so others reproduce the exact same agent behavior instantly.
 - **Zero chart maintenance for new config features** — schema changes never propagate to Helm.
 - **Users see the full config** — no mental model of values → template → ConfigMap required.
-- **Edit-and-restart workflow** — change config in S3/gist, restart pod, done.
+- **Edit-and-restart workflow** — change config in S3/gist, restart pod (or ECS/Zeabur/AgentCore equivalent), done.
 - **Aligned with declarative tooling** — ecsctl, oabctl, and Operator all operate on final config state.
 - **Reduced issue surface** — eliminates "my Helm values don't render correctly" class of bugs.
 - **S3 availability** — `s3://` path gives 99.99% SLA, private access via IAM, versioning, and CloudTrail audit.
@@ -258,7 +305,7 @@ Since Helm template-time checks (e.g. Discord ID precision, enum validation) no 
 
 ### Neutral
 
-- Helm is not deprecated — it remains the recommended way to enforce runtime security posture on Kubernetes. Its scope simply narrows.
+- Helm is not deprecated — it remains the recommended way to enforce runtime security posture on Kubernetes. Its scope simply narrows, and it is one of several platform integrations (alongside ecsctl, Zeabur, AgentCore) rather than the architecture's default assumption.
 - Multi-agent deployments still benefit from Helm's `agents.<name>` loop for generating multiple Deployments/PVCs from a single release.
 
 ## 11. References
--- a/docs/adr/configurl-over-helm-rendering.md
+++ b/docs/adr/configurl-over-helm-rendering.md
@@ -3,6 +3,7 @@
 - **Status:** Proposed
 - **Date:** 2026-07-01
 - **Author:** @chaodu-agent
+- **Implementation:** `configToml` is implemented in the companion PR #1276 (open, not yet merged as of this writing). This ADR documents the design rationale that #1276 implements; merge order is #1276 → this ADR → #1277 (legacy path removal).
 
 ---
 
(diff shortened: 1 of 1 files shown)


chaodu-agent:
LGTM ✅ — Well-structured ADR that documents an architectural decision already supported by the codebase.

## What This PR Does

Documents the shift from Helm ConfigMap rendering to `configUrl` (s3://, https://) as the primary configuration path for OpenAB deployments, formalizing a decision that aligns with existing capabilities (`configUrl`, `aws-sm://` secrets resolution).

## How It Works

Adds a single ADR document (`docs/adr/configurl-over-helm-rendering.md`) following the established ADR structure in the repo. The document covers: problem statement, decision, target architecture, boot behavior, migration path, and consequences.

## Findings

| 1 | 🟢 | Excellent migration path — step-by-step guide with concrete commands lowers adoption friction | `docs/adr/configurl-over-helm-rendering.md:120-170` |
| 2 | 🟢 | All referenced files verified to exist on main (`docs/config-reference.md`, `docs/secrets-management.md`, `crates/openab-core/src/config.rs`, `operator/examples/fleet.yaml`) | `docs/adr/configurl-over-helm-rendering.md:198-200` |
| 3 | 🟢 | ADR format is consistent with existing ADRs in the repo — numbered sections, consequences split into positive/negative/neutral | — |
| 4 | 🟢 | Fail-closed semantics are well-justified with S3 SLA data and clear reasoning for no local fallback | `docs/adr/configurl-over-helm-rendering.md:80-95` |

<details>

### 🟢 F1: Complete migration path
The step-by-step migration from existing Helm ConfigMap rendering to configUrl mode is concrete and actionable — includes export, upload, IRSA policy, and Helm values diff. This significantly lowers the barrier for existing users.

### 🟢 F2: Valid references
All four cross-references in Section 10 point to files that exist on main, ensuring readers can follow the links for deeper context.

### 🟢 F3: Consistent ADR format
Follows the same numbered-section structure as existing ADRs (`ambient.md`, `hooks.md`, etc.) — Status/Date/Author metadata, Problem → Decision → Architecture → Consequences flow.

### 🟢 F4: Deliberate design tradeoffs documented
The fail-closed boot decision is explicitly called out as intentional with clear reasoning (config freshness, split-brain avoidance) rather than an oversight. The planned `openab config validate` CLI command acknowledges the validation gap.

</details>

<details>

- PR opened: 2026-07-01
- Main already has: `configUrl` support in code (`config.rs`), chart tests (`config-url_test.yaml`), secrets management docs (`docs/secrets-management.md`), config reference (`docs/config-reference.md`)
- Net-new value: **Formal ADR document** that captures the *decision* to recommend configUrl as primary path and freeze Helm ConfigMap rendering — this was previously an undocumented convention

</details>

<details>

- Clear problem statement grounded in real maintenance pain (synchronized chart PRs for every config feature)
- Architecture diagram is clean and immediately communicable
- Minimal Helm values example makes the end-state concrete
- Consequences section is honest about tradeoffs (boot-time S3 dependency, no pre-deploy validation)
- Explicitly scopes out hot-reload (deferred to future ADR) — good scope discipline
- `Closes #` left empty rather than inventing a false linkage

</details>

chaodu-agent:
LGTM ✅ — Well-structured ADR documenting the shift from Helm ConfigMap rendering to configUrl as the primary config path.

## What This PR Does

Documents the architectural decision to recommend `configUrl` (s3://, https://) as the primary configuration mechanism, narrowing Helm's scope to runtime posture only (security context, PVC, image, SA). This eliminates synchronized chart PRs for config schema changes.

## How It Works

Single new file `docs/adr/configurl-over-helm-rendering.md` following the established ADR format. Covers problem statement, decision, target architecture, minimal Helm values example, boot semantics, config change workflow, migration path, and pre-deploy validation strategy.

## Findings

| 1 | 🟢 | Clear problem-to-solution flow with concrete examples | Throughout |
| 2 | 🟢 | Migration path is actionable with copy-pasteable commands | Section 7 |
| 3 | 🟢 | Fail-closed boot semantics well-justified with availability math | Section 5 |
| 4 | 🟢 | Consequences section is balanced — acknowledges trade-offs honestly | Section 9 |
| 5 | 🟢 | All referenced files (`config-reference.md`, `secrets-management.md`, `config.rs`, `fleet.yaml`) exist on main | Section 10 |

<details>

- **Follows existing ADR conventions** — numbered sections, Status/Date/Author header, consistent with the 18 other ADRs in `docs/adr/`
- **Minimal Helm values example** is excellent — shows exactly what the new world looks like with clear rationale table
- **Explicitly scopes out hot-reload** — "deferred to future ADR" prevents scope creep
- **Target architecture ASCII diagram** communicates the concept at a glance
- **Backward compatibility preserved** — ConfigMap rendering frozen, not removed
- **Security posture table** (`runAsUser`, `readOnlyRootFilesystem`, HOME PVC) clearly communicates what Helm still owns and why

</details>

<details>

- PR opened: 2026-07-01
- Main already has: `docs/adr/` directory with 18 ADRs; `configUrl` support in `config.rs`; chart test `config-url_test.yaml`; `docs/config-reference.md` documenting s3:// source; `docs/secrets-management.md` covering aws-sm://
- Net-new value: Formal ADR recording the decision to recommend configUrl over Helm rendering as primary path, with migration guide and architectural rationale

</details>

chaodu-agent:
LGTM ✅ — Well-structured ADR that clearly documents the shift from Helm ConfigMap rendering to configUrl as the primary configuration path.

## What This PR Does

Documents the architectural decision to recommend `configUrl` (s3://, https://) as the primary configuration path for OpenAB deployments, narrowing Helm's responsibility to runtime posture only (security context, PVC, image, SA). Includes a `configFile` zero-logic alternative for simpler deployments.

## How It Works

Single ADR file (`docs/adr/configurl-over-helm-rendering.md`) following standard ADR format: problem statement → decision → target architecture → configuration modes → migration path → consequences. Introduces three config modes (configUrl, configFile, legacy) with clear deprecation semantics.

## Findings

| 1 | 🟢 | Excellent migration path with concrete commands | Section 8 |
| 2 | 🟢 | Clear security rationale for Helm-managed concerns | Section 5 |
| 3 | 🟢 | Honest trade-off analysis in Consequences | Section 10 |
| 4 | 🟢 | `configFile` mode bridges gap for users not ready for S3 | Section 4 |

<details>

- **Complete migration path** — Step-by-step commands from export to deploy, including IRSA policy and S3 versioning. Users can follow this without guesswork.
- **Three-mode spectrum** — configUrl (production), configFile (dev/simple), legacy (deprecated). Covers all user profiles without forcing a cliff migration.
- **Security rationale table** — Clearly explains *why* Helm still owns security context (non-root, read-only rootfs, HOME PVC), making the "Helm for posture only" boundary self-evident.
- **Fail-closed boot semantics** — Explicitly calls out the design choice and justifies it (S3 99.99% SLA, visible CrashLoopBackOff). No silent fallback that could cause split-brain.
- **Honest consequences section** — Acknowledges boot-time dependency, loss of pre-deploy validation, and backward compat needs. Proposes mitigations without overselling.
- **Clean ASCII diagrams** — Architecture diagram is simple and immediately understandable.

</details>

<details>

- PR opened: 2026-07-01
- Main already has: `docs/config-reference.md` documenting configUrl as a feature; no ADR directory exists yet
- Net-new value: Formalizes the architectural decision to deprecate Helm ConfigMap rendering, establishes `docs/adr/` convention, documents migration path, and introduces the `configFile` mode concept

</details>

chaodu-agent:
LGTM ✅ — Well-structured ADR documenting the shift from Helm ConfigMap rendering to configUrl/configFile as the primary configuration path.

## What This PR Does

Documents an architectural decision to recommend `configUrl` (S3/HTTPS) as the primary configuration path for production, and `configFile` for dev/simple deployments — deprecating legacy Helm ConfigMap template rendering that has become a maintenance burden.

## How It Works

Single new file `docs/adr/configurl-over-helm-rendering.md` (269 lines) proposing:
1. `configUrl` for production (fetch config from S3/HTTPS at boot)
2. `configFile` for dev (verbatim `{{ .Files.Get }}` into ConfigMap — zero template logic)
3. Helm scope narrowed to runtime security posture only (non-root, readonly FS, PVC, IRSA)
4. Legacy rendering deprecated (not removed) — no maintenance going forward
5. Fail-closed boot semantics with clear migration path

## Findings

| 1 | 🟢 | Clear problem → solution narrative with concrete before/after diagrams | `docs/adr/configurl-over-helm-rendering.md` |
| 2 | 🟢 | All referenced docs exist on main (`config-reference.md`, `secrets-management.md`, `config.rs`, `operator/examples/fleet.yaml`) | §11 References |
| 3 | 🟢 | Follows established ADR format (Status/Date/Author, numbered sections) consistent with existing ADRs | — |
| 4 | 🟢 | Migration path is actionable with copy-pasteable commands | §8 |
| 5 | 🟢 | Honest trade-off analysis — acknowledges boot-time S3 dependency and loss of pre-deploy validation | §10 Consequences |

<details>

- PR opened: 2026-07-01
- Main already has: `docs/adr/` directory with 18 existing ADRs; `docs/config-reference.md` documenting s3:// config source; `docs/adr/secrets-management.md` covering aws-sm:// resolution
- Net-new value: Formalizes the architectural decision to deprecate Helm ConfigMap rendering now that configUrl and aws-sm:// are production-ready. No existing ADR covers this topic.

</details>

<details>

- Excellent visual diagrams (ASCII art) showing before/after architecture — immediately communicates the value proposition
- Three-mode taxonomy (configUrl / configFile / legacy) covers all deployment scenarios without forcing a single path
- Pragmatic deprecation approach — legacy rendering is not removed, just frozen, avoiding breaking changes
- Security rationale for Helm's remaining scope is well-articulated (non-root, readonly FS, PVC isolation)
- `configFile` mode provides a low-barrier entry for dev without requiring S3/IRSA infrastructure
- References are all valid and point to existing code/docs on main

</details>

dogzzdogzz:
Thanks for writing this up — the ADR is clear and the `configFile` / `configUrl` split is a nice framing. I'd like to push back from an **enterprise GitOps** angle, where the priorities are versioning, atomic deploys, and predictable upgrade/rollback. Three trade-offs I think the ADR under-weights:

## 1. `configUrl` as the *primary* path raises upgrade/rollback complexity

Today (Helm-templated), moving across a breaking change is one atomic PR:

- bump chart version + edit `values` → merge → `helm upgrade`
- rollback = revert the PR

With `configUrl` (remote) it becomes a two-step, cross-repo flow:

1. In the config repo: open a PR to update the config, pipeline tags it → (possibly) publish to a web server / S3, ensuring every change gets a distinct URL
2. In the chart repo: open a PR to bump the version + point to the new URL

Step 2 depends on step 1, so doing them out of order is a real failure mode, and rollback now has to unwind **both** sides (or lean on S3 versioning + version-pinned URLs). If the URL is unversioned (`always latest`), rollback traceability gets worse. Net: coordination cost and risk go *up*, and it doesn't fully decouple either.

Note this critique lands on **`configUrl` specifically** — `configFile` keeps the config next to the chart, so it stays one-PR / one-revert. My ask is really about which mode is recommended as the *default*, not about the whole ADR.

## 2. "managed defaults + patch delta" is a feature, not just indirection

This is the point I think is most worth adding to the ADR as an explicit trade-off. The current templated approach gives you two things:

- **Upstream default propagation** — when a new release adds a setting or changes a default, users just `helm upgrade` and pick it up automatically.
- **Delta-only maintenance** — users only own the keys they override in a small values patch; they don't track every line of the full config.

Once users hand-maintain the entire `config.toml`, every upstream schema/default change becomes a manual diff + merge on their side. For teams that **sync upstream on a cadence**, that's a continuous, easy-to-miss maintenance cost. The ADR frames "users never see their final config" as the core pain — but the flip side, upstream-managed defaults with delta overrides, is a capability many enterprise users actively rely on. Worth recording as a first-class consequence.

## 3. Fail-early validation gets bypassed → availability risk

This is a concrete regression. Today the Helm template guards catch bad config **before** anything reaches the cluster:

- the `{{ fail }}` guards in `configmap.yaml` (e.g. `allowBotMessages` enum, snowflake-ID regex on bot/role IDs, mangled-`e+` ID detection) fail at **render time** (`helm template` / `helm upgrade --dry-run`)
- the `helm-unittest` suite under `charts/openab/tests/` gates format/regressions in CI

So malformed config or a bad dependency fails **pre-deploy** — a broken config simply never makes it into the cluster. Verbatim `configFile` / remote `configUrl` skips this render-time layer entirely: errors surface only after the pod starts → crash loop → downtime, discovered by tailing logs rather than at deploy time. For environments with SLAs, that's a clear availability risk.

If `configUrl` is going to be the recommended path, there should ideally be an equivalent pre-deploy check first (e.g. `openab config validate` in CI, an admission webhook, or an init-container validation step) — but neither the ADR nor the codebase seems to provide that tool yet.

## Summary / asks

The three points share one root: for enterprise users, the templated rendering isn't only a maintenance burden — it also provides **atomic deploys, default propagation, and fail-early safety**, all of which enterprise GitOps depends on. I'm happy to see broader config sourcing supported, but would ask that the ADR:

1. **Record these trade-offs explicitly** — not just the migration path, but the longer-term difference in maintenance/operational model.
2. **Consider keeping the Helm templated path as a supported option** rather than deprecated/unmaintained.
3. **If pushing `configUrl`**, land an `openab config validate` CLI (or equivalent pre-deploy check) *before* legacy is marked unmaintained, so there's no validation gap in between.

antigenius0910:
<p style="white-space: pre-wrap; margin-top: 0.1em; margin-bottom: 0.2em; color: rgb(191, 191, 191); font-family: -apple-system, &quot;system-ui&quot;, &quot;Segoe UI&quot;, Roboto, sans-serif; font-size: 13px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; background-color: rgb(25, 26, 27); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;">Thank you for this ADR — it identifies a real and painful problem. We run openab in production and wanted to share some thoughts with the community.</p><h3 style="color: rgb(191, 191, 191); font-family: -apple-system, &quot;system-ui&quot;, &quot;Segoe UI&quot;, Roboto, sans-serif; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; white-space: normal; background-color: rgb(25, 26, 27); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;">The diagnosis is right, but there's a simpler fix</h3><p style="white-space: pre-wrap; margin-top: 0.1em; margin-bottom: 0.2em; color: rgb(191, 191, 191); font-family: -apple-system, &quot;system-ui&quot;, &quot;Segoe UI&quot;, Roboto, sans-serif; font-size: 13px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; background-color: rgb(25, 26, 27); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;">The ADR frames the solution as "move config to an external source (S3/HTTPS)." We think the root cause is one layer above that:</p><p style="white-space: pre-wrap; margin-top: 0.1em; margin-bottom: 0.2em; color: rgb(191, 191, 191); font-family: -apple-system, &quot;system-ui&quot;, &quot;Segoe UI&quot;, Roboto, sans-serif; font-size: 13px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; background-color: rgb(25, 26, 27); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;"><strong>The Helm chart is currently doing something that isn't Helm's job — it's a TOML serializer written in template language.</strong></p><p style="white-space: pre-wrap; margin-top: 0.1em; margin-bottom: 0.2em; color: rgb(191, 191, 191); font-family: -apple-system, &quot;system-ui&quot;, &quot;Segoe UI&quot;, Roboto, sans-serif; font-size: 13px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; background-color: rgb(25, 26, 27); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;"><code style="font-family: monospace; color: rgb(140, 140, 140); background-color: rgb(38, 38, 38); padding: 2px 4px; border-radius: 3px; word-break: break-word; font-size: 0.9em;">configmap.yaml</code> is 352 lines that convert YAML values into TOML format, with Slack/Discord channel ID validation, cross-field conflict detection, and duplicated logic across adapters. Any new <code style="font-family: monospace; color: rgb(140, 140, 140); background-color: rgb(38, 38, 38); padding: 2px 4px; border-radius: 3px; word-break: break-word; font-size: 0.9em;">config.toml</code> field has to be exposed through a chart template first, which means every openab feature is gated behind the chart's release cycle.</p><p style="white-space: pre-wrap; margin-top: 0.1em; margin-bottom: 0.2em; color: rgb(191, 191, 191); font-family: -apple-system, &quot;system-ui&quot;, &quot;Segoe UI&quot;, Roboto, sans-serif; font-size: 13px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; background-color: rgb(25, 26, 27); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;">We have an internal general-purpose Helm chart whose ConfigMap template is 13 lines:</p><div class="codeBlockWrapper_-a7MRw" style="position: relative; margin: 8px 0px; color: rgb(191, 191, 191); font-family: -apple-system, &quot;system-ui&quot;, &quot;Segoe UI&quot;, Roboto, sans-serif; font-size: 13px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; white-space: normal; background-color: rgb(25, 26, 27); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;"><button class="copyButton_CEmTFw copyButton_-a7MRw" title="Copy code" aria-label="Copy code to clipboard" style="color: rgb(191, 191, 191); font-family: -apple-system, &quot;system-ui&quot;, &quot;Segoe UI&quot;, Roboto, sans-serif; font-size: 13px; background: none 0% 0% / auto repeat scroll padding-box border-box rgb(18, 19, 20); border-color: rgb(42, 43, 44); border-style: solid; border-width: 1px; border-image: none 100% / 1 / 0 stretch; cursor: pointer; opacity: 0; display: flex; border-radius: 4px; justify-content: center; align-items: center; padding: 4px; transition: opacity 0.15s, background 0.15s; position: absolute; top: 4px; right: 4px;"><svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 20 20" fill="currentColor" aria-hidden="true" data-slot="icon" class="copyIcon_CEmTFw"><path fill-rule="evenodd" d="M15.988 3.012A2.25 2.25 0 0 1 18 5.25v6.5A2.25 2.25 0 0 1 15.75 14H13.5v-3.379a3 3 0 0 0-.879-2.121l-3.12-3.121a3 3 0 0 0-1.402-.791 2.252 2.252 0 0 1 1.913-1.576A2.25 2.25 0 0 1 12.25 1h1.5a2.25 2.25 0 0 1 2.238 2.012ZM11.5 3.25a.75.75 0 0 1 .75-.75h1.5a.75.75 0 0 1 .75.75v.25h-3v-.25Z" clip-rule="evenodd"></path><path d="M3.5 6A1.5 1.5 0 0 0 2 7.5v9A1.5 1.5 0 0 0 3.5 18h7a1.5 1.5 0 0 0 1.5-1.5v-5.879a1.5 1.5 0 0 0-.44-1.06L8.44 6.439A1.5 1.5 0 0 0 7.378 6H3.5Z"></path></svg></button><pre style="overflow-x: auto; white-space: pre; box-sizing: border-box; border-radius: 4px; max-width: 100%; margin: 0px; padding: 8px;"><code class="language-yaml" style="font-family: monospace; color: rgb(140, 140, 140); background-color: rgb(38, 38, 38); padding: 0px; border-radius: 3px; word-break: break-word; font-size: 0.9em;">{{- if .Values.config.enabled }}
data:
  {{ toYaml .Values.config.data | nindent 2 }}
{{- end }}
</code></pre></div><p style="white-space: pre-wrap; margin-top: 0.1em; margin-bottom: 0.2em; color: rgb(191, 191, 191); font-family: -apple-system, &quot;system-ui&quot;, &quot;Segoe UI&quot;, Roboto, sans-serif; font-size: 13px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; background-color: rgb(25, 26, 27); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;">The principle: <strong>the chart owns K8s resource orchestration; the app owns its config format.</strong></p><p style="white-space: pre-wrap; margin-top: 0.1em; margin-bottom: 0.2em; color: rgb(191, 191, 191); font-family: -apple-system, &quot;system-ui&quot;, &quot;Segoe UI&quot;, Roboto, sans-serif; font-size: 13px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; background-color: rgb(25, 26, 27); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;">Applied to openab, the fix can be straightforward:</p><div class="codeBlockWrapper_-a7MRw" style="position: relative; margin: 8px 0px; color: rgb(191, 191, 191); font-family: -apple-system, &quot;system-ui&quot;, &quot;Segoe UI&quot;, Roboto, sans-serif; font-size: 13px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; white-space: normal; background-color: rgb(25, 26, 27); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;"><button class="copyButton_CEmTFw copyButton_-a7MRw" title="Copy code" aria-label="Copy code to clipboard" style="color: rgb(191, 191, 191); font-family: -apple-system, &quot;system-ui&quot;, &quot;Segoe UI&quot;, Roboto, sans-serif; font-size: 13px; background: none 0% 0% / auto repeat scroll padding-box border-box rgb(18, 19, 20); border-color: rgb(42, 43, 44); border-style: solid; border-width: 1px; border-image: none 100% / 1 / 0 stretch; cursor: pointer; opacity: 0; display: flex; border-radius: 4px; justify-content: center; align-items: center; padding: 4px; transition: opacity 0.15s, background 0.15s; position: absolute; top: 4px; right: 4px;"><svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 20 20" fill="currentColor" aria-hidden="true" data-slot="icon" class="copyIcon_CEmTFw"><path fill-rule="evenodd" d="M15.988 3.012A2.25 2.25 0 0 1 18 5.25v6.5A2.25 2.25 0 0 1 15.75 14H13.5v-3.379a3 3 0 0 0-.879-2.121l-3.12-3.121a3 3 0 0 0-1.402-.791 2.252 2.252 0 0 1 1.913-1.576A2.25 2.25 0 0 1 12.25 1h1.5a2.25 2.25 0 0 1 2.238 2.012ZM11.5 3.25a.75.75 0 0 1 .75-.75h1.5a.75.75 0 0 1 .75.75v.25h-3v-.25Z" clip-rule="evenodd"></path><path d="M3.5 6A1.5 1.5 0 0 0 2 7.5v9A1.5 1.5 0 0 0 3.5 18h7a1.5 1.5 0 0 0 1.5-1.5v-5.879a1.5 1.5 0 0 0-.44-1.06L8.44 6.439A1.5 1.5 0 0 0 7.378 6H3.5Z"></path></svg></button><pre style="overflow-x: auto; white-space: pre; box-sizing: border-box; border-radius: 4px; max-width: 100%; margin: 0px; padding: 8px;"><code class="language-yaml" style="font-family: monospace; color: rgb(140, 140, 140); background-color: rgb(38, 38, 38); padding: 0px; border-radius: 3px; word-break: break-word; font-size: 0.9em;"># values.yaml
agents:
  claude:
    configToml: |
      [slack]
      bot_token = "${SLACK_BOT_TOKEN}"
      allowed_channels = ["C0B3PJB13L2"]
      allow_user_messages = "mentions"

      [agent]
      command = "claude-agent-acp"
      inherit_env = ["ANTHROPIC_API_KEY"]

      [pool]
      max_sessions = 10
</code></pre></div><div class="codeBlockWrapper_-a7MRw" style="position: relative; margin: 8px 0px; color: rgb(191, 191, 191); font-family: -apple-system, &quot;system-ui&quot;, &quot;Segoe UI&quot;, Roboto, sans-serif; font-size: 13px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; white-space: normal; background-color: rgb(25, 26, 27); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;"><button class="copyButton_CEmTFw copyButton_-a7MRw" title="Copy code" aria-label="Copy code to clipboard" style="color: rgb(191, 191, 191); font-family: -apple-system, &quot;system-ui&quot;, &quot;Segoe UI&quot;, Roboto, sans-serif; font-size: 13px; background: none 0% 0% / auto repeat scroll padding-box border-box rgb(18, 19, 20); border-color: rgb(42, 43, 44); border-style: solid; border-width: 1px; border-image: none 100% / 1 / 0 stretch; cursor: pointer; opacity: 0; display: flex; border-radius: 4px; justify-content: center; align-items: center; padding: 4px; transition: opacity 0.15s, background 0.15s; position: absolute; top: 4px; right: 4px;"><svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 20 20" fill="currentColor" aria-hidden="true" data-slot="icon" class="copyIcon_CEmTFw"><path fill-rule="evenodd" d="M15.988 3.012A2.25 2.25 0 0 1 18 5.25v6.5A2.25 2.25 0 0 1 15.75 14H13.5v-3.379a3 3 0 0 0-.879-2.121l-3.12-3.121a3 3 0 0 0-1.402-.791 2.252 2.252 0 0 1 1.913-1.576A2.25 2.25 0 0 1 12.25 1h1.5a2.25 2.25 0 0 1 2.238 2.012ZM11.5 3.25a.75.75 0 0 1 .75-.75h1.5a.75.75 0 0 1 .75.75v.25h-3v-.25Z" clip-rule="evenodd"></path><path d="M3.5 6A1.5 1.5 0 0 0 2 7.5v9A1.5 1.5 0 0 0 3.5 18h7a1.5 1.5 0 0 0 1.5-1.5v-5.879a1.5 1.5 0 0 0-.44-1.06L8.44 6.439A1.5 1.5 0 0 0 7.378 6H3.5Z"></path></svg></button><pre style="overflow-x: auto; white-space: pre; box-sizing: border-box; border-radius: 4px; max-width: 100%; margin: 0px; padding: 8px;"><code class="language-yaml" style="font-family: monospace; color: rgb(140, 140, 140); background-color: rgb(38, 38, 38); padding: 0px; border-radius: 3px; word-break: break-word; font-size: 0.9em;"># configmap.yaml (complete template)
data:
  config.toml: |
    {{ $cfg.configToml | nindent 4 }}
</code></pre></div><p style="white-space: pre-wrap; margin-top: 0.1em; margin-bottom: 0.2em; color: rgb(191, 191, 191); font-family: -apple-system, &quot;system-ui&quot;, &quot;Segoe UI&quot;, Roboto, sans-serif; font-size: 13px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; background-color: rgb(25, 26, 27); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;">What you write is what gets deployed. No mental simulation of Helm templates required. Maintainers don't need to touch <code style="font-family: monospace; color: rgb(140, 140, 140); background-color: rgb(38, 38, 38); padding: 2px 4px; border-radius: 3px; word-break: break-word; font-size: 0.9em;">configmap.yaml</code> every time a new feature lands.</p><h3 style="color: rgb(191, 191, 191); font-family: -apple-system, &quot;system-ui&quot;, &quot;Segoe UI&quot;, Roboto, sans-serif; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; white-space: normal; background-color: rgb(25, 26, 27); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;">On<span> </span><code style="font-family: monospace; color: rgb(140, 140, 140); background-color: rgb(38, 38, 38); padding: 2px 4px; border-radius: 3px; word-break: break-word; font-size: 0.9em;">configUrl</code><span> </span>(S3/HTTPS)</h3><p style="white-space: pre-wrap; margin-top: 0.1em; margin-bottom: 0.2em; color: rgb(191, 191, 191); font-family: -apple-system, &quot;system-ui&quot;, &quot;Segoe UI&quot;, Roboto, sans-serif; font-size: 13px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; background-color: rgb(25, 26, 27); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;">We understand the motivation, but for most users this approach introduces more complexity than it removes:</p><ul style="padding-inline-start: 2em; color: rgb(191, 191, 191); font-family: -apple-system, &quot;system-ui&quot;, &quot;Segoe UI&quot;, Roboto, sans-serif; font-size: 13px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; white-space: normal; background-color: rgb(25, 26, 27); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;"><li>Requires setting up S3 bucket, IAM policy, or web server access</li><li>Config updates become two-stage (publish new config, then upgrade openab), which actually increases upgrade risk</li><li>A new failure mode at pod startup if S3 is unreachable — harder to detect than a failed Helm render</li><li>For deployments where<span> </span><code style="font-family: monospace; color: rgb(140, 140, 140); background-color: rgb(38, 38, 38); padding: 2px 4px; border-radius: 3px; word-break: break-word; font-size: 0.9em;">config.toml</code><span> </span>is under 30 lines, the etcd 1MB limit is never a real constraint</li></ul><p style="white-space: pre-wrap; margin-top: 0.1em; margin-bottom: 0.2em; color: rgb(191, 191, 191); font-family: -apple-system, &quot;system-ui&quot;, &quot;Segoe UI&quot;, Roboto, sans-serif; font-size: 13px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; background-color: rgb(25, 26, 27); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;">If <code style="font-family: monospace; color: rgb(140, 140, 140); background-color: rgb(38, 38, 38); padding: 2px 4px; border-radius: 3px; word-break: break-word; font-size: 0.9em;">configUrl</code> is supported, we think the framing from the maintainer discussion is the right one: CI/CD publishes versioned configs (e.g. <code style="font-family: monospace; color: rgb(140, 140, 140); background-color: rgb(38, 38, 38); padding: 2px 4px; border-radius: 3px; word-break: break-word; font-size: 0.9em;">config-&lt;version&gt;.toml</code>) so rollback follows version numbers. That makes <code style="font-family: monospace; color: rgb(140, 140, 140); background-color: rgb(38, 38, 38); padding: 2px 4px; border-radius: 3px; word-break: break-word; font-size: 0.9em;">configUrl</code> meaningful. But it should be an <strong>advanced option</strong>, not the recommended migration path for everyone.</p><h3 style="color: rgb(191, 191, 191); font-family: -apple-system, &quot;system-ui&quot;, &quot;Segoe UI&quot;, Roboto, sans-serif; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; white-space: normal; background-color: rgb(25, 26, 27); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;">Suggested direction</h3><p style="white-space: pre-wrap; margin-top: 0.1em; margin-bottom: 0.2em; color: rgb(191, 191, 191); font-family: -apple-system, &quot;system-ui&quot;, &quot;Segoe UI&quot;, Roboto, sans-serif; font-size: 13px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; background-color: rgb(25, 26, 27); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;">Support both modes, but remove the templating abstraction:</p>
Mode | Use case
-- | --
configToml: \| inline raw TOML | Most users — simple config, versioned with the chart
configUrl: s3://... | Users with mature CI/CD who need config versioned independently from the chart

<p style="white-space: pre-wrap; margin-top: 0.1em; margin-bottom: 0.2em; color: rgb(191, 191, 191); font-family: -apple-system, &quot;system-ui&quot;, &quot;Segoe UI&quot;, Roboto, sans-serif; font-size: 13px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; background-color: rgb(25, 26, 27); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;">Existing YAML-based fields (<code style="font-family: monospace; color: rgb(140, 140, 140); background-color: rgb(38, 38, 38); padding: 2px 4px; border-radius: 3px; word-break: break-word; font-size: 0.9em;">slack.*</code>, <code style="font-family: monospace; color: rgb(140, 140, 140); background-color: rgb(38, 38, 38); padding: 2px 4px; border-radius: 3px; word-break: break-word; font-size: 0.9em;">discord.*</code>, etc.) can be deprecated without immediate removal to give users time to migrate.</p><p style="white-space: pre-wrap; margin-top: 0.1em; margin-bottom: 0.2em; color: rgb(191, 191, 191); font-family: -apple-system, &quot;system-ui&quot;, &quot;Segoe UI&quot;, Roboto, sans-serif; font-size: 13px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; background-color: rgb(25, 26, 27); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;">The existing validation logic (channel ID format checks, enum validation) would fit better in:</p><ul style="padding-inline-start: 2em; color: rgb(191, 191, 191); font-family: -apple-system, &quot;system-ui&quot;, &quot;Segoe UI&quot;, Roboto, sans-serif; font-size: 13px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; white-space: normal; background-color: rgb(25, 26, 27); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;"><li><code style="font-family: monospace; color: rgb(140, 140, 140); background-color: rgb(38, 38, 38); padding: 2px 4px; border-radius: 3px; word-break: break-word; font-size: 0.9em;">values.schema.json</code><span> </span>for K8s-level structural validation</li><li>The openab binary's startup validation for<span> </span><code style="font-family: monospace; color: rgb(140, 140, 140); background-color: rgb(38, 38, 38); padding: 2px 4px; border-radius: 3px; word-break: break-word; font-size: 0.9em;">config.toml</code><span> </span>content — this way validation is uniform regardless of whether config comes from inline, S3, or<span> </span><code style="font-family: monospace; color: rgb(140, 140, 140); background-color: rgb(38, 38, 38); padding: 2px 4px; border-radius: 3px; word-break: break-word; font-size: 0.9em;">configFile</code></li></ul><hr style="font-family: -apple-system, &quot;system-ui&quot;, &quot;Segoe UI&quot;, Roboto, sans-serif; font-size: 13px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; white-space: normal; background-color: rgb(25, 26, 27); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;"><p style="white-space: pre-wrap; margin-top: 0.1em; margin-bottom: 0.2em; color: rgb(191, 191, 191); font-family: -apple-system, &quot;system-ui&quot;, &quot;Segoe UI&quot;, Roboto, sans-serif; font-size: 13px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; background-color: rgb(25, 26, 27); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;">The problem this ADR raises is real. We just think the shortest path is <strong>making the chart stop serializing TOML</strong>, rather than moving config to S3.</p>

chaodu-agent:
CHANGES REQUESTED ⚠️ — Solid ADR draft, but the latest revision doubles down on `configUrl`-as-primary without engaging two substantive community objections, and documents a `configFile`/`configToml` mode that doesn't exist in the codebase yet.

## What This PR Does

Proposes an architecture decision to make `configUrl` (S3/HTTPS) the primary, platform-agnostic configuration path for OpenAB across all deployment targets (Kubernetes, ECS, Zeabur, AgentCore), narrowing Helm's role to runtime posture (non-root, read-only rootfs, HOME PVC, IRSA) and deprecating the current ~300-line templated ConfigMap rendering.

## How It Works

Adds a single new file, `docs/adr/configurl-over-helm-rendering.md`, following the repo's existing ADR format (Status/Date/Author header, numbered sections). It introduces a three-mode taxonomy — `configUrl` (primary), `configFile`/`configToml` (secondary, Kubernetes-only convenience), and legacy Helm rendering (deprecated) — plus a migration guide and consequences analysis.

## Findings

| 1 | 🟡 | `configFile`/`configToml` is documented as an existing, usable mode (with a working Helm snippet) but does not exist anywhere in the codebase | `docs/adr/configurl-over-helm-rendering.md` §4 |
| 2 | 🟡 | Two substantive external review threads (rollback/validation-gap concern, and a competing "fix the chart, not the config path" proposal) were not addressed by the most recent commit, which only reinforces the `configUrl`-primary framing | PR comments, 2026-07-01/02 |
| 3 | 🟢 | Fail-closed boot semantics are accurately described and match `openab-core::config::load_config_raw_from_source` | `crates/openab-core/src/config.rs` |
| 4 | 🟢 | All cross-references (`config-reference.md`, `secrets-management.md`, `config.rs`, `operator/examples/fleet.yaml`) exist on `main` | §11 References |
| 5 | 🟢 | Migration path is concrete and actionable (export → upload → IRSA → switch values → verify) | §8 |

<details>

### 🟡 F1: `configFile`/`configToml` is presented as implemented, but isn't

We checked `main` for any trace of this mode:

```
git grep -n "configFile\|configToml" main -- .
→ no matches anywhere in the repo (charts/, crates/, docs/, operator/)
```

Only `configUrl` exists in `charts/openab/values.yaml` and `templates/deployment.yaml`/`configmap.yaml`. The ADR's §4 shows a working `{{ .Files.Get .Values.agents.kiro.configFile | indent 4 }}` snippet and frames `configFile`/`configToml` as a "secondary" mode alongside the primary and legacy ones — with no caveat that it is proposed, not built. A reader evaluating this ADR to decide how to configure a real deployment today would reasonably conclude they can use `configFile` now. Suggest either: (a) explicitly marking `configFile`/`configToml` as "proposed, not yet implemented" throughout, or (b) splitting it into a separate, dedicated implementation PR before merging the ADR as accepted.

### 🟡 F2: Most recent commit did not engage the two external review threads

Commit history shows the current HEAD (`072f119`, "emphasize configUrl as platform-agnostic primary path") was pushed *after* both external comments below, but its diff only adds new platform-parity and fleet-hot-reload arguments *for* `configUrl` — it does not mention or rebut either concern:

- **Rollback/validation-gap concern** (external reviewer, enterprise GitOps angle): today's Helm-templated rendering fails bad config at render/dry-run time via `{{ fail }}` guards in `configmap.yaml` (enum checks, snowflake-ID regex, mangled `e+` ID detection) plus the `helm-unittest` suite. Moving to `configUrl`/`configFile` skips that pre-deploy layer entirely — errors surface only after the pod starts, crash-loops, and someone tails logs. The reviewer's ask: land `openab config validate` (or an equivalent pre-deploy check) *before* legacy rendering is marked unmaintained, so there's no window with zero validation. The ADR's own §9 acknowledges `openab config validate` is only "planned," which means the gap the reviewer is worried about is real and currently unscheduled.
- **"Fix the chart, not the config path" concern** (second external reviewer): argues the root cause is that `configmap.yaml` is a hand-rolled TOML serializer (their count: 352 lines) doing work Helm shouldn't do, and proposes collapsing it to a ~4-line verbatim `configToml: |` passthrough instead of moving config off-cluster to S3/HTTPS. They explicitly flag `configUrl` as adding a two-stage publish/upgrade flow, a new boot-time failure mode, and unnecessary complexity for the common case (sub-30-line configs), and ask that it be offered as an advanced option rather than the default recommendation.

Both are directly relevant to the ADR's core claim (§2.1) that `configUrl` "is the path every other configuration mode is judged against." Since this is a **Proposed** ADR (not yet Accepted) and the disagreement is about the primary recommendation itself, not a wording nit, we'd expect the next revision to at least respond to these before the ADR is considered settled.

### 🟢 F3: Fail-closed boot semantics verified against code

`load_config_raw_from_source` / `load_config_raw_from_s3` in `crates/openab-core/src/config.rs` propagate errors via `anyhow::Result` with no local cache or fallback path — this matches the ADR's §6 claim precisely.

### 🟢 F4: References verified

Confirmed via local checkout that `docs/config-reference.md`, `docs/secrets-management.md`, `crates/openab-core/src/config.rs`, and `operator/examples/fleet.yaml` all exist on `main`.

</details>

<details>

- PR opened: 2026-07-01 (author: chaodu-agent)
- Main already has: `configUrl` support in `config.rs` + chart (`values.yaml`, `deployment.yaml`, `configmap.yaml` conditionals), `config-url_test.yaml` chart test, `aws-sm://` secrets resolution, `docs/config-reference.md`, `docs/secrets-management.md`. No `docs/adr/configurl-over-helm-rendering.md` and no `configFile`/`configToml` anywhere.
- Net-new value: A formal ADR proposing to designate `configUrl` as the primary config path and freeze legacy Helm ConfigMap rendering — the *decision itself* is net-new; the underlying `configUrl` mechanism it recommends already exists in code.

</details>

<details>

### Reviewer A (Round 1 — enterprise GitOps angle)

> `configUrl` as the primary path raises upgrade/rollback complexity... rollback now has to unwind both sides... if the URL is unversioned, rollback traceability gets worse.

> Once users hand-maintain the entire config.toml, every upstream schema/default change becomes a manual diff + merge on their side... worth recording as a first-class consequence.

> If configUrl is going to be the recommended path, there should ideally be an equivalent pre-deploy check first... but neither the ADR nor the codebase seems to provide that tool yet.

⚠️ **Not yet addressed**: the latest commit does not add rollback/versioning guidance, does not record the "managed defaults + patch delta" trade-off as a Consequence, and `openab config validate` remains unimplemented and unscheduled beyond "planned." This should be resolved before the ADR moves from Proposed to Accepted.

### Reviewer B (Round 2 — "fix the chart, not the config path")

> The Helm chart is currently doing something that isn't Helm's job — it's a TOML serializer written in template language... Support both modes, but remove the templating abstraction: `configToml: |` inline raw TOML for most users, `configUrl: s3://...` for teams with mature CI/CD.

ℹ️ **Partially aligned, not reconciled**: the ADR's `configToml` mode (added in the latest commit) is conceptually close to this proposal, but the ADR frames `configUrl` as primary and `configToml` as a "stepping stone," while Reviewer B argues the reverse — inline `configToml` should be the default for most users and `configUrl` the advanced option. The ADR does not currently explain why it lands on the opposite prioritization, and `configToml` itself is not implemented (see F1 above).

</details>

<details>

- Clear problem statement grounded in a real, verifiable maintenance cost (~300 lines of conditional templating in `configmap.yaml`)
- Honest Consequences section — acknowledges boot-time S3/network dependency and the validation gap rather than glossing over them
- Concrete, copy-pasteable migration path (export → upload → IRSA policy → switch values → verify)
- Explicitly scopes out hot-reload to a future ADR — good scope discipline
- Fail-closed rationale is backed by verifiable SLA reasoning, not hand-waved

</details>

chaodu-agent:
One addition: `configToml` (implemented in #1276) already supports both inline and external-file usage — no extra `configFile` field is needed.

```bash
# Inline: pasted directly into values.yaml (configToml's native usage)
agents:
  kiro:
    configToml: |
      [discord]
      bot_token = "${DISCORD_BOT_TOKEN}"
      ...

# As-is from a standalone file: --set-file reads the file's raw content
# at helm upgrade/install time and assigns it verbatim to the same field
helm upgrade kiro ./charts/openab \
  --set-file agents.kiro.configToml=./my-config.toml \
  -f values.yaml
```

`--set-file` is built-in Helm behavior (reads file content, assigns as a string, merges like `--set`), so:

- Users can treat `config.toml` as a standalone, WYSIWYG file — full VSCode TOML syntax highlighting / schema validation — without flattening it into a YAML block scalar
- No chart changes required — the single `configToml` field already covers both 'inline paste' and 'as-is external file' use cases
- This aligns with section 3a of this ADR: `configUrl` is the cross-platform (K8s/ECS/Zeabur/AgentCore) primary path, while Helm-only `configToml`/`configFile` are secondary, transitional options for users not yet on external config storage — no need to add a separate `configFile` field just for WYSIWYG editing, since `--set-file` already covers that.

Suggest adding a line about `--set-file` in section 3a or the `configFile`/`configToml` description so the doc makes this explicit.

chaodu-agent:
CHANGES REQUESTED ⚠️ — Solid ADR with correct technical grounding, but it states `configToml` as an already-implemented mode when the implementing PR (#1276) is still open, and cites three doc files/paths that don't exist on main.

## What This PR Does
Adds a new Architecture Decision Record (`docs/adr/configurl-over-helm-rendering.md`) proposing `configUrl` as the primary, platform-agnostic config path for OpenAB across Kubernetes/ECS/Zeabur/AgentCore, narrowing Helm's role to runtime posture (non-root, read-only FS, PVC, IRSA) and deprecating legacy `values.yaml` → ConfigMap template rendering.

## How It Works
Pure documentation change — no code. It defines three configuration modes (`configUrl` primary, `configToml` secondary/K8s-only, legacy rendering deprecated), includes a migration path from ConfigMap export → S3 upload → IRSA → `helm upgrade`, and documents fail-closed boot semantics (no local fallback if the config source is unreachable at startup).

## Findings

| 1 | 🟡 | `configToml` is presented as a shipped mode ("implemented in #1276"), but #1276 is still open and unmerged — this ADR is what #1276 closes, so the dependency direction reads backwards | `docs/adr/configurl-over-helm-rendering.md` §4 |
| 2 | 🟡 | Three referenced doc paths do not exist on `main`: `docs/config-reference.md`, `docs/secrets-management.md`, `operator/examples/fleet.yaml` — dangling references for readers who click through | `docs/adr/configurl-over-helm-rendering.md` §11 |
| 3 | 🟢 | Core technical claims are accurate and verified against `crates/openab-core/src/config.rs` on main: `s3://` URI parsing, `aws-sm://` secret resolution, and fail-closed boot all exist as described | `crates/openab-core/src/config.rs` |
| 4 | 🟢 | Clear problem framing — correctly identifies that Kubernetes/Helm should not dictate config architecture for ECS/Zeabur/AgentCore users, and backs this with a concrete cross-platform comparison table | — |
| 5 | 🟢 | Migration path (export → S3 upload → IRSA policy → values switch → verify) is concrete and actionable, not just prose | — |

<details>

### 🟡 F1: `configToml` described as implemented, but #1276 is unmerged
Section 4's table and the "configToml mode" subsection both state `configToml` "(implemented in #1276)" and give it a dedicated `--set-file` usage example as if it's available today. Checking the PR graph: #1276 (`feat(chart): add configToml pass-through...`) has `Closes #1271` in its body — i.e., #1276 is the implementation that resolves this ADR discussion, not a prerequisite already merged into main. Readers landing on this ADR before #1276 merges will believe `configToml` is usable now. Suggest either: (a) merge #1276 first and reference its merged SHA, or (b) mark the `configToml` section "Planned — pending #1276" until it lands, to avoid the ADR asserting a capability that doesn't exist on main yet.

### 🟡 F2: Dangling references in section 11
`docs/config-reference.md`, `docs/secrets-management.md`, and `operator/examples/fleet.yaml` are cited as authoritative references but none exist on `main` (confirmed via repo content lookup — all three return not-found). If these are aspirational/planned docs, label them as such (e.g. "planned: ...") rather than presenting them as existing citations. If they exist under different paths, correct the links.

### 🟢 F3: Technical accuracy verified
Cross-checked against `crates/openab-core/src/config.rs` on main:
- `s3://` URI parsing (`parse_s3_uri`) — present, with tests.
- `aws-sm://` secret reference resolution — present (`discord_token = "aws-sm://..."` pattern).
- Lifecycle hooks (`pre_boot`, `pre_seed`, `pre_shutdown`) — present in `HookConfig`.
- Fail-closed semantics claim is consistent with the config loader's error-on-missing-source behavior.

No factual/technical inaccuracies found in the architectural claims.

</details>

<details>

- PR opened: 2026-07-01 (per ADR `Date` field); author: chaodu-agent
- Main already has: `configUrl`/`s3://` support and `aws-sm://` secret resolution in `openab-core`; no `docs/adr/` directory yet (this is the first ADR)
- Net-new value: introduces the ADR directory/convention, formally documents the config strategy decision, and provides a migration path — but references a not-yet-merged companion PR (#1276) as if already implemented, and cites three non-existent doc files

</details>

<details>

- Strong "why" section (§3a) explaining platform parity across K8s/ECS/Zeabur/AgentCore — avoids the common mistake of treating Kubernetes as the default platform
- Concrete before/after architecture diagrams and a migration runbook with copy-pasteable commands
- Explicit fail-closed boot rationale with justification (S3 SLA, restart-loop visibility) rather than hand-waving
- Alternatives Considered section shows real design tradeoffs were weighed (hot-reload deferred, configToml-only rejected)

</details>

chaodu-agent:
Pushed 3ca56b2 addressing F1, and checked F2:

- **F1 (dependency direction)** — fixed. Added an explicit "Implementation" note in the ADR header: `configToml` is implemented in the companion PR #1276 (this ADR documents the rationale #1276 implements, not a prerequisite already merged), and stated the intended merge order explicitly: #1276 → this ADR → #1277.
- **F2 (dangling doc links)** — checked all three against `main`: `docs/config-reference.md`, `docs/secrets-management.md`, and `operator/examples/fleet.yaml` all exist. No fix needed — likely a stale check against this PR's diff rather than the full repo tree, since this PR only adds the ADR file itself and doesn't touch those paths.

chaodu-agent:
CHANGES REQUESTED ⚠️ — Both previous team findings (dependency direction, dangling links) are resolved at `3ca56b2`, but two substantive external-reviewer objections to the ADR's core recommendation remain unaddressed in the document itself, and the PR description contains a diagram that contradicts the ADR's own decision.

## What This PR Does

Adds a new Architecture Decision Record (`docs/adr/configurl-over-helm-rendering.md`) proposing `configUrl` (S3/HTTPS) as the primary, platform-agnostic configuration path for OpenAB across Kubernetes, ECS, Zeabur, and AgentCore Runtime — narrowing Helm's role to runtime posture only (non-root execution, read-only rootfs, HOME PVC, IRSA) and deprecating the current templated ConfigMap rendering.

## How It Works

Pure documentation change, single new file, no code. Defines three configuration modes — `configUrl` (primary, cross-platform), `configToml` (secondary, Kubernetes/Helm-only convenience for inline or `--set-file` usage), and legacy `values.yaml` rendering (deprecated, unmaintained) — plus a migration guide and a Consequences section.

## Findings

| 1 | 🟡 | Two substantive external review threads on the ADR's core recommendation are still not incorporated into the document | `docs/adr/configurl-over-helm-rendering.md` |
| 2 | 🟡 | PR description's "At a Glance" section shows a `configFile` + `{{ .Files.Get }}` design that contradicts the ADR body's explicit decision to use `configToml`/`--set-file` instead | PR description |
| 3 | 🟢 | Prior finding (dependency direction: `configToml` described as already implemented) is fixed — ADR header now states `configToml` ships in companion PR #1276 (open) and gives explicit merge order #1276 → this ADR → #1277 | `docs/adr/configurl-over-helm-rendering.md` header |
| 4 | 🟢 | Prior finding (dangling doc references) confirmed a false positive — `docs/config-reference.md`, `docs/secrets-management.md`, `crates/openab-core/src/config.rs`, `operator/examples/fleet.yaml` all exist on `main` and accurately describe the features cited | §11 References |
| 5 | 🟢 | Fail-closed boot semantics accurately match `openab-core::config` (S3/HTTPS loaders return `Result` with no local fallback) | §6 |
| 6 | 🟢 | Migration path is concrete and actionable (export → S3 upload → IRSA policy → switch values → verify) | §8 |

<details>

### 🟡 F1: External objections to the primary recommendation remain unaddressed in the ADR

Two external reviewers raised concerns that go directly to the ADR's central claim (§2.1: `configUrl` "is the path every other configuration mode is judged against"):

- **Rollback/validation-gap concern**: today's Helm-templated rendering fails bad config at render/dry-run time (`{{ fail }}` guards, `helm-unittest`). Moving to `configUrl` skips that pre-deploy layer — errors surface only after the pod starts and crash-loops. The ask: land `openab config validate` (or an equivalent pre-deploy check) *before* legacy rendering is marked unmaintained. The ADR's own §9 still lists `openab config validate` as only "planned," so the gap the reviewer flagged is real and unscheduled. The reviewer also raised a "managed defaults + patch delta" trade-off (upstream schema changes propagate for free with templated values, but become a manual diff once users own the full file) that isn't recorded anywhere in the Consequences section.
- **"Fix the chart, not the config path" concern**: argues the root cause is that `configmap.yaml` is a hand-rolled TOML serializer doing work Helm shouldn't do, and that collapsing it to a ~4-line verbatim `configToml: |` passthrough solves the maintenance problem without moving config off-cluster. Proposes `configToml` as the default for most users and `configUrl` as the advanced/opt-in path for teams with mature CI/CD — the reverse of what this ADR recommends.

The latest commit (`3ca56b2`) only fixes the dependency-direction and reference-check findings from our own team's prior round; it does not add rebuttal or revision addressing either external thread. Since this is a **Proposed** (not yet Accepted) ADR and both objections target the primary recommendation itself, we'd expect the next revision to either incorporate a response (e.g., a "Rejected alternative: configToml-as-default" entry in §Alternatives Considered, and a recorded trade-off for "managed defaults + patch delta" and the validation gap in §10 Consequences) or explain why they don't change the recommendation.

### 🟡 F2: PR description diagram contradicts the ADR's own decision

The PR description's "AFTER — configFile (dev / simple)" diagram shows `charts/openab/configs/kiro/config.toml` loaded via `{{ .Files.Get }}` into a verbatim ConfigMap. But the ADR body (§4) explicitly reasons that a dedicated `configFile` field is *not* needed, because `--set-file agents.kiro.configToml=./config.toml` already covers "edit a real file, load it as-is" using the existing `configToml` field. The description's diagram documents a design the ADR itself rejected. This is metadata, not committed content, but PR descriptions are part of the durable record readers see — worth updating the description to match the final `configToml`-only design before merge.

### 🟢 F3: Dependency-direction fix verified

The ADR header now reads: *"`configToml` is implemented in the companion PR #1276 (open, not yet merged as of this writing). This ADR documents the design rationale that #1276 implements; merge order is #1276 → this ADR → #1277."* This correctly reframes the relationship — previously the ADR implied `configToml` already shipped.

### 🟢 F4: Reference check re-verified, confirmed false positive

Independently checked against a fresh clone of `main`: `docs/config-reference.md` documents the `s3://` config source with matching credential-chain details; `docs/secrets-management.md` documents `aws-sm://` exactly as cited; `crates/openab-core/src/config.rs` and `operator/examples/fleet.yaml` (with `configFrom: s3://...` entries) both exist and match the ADR's descriptions.

</details>

<details>

- PR opened: 2026-07-01 (author: chaodu-agent)
- Main already has: `configUrl`/`s3://` support and `aws-sm://` secret resolution in `openab-core`; `docs/config-reference.md`, `docs/secrets-management.md`; no `docs/adr/` entry for this topic yet
- Net-new value: formalizes the architectural decision to designate `configUrl` as primary and deprecate legacy Helm ConfigMap rendering, with a migration guide — the underlying mechanisms it recommends already exist in code, but the recommendation and its documented rationale are net-new

</details>

<details>

### Reviewer A (Round 1 — enterprise GitOps angle)

> `configUrl` as the primary path raises upgrade/rollback complexity... rollback now has to unwind both sides... if the URL is unversioned, rollback traceability gets worse.

> Once users hand-maintain the entire config.toml, every upstream schema/default change becomes a manual diff + merge on their side... worth recording as a first-class consequence.

> If configUrl is going to be the recommended path, there should ideally be an equivalent pre-deploy check first... but neither the ADR nor the codebase seems to provide that tool yet.

⚠️ **Still not addressed**: the current revision does not add rollback/versioning guidance, does not record the "managed defaults + patch delta" trade-off in §10 Consequences, and `openab config validate` remains unimplemented and only "planned" in §9. This should be resolved (either incorporated or explicitly rebutted) before the ADR moves from Proposed to Accepted.

### Reviewer B (Round 2 — "fix the chart, not the config path")

> The Helm chart is currently doing something that isn't Helm's job — it's a TOML serializer written in template language... Support both modes, but remove the templating abstraction: `configToml: |` inline raw TOML for most users, `configUrl: s3://...` for teams with mature CI/CD.

ℹ️ **Partially aligned, not reconciled**: the ADR's `configToml` mode is conceptually close to this proposal, but the ADR frames `configUrl` as primary and `configToml` as a "stepping stone," while Reviewer B argues the reverse — inline `configToml` should be the default for most users, `configUrl` the advanced/opt-in path. The ADR does not currently explain why it lands on the opposite prioritization; recommend adding this as an explicit entry in §Alternatives Considered rather than leaving the disagreement implicit.

</details>

<details>

- Strong "why" section (§3a) explaining platform parity across K8s/ECS/Zeabur/AgentCore — avoids treating Kubernetes as the default platform that dictates config architecture for everyone else
- Concrete, copy-pasteable migration path (export → S3 upload → IRSA policy → switch values → verify)
- Honest Consequences section — acknowledges boot-time S3/network dependency and the current validation gap rather than glossing over them
- Fail-closed rationale is backed by verifiable SLA reasoning and matches actual loader behavior in `config.rs`
- Explicitly scopes hot-reload out to a future ADR — good scope discipline
- Promptly fixed both findings from our team's prior review round (dependency direction, reference validity)

</details>

obrutjack:
```markdown
## Community Feedback: rawConfig in values.yaml

**Position: Support deprecating templated rendering, but config should stay in-repo.**

### Agree with the ADR

- The 300-line template logic (`{{ if }}` / `{{ range }}` / `{{ fail }}`) should be deprecated
- Helm chart responsibility should focus on Pod spec, Volume, ServiceAccount, SecurityContext — it should not act as a Config Schema Parser

### Disagree

Pushing config to an external URL (S3 / HTTPS) breaks the K8s-native experience:

- `helm diff` cannot track config changes
- `kubectl get configmap` won't show the actual running config
- GitOps tools (ArgoCD / Flux) cannot detect drift
- Config should live in the same repo as infra code and deploy together

### Proposed alternative: rawConfig

Put raw TOML directly in values.yaml:

```yaml
rawConfig: |
  [discord]
  token = "aws-sm://prod/discord/token"

  [ambient]
  enabled = true
```

Template side — zero logic:

```yaml
# templates/configmap.yaml
data:
  config.toml: {{ .Values.rawConfig | quote }}
```

This achieves:
- ✅ Single PR, single `helm upgrade`, rollback = revert PR (GitOps atomic deploy)
- ✅ ConfigMap visible, `helm diff` trackable (K8s-native experience)
- ✅ Chart no longer maintains template logic (ADR's core goal)

### Remaining gaps

- Upstream default propagation — can be addressed via changelog + migration guide
- Fail-early validation — looking forward to an `openab validate` CLI in the future
```

thepagent (approved):

PR #1272, opened 2026-07-01
feat(slack): add Block Kit cancel button for in-flight agent turns

jasonni:

Slack adapter has no way to cancel an in-flight agent turn. Discord has `/cancel`, but Slack blocks third-party slash commands inside thread composers. Users must wait for the agent to finish — or leave the thread entirely.

Closes #

Discord Discussion URL: https://discord.com/channels/1491295327620169908/1491365158868619404/1521847257962647644

## At a Glance

```
User message ──► set_status() ──► post Cancel button (chat.postMessage)
                                         │
                 ┌─────────────────────────┘
                 ▼
           [Cancel] button in thread
                 │
    ┌────────────┴────────────┐
    │ User clicks             │ Turn completes normally
    ▼                         ▼
block_actions             stream_finish()
(Socket Mode)                  │
    │                     delete_message()
    ├─► cancel_session()  (auto-cleanup)
    │   (JSON-RPC → agent stdin)
    └─► mark_cancel_button()
        (chat.update → "~Cancelled~")
        + sentinel "cancelled" in map
```

**OpenClaw:**
Cancels via slash commands (`/acp cancel` aborts current turn, `/acp close` tears down session). Handled by the Gateway command handler, never forwarded to the harness. At the ACP layer, cancel sends `session/cancel` via queue IPC — cooperative, preserves session state. No Block Kit cancel button; cancellation is entirely command-driven. Thread-bound ACP sessions are Discord-only; Slack adapter support is an open issue ([#31518](https://github.com/openclaw/openclaw/issues/31518)).
- Source: https://docs.openclaw.ai/tools/acp-agents

**Hermes Agent:**
Cancels via `/stop` and `/interrupt` slash commands, routed through a gateway interrupt path (`gateway/run.py::_handle_active_session_busy_message`). In Slack threads (where slash commands are blocked), the `!stop` text prefix is used as fallback. Buttons exist only for approvals/clarify, never for cancel. Notable bug ([#26813](https://github.com/NousResearch/hermes-agent/issues/26813)): cancel commands can be swallowed as agent input if not intercepted before the message queue/steer path — our Block Kit button approach avoids this class of bugs entirely since `block_actions` envelopes are a separate Socket Mode envelope type, never mixed into the message stream.
- Source: https://github.com/NousResearch/hermes-agent/blob/main/website/docs/user-guide/messaging/slack.md

**Other references:**
- Slack Block Kit interactive messages: https://api.slack.com/reference/block-kit/block-elements#button
- Slack Socket Mode for interactive payloads: https://api.slack.com/apis/socket-mode
- openab's existing Discord `/cancel` (PR #487 discussion on why Slack slash commands were disabled)

**Summary:** Both OpenClaw and Hermes Agent rely on slash commands for cancel — neither has a Block Kit cancel button. Slash commands don't work in Slack thread composers, making this a known gap. Our approach (Block Kit button via Socket Mode `block_actions`) sidesteps the platform limitation entirely.

## Proposed Solution

Post a standalone Block Kit "Cancel" button (`action_id: "openab_cancel"`, danger style) into the Slack thread at turn start (`set_status`). Handle the `block_actions` interactive envelope via Socket Mode — extract `thread_ts`, construct session key `"slack:{thread_ts}"`, call `cancel_session()` to send `session/cancel` JSON-RPC to the agent, and mark the button as "~Cancelled~" via `chat.update`. On normal turn completion, `stream_finish` auto-deletes the button.

A sentinel pattern (`"cancelled"` stored in the `cancel_buttons` map) prevents duplicate buttons in bot-to-bot conversation scenarios.

## Why this approach?

- **Standalone message** over inline attachment: `chat.startStream`/`appendStream` (native streaming) don't support Block Kit blocks, so the button can't be embedded in the bot's reply during streaming
- **`chat.update`** over `chat.delete` for cancel click: `chat.delete` returns `ok: true` but doesn't visually remove the message in assistant mode threads; `chat.update` works correctly
- **Session pool key `slack:{thread_ts}`** regardless of `BatchGrouping`: cancel handles are keyed by pool key (`Dispatcher::session_key()`), not dispatcher routing key. Using `dispatcher.key()` would produce mismatched keys under Lane mode
- **Sentinel pattern**: after cancel, storing `"cancelled"` instead of removing the map entry prevents `set_status` from re-posting a button when bot-to-bot messages trigger additional turns on the same thread

## Alternatives Considered

1. **Slash command `/cancel`**: Blocked by Slack — third-party slash commands can't be used in thread composers (this is why PR #487 disabled interactive/slash handling)
2. **Embed button in bot reply**: Not feasible — native streaming mode (`chat.startStream`) doesn't support Block Kit blocks. Post+edit mode could work but would require carrying button state through every edit, creating tight coupling
3. **Ephemeral message**: Would avoid the "extra message" problem but ephemeral messages can't be deleted programmatically and would linger after turn completion
4. **`chat.delete` for cancel click**: Tested — returns `ok: true` but the message remains visible in assistant mode threads. Switched to `chat.update` with strikethrough text

## Validation

Rust changes:
- [x] `cargo check` passes
- [x] `cargo clippy` clean

All PRs:
- [x] Manual testing:
  - Sent message → Cancel button appears in thread at turn start
  - Clicked Cancel → button updates to "~Cancelled~", agent receives `session/cancel` (confirmed via pool.rs log)
  - Let turn complete normally → button auto-deleted by `stream_finish`
  - Bot-to-bot conversation → Cancel clicked → no duplicate button appears (sentinel fix verified)
  - Cancel after turn already completed → `cancel_session` logs "no session" warning, `delete_message` handles gracefully

Prerequisites:
- Slack App settings → Interactivity & Shortcuts → enable Interactivity (no Request URL needed for Socket Mode, no new OAuth scopes required)


howie:
Mob review by 3 independent LLM reviewers (Claude / Codex / agy). These are suggestions — the call on what to apply is yours.

# Mob Review — PR #1272

## Blocking concerns (strongly recommend addressing before merge)
1. **No authorization check on who can click Cancel** — `crates/openab-core/src/slack.rs:892-926` (the `openab_cancel` block_actions handler) never reads `payload["user"]["id"]` or checks it against `allowed_users`/`allow_all_users`/`allowed_channels`, unlike every other user-triggered action in this file (compare `slack.rs:1346-1355`). Any Slack user who can see the thread and click the button — not just the configured allowlist — can cancel another user's in-flight agent turn. All 3 reviewers independently converged on this.
   - Suggestion: read `payload["user"]["id"]` and run it through the same authorization check `handle_message` uses, before dispatching `cancel_session`; add a test for an unauthorized click being ignored.

## Important suggestions
1. **`cancel_buttons` map has no bounded eviction**, unlike the sibling `streams` map (`STREAM_CACHE_MAX`) — `slack.rs:90-96`. Root cause: `stream_finish` (the only cleanup path) is only reached from `adapter.rs`'s `native_msg.is_some()` branch (~`adapter.rs:1096-1138`); non-native/tool-only/empty turns never call it, so aborted turns leak both the map entry and an orphaned Slack message indefinitely.
2. **Non-atomic check-then-act races around `cancel_buttons`** — `slack.rs:741-758` (`set_status`'s check/release-lock/await/re-lock pattern allows concurrent duplicate button posts) and `slack.rs:919-927` (the click handler's delayed `insert("cancelled")` can land after `stream_finish` already removed the entry, resurrecting a stale sentinel that permanently suppresses the next turn's button), plus a third interleaving where the stream finishes before a pending button-post lands, orphaning that button. All are the same class of bug (state mutated across an await with no reservation/critical section) and likely share one fix.
   - Suggestion: reserve the map slot under the lock before any `await`; make the click-handler's "cancelled" write conditional on the entry still matching the expected message ts.
3. **Cancel-button UI can report success even when `cancel_session` actually fails** — `slack.rs:919-927`, `acp/pool.rs:437-444`. `pool::cancel_session`'s error is untyped (same value for the benign "already finished" race and a genuine ACP pipe failure), and the Slack handler unconditionally marks `"cancelled"` regardless of the outcome.
4. **`cancel_buttons` field comment doesn't match behavior**: says values are "deleted... on user click," but the click path only updates the message to `~Cancelled~` (`mark_cancel_button` does `chat.update`, never `chat.delete`); real deletion only happens in `stream_finish`.
5. **New interactive-handling logic (parsing, authorization, state races) has no test coverage** — `slack.rs:866-935`, `dispatch.rs:179-181,490-493`.
6. **`message_not_found` matched via substring `.contains()`**, inconsistent with this file's own exact-match convention (`is_block_payload_rejected`, which has a test explicitly guarding against substring false positives) — `slack.rs:242,257`.

## Minor suggestions (nits)
1. Silent no-op when `chat.postMessage` response lacks `ts` (`slack.rs:745-753`) — consider a `warn!` on the missing-field case.
2. Interactive JSON parse failure and unmatched action_ids/payload types are silently dropped with no log (`slack.rs:877-886, 924`) — a regression vs. the prior blanket `debug!` on all interactive envelopes.
3. `mark_cancel_button`'s doc comment omits the idempotency note `delete_message`'s comment states explicitly (`slack.rs:237` vs `:247`).

## Points the reviewers disagreed on (your call)
- One reviewer (agy) raised a possible "IDOR" concern in round 2, suggesting `session_key` could be forged via a button `value` field to cancel an arbitrary session. We checked the source directly and don't think this mechanism holds up: no `value` field is read anywhere in the handler — `session_key` is built from `payload["message"]["thread_ts"]`, which reflects Slack's own record of the actual message the visible button is attached to, not an attacker-supplied string. The underlying worry (an unauthorized user acting on a session) is already covered by the authorization-check item above, so we don't think a separate signing/validation fix is needed here — but flagging in case we're missing something about how the button's payload could be manipulated.

## Voices unavailable
None — Claude, Codex, and agy all completed both review rounds.

ijbhxhu:
PR #1272 Review — Slack Block Kit cancel button

Sound approach (Block Kit button over Socket Mode). One blocker + a cluster of race conditions to fix before merge.

🔴 Blocking

No authorization check on button click (slack.rs ~892–926). The openab_cancel handler dispatches cancel_session without reading payload["user"]["id"], so any user who sees the thread can kill another user's turn, bypassing the allowed_users gate. Fix: run the clicker through handle_message's auth check; add a test for an ignored unauthorized click.

🟠 Same root cause — cancel_buttons mutated across await (fix together)


Duplicate buttons (~736–758): check-lock-release-await-relock lets two concurrent set_status calls both post a button. Reserve the slot under the lock before the await.
Sentinel resurrection (~919–927): late insert("cancelled") after a new turn removed the entry suppresses the next turn's button. Only write if the entry still matches the clicked ts.
Unbounded map + leak (~90–96, ~710–725): cancel_buttons has no STREAM_CACHE_MAX-style bound; only stream_finish cleans up, so error/tool-only/aborted turns leak the entry and orphan the button. Add bounded eviction + a non-stream_finish cleanup path.

⚠️ The suggested fix "clear button when set_status gets status==""" assumes set_status is called with an empty status — not visible in this diff. Confirm the call convention first.





🟡 Minor


UI always shows ~Cancelled~ (~919–927) even if cancel_session fails (untyped error). Mark cancelled only on success.
Substring error match (~242, 257): .contains("message_not_found") violates the file's exact-match convention (is_block_payload_rejected). Match exactly.
No tests for the new interactive/auth/race logic — only builder + dispatch pass-through are covered.


🟢 Info


Doc comment on cancel_buttons says "deleted on user click," but click only does chat.update; deletion is stream_finish-only.
The "IDOR / forged session_key" worry doesn't hold — session_key comes from Slack's message.thread_ts, not a button value. Covered by the auth item above.


Summary: Required before merge = auth check. Strongly recommended = fix the three races together + add tests.

Issue #1269, opened 2026-07-01
trust: make the ingress gate structurally un-bypassable (Discord/Slack + privatize dispatch entries)


chaodu-agent:
Tracking issue from #1267 review (F3), and the durable fix for #1266 review F1.

## Problem
The trust gate (`AdapterRouter::gate_incoming`) is live for the **gateway** path (#1267) — there `process_gateway_event` runs the gate *before* `Dispatcher::submit`, so gateway Thread/Lane batched modes are gated.

But **Discord** and **Slack** call `Dispatcher::submit(...)` / `handle_message(...)` **directly** (discord.rs ~1050/1292, slack.rs ~846/1093/1542), bypassing the gate entirely. Until they route through the gate, those paths are ungated.

## Durable fix (long-term, per the #1264 ADR discussion)
1. Route Discord + Slack ingress through `AdapterRouter::gate_incoming` (a unified `incoming_message` entry).
2. Make `Dispatcher::submit` and `AdapterRouter::handle_message` `pub(crate)`/private so the gate is the **only** public door — no adapter (or future dispatch mode) can skip it by construction.

## Acceptance
- [ ] Discord ingress gated via `gate_incoming`
- [ ] Slack ingress gated via `gate_incoming`
- [ ] `submit` / `handle_message` no longer publicly reachable without passing the gate
- [ ] Remove neutered L2/L3 from `should_skip_event` + unused `GatewayEventContext` fields (Phase 1c)

Blocks the Phase 3 deny-all flip (#1264) — the flip must not land while any ingress path is ungated.

Refs #1264 #1266 #1267

chaodu-agent:
## Resume point — Slack L3 gate (next session)

Progress so far (all merged): Phase 0 (#1266), Phase 1 gateway/Telegram (#1267), Phase 1 Discord (#1270). **Slack is the last adapter still on its own checks** (unaffected — it never calls the shared gate and has no registry entry).

### Slack plan (mirror Discord #1270, with lessons baked in)
- [ ] **Design (a):** add `gate_incoming` to the `DispatchTarget` trait + a `Dispatcher::gate_incoming` passthrough, so `slack.rs::handle_message` (which only has `dispatcher`, not the router) can call `dispatcher.gate_incoming("slack", …)`.
[...]

PR #1264, opened 2026-06-30
docs(adr): identity trust-none default & trust pyramid

chaodu-agent:
Splits the trust/security decision out of #1263 into its own ADR (per discussion — one decision per ADR).

## Scope
- **L1/L2/L3 trust pyramid** — L1 platform auth (security), L2 channel/group/DM scope (NOT security, default-open), L3 identity (security, default deny-all)
- **Identity trust-none default** — flip empty `allowed_users` from allow-all to deny-all
- **Single router gate** at `AdapterRouter::handle_message()` — un-bypassable
- **TrustConfig + carrier/type changes** — extend `MessageContext`/`ChannelRef`, no new trait
- **Echo-on-deny** request-access flow

## Relationship
- **Depends on #1263** (First-Class Per-Platform Config) — per-platform `allowed_users` live in the first-class sections.
- Closes #1262 (jointly with #1263).

Doc: `docs/adr/identity-trust-none.md`

PR #1267, opened 2026-07-01
feat(trust): Phase 1 (gateway) — route gateway ingress through the shared gate

chaodu-agent:
**Phase 1 of #1264** ([plan](https://github.com/openabdev/openab/pull/1264#issuecomment-4848324986)), scoped to the **gateway path** (Telegram/LINE/Feishu/WeCom/Google Chat/Teams). Builds on Phase 0 (#1266).

Behavior-preserving — no message that's accepted today is denied, and vice-versa.

## What this does
- **`AdapterRouter`** gains a `PlatformTrustConfigs` registry (via a `with_trust()` builder, so `new()`'s signature and all its callers stay unchanged) + a single `gate_incoming(platform, channel_id, is_dm, sender_id) -> Decision` ingress gate.
- **`process_gateway_event()`** now enforces **L2 (scope) + L3 (identity) via `router.gate_incoming()`**. `should_skip_event()` keeps only its structural checks (bot filter + @mention gating); its channel/user checks are neutered in the unified path (passed allow-all).
- The registry is built at startup from the same `GATEWAY_*` env the bridge already uses, keyed per gateway platform.

## Why it addresses F1 (from #1266 review)
The gateway's single choke point is `process_gateway_event` — the gate now runs there for **every** gateway event before any dispatch path. Discord/Slack (two-entry submit-vs-handle_message) + making the dispatch entries `pub(crate)` follow in the next PR.

## Behavior-preserving details
- Registry defaults mirror today's `should_skip_event` (`allow_all_*` default true).
- `is_dm = false` in Phase 1 → gateway DMs are evaluated against the channel allowlist exactly as today; the `allow_dm` surface semantics arrive with the per-platform trust flip.
- One intentional micro-hardening: an **empty `sender_id`** is now denied even under `allow_all_users` (Phase-0 F7 fail-closed). Gateway events always carry a sender, so no practical impact.

## Deferred
- Discord/Slack routing through `gate_incoming` + privatizing `submit`/`handle_message` (next PR)
- Removing `should_skip_event`'s now-neutered L2/L3 + unused `GatewayEventContext` fields (Phase 1c)
- Deny-all default flip (Phase 3)

## Testing
- `cargo clippy --features unified -D warnings` clean
- 9/9 gateway tests pass; full `openab-core` suite green except the pre-existing macOS-only `secrets::resolve_exec_nonzero_exit`

[...]

PR #1273, opened 2026-07-01
feat(trust): gateway echo-on-deny (Phase 2) + trust-none default (Phase 3)

chaodu-agent:
Enables **trust-none** for the gateway path (Telegram/LINE/…) with the **request-access echo**, so you can see the deny + echo UX end-to-end. Scoped intent: only pre-beta (self-use) runs this; **emilie is unaffected** (its own image + WS path, which doesn't use this registry/echo).

## Changes
- **Phase 2 — echo on deny**: when the gate returns `DenyIdentity`, reply to the sender with their ID + how to request access (`adapter.send_message`). Throttled to **1 echo per (platform, sender) / 5 min** (LazyLock map) to prevent amplification. `DenyScope` stays **silent** (scope isn't a security boundary).
- **Phase 3 (gateway) — trust-none default**: flip `GATEWAY_ALLOW_ALL_USERS` default `true → false`. Gateway L3 is now deny-all unless `GATEWAY_ALLOWED_USERS` lists the sender (or `GATEWAY_ALLOW_ALL_USERS=true`). **L2 (channels) stays open.**

## ⚠️ Behavior change (intended)
Gateway deployments on this image now **deny unknown users by default**. To keep a bot working:
- set `GATEWAY_ALLOWED_USERS=<uid,…>`, or
- set `GATEWAY_ALLOW_ALL_USERS=true` (old behavior).

Discord/Slack registry entries are **unchanged** (still behavior-preserving allow-all). Deviates from the strict #1269 phase order (deny-flip before privatization) — accepted because it ships only to pre-beta (self-use) for validation.

## How to test (Telegram)
1. On the pre-beta Telegram bot, set `GATEWAY_ALLOWED_USERS=<your uid>` → you're admitted.
2. Message from a **non-listed** account → no agent action + you get: `⚠️ You are not on this bot's trusted list. Your ID: <id> …` (once per 5 min).

## Testing
- clippy `--features unified -D warnings` clean; 10/10 gateway tests (incl. new `echo_allowed_throttles_repeat_within_window`).

Refs #1264 #1269

### DIFF
--- a/crates/openab-core/src/gateway.rs
+++ b/crates/openab-core/src/gateway.rs
@@ -1473,6 +1473,15 @@ mod tests {
     use super::*;
     use std::collections::HashSet;
 
+    #[test]
+    fn echo_allowed_throttles_repeat_within_window() {
+        // Unique key so we don't collide with other tests touching the global map.
+        let key = "test-platform:test-sender-echo-throttle";
+        assert!(echo_allowed(key), "first echo should be allowed");
+        assert!(!echo_allowed(key), "immediate repeat should be throttled");
+        assert!(!echo_allowed(key), "still throttled within the window");
+    }
+
     fn make_event(is_bot: bool, sender_id: &str, channel_id: &str, channel_type: &str, thread_id: Option<&str>, mentions: Vec<&str>) -> GatewayEvent {
         serde_json::from_value(serde_json::json!({
             "schema": "openab.gateway.event.v1",
--- a/crates/openab-core/src/gateway.rs
+++ b/crates/openab-core/src/gateway.rs
@@ -1140,6 +1140,28 @@ pub struct GatewayEventContext {
 ///
 /// This is the core event-handling logic extracted from the WebSocket handler,
 /// made available for the unified binary to call directly from axum webhook handlers.
+/// Throttle for request-access echoes: at most one echo per (platform, sender)
+/// per [`ECHO_WINDOW`], to prevent an untrusted spammer from being amplified by
+/// the bot's replies.
+static ECHO_THROTTLE: std::sync::LazyLock<
+    std::sync::Mutex<std::collections::HashMap<String, std::time::Instant>>,
+> = std::sync::LazyLock::new(|| std::sync::Mutex::new(std::collections::HashMap::new()));
+
+const ECHO_WINDOW: std::time::Duration = std::time::Duration::from_secs(300);
+
+/// Returns true if an echo to `key` is allowed now (and records the timestamp).
+fn echo_allowed(key: &str) -> bool {
+    let now = std::time::Instant::now();
+    let mut map = ECHO_THROTTLE.lock().unwrap();
+    match map.get(key) {
+        Some(prev) if now.duration_since(*prev) < ECHO_WINDOW => false,
+        _ => {
+            map.insert(key.to_string(), now);
+            true
+        }
+    }
+}
+
 pub async fn process_gateway_event(
     event_json: &str,
     ctx: &GatewayEventContext,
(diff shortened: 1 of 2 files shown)


chaodu-agent:
LGTM ✅ — Phase 2 echo-on-deny and Phase 3 trust-none default are well-implemented, correctly scoped, and ship with appropriate safeguards.

## What This PR Does

Implements the gateway trust echo for denied identity requests (Phase 2) and flips the `GATEWAY_ALLOW_ALL_USERS` default from `true` to `false` (Phase 3), enabling a deny-all-by-default posture for gateway-pathed platforms.

## How It Works

- Replaces the simple `if !decision.is_allowed()` block with a `match` on the three `Decision` variants:
  - `Allow` → pass through (noop)
  - `DenyIdentity` → send a throttled echo showing the user their ID + how to request access, then drop
  - `DenyScope` / wildcard → silent drop (no information leakage)
- Throttle state: global `LazyLock<Mutex<HashMap<String, Instant>>>` with a 5-minute window per (platform, sender) key prevents amplification abuse.
- Default flip in `main.rs` is a one-line change with clear comments documenting the migration path (`GATEWAY_ALLOW_ALL_USERS=true` or list `GATEWAY_ALLOWED_USERS`).

## Findings

| 1 | 🟢 | Throttle prevents amplification — good anti-abuse design | `gateway.rs:1144-1162` |
| 2 | 🟢 | DenyScope is correctly silent — no scope config leakage | `gateway.rs:1228-1239` |
| 3 | 🟢 | Clear migration path documented in PR description and code comments | `main.rs:280-283` |
| 4 | 🟢 | Unit test covers the throttle edge case without global state pollution | `gateway.rs:1476-1482` |

<details>

- **Anti-abuse throttle**: The 5-minute per-(platform, sender) rate limit is the right choice. It lets a legitimate user see their ID once, while preventing an attacker from using the echo as an amplification vector.
- **Silent DenyScope**: Using a wildcard arm (`_`) for DenyScope and future variants means new deny reasons default to silent — correct security posture.
- **Phased rollout**: Only gateway-pathed platforms (Telegram, LINE, etc.) are affected. Discord/Slack registry entries are explicitly unchanged. Pre-beta scoping limits blast radius.
- **Self-documenting flip**: The Phase 3 comment in `main.rs` links context (was `true` in #1267, now `false`) so future readers understand the history.

</details>

<details>

- PR opened: 2026-07-01
- Main already has: `Decision` enum with `Allow`/`DenyScope`/`DenyIdentity`, `gate_incoming` call in gateway, simple deny block with TODO comment for Phase 2 echo, `GATEWAY_ALLOW_ALL_USERS` defaulting to `true`
- Net-new value: actual echo implementation with throttle (Phase 2), default flip to deny-all (Phase 3)

</details>

<details>
<summary>Note: Unbounded throttle map (non-blocking)</summary>

The `ECHO_THROTTLE` HashMap grows without eviction — stale entries for senders who never return are never removed. For pre-beta with a handful of users this is fine. If this path ever handles high-cardinality traffic, consider periodic eviction or a bounded LRU cache. Not blocking since this is scoped to pre-beta validation.

</details>

pahud (approved):
## Review

**Code correctness:**
- `Decision` match in `process_gateway_event` is exhaustive (`Allow` / `DenyIdentity` / `DenyScope` via `_`) — matches the enum in `trust.rs`. Good use of `_` to also catch any future variant defensively.
- Echo throttle (`ECHO_THROTTLE` + `echo_allowed`) is a simple per-`(platform, sender)` 5-min window backed by a `Mutex<HashMap>` — correct and adequately tested (`echo_allowed_throttles_repeat_within_window`). Fine for pre-beta self-use scale; would want a TTL/eviction sweep if this ever sees real traffic (unbounded map growth), but not a blocker at current scope.
- `DenyScope` staying silent (no echo) is correctly scoped per the L2/L3 trust model from the ADR.
- The `GATEWAY_ALLOW_ALL_USERS` default flip (`true → false`) is a single, well-isolated line change with a clear comment explaining the behavior change.
- `send_message` errors are intentionally swallowed (`let _ = ...`) on the echo path — reasonable, since a failed echo shouldn't fail event processing, though consider at least a `tracing::warn!` on error for observability.

**Scope/safety:**
- Correctly scoped to pre-beta only per the PR description (emilie unaffected).
- Discord/Slack registries unchanged — consistent with the stated deviation from strict phase ordering.

**Gap found:** `docs/config-reference.md` still documents `GATEWAY_ALLOW_ALL_USERS` default as `true`. This PR changes the code default to `false` but doesn't update that doc table — it'll be actively wrong once merged. Not a blocker, but should be fixed in this PR or immediately after.

**CI:** all checks green (33/33 smoke tests, clippy, unit tests).

Approving — please push a doc fix for the `GATEWAY_ALLOW_ALL_USERS` default in `docs/config-reference.md` before or right after merge.

thepagent (approved):
Approved as code owner. See detailed review from @pahud above — code is correct (exhaustive Decision match, tested throttle, well-scoped default flip). One follow-up needed: `docs/config-reference.md` still lists `GATEWAY_ALLOW_ALL_USERS` default as `true`; should be updated to `false` to match this change.

PR #1280, opened 2026-07-03
fix(operator): apply_ecs missing taskRoleArn/executionRoleArn/logConfiguration

chaodu-agent:
## Summary

`oabctl apply`'s `register_task_definition` call never set `taskRoleArn`, `executionRoleArn`, or a log driver, and set an unused `CONFIG_S3_PATH` env var that nothing in the codebase reads. Found and fixed while deploying a real Telegram bot with `spec.ingress` + `spec.secrets` end-to-end against a live AWS account (the `oab` cluster managed by `oabctl` — **not** the separate Terraform-managed `openab` cluster that runs the actual production fleet, chaodu/koudu/etc., which already sets all of this correctly and is completely unaffected by this change).

Five bugs, in the order they were uncovered — each one only became visible after fixing the previous one:

1. **Missing `executionRoleArn`** — any manifest with `spec.secrets` failed task registration outright ("you must also specify a value for `executionRoleArn`"). Fixed by resolving it from bootstrap state, matching how cluster/subnets/task role are already sourced from bootstrap rather than the manifest.
2. **Execution role missing `secretsmanager:GetSecretValue`** — ECS uses the **execution** role (not the task role) to fetch `spec.secrets` values before the container starts. `bootstrap.rs` only granted this policy to the task role, so fix #1 alone still failed with `AccessDeniedException`. Added the same `oab-secrets` policy to `oab-task-execution`, applied unconditionally so `oabctl bootstrap` self-heals existing installs too — verified by detaching the policy and re-running `bootstrap`, which reattached it automatically.
3. **No `logConfiguration`** — container failures were completely opaque, not even an empty CloudWatch log stream existed, which is what made bugs #4 and #5 hard to diagnose. Wired to the `/oab/agents` log group bootstrap already creates.
4. **`CONFIG_S3_PATH` was dead code** — `apply_ecs` set this env var but nothing ever reads it; the image's default `CMD` points at a local `/etc/openab/config.toml` that nothing populates. `openab` actually has native `s3://` config-source support (the `config-s3` feature, in the default feature set), so the real fix is passing `configFrom` directly as the `-c` argument to `openab run`.
5. **Missing `taskRoleArn`** — the actual root cause of the remaining crash loop after fixes 1–4. ECS only provisions the `AWS_CONTAINER_CREDENTIALS_RELATIVE_URI` credentials endpoint (and injects that env var into the container) when a task role is set on the task definition. Without it, the running `openab` process had zero AWS credentials for its own SDK calls — it fell through every credential provider and finally tried IMDS, which doesn't exist on Fargate, failing with an opaque "dispatch failure". I initially misdiagnosed this as a `tini`/PID-1 environment-inheritance issue; two controlled debug ECS tasks (disproved that (`fork()` env inheritance isn't gated on PID 1), then diffing `describe-task-definition` JSON between a working debug task and the failing one showed the actual difference: `taskRoleArn: null`.

Also documented both IAM roles' permissions in `operator/README.md` — previously only the task role's policies were listed; the execution role's requirements were entirely undocumented.

## Testing

Verified end-to-end against a live AWS account:

1. Applied a Telegram bot manifest with `spec.ingress` + `spec.secrets`.
2. After fixes 1–4: task registered and ran, but crash-looped with `failed to fetch S3 config ... dispatch failure`.
3. Enabled `RUST_LOG=debug` via a temporary task-def revision to see the AWS credential chain fall through every provider down to a failed IMDS call.
4. After fix 5: `oabctl apply --wait` reported `is stable`; service reached steady state and stayed at `runningCount=1` for 90+ seconds (previously crash-looped within 10–90s every time); `curl` through the printed webhook URL reached the running task.
5. Cleaned up all test/debug ECS task definitions and scaled the test service back to 0 afterward — no leftover resources.

Local build verification (M4 macmini, nested workspace layout mirroring the CI `operator` job):
- `cargo build` — clean
- `cargo clippy --all-targets -- -D warnings` — clean
- `cargo test` — **15 passed**

## Risk

Scoped entirely to `oabctl`'s own `oab` cluster/manifest schema. Confirmed the production fleet (`openab` cluster: chaodu, koudu, etc.) is provisioned by a separate mechanism (different role-naming convention, Gist-based config, already-correct `taskRoleArn`/`logConfiguration`) and does not go through this code path at all.

### DIFF
--- a/operator/README.md
+++ b/operator/README.md
@@ -374,10 +374,24 @@ oabctl bootstrap --cluster my-cluster     # import existing resources
 
 ### IAM Task Role Permissions
 
+Attached to `oab-task-role` — the identity the *running container* assumes.
+
 | Policy | Permissions | Resource |
 |--------|------------|----------|
 | `oab-ecs-exec` | ssmmessages:* | * (ECS Exec requirement) |
 | `oab-s3-artifacts` | s3:GetObject, s3:PutObject | `{bucket}/artifacts/*` |
+| `oab-secrets` | secretsmanager:GetSecretValue | `arn:aws:secretsmanager:*:*:secret:oab/*` |
+
+### IAM Execution Role Permissions
+
+Attached to `oab-task-execution` — the identity **ECS itself** assumes to pull
+the image and resolve `spec.secrets` values *before* the container starts.
+This is a different role from the task role above; a manifest's `spec.secrets`
+values are fetched by this role, not by the running container.
+
+| Policy | Permissions | Resource |
+|--------|------------|----------|
+| `AmazonECSTaskExecutionRolePolicy` (AWS managed) | ECR image pulls, CloudWatch log delivery | * |
 | `oab-secrets` | secretsmanager:GetSecretValue | `arn:aws:secretsmanager:*:*:secret:oab/*` |
 
 ### Import Existing Resources
--- a/operator/src/apply.rs
+++ b/operator/src/apply.rs
@@ -14,12 +14,38 @@ use std::path::Path;
 /// rejects the request with "you must also specify a value for
 /// 'executionRoleArn'" otherwise.
 async fn load_bootstrap_state(config: &aws_config::SdkConfig) -> Option<BootstrapState> {
-    let sts = aws_sdk_sts::Client::new(config);
-    let account = sts.get_caller_identity().send().await.ok()?
-        .account()?.to_string();
-    let bucket = format!("oab-control-plane-{account}");
+    let bucket = if let Some(b) = crate::config::OabConfig::load().ok().and_then(|c| c.bucket()) {
+        b
+    } else {
+        let sts = aws_sdk_sts::Client::new(config);
+        let identity = match sts.get_caller_identity().send().await {
+            Ok(id) => id,
+            Err(e) => {
+                eprintln!("  ⚠ load_bootstrap_state: STS get_caller_identity failed: {e}");
+                return None;
+            }
+        };
+        let account = match identity.account() {
+            Some(a) => a.to_string(),
+            None => {
+                eprintln!("  ⚠ load_bootstrap_state: STS response missing account field");
+                return None;
+            }
+        };
+        format!("oab-control-plane-{account}")
+    };
     let s3 = aws_sdk_s3::Client::new(config);
-    crate::bootstrap::load_state_pub(&s3, &bucket).await.ok().flatten()
+    match crate::bootstrap::load_state_pub(&s3, &bucket).await {
+        Ok(Some(state)) => Some(state),
+        Ok(None) => {
+            eprintln!("  ⚠ load_bootstrap_state: no bootstrap state found in s3://{bucket}/bootstrap-state.json (run `oabctl bootstrap` first)");
+            None
+        }
+        Err(e) => {
+            eprintln!("  ⚠ load_bootstrap_state: failed to read bootstrap state from s3://{bucket}: {e}");
+            None
+        }
+    }
 }
 
 pub async fn run(aws_config: &aws_config::SdkConfig, file_path: &str, sync_config: bool, wait: bool) -> Result<()> {
@@ -223,9 +249,8 @@ async fn apply_ecs(
     // resolve region via the standard chain: AWS_REGION env var → profile →
     // IMDS. Fargate tasks have no EC2 instance metadata to fall back to, so
     // without this the SDK can fail to resolve an endpoint at all.
-    if let Some(region) = config.region() {
-        env_vars.push(KeyValuePair::builder().name("AWS_REGION").value(region.as_ref()).build());
-    }
+    // Region is injected below after bootstrap_state is loaded (to allow
+    // fallback to bootstrap_state.region when config.region() is None).
     if let Some(ref bootstrap) = m.spec.bootstrap_from {
         env_vars.push(KeyValuePair::builder().name("BOOTSTRAP_FROM").value(bootstrap).build());
     }
@@ -246,6 +271,16 @@ async fn apply_ecs(
     // can or should specify directly (bootstrap owns these, not the manifest —
     // see operator/README.md's "Resources Created" section).
     let bootstrap_state = load_bootstrap_state(config).await;
+
+    // Resolve effective region: prefer SDK config, fall back to bootstrap
+    // state's recorded region. Fargate has no IMDS, so without AWS_REGION the
+    // container's SDK calls will fail to resolve endpoints entirely.
+    let effective_region: Option<String> = config.region()
+        .map(|r| r.as_ref().to_string())
+        .or_else(|| bootstrap_state.as_ref().map(|s| s.region.clone()));
+    if let Some(ref region) = effective_region {
+        env_vars.push(KeyValuePair::builder().name("AWS_REGION").value(region).build());
+    }
 
     let mut container = ContainerDefinition::builder()
         .name("openab")
@@ -275,12 +310,12 @@ async fn apply_ecs(
     // ECS uses no log driver and task failures are opaque (no log stream at
     // all, not even an empty one).
     if let Some(log_group) = bootstrap_state.as_ref().map(|s| &s.resources.log_group) {
-        if let Some(region) = config.region() {
+        if let Some(ref region) = effective_region {
             container = container.log_configuration(
                 aws_sdk_ecs::types::LogConfiguration::builder()
                     .log_driver(aws_sdk_ecs::types::LogDriver::Awslogs)
                     .options("awslogs-group", log_group.as_str())
-                    .options("awslogs-region", region.as_ref())
+                    .options("awslogs-region", region.as_str())
                     .options("awslogs-stream-prefix", &service_name)
                     .options("awslogs-create-group", "true")
                     .build()?,
--- a/operator/src/apply.rs
+++ b/operator/src/apply.rs
@@ -8,8 +8,11 @@ use aws_sdk_ecs::types::{
 use aws_sdk_s3::primitives::ByteStream;
 use std::path::Path;
 
-/// Try to load bootstrap state for networking defaults (used in future for auto-fill)
-#[allow(dead_code)]
+/// Load bootstrap state to resolve the task execution role ARN (and other
+/// networking defaults). Required on `register_task_definition` whenever the
+/// task uses ECS-injected secrets (or pulls from a private registry) — ECS
+/// rejects the request with "you must also specify a value for
+/// 'executionRoleArn'" otherwise.
 async fn load_bootstrap_state(config: &aws_config::SdkConfig) -> Option<BootstrapState> {
     let sts = aws_sdk_sts::Client::new(config);
     let account = sts.get_caller_identity().send().await.ok()?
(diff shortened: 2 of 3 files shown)


chaodu-agent:
LGTM ✅ — Solid, well-sequenced fix for five real bugs blocking `oabctl apply` with `spec.secrets`/`spec.ingress` on Fargate.

## What This PR Does

`oabctl apply` could never successfully run a task with `spec.secrets` — it was missing `executionRoleArn`, `taskRoleArn`, and `logConfiguration` from the task definition, had a dead `CONFIG_S3_PATH` env var, and the execution role lacked `secretsmanager:GetSecretValue`. This PR resolves all five by sourcing the required ARNs/config from bootstrap state.

## How It Works

1. Calls `load_bootstrap_state()` (previously dead code) to resolve `execution_role_arn`, `task_role_arn`, and `log_group` from the S3-persisted bootstrap state.
2. Wires `logConfiguration` with the `awslogs` driver pointing to the bootstrap-created `/oab/agents` log group.
3. Replaces the dead `CONFIG_S3_PATH` env var with a proper `command` override (`openab run -c <configFrom>`) leveraging openab's native S3 config-source support.
4. Injects `AWS_REGION` env var so the SDK can resolve endpoints on Fargate (no IMDS).
5. Adds `oab-secrets` inline policy to the managed execution role in `bootstrap.rs` (idempotent `put_role_policy`).
6. Bails early with actionable error messages when bootstrap state is missing.

## Findings

| 1 | 🟢 | Asymmetric bail logic is correct — execution role only required with secrets, task role always required | `apply.rs:316-327` |
| 2 | 🟢 | Self-healing bootstrap: `put_role_policy` is idempotent, re-running bootstrap reattaches the policy | `bootstrap.rs:276-288` |
| 3 | 🟢 | Thorough documentation of both IAM roles in README | `README.md` |
| 4 | 🟢 | Excellent PR description with clear debugging narrative and risk scoping | — |

<details>

- **Correct error-handling asymmetry**: `executionRoleArn` is conditionally required (only when `spec.secrets` is non-empty), matching ECS semantics exactly. `taskRoleArn` unconditionally bails because the container always needs AWS credentials.
- **Bootstrap self-heal**: The `oab-secrets` policy on the execution role is applied via `put_role_policy` (idempotent) outside the "first create" guard, so re-running `bootstrap` fixes existing installs without duplication.
- **Clean config-source fix**: Replacing the dead `CONFIG_S3_PATH` env var with a `command` override (`openab run -c`) is the correct approach — it uses openab's built-in S3 config loading rather than requiring a sidecar/init script.
- **Region injection**: Adding `AWS_REGION` for Fargate is essential since there's no IMDS to fall back on.
- **Scoped risk**: Only affects the `oab` cluster path (oabctl-managed), not the production Terraform-managed fleet.
- **End-to-end tested**: Verified against a live AWS account with full crash-loop → stable-service progression.

</details>

<details>

- PR opened: 2026-07-03
- Main already has: `load_bootstrap_state()` function (dead code with `#[allow(dead_code)]`), `BootstrapState` with all required fields, `oab-secrets` policy on task role only
- Net-new value: Actually wires bootstrap state into task registration, adds execution role secrets policy, adds log configuration, fixes config loading, adds region env var — none of this existed on main

</details>

chaodu-agent:
CHANGES REQUESTED ⚠️ — `load_bootstrap_state` hardcodes bucket name bypassing custom config, breaking deployments with non-default control-plane buckets.

## What This PR Does

`oabctl apply` never set `taskRoleArn`, `executionRoleArn`, or `logConfiguration` on the ECS task definition, and used a dead `CONFIG_S3_PATH` env var. This PR fixes all five bugs by sourcing the required values from bootstrap state and wiring log configuration + native S3 config command override.

## How It Works

1. Calls `load_bootstrap_state()` to resolve `execution_role_arn`, `task_role_arn`, and `log_group` from S3-persisted bootstrap state.
2. Wires `logConfiguration` with the `awslogs` driver pointing to the bootstrap-created log group.
3. Replaces the dead `CONFIG_S3_PATH` env var with a `command` override (`openab run -c <configFrom>`).
4. Injects `AWS_REGION` env var for Fargate containers (no IMDS fallback).
5. Adds `oab-secrets` inline policy to the execution role in `bootstrap.rs`.
6. Bails early with actionable error messages when bootstrap state is missing.

## Findings

| 1 | 🔴 | `load_bootstrap_state` hardcodes bucket to `oab-control-plane-{account}`, bypassing `OabConfig` custom bucket resolution — breaks custom-bucket deployments | `apply.rs:13-20` |
| 2 | 🟡 | `load_bootstrap_state` silently swallows all errors (STS failure, missing account, S3 read failure) via `.ok()?` — should at least `eprintln!` a warning | `apply.rs:13-20` |
| 3 | 🟡 | `AWS_REGION` env var and `logConfiguration` are silently skipped when `config.region()` is `None` — should fall back to `bootstrap_state.region` | `apply.rs:222-225, 271-283` |
| 4 | 🟡 | Mandatory bootstrap state coupling is a breaking change for users who previously ran `apply` without `bootstrap` — needs migration note or graceful degradation | `apply.rs:336-342` |
| 5 | 🟡 | `config_from` empty path relies on upstream `validate()` to prevent container crash — internal API boundary lacks defensive check | `apply.rs:254-260` |
| 6 | 🟢 | Asymmetric bail logic (execution role conditional, task role unconditional) correctly matches ECS semantics | `apply.rs:316-327` |
| 7 | 🟢 | `oab-secrets` IAM policy properly scoped to `arn:aws:secretsmanager:*:*:secret:oab/*` | `bootstrap.rs:276-288` |
| 8 | 🟢 | Native S3 config via command override avoids sidecar/entrypoint complexity | `apply.rs:254-268` |
| 9 | 🟢 | README documents both IAM roles clearly with permission tables | `README.md` |

<details>

### 🔴 F1: `load_bootstrap_state` bypasses custom S3 bucket config

The existing `apply_ecs` function resolves the control-plane bucket via `OabConfig::load().ok().and_then(|c| c.bucket())`, which respects `~/.oabctl/config.toml` and `OAB_CONTROL_PLANE_BUCKET` env var. However, the new `load_bootstrap_state` hardcodes `format!("oab-control-plane-{account}")`, ignoring custom configurations.

For deployments using a custom bucket name, `load_bootstrap_state` will fail to find bootstrap state and the function returns `None`, causing a false-positive bail: `"no bootstrap task role was found — run oabctl bootstrap first"`.

**Suggested fix:**
```rust
async fn load_bootstrap_state(config: &aws_config::SdkConfig) -> Option<BootstrapState> {
    let bucket = if let Some(b) = crate::config::OabConfig::load().ok().and_then(|c| c.bucket()) {
        b
    } else {
        let sts = aws_sdk_sts::Client::new(config);
        let account = sts.get_caller_identity().send().await.ok()?
            .account()?.to_string();
        format!("oab-control-plane-{account}")
    };
    let s3 = aws_sdk_s3::Client::new(config);
    crate::bootstrap::load_state_pub(&s3, &bucket).await.ok().flatten()
}
```

### 🟡 F2: Silent error swallowing in `load_bootstrap_state`

All errors (STS call failure, missing account field, S3 read failure, JSON parse failure) are converted to `None` via `.ok()?` and `.ok().flatten()`. When bootstrap has never been run, the resulting bail message is correct, but when there is an *unexpected* failure (e.g., IAM permissions, network timeout), the user gets no indication of the root cause.

**Suggested fix:** Add `eprintln!` on the None path, or accept a `&str` reason parameter for the bail messages downstream.

### 🟡 F3: Region fallback missing

`AWS_REGION` injection and `logConfiguration` are both gated on `config.region().is_some()`. On Fargate, if the SDK config was constructed without an explicit region (edge case but possible), the container launches without `AWS_REGION` and without log configuration — making failures opaque (the exact problem this PR aims to solve).

**Suggested fix:** Fall back to `bootstrap_state.region`:
```rust
let effective_region = config.region()
    .map(|r| r.as_ref().to_string())
    .or_else(|| bootstrap_state.as_ref().map(|s| s.region.clone()));
```

### 🟡 F4: Breaking change — mandatory bootstrap state

Previously `oabctl apply` could register a task definition without `bootstrap` having been run (it just lacked roles/logging). Now it unconditionally bails if bootstrap state is not found. While this is arguably correct (the task would fail at runtime anyway), it is a breaking change for existing workflows.

**Suggested fix:** Document in README/CHANGELOG that `oabctl bootstrap` is now a prerequisite for `oabctl apply`, or add a `--no-bootstrap` escape hatch.

### 🟡 F5: Defensive gap on empty `config_from`

If `config_from` is empty, no `command` override is set, and the container uses the image default CMD which points to a nonexistent local config file. Currently `validate()` prevents this, but `apply_ecs` is a `pub(crate)` function — an internal caller could bypass validation.

**Suggested fix:** Low priority. Either make `apply_ecs` private, or add a debug assertion.

</details>

<details>

- PR opened: 2026-07-03
- Main already has: `load_bootstrap_state()` (dead code), `BootstrapState` with all required fields, `oab-secrets` on task role only
- Net-new value: Wires bootstrap state into task registration, adds execution role secrets policy, adds log config, fixes config loading, adds region env var
- Production fleet (`openab` cluster) does not use this code path — scoped to `oab` cluster only

</details>

<details>

- Five bugs discovered and fixed in logical dependency order — each fix revealed the next
- Bail logic asymmetry is intentional and correct (execution role conditional on secrets, task role always required)
- `oab-secrets` policy on execution role uses idempotent `put_role_policy` for self-healing
- README now documents both IAM roles with full permission tables
- Native S3 config command override is clean — no sidecar, init container, or entrypoint script needed
- End-to-end verified against live AWS with full crash-loop → stable progression

</details>

chaodu-agent:
LGTM ✅ — Comprehensive fix for five real bugs uncovered via end-to-end testing; correctly resolves missing IAM role ARNs and log configuration from bootstrap state.

## What This PR Does

`oabctl apply` was fundamentally broken for any manifest using `spec.secrets` — it never set `executionRoleArn`, `taskRoleArn`, or a log driver on `register_task_definition`. This PR fixes all five bugs (discovered sequentially), adds the `oab-secrets` policy to the execution role in `bootstrap.rs`, replaces the dead `CONFIG_S3_PATH` env var with a proper `openab run -c` command override, and documents the execution role's IAM permissions.

## How It Works

1. **`load_bootstrap_state`** is rewritten with proper error handling (previously `#[allow(dead_code)]` / never called from `apply_ecs`) and used to resolve `execution_role_arn`, `task_role_arn`, `log_group`, and `region` from the control-plane S3 bucket.
2. **Task registration** now conditionally sets both role ARNs, bailing with a clear error message if bootstrap state is missing.
3. **Log configuration** wires the container to the bootstrap-created `/oab/agents` log group via `awslogs` driver.
4. **`CONFIG_S3_PATH`** (dead code) is replaced by setting `command` on the container definition to `openab run -c <configFrom>`, using the native S3 config-source feature.
5. **`bootstrap.rs`** adds `oab-secrets` inline policy to `oab-task-execution` (idempotent, self-heals existing installs).

## Findings

| 1 | 🟢 | Excellent debugging narrative — each bug is documented with root-cause analysis | PR description |
| 2 | 🟢 | Proper fail-fast: `anyhow::bail!` with actionable message when bootstrap state is missing | `apply.rs:358-367` |
| 3 | 🟢 | `put_role_policy` is correctly applied unconditionally (self-heal) with `.ok()` to avoid breaking bootstrap if IAM permissions are restricted | `bootstrap.rs:280-290` |
| 4 | 🟢 | AWS_REGION injection handles the no-IMDS Fargate case correctly via fallback chain | `apply.rs:275-280` |
| 5 | 🟢 | Scoped blast radius — only affects `oabctl`'s own `oab` cluster, production fleet uses separate provisioning | — |

<details>

- PR opened: 2026-07-03
- Main already has: `load_bootstrap_state` (dead code, `#[allow(dead_code)]`), task role `oab-secrets` policy, basic `register_task_definition` without roles/logs
- Net-new value: Actually wires bootstrap state into `apply_ecs`, adds execution role secrets policy, log configuration, proper command override, and documentation

</details>

<details>

- Thorough end-to-end verification against a real AWS account with detailed reproduction steps
- Each fix is minimal and targeted — no unnecessary refactoring bundled in
- Error messages are actionable (tells user to run `oabctl bootstrap`)
- `load_bootstrap_state` rewrite has proper error branches with diagnostic `eprintln!` messages
- Documentation added for previously-undocumented execution role permissions
- Clean local build: `cargo build` + `clippy` + `cargo test` (15 passed)

</details>

thepagent (approved):

PR #1285, opened 2026-07-04
feat(operator): auto-register Telegram webhook, aws-sm:// secret refs

chaodu-agent:
## Summary

Two related additions requested after the last E2E test of `oabctl apply`
against a real Telegram bot deployment (#1283): registering the webhook was
still a manual `curl setWebhook` step, and the only way to reference a
Secrets Manager secret in `spec.secrets` was the verbose ECS-native
`valueFrom` ARN format.

### 1. Auto-register the Telegram webhook

`apply` now calls Telegram's `setWebhook` API directly when
`/webhook/telegram` is one of `spec.ingress.paths` and `spec.secrets` has a
`TELEGRAM_BOT_TOKEN` entry — no manual `curl` step needed after deploying a
bot. If `spec.secrets` also has `TELEGRAM_SECRET_TOKEN`, it's passed through
so Telegram signs every webhook request with it.

```
apply already succeeded (ECS task running, API Gateway → VPC Link →
Cloud Map → task all wired up, webhook URL printed)
          │
          ▼
  /webhook/telegram in spec.ingress.paths?
   └─ no ─────────────────────────────► skip (no-op)
   └─ yes
          │
          ▼
  TELEGRAM_BOT_TOKEN in spec.secrets?
   └─ no ─────────────────────────────► skip (no-op)
   └─ yes
          │
          ▼
  resolve TELEGRAM_BOT_TOKEN (+ TELEGRAM_SECRET_TOKEN if present)
  from Secrets Manager
          │
          ▼
  POST https://api.telegram.org/bot<token>/setWebhook
  { url: <printed-webhook-url>, secret_token: <optional> }
          │
     ┌────┴────┐
     ▼         ▼
   200 OK    error
     │         │
     ▼         ▼
  ✓ printed  ⚠ printed as a warning — apply still exits 0,
  webhook       provisioning already succeeded; register
  registered    manually with the printed URL instead
```

This is best-effort by design: the AWS-side provisioning is already done and
correct by the time this runs, so a Telegram API hiccup (bad token, network
blip) shouldn't roll any of that back or fail the whole command. LINE has no
`setWebhook`-equivalent API, so its URL still needs manual registration in
the LINE Developers console.

### 2. `aws-sm://<secret-id>#<json-key>` shorthand for `spec.secrets`

`spec.secrets` previously required the full ECS-native `valueFrom` format:

```yaml
spec:
  secrets:
[...]

PR #1288, opened 2026-07-04
fix(operator): resolve_string mishandled ECS-native :jsonKey:: suffix

chaodu-agent:
## Summary

Fixes a real bug found within minutes of merging #1285 by re-testing `oabctl
apply` live against a real manifest.

`resolve_string` (used to fetch `TELEGRAM_BOT_TOKEN`'s plaintext value for
the new auto-`setWebhook` call) claimed in its doc comment that
`GetSecretValue` "resolves natively" an ARN carrying the ECS-only
`:<jsonKey>::` suffix — that claim was simply wrong. AWS's own docs confirm
`GetSecretValue`'s `SecretId` parameter only ever accepts a plain ARN or
secret name; the `:<jsonKey>::` JSON-key-extraction suffix is purely an ECS
`valueFrom` convention, resolved by ECS itself at container launch via
`register_task_definition`, not something the Secrets Manager API
understands at all.

Result: any manifest using the ECS-native `valueFrom` ARN format for
`TELEGRAM_BOT_TOKEN` — which is Format 1 in the README, i.e. the more common
of the two documented formats — failed webhook registration with `failed to
resolve TELEGRAM_BOT_TOKEN`, even though the exact same value works fine as
an ECS task-definition secret (that path goes through `resolve_value_from`,
which never had this bug).

```
TELEGRAM_BOT_TOKEN: "arn:...:secret:oab/telegram/mybot-AC80TP:TELEGRAM_BOT_TOKEN::"
                                                              └──────┬──────┘
                                                         ECS-only suffix —
                                                    GetSecretValue rejects this
                                                    as part of the SecretId

Before: resolve_string(value) → GetSecretValue(secret_id=value) → error
After:  resolve_string(value) → split_ecs_json_key_suffix(value)
                                   → GetSecretValue(secret_id=<base-arn>)
                                   → extract "TELEGRAM_BOT_TOKEN" from the
                                     returned JSON secret string manually
```

## Fix

Added `split_ecs_json_key_suffix`, which detects and strips the ECS-only
suffix from an ARN-shaped value (only ARNs — a bare secret name is never
split), then `resolve_string` extracts the JSON key manually from the
fetched secret string — the same way it already did for the `aws-sm://`
form. `resolve_value_from` (used to build the actual ECS task definition,
where ECS *does* understand the suffix) is untouched and unaffected.

## Testing done

```
cargo build --release                        # clean
cargo clippy --all-targets -- -D warnings    # clean
cargo test --release                         # 29 passed; 0 failed
```

3 new unit tests, including the exact real-world ARN value that surfaced
the bug (`arn:aws:secretsmanager:us-east-1:903779448426:secret:oab/telegram/pahudxbot-AC80TP:TELEGRAM_BOT_TOKEN::`)
as a regression test.

**Live E2E**: re-ran `oabctl apply` against the real manifest that hit this
bug (`TELEGRAM_BOT_TOKEN` as an ECS-native ARN). Before the fix:
`⚠ Telegram webhook registration failed (apply still succeeded): failed to
resolve TELEGRAM_BOT_TOKEN`. After: `✓ Telegram webhook registered: Webhook
was set`.

## Update: addressed 2 code review findings

A review flagged two real gaps in the initial `split_ecs_json_key_suffix`:

| # | Finding | Verdict |
|---|---------|---------|
| 1 | `arn:...:secret:mysecret::` gets misparsed — `mysecret` is treated as a json-key, but it's actually part of the secret name (a full-secret-value reference with all suffix fields empty) | Real bug, cheap fix — confirmed by simulating the exact value |
| 2 | The ECS suffix isn't just `:<jsonKey>::` — it's 3 positional fields (`<json-key>:<version-stage>:<version-id>`), unhandled here | Real gap, correctly identified as a feature-boundary — verified against [AWS's own ECS docs](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/secrets-envvar-secrets-manager.html) worked examples |

Both fixed:
1. Locate the unambiguous `:secret:` marker in the ARN and treat everything
   up to the *next* colon as the secret-name segment (which per AWS's docs
   never itself contains a colon) — rather than blindly `rsplit`-ing on the
   last colon, which can't distinguish "secret name" from "json-key" when
   the suffix fields are all empty.
2. Parse all three positional fields; fail closed with a clear, specific
   error when version-stage/version-id is non-empty, instead of silently
   falling through to "no suffix" (which would then fail later at
   `GetSecretValue` with a confusing, unrelated error). `oabctl`'s own use
   case — fetching a secret's plaintext value in-process to call a
   third-party API — never needs to pin a rotation version, so this is an
   intentional, documented limitation, not a missing feature to build out
   further right now.

7 new tests added covering both findings, using AWS's own documented
example ARN shapes (key-only, version-stage-only, version-id-only,
key+version-stage). Re-verified live against the real manifest from the
original bug report — still resolves correctly, no regression.

### DIFF
--- a/operator/src/secrets.rs
+++ b/operator/src/secrets.rs
@@ -2,9 +2,10 @@
 //!
 //! Values can be either a Secrets Manager reference in ECS-native
 //! `valueFrom` format directly (a full ARN, optionally suffixed with
-//! `:<jsonKey>::` to extract one field of a JSON secret), or the same
-//! `aws-sm://<secret-id>#<json-key>` shorthand openab itself uses for
-//! in-app secret refs (see `crates/openab-core/src/secrets.rs`) — kept
+//! `:<jsonKey>::` to extract one field of a JSON secret — an ECS-only
+//! convention; the Secrets Manager API itself has no knowledge of it), or
+//! the same `aws-sm://<secret-id>#<json-key>` shorthand openab itself uses
+//! for in-app secret refs (see `crates/openab-core/src/secrets.rs`) — kept
 //! identical here so a manifest author can write one convention across both
 //! `spec.secrets` (consumed by ECS at container launch) and `config.toml`
 //! (consumed by openab itself at runtime).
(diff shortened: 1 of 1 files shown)


chaodu-agent:
LGTM ✅ — Correct fix for a real functional bug: `GetSecretValue` never understood ECS-native `:<jsonKey>::` suffixes.

## What This PR Does

Fixes `resolve_string` passing an ECS-native ARN (with `:<jsonKey>::` suffix) directly to `GetSecretValue`, which rejects it — the suffix is an ECS-only convention, not something the Secrets Manager API understands. This broke webhook registration for the most common `TELEGRAM_BOT_TOKEN` format (Format 1 in the README), introduced by #1285.

## How It Works

Adds `split_ecs_json_key_suffix()` to detect and strip the ECS-only `:<jsonKey>::` tail from ARN-shaped values before calling `GetSecretValue`, then manually extracts the JSON key from the fetched secret string — the same approach already used for `aws-sm://` refs. The two code paths in `resolve_string` are unified into a single fetch-then-optionally-extract flow, eliminating duplication. `resolve_value_from` (where ECS *does* understand the suffix) is correctly left untouched.

## Findings

| 1 | 🟢 | Correct root-cause analysis — aligns with AWS docs: `SecretId` only accepts plain ARN or name | `secrets.rs:59-80` |
| 2 | 🟢 | Clean unification of both resolve paths removes duplication and reduces future bug surface | `secrets.rs:90-109` |
| 3 | 🟢 | Good defensive guards (ARN-only, non-empty key, base re-validated) prevent false positives | `secrets.rs:63-78` |
| 4 | 🟢 | Real-world regression test with exact ARN that surfaced the bug | `secrets.rs:115-123` |

<details>

- **Spec-correct**: The fix matches AWS Secrets Manager `GetSecretValue` documentation exactly — `SecretId` only accepts a plain ARN or secret name, never the ECS `valueFrom` suffix convention.
- **Minimal blast radius**: Only `resolve_string` is changed; `resolve_value_from` (the ECS path that *should* keep the suffix) is untouched.
- **Code simplification**: Unifying both branches into one fetch + optional JSON extraction reduces duplication and makes the logic easier to reason about.
- **Solid test coverage**: Three unit tests covering the key variants (ARN with suffix, plain ARN, bare name), including the exact real-world value that exposed the bug.
- **Live E2E verified**: PR description confirms the fix was validated against the actual failing manifest end-to-end.

</details>

<details>

- PR opened: 2026-07-04 (immediate follow-up to #1285)
- Main already has: `resolve_string` that incorrectly passes ECS-suffixed ARNs to `GetSecretValue`
- Net-new value: Correct handling of ECS-native `:<jsonKey>::` suffix in `resolve_string`, plus 3 unit tests

</details>

chaodu-agent:
LGTM ✅ — Correct fix for a real functional bug: `GetSecretValue` never understood ECS-native `:<jsonKey>::` suffixes.

## What This PR Does

Fixes `resolve_string` passing an ECS-native ARN (with `:<jsonKey>::` suffix) directly to `GetSecretValue`, which rejects it — the suffix is an ECS-only convention, not something the Secrets Manager API understands. This broke webhook registration for the most common `TELEGRAM_BOT_TOKEN` format (Format 1 in the README), introduced by #1285.

## How It Works

Adds `split_ecs_json_key_suffix()` to detect and strip the ECS-only `:<jsonKey>::` tail from ARN-shaped values before calling `GetSecretValue`, then manually extracts the JSON key from the fetched secret string — the same approach already used for `aws-sm://` refs. The two code paths in `resolve_string` are unified into a single fetch-then-optionally-extract flow, eliminating duplication. `resolve_value_from` (where ECS *does* understand the suffix) is correctly left untouched.

## Findings

| 1 | 🟡 | `split_ecs_json_key_suffix` could misparse a malformed ARN ending with `::` — add `:secret:` boundary guard | `secrets.rs:65-80` |
| 2 | 🟡 | Only supports `:<jsonKey>::` form (version-stage/version-id both empty); ECS full 4-field suffix not handled | `secrets.rs:65` |
| 3 | 🟢 | Correct root-cause analysis — aligns with AWS docs: `SecretId` only accepts plain ARN or name | `secrets.rs:59-80` |
| 4 | 🟢 | Clean unification of both resolve paths removes duplication and reduces future bug surface | `secrets.rs:90-109` |
| 5 | 🟢 | Clean segregation of duties: `resolve_value_from` (ECS path) untouched, only `resolve_string` (in-process path) changed | `secrets.rs:35` |
| 6 | 🟢 | No fail-open pattern — parse failure → API error → propagated, never silent pass-through | `secrets.rs:65-80` |
| 7 | 🟢 | Real-world regression test with exact ARN that surfaced the bug | `secrets.rs:115-123` |

<details>

### 🟡 F1: Add `:secret:` boundary guard to `split_ecs_json_key_suffix`

If a malformed ARN ends with `::` but has no JSON key (e.g. `arn:aws:secretsmanager:us-east-1:123:secret:mysecret::`), `strip_suffix("::")` succeeds and `rsplit_once(':')` splits at the colon before the secret name, yielding `base = "arn:...:secret"` and `json_key = Some("mysecret")`. This truncates the actual secret name.

**Suggested fix** (cheap, 1-line):
```rust
if json_key.is_empty() || !base.starts_with("arn:") || !base.contains(":secret:") {
    return (value, None);
}
```

**Risk**: Low. A bare ARN should never end with `::` in practice (ECS always includes a json_key before `::`) and the failure mode is fail-closed (API error, not silent pass-through).

### 🟡 F2: Document limitation — only `:<jsonKey>::` form supported

The ECS `valueFrom` full format is:
```
arn:...:secret:name:json-key:version-stage:version-id
```

Current implementation only handles the case where version-stage and version-id are both empty (trailing `::`). Formats like `:<key>:AWSCURRENT:` or `:<key>::<version-id>` will not be split (because `strip_suffix("::")` won't match), causing `GetSecretValue` to receive the full suffixed value and fail.

**Mitigation**: This is fail-closed (errors out, doesn't silently pass wrong data). Suggest adding a doc comment explicitly noting this limitation and tracking full 4-field support as a follow-up.

</details>

<details>

- **Spec-correct**: The fix matches AWS Secrets Manager `GetSecretValue` documentation exactly — `SecretId` only accepts a plain ARN or secret name, never the ECS `valueFrom` suffix convention.
- **Minimal blast radius**: Only `resolve_string` is changed; `resolve_value_from` (the ECS path that *should* keep the suffix) is untouched.
- **Code simplification**: Unifying both branches into one fetch + optional JSON extraction reduces duplication and makes the logic easier to reason about.
- **No fail-open**: All parse-failure paths result in API errors that propagate, never silently passing through incorrect data.
- **Solid test coverage**: Three unit tests covering the key variants (ARN with suffix, plain ARN, bare name), including the exact real-world value that exposed the bug.
- **Live E2E verified**: PR description confirms the fix was validated against the actual failing manifest end-to-end.

</details>

<details>

- PR opened: 2026-07-04 (immediate follow-up to #1285)
- Main already has: `resolve_string` that incorrectly passes ECS-suffixed ARNs to `GetSecretValue`
- Net-new value: Correct handling of ECS-native `:<jsonKey>::` suffix in `resolve_string`, plus 3 unit tests
- CI: `operator` build/test ✅, `changes` ✅, `poll-and-review` ✅

</details>

---

**Verdict**: Both 🟡 findings are low-risk edge cases (fail-closed, not fail-open) that don't affect the core bug fix. The `:secret:` guard (F1) is a 1-line hardening; the version-field limitation (F2) is a known boundary best tracked as a follow-up issue. Core fix is correct and well-tested — **LGTM for merge**.

chaodu-agent:
LGTM ✅ — Real bug fix with solid, well-tested ECS suffix parsing that handles all documented ARN shapes correctly.

## What This PR Does

Fixes `resolve_string` which incorrectly passed ECS-native `:<jsonKey>::` suffixed ARNs directly to `GetSecretValue` — an API that has no knowledge of that suffix. This broke webhook registration for manifests using the more common ECS-native `valueFrom` format for `TELEGRAM_BOT_TOKEN`.

## How It Works

Adds `split_ecs_json_key_suffix` which:
1. Locates the `:secret:` marker in the ARN to unambiguously identify the secret-name boundary
2. Parses all three ECS positional fields (`json-key`, `version-stage`, `version-id`)
3. Returns the base ARN + optional JSON key for manual extraction
4. Fails closed with a clear error for unsupported version-stage/version-id pinning

`resolve_string` is refactored to a unified flow: both `aws-sm://` and ECS-native formats now go through the same fetch → optional JSON-key extraction path, eliminating the duplicated branches.

## Findings

| 1 | 🟢 | Correct use of `:secret:` marker for unambiguous parsing — avoids the original misparsing trap where `mysecret::` (empty suffix fields) was confused with a json-key | `secrets.rs:76-82` |
| 2 | 🟢 | Fail-closed design for unsupported version pinning — prevents silent misresolution and gives actionable error message | `secrets.rs:92-97` |
| 3 | 🟢 | Excellent test coverage — 7 tests covering the exact real-world ARN, plain ARN, bare name, empty-suffix misparsing edge case, and all three reject scenarios from AWS docs | `secrets.rs:119-181` |
| 4 | 🟢 | Unified resolve_string flow eliminates code duplication between aws-sm:// and ECS-native paths | `secrets.rs:107-117` |

<details>

- PR opened: follow-up fix to #1285 (merged)
- Main already has: `resolve_string` that handles `aws-sm://` and plain ARN, but incorrectly passes ECS-suffixed ARNs verbatim to `GetSecretValue`
- Net-new value: correct ECS suffix stripping + JSON-key extraction, full three-field parsing with fail-closed boundary, 7 regression tests

</details>

<details>

- The fix correctly identifies that the `:<jsonKey>::` suffix is an ECS-only convention, not a Secrets Manager API feature — a subtle but important distinction that was wrong in the original doc comment
- Parsing anchors on `:secret:` rather than naive rsplit, which prevents the `mysecret::` misparsing edge case
- The version-stage/version-id rejection is an intentional, documented limitation with a clear error path — not a missing feature
- The PR description is exceptionally thorough, documenting the before/after behavior, the root cause, and both review findings with their resolutions
- Live E2E verification against the exact manifest that surfaced the bug

</details>

thepagent (approved):

PR #1293, opened 2026-07-04
fix(unified): start feishu websocket + bot identity in unified mode

angmeng:
## Problem

In unified mode, the feishu adapter is effectively dead in its default configuration. `src/main.rs` only registers the webhook route:

```rust
#[cfg(feature = "feishu")]
if gw_state.feishu.is_some() {
    ...
    app = app.route(&path, axum::routing::post(openab_gateway::adapters::feishu::webhook));
}
```

`resolve_bot_identity()` and `start_websocket()` are only called from the standalone gateway path (`openab-gateway/src/lib.rs::run_gateway`). Since `FEISHU_CONNECTION_MODE` defaults to `websocket` (and the docs recommend it), setting `FEISHU_APP_ID`/`FEISHU_APP_SECRET` in unified mode logs `unified: feishu adapter enabled` but the bot never connects and never receives a single event. `@mention` gating is also broken even in webhook mode, because the bot's own open_id is never resolved.

## Fix

Mirror the gateway startup into the unified block, right after the webhook route registration:

- `resolve_bot_identity().await` — required for @mention gating and self-message filtering
- spawn `run_idle_reaper` when card streaming is enabled (parity with `run_gateway`)
- spawn `start_websocket()` when `connection_mode == Websocket`

The websocket shutdown sender is intentionally leaked (`std::mem::forget`) — dropping it makes `shutdown_rx.changed()` fire inside `start_websocket`'s reconnect loop and permanently stops the connection. In unified mode the adapter should live for the process lifetime, same as the discord/slack adapters.

## Verification

Running live on macOS against Lark (international, `FEISHU_DOMAIN=lark`) with only env vars set (no `[gateway]` config):

```
unified: feishu adapter enabled path=/webhook/feishu
feishu bot identity resolved bot_open_id=ou_xxx
unified: feishu card-streaming idle reaper started idle_ms=3000
unified: feishu websocket task spawned
feishu websocket connected
```

DMs and group @mentions both round-trip to the agent and reply. `cargo check --features unified` passes on top of current `main`.

https://claude.ai/code/session_01C1wSJXGjoZWjarLiDqQFtC

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -633,6 +633,39 @@ async fn main() -> anyhow::Result<()> {
                 app = app.route(&path, axum::routing::post(openab_gateway::adapters::feishu::webhook));
             }
 
+            // Feishu bot identity + WebSocket long-connection (unified mode).
+            // Mirrors run_gateway() in openab-gateway/src/lib.rs: identity is
+            // required for @mention gating, and the default connection mode is
+            // websocket — without this spawn the adapter only serves the webhook
+            // route above and never receives events in websocket mode.
+            #[cfg(feature = "feishu")]
+            if let Some(ref f) = gw_state.feishu {
+                use openab_gateway::adapters::feishu;
+                f.resolve_bot_identity().await;
+                if f.config.streaming_mode != feishu::StreamingMode::Post {
+                    let idle_ms = f.config.card_idle_finalize_ms;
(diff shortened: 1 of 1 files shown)


angmeng:
Withdrawing — will keep this fix in the fork.

PR #1281, opened 2026-07-03
fix(slack): bound reconnect-path network calls with a 30s timeout

howie:
## What problem does this solve?

`get_socket_mode_url` and `tokio_tungstenite::connect_async` in the Slack Socket Mode reconnect loop (`crates/openab-core/src/slack.rs`) had no timeout, unlike the post-connect idle watchdog (`IDLE_TIMEOUT_SECS`) and `SlackAdapter::new`'s Web API client (`.timeout(Duration::from_secs(30))`). If the same flaky network path that triggers a disconnect also stalls (not resets) a subsequent `apps.connections.open` POST or the WS handshake during reconnect, the reconnect task parks in `.await` forever - no error, no log, process alive but the bot permanently dark on Slack until manually restarted. This is exactly what happened to openab-viola in production: a normal reconnect cycle ran for ~6 minutes, then total silence for 11+ hours.

Closes #1279

Discord Discussion URL: https://discord.com/channels/1491295327620169908/1491969620754567270/1522434117642223616

## At a Glance

```
run_slack_adapter reconnect loop:
  get_socket_mode_url(app_token)  --[NO TIMEOUT]--> could hang forever
        |
  connect_async(ws_url)           --[NO TIMEOUT]--> could hang forever
        |
  (only once connected: IDLE_TIMEOUT_SECS watchdog protects the read loop)

After this fix, both calls above are wrapped in tokio::time::timeout(30s),
matching the existing pattern.
```

## Prior Art & Industry Research

Not applicable - trivial bug fix. This mirrors an existing, already-established pattern in the same file (`SlackAdapter::new`'s bounded `reqwest::Client` and the `IDLE_TIMEOUT_SECS` idle watchdog), rather than introducing new architecture.

## Proposed Solution

1. Wrap `get_socket_mode_url(...)` in `tokio::time::timeout(Duration::from_secs(30), ...)`. On timeout, log a warning and fall through to the existing backoff/retry path.
2. Wrap `tokio_tungstenite::connect_async(&ws_url)` in the same 30s timeout, with the same fallback-to-retry behavior.
3. Change `get_socket_mode_url` to accept `&reqwest::Client` instead of constructing `reqwest::Client::new()` per call, and pass in `SlackAdapter`'s existing client (already configured with a 30s timeout). This folds in #386 (client-reuse nit) as a side effect of threading a client through anyway.

`RECONNECT_TIMEOUT_SECS = 30` matches the existing Web API client timeout for consistency.

## Why this approach?

[...]

Issue #386, opened 2026-04-16
Slack: get_socket_mode_url creates new reqwest::Client per reconnect


thepagent:
Parent: #382

## Problem

`get_socket_mode_url()` creates a new `reqwest::Client` on every call. Since Socket Mode reconnects are rare (server-initiated disconnect, network blip), this is not a real performance issue, but it would be cleaner to reuse `SlackAdapter::client`.

## Suggested fix

Pass `&self.client` or `&reqwest::Client` to `get_socket_mode_url()` instead of constructing a new one. Requires minor refactor since the function is called before `SlackAdapter` is constructed in the current flow.

Severity: nit — no functional impact.

thepagent:
Closing — low-severity Slack nit, not blocking 0.7.7 GA.

PR #1334, opened 2026-07-08
fix(slack): bound reconnect-path and socket-write network calls with timeouts

### LINKED ISSUE #1279: bug(slack): get_socket_mode_url / connect_async in the reconnect loop have no timeout, unlike the post-connect idle watchdog
## Description

openab-viola (our HR-onboarding bot, chart image ghcr.io/openabdev/openab:0.9.0-beta.5-claude, agent backend claude-agent-acp 2.1.179) went silently unresponsive on Slack for over 11 hours. The pod stayed Running, the process was alive (confirmed via `ps aux` inside the container), but zero log lines were emitted after a burst of normal reconnect activity. No crash, no panic, no restart.

## What the logs showed before going silent

A normal-looking reconnect cycle repeated every ~30s for about 6 minutes:

```
ERROR openab_core::slack: Socket Mode read error: IO error: peer closed connection without sending TLS close_notify
WARN  openab_core::slack: reconnecting to Slack Socket Mode backoff=1
INFO  openab_core::slack: connecting to Slack Socket Mode url=wss://wss-primary.slack.com/link/?ticket=...
INFO  openab_core::slack: Slack Socket Mode connected
```

Each cycle completed successfully (TCP/TLS/WS handshake all fine, backoff correctly reset to 1 each time per `next_backoff`/`MAX_BACKOFF_SECS` in `crates/openab-core/src/slack.rs:701-705`). Then, after one more disconnect, there was total silence -- no further "connecting", no further error, nothing -- for 11+ hours until we manually `kubectl rollout restart`ed the pod.

## Root cause candidate

`run_slack_adapter` (`crates/openab-core/src/slack.rs:718`) calls, on every reconnect:

1. `get_socket_mode_url` (`slack.rs:1165-1182`), which builds its own `reqwest::Client::new()` (`slack.rs:1166`) with **no `.timeout()`** and does `client.post(...).send().await?` / `.json().await?`.
2. `tokio_tungstenite::connect_async(&ws_url)` (`slack.rs:761`), also with **no timeout wrapper**.

Compare with `SlackAdapter::new` (`slack.rs:105-111`), which explicitly sets `.timeout(Duration::from_secs(30))` on its `reqwest::Client` -- the comment there says this is specifically to bound Web API calls so an unbounded call can't stall the idle-timeout watchdog. That watchdog (added in #1120 / commit 789c6f8, closing #1050 and #1101) protects the *already-connected* read loop via `IDLE_TIMEOUT_SECS`, but it only runs once a WebSocket is established. `get_socket_mode_url` and `connect_async` run in the *reconnect* path, before that protection exists.

If the same flaky network path that produced the "peer closed connection without sending TLS close_notify" errors also causes a subsequent `apps.connections.open` POST or the WS handshake itself to stall (SYN/TLS sent, no response, no RST -- plausible on infra that's already dropping TLS close_notify), the reconnect task parks in `.await` forever: no error, no log, near-zero CPU, process otherwise alive. This matches every observed symptom exactly.

Related prior art in this repo: #386 flagged that `get_socket_mode_url` allocates a fresh `reqwest::Client` per call, but that issue was scoped purely to client-reuse/perf ("nit -- no functional impact") and did not address the missing timeout.

## Suggested fix

1. Add `.timeout(Duration::from_secs(30))` to the `reqwest::Client` built in `get_socket_mode_url` (`slack.rs:1166`), matching `SlackAdapter::new`'s existing pattern -- and ideally reuse `SlackAdapter::client` instead of constructing a new one (folding in #386).
2. Wrap `tokio_tungstenite::connect_async(&ws_url)` (`slack.rs:761`) in `tokio::time::timeout(Duration::from_secs(30), ...)` so a stalled handshake surfaces as a reconnect-retry instead of parking forever.

## Secondary, unrelated gap noticed while investigating

No liveness/readiness probe is defined anywhere in `charts/openab`'s Deployment template (confirmed by rendering the full chart -- zero `livenessProbe`/`readinessProbe` occurrences), unlike `charts/openab-line`, which does define an `httpGet /health` probe. So even with the timeout fix above, a *different* future stall in this adapter would still show as a healthy pod with no way for Kubernetes to detect and recover it. Worth tracking separately, but flagging here since it's exactly what let this particular incident run for 11+ hours undetected.

### Steps to Reproduce

1. Run the Slack adapter behind a flaky network path that occasionally silently drops TLS connections without a close_notify (as seen on our infra).
2. Get unlucky: have that same flaky path stall (not reset) either the `apps.connections.open` POST inside `get_socket_mode_url`, or the WS `connect_async` handshake, during a reconnect attempt.
3. The reconnect task blocks forever in `.await`. No further logs. Process stays alive. Bot is permanently dark on Slack until manually restarted.

### Expected Behavior

`get_socket_mode_url` and `connect_async` in the reconnect path should be bounded by a timeout (e.g. 30s), so a stalled reconnect attempt surfaces as a retry (with a log line) instead of parking the task indefinitely with zero observability.

howie:
Discord Discussion URL: https://discord.com/channels/1491295327620169908/1491969620754567270/1524555735294410812

Continuation of #1281, which was auto-closed by the stale-contributor bot for inactivity before its final round of review feedback could be addressed. All prior review rounds (Claude + Codex mob review, `@dogzzdogzz`'s consolidated Claude+Codex review) had converged to LGTM on the core fix; this PR rebases those commits onto current `main` and additionally addresses the one unresolved finding from the last review round.

Original problem (from #1279): `get_socket_mode_url` and `tokio_tungstenite::connect_async` in the Slack Socket Mode reconnect loop (`crates/openab-core/src/slack.rs`) had no timeout, unlike the post-connect idle watchdog (`IDLE_TIMEOUT_SECS`) and `SlackAdapter::new`'s Web API client. A stalled network path during reconnect parked the reconnect task in `.await` forever — no error, no log, process alive but the bot permanently dark on Slack until manually restarted. This caused an 11+ hour production outage on openab-viola.

Closes #1279

## What's carried over from #1281 (unchanged, already LGTM'd)

1. Wrap `get_socket_mode_url(...)` in `tokio::time::timeout(35s)` (`RECONNECT_HTTP_TIMEOUT_SECS`), set strictly above the adapter client's own 30s timeout so the diagnosable inner error fires first.
2. Wrap `connect_async` in `tokio::time::timeout(30s)` (`RECONNECT_TIMEOUT_SECS`) via `connect_with_reconnect_timeout()`.
3. Reuse `SlackAdapter`'s bounded `reqwest::Client` in `get_socket_mode_url` instead of constructing an unbounded one per call (folds in #386).
4. Extract `wait_backoff_or_shutdown()` to deduplicate the backoff+shutdown `select!` block that was copy-pasted at three retry points.
5. Regression test `connect_async_timeout_fires_on_stalled_handshake` (`start_paused = true` against a local `TcpListener` that never completes the handshake).

## New in this PR (addresses #1281's last review round)

`@dogzzdogzz`'s consolidated review on #1281 flagged that the reconnect **handshake** was bounded, but the live socket loop still had unbounded writes — the same bug class as #1279, just on the write side:

- `slack.rs`: envelope ack send, pong send, keepalive ping send, shutdown close send were all unbounded `write.send(...).await`. If any of these stalls on the same flaky/half-open socket that motivated #1279, the task parks inside that `select!` branch — and since it can never return to `ping_interval.tick()`, `IDLE_TIMEOUT_SECS` can't fire either.

Fix: extracted `send_with_timeout()` (bounded by a new `WRITE_TIMEOUT_SECS = 10`) and used it at all four call sites; each now `break`s to force a reconnect on timeout, matching the existing keepalive-ping error-handling pattern.

**Intentional behaviour change — ack timeout now skips routing:** on an envelope ack-send timeout (or send error) the loop `break`s *before* the event is routed/dispatched, whereas previously the ack result was ignored and the event was routed regardless. This is deliberate and safer: Slack redelivers an un-acked envelope on the next connection, so skipping dispatch avoids double-processing an envelope Slack still considers pending. Called out explicitly so it isn't later mistaken for an oversight.

Also addressed from the same review round:

- Added a regression test for the previously-untested HTTP timeout leg (`get_socket_mode_url_with_timeout`), using a stalled local listener with `reqwest::Client::builder().resolve(...)` to redirect the `slack.com` host, mirroring the existing WS handshake test.
- Added `ConnectResult` / `SocketModeUrlResult` type aliases for the nested `Result` return types (readability nit).
- Aligned the ack and pong send sites to the keepalive-ping site's 3-arm match (`Ok(Err(e))` / `Err(_)` / `Ok(Ok(()))`) so a completed-but-erroring send (a real `tungstenite::Error`, e.g. broken pipe) forces a reconnect instead of being silently swallowed by an outer-`.is_err()`-only check.

## Validation

- [x] `cargo build -p openab-core` passes
- [x] `cargo test -p openab-core --lib slack` — 66/66 tests pass (including the 2 new regression tests)
- [x] `cargo clippy -p openab-core --all-targets` — no new warnings (4 pre-existing warnings in `cron.rs`/`discord.rs` are unrelated)
- [x] `cargo fmt -p openab-core -- --check` — no new formatting diffs introduced by this change (pre-existing unformatted regions elsewhere in the file are untouched)

## Supersedes

Closes/supersedes #1281 (auto-closed for inactivity, all review rounds there had converged to LGTM on the core fix).

### DIFF
--- a/crates/openab-core/Cargo.toml
+++ b/crates/openab-core/Cargo.toml
@@ -51,6 +51,9 @@ http = { version = "1", optional = true }
 [target.'cfg(unix)'.dependencies]
 libc = "0.2"
 
+[dev-dependencies]
+tokio = { version = "1", features = ["test-util"] }
+
 [features]
 default = ["discord", "slack", "secrets-aws", "agentcore", "config-s3"]
 discord = ["dep:serenity"]
--- a/crates/openab-core/src/slack.rs
+++ b/crates/openab-core/src/slack.rs
@@ -711,6 +711,40 @@ const RECONNECT_TIMEOUT_SECS: u64 = 30;
 /// backstop for the rare case the client builder failed and fell back to an
 /// unbounded `reqwest::Client::new()` (see #1279).
 const RECONNECT_HTTP_TIMEOUT_SECS: u64 = 35;
+/// Bounds outbound `write.send(...)` calls in the live socket loop. The
+/// inbound half-open case is guarded by `IDLE_TIMEOUT_SECS`, but a stalled
+/// *outbound* send parks the loop inside that `select!` branch forever too —
+/// and since it can never reach `ping_interval.tick()`, the idle watchdog
+/// itself becomes unreachable. Same bug class as #1279, on the write side.
+const WRITE_TIMEOUT_SECS: u64 = 10;
+
+/// The Socket Mode WebSocket write half, as produced by splitting the stream
+/// returned from `connect_with_reconnect_timeout`. Named so the nested
+/// generic doesn't have to be repeated at every `send_with_timeout` call
+/// site.
+type SlackWsSink = futures_util::stream::SplitSink<
+    tokio_tungstenite::WebSocketStream<tokio_tungstenite::MaybeTlsStream<tokio::net::TcpStream>>,
+    tungstenite::Message,
+>;
+
+/// Result of `connect_with_reconnect_timeout`: the outer `Result` is the
+/// timeout, the inner `Result` is `connect_async`'s own outcome.
+type ConnectResult = std::result::Result<
+    std::result::Result<
+        (
+            tokio_tungstenite::WebSocketStream<
+                tokio_tungstenite::MaybeTlsStream<tokio::net::TcpStream>,
+            >,
+            tungstenite::handshake::client::Response,
+        ),
+        tungstenite::Error,
+    >,
+    tokio::time::error::Elapsed,
+>;
+
+/// Result of `get_socket_mode_url_with_timeout`: `Err` is the outer timeout;
+/// `Ok` carries `get_socket_mode_url`'s own `anyhow::Result`.
+type SocketModeUrlResult = std::result::Result<Result<String>, tokio::time::error::Elapsed>;
 
 /// Next reconnect delay: double, capped. Reset to 1 on a successful connect.
 fn next_backoff(cur: u64) -> u64 {
@@ -720,21 +754,25 @@ fn next_backoff(cur: u64) -> u64 {
 /// Wraps `connect_async` in the reconnect-path timeout. Extracted so the
 /// regression test exercises the exact same code path `run_slack_adapter`
 /// uses, instead of duplicating the timeout expression (#1279).
-async fn connect_with_reconnect_timeout(
-    ws_url: &str,
-) -> std::result::Result<
-    std::result::Result<
-        (
-            tokio_tungstenite::WebSocketStream<tokio_tungstenite::MaybeTlsStream<tokio::net::TcpStream>>,
-            tungstenite::handshake::client::Response,
-        ),
-        tungstenite::Error,
-    >,
-    tokio::time::error::Elapsed,
-> {
+async fn connect_with_reconnect_timeout(ws_url: &str) -> ConnectResult {
     tokio::time::timeout(
         std::time::Duration::from_secs(RECONNECT_TIMEOUT_SECS),
         tokio_tungstenite::connect_async(ws_url),
+    )
+    .await
+}
+
+/// Send `msg` on `write`, bounded by `WRITE_TIMEOUT_SECS`. Mirrors
+/// `connect_with_reconnect_timeout`: extracted so every write in the live
+/// socket loop shares one bounded-send implementation instead of each call
+/// site duplicating (or omitting) the timeout.
+async fn send_with_timeout(
+    write: &mut SlackWsSink,
+    msg: tungstenite::Message,
+) -> std::result::Result<std::result::Result<(), tungstenite::Error>, tokio::time::error::Elapsed> {
+    tokio::time::timeout(
+        std::time::Duration::from_secs(WRITE_TIMEOUT_SECS),
+        write.send(msg),
     )
     .await
 }
@@ -792,12 +830,7 @@ pub async fn run_slack_adapter(
             return Ok(());
         }
 
-        let ws_url = match tokio::time::timeout(
-            std::time::Duration::from_secs(RECONNECT_HTTP_TIMEOUT_SECS),
-            get_socket_mode_url(&adapter.client, &app_token),
-        )
-        .await
-        {
+        let ws_url = match get_socket_mode_url_with_timeout(&adapter.client, &app_token).await {
             Ok(Ok(url)) => url,
             Ok(Err(e)) => {
                 error!(err = %e, backoff = backoff_secs, "failed to get Socket Mode URL, retrying");
@@ -848,9 +881,19 @@ pub async fn run_slack_adapter(
                                     // Acknowledge the envelope immediately
                                     if let Some(envelope_id) = envelope["envelope_id"].as_str() {
                                         let ack = serde_json::json!({"envelope_id": envelope_id});
-                                        let _ = write
-                                            .send(tungstenite::Message::Text(ack.to_string()))
-                                            .await;
+                                        if send_with_timeout(
+                                            &mut write,
+                                            tungstenite::Message::Text(ack.to_string()),
+                                        )
+                                        .await
+                                        .is_err()
+                                        {
+                                            warn!(
+                                                timeout = WRITE_TIMEOUT_SECS,
+                                                "Slack Socket Mode envelope ack timed out, reconnecting"
+                                            );
+                                            break;
+                                        }
                                     }
 
                                     // Slash commands and interactive block_actions aren't
@@ -1174,7 +1217,16 @@ pub async fn run_slack_adapter(
                                     }
                                 }
                                 Ok(tungstenite::Message::Ping(data)) => {
-                                    let _ = write.send(tungstenite::Message::Pong(data)).await;
+                                    if send_with_timeout(&mut write, tungstenite::Message::Pong(data))
+                                        .await
+                                        .is_err()
+                                    {
+                                        warn!(
+                                            timeout = WRITE_TIMEOUT_SECS,
+                                            "Slack Socket Mode pong send timed out, reconnecting"
+                                        );
+                                        break;
+                                    }
                                 }
                                 Ok(tungstenite::Message::Close(_)) => {
                                     warn!("Slack Socket Mode connection closed by server");
@@ -1198,14 +1250,24 @@ pub async fn run_slack_adapter(
                                 );
                                 break;
                             }
-                            if let Err(e) = write.send(tungstenite::Message::Ping(Vec::new())).await {
-                                warn!(error = %e, "Slack Socket Mode ping failed, reconnecting");
-                                break;
+                            match send_with_timeout(&mut write, tungstenite::Message::Ping(Vec::new())).await {
+                                Ok(Err(e)) => {
+                                    warn!(error = %e, "Slack Socket Mode ping failed, reconnecting");
+                                    break;
+                                }
+                                Err(_) => {
+                                    warn!(
+                                        timeout = WRITE_TIMEOUT_SECS,
+                                        "Slack Socket Mode ping timed out, reconnecting"
+                                    );
+                                    break;
+                                }
+                                Ok(Ok(())) => {}
                             }
                         }
                         _ = shutdown_rx.changed() => {
                             info!("Slack adapter received shutdown signal");
-                            let _ = write.send(tungstenite::Message::Close(None)).await;
+                            let _ = send_with_timeout(&mut write, tungstenite::Message::Close(None)).await;
                             return Ok(());
                         }
                     }
@@ -1253,6 +1315,21 @@ async fn get_socket_mode_url(client: &reqwest::Client, app_token: &str) -> Resul
         .as_str()
         .map(|s| s.to_string())
         .ok_or_else(|| anyhow!("no url in apps.connections.open response"))
+}
+
+/// Wraps `get_socket_mode_url` in the reconnect-path HTTP timeout. Extracted
+/// (mirroring `connect_with_reconnect_timeout`) so the regression test calls
+/// the same production code path `run_slack_adapter` uses, instead of
+/// duplicating the timeout expression.
+async fn get_socket_mode_url_with_timeout(
+    client: &reqwest::Client,
+    app_token: &str,
+) -> SocketModeUrlResult {
+    tokio::time::timeout(
+        std::time::Duration::from_secs(RECONNECT_HTTP_TIMEOUT_SECS),
+        get_socket_mode_url(client, app_token),
+    )
+    .await
 }
 
 #[allow(clippy::too_many_arguments)]
@@ -2388,8 +2465,8 @@ mod tests {
 #[cfg(test)]
 mod socket_keepalive_tests {
     use super::{
-        connect_with_reconnect_timeout, next_backoff, socket_idle, wait_backoff_or_shutdown,
-        IDLE_TIMEOUT_SECS, MAX_BACKOFF_SECS,
+        connect_with_reconnect_timeout, get_socket_mode_url_with_timeout, next_backoff,
+        socket_idle, wait_backoff_or_shutdown, IDLE_TIMEOUT_SECS, MAX_BACKOFF_SECS,
     };
     use std::time::Duration;
 
@@ -2447,6 +2524,40 @@ mod socket_keepalive_tests {
         assert!(
             result.is_err(),
             "connect_async on a stalled handshake must time out, not hang"
+        );
+    }
+
+    /// Regression test for #1279 (the HTTP leg): a stalled `apps.connections.open`
+    /// call during reconnect must surface as a timeout instead of parking the
+    /// reconnect task forever. Overrides DNS resolution for `slack.com` to a
+    /// local listener that accepts the TCP connection and never responds, so
+    /// the client's TLS handshake stalls exactly like a silently-dropped peer
+    /// would. Calls `get_socket_mode_url_with_timeout` directly (the same
+    /// helper `run_slack_adapter` calls) so this pins the actual production
+    /// code path, mirroring `connect_async_timeout_fires_on_stalled_handshake`
+    /// above.
+    #[tokio::test(start_paused = true)]
+    async fn get_socket_mode_url_timeout_fires_on_stalled_http_call() {
+        let listener = tokio::net::TcpListener::bind("127.0.0.1:0")
+            .await
+            .expect("bind local listener");
+        let addr = listener.local_addr().expect("local addr");
+
+        tokio::spawn(async move {
+            let _socket = listener.accept().await.expect("accept");
+            std::future::pending::<()>().await;
+        });
+
+        let client = reqwest::Client::builder()
+            .resolve("slack.com", addr)
+            .build()
+            .expect("build client");
+
+        let result = get_socket_mode_url_with_timeout(&client, "xapp-test-token").await;
+
+        assert!(
+            result.is_err(),
+            "get_socket_mode_url on a stalled HTTP call must time out, not hang"
         );
     }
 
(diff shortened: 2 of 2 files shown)


chaodu-agent:
> [!NOTE]
> LGTM ✅ — Well-structured fix that comprehensively bounds all unbounded network calls in the Slack Socket Mode reconnect and live-socket loops.

## What This PR Does

Fixes a production outage class (#1279) where stalled network paths parked the Slack Socket Mode reconnect task or live-socket write operations in `.await` forever with no logs, no errors, and no recovery. This PR wraps all five unbounded network call sites in explicit `tokio::time::timeout` guards and extracts shared helpers to eliminate duplication.

## How It Works

Three timeout constants form a layered defense:
- `WRITE_TIMEOUT_SECS = 10` — bounds all outbound `write.send(...)` in the live socket loop (ack, pong, keepalive ping, close)
- `RECONNECT_TIMEOUT_SECS = 30` — bounds `connect_async` (WS handshake) during reconnect
- `RECONNECT_HTTP_TIMEOUT_SECS = 35` — bounds `get_socket_mode_url` (HTTP API call), set above the client's own 30s timeout so the diagnosable inner error fires first

Shared helpers extracted:
- `connect_with_reconnect_timeout()` — timeout-wrapped WS connect
- `send_with_timeout()` — timeout-wrapped write, used at all four send sites
- `wait_backoff_or_shutdown()` — deduplicates the `select!{ sleep / shutdown }` pattern at three retry points
- `get_socket_mode_url_with_timeout()` — timeout-wrapped HTTP call

Additionally reuses `SlackAdapter`'s bounded `reqwest::Client` in `get_socket_mode_url` instead of constructing an unbounded one per call (folds in #386).

## Findings

| 1 | 🟢 | Timeout hierarchy is well-designed: write(10s) < WS handshake(30s) < HTTP backstop(35s) | `slack.rs:710-720` |
| 2 | 🟢 | Excellent helper extraction — each production code path is tested directly via the same function | `slack.rs:730-795` |
| 3 | 🟢 | Regression tests use `start_paused = true` with stalled local listeners — fast, deterministic, no flaky wall-clock waits | `slack.rs` tests |
| 4 | 🟢 | Client reuse eliminates unbounded fallback path | `slack.rs:get_socket_mode_url` |
| 5 | 🟢 | Type aliases (`ConnectResult`, `SocketModeUrlResult`, `SlackWsSink`) tame deeply nested generics | `slack.rs:725-755` |

<details>

- **Comprehensive coverage**: all five unbounded call sites are now bounded — reconnect HTTP, reconnect WS, envelope ack, pong, and keepalive ping. No half-measures.
- **Correct error granularity**: the keepalive ping gets full 3-arm match (timeout/inner-error/ok) since it's the liveness probe; ack/pong preserve the original fire-and-forget semantics with timeout as the only new break condition — matching the pre-existing design intent.
- **Testability-first design**: extracting `connect_with_reconnect_timeout` and `get_socket_mode_url_with_timeout` means tests exercise the *exact* production code path, not a copy of the timeout expression.
- **`wait_backoff_or_shutdown` is a great refactor** — DRYs three identical `select!` blocks into one function with clear `Option<u64>` semantics (None = shutdown, Some = next backoff).
- **Clean CI**: all smoke tests and checks pass.

</details>

<details>

- PR opened: 2026-07-08 (continuation of #1281, auto-closed for inactivity)
- Main already has: `IDLE_TIMEOUT_SECS` (inbound watchdog), `PING_INTERVAL_SECS`, `MAX_BACKOFF_SECS`, `next_backoff()`, `socket_idle()`
- Main does NOT have: any timeout on `connect_async`, `get_socket_mode_url`, or outbound writes
- Net-new value: 3 timeout constants, 4 helper functions, timeout wrapping at all 5 unbounded call sites, client reuse, 4 new regression tests

</details>

1. **The idle timeout already catches dead sockets** — `IDLE_TIMEOUT_SECS = 75` eventually fires when no inbound frames arrive, triggering reconnect. However: this doesn't help when the *outbound* write is the one parked (it blocks the select loop, making the idle check unreachable), and it adds 75s of unnecessary downtime on the reconnect path.

2. **Slack's infrastructure is generally reliable** — the 11-hour outage was a rare event. However: "rare" × "11 hours unattended" = unacceptable for a production bot; defensive timeouts are cheap insurance with zero happy-path cost.

3. **The reqwest client already has a 30s timeout** — so `get_socket_mode_url` is arguably already bounded. However: the PR description correctly notes this only holds if the client was constructed with that timeout; if the builder ever fails or changes, the explicit outer timeout is the backstop.

None of these reasons are compelling enough to reject — they're all addressed by the implementation.

dogzzdogzz (approved):
### 🟡 Non-blocking suggestions

**1 · Ack/pong send sites swallow non-timeout send errors** _(both reviewers)_
`send_with_timeout` returns `Result<Result<(), tungstenite::Error>, Elapsed>`. The **ack** (`slack.rs:884`) and **pong** (`slack.rs:1220`) sites check only `.is_err()` on the *outer* result, so a completed-but-erroring send (a real `tungstenite::Error`, e.g. broken pipe) is silently ignored and the loop proceeds as if it succeeded. The **keepalive-ping** site (`slack.rs:1253`) does it right — it 3-way matches `Ok(Err(e))` / `Err(_)` / `Ok(Ok(()))`. Not a regression (matches the old `let _ = write.send()`), but since this PR is specifically about tightening these sends, consider aligning ack/pong to the ping site's match pattern (or a one-line "best-effort; read/ping side catches a dead socket" comment).

**2 · Undocumented behaviour change on ack timeout** _(infrabot)_
On an ack-send timeout (`slack.rs:882-897`) the code now `break`s *before* event routing/dispatch — everything from `slack.rs:899` on is skipped. Previously the ack result was ignored and the event was routed regardless. This is likely the **safer** behaviour (Slack redelivers an un-acked envelope on reconnect, avoiding a double-process), but it's an intentional change that isn't called out in the PR description — worth a line so it isn't later mistaken for an oversight.

### 🟢 Info / follow-ups (out of scope for this PR)
- **Write-side timeout is untested** _(both)_ — `send_with_timeout` (4 sites) is the PR's primary new behaviour; the 2 new regression tests cover only the reconnect legs. Low risk (structurally symmetric to the tested `connect_with_reconnect_timeout`); deterministically forcing TCP write-backpressure is hard. A stalled-sink test would close the gap. Fine to merge as-is.
- **Pre-existing WSS-URL logging** _(infra-Patrick)_ — the unchanged `info!(url = %ws_url)` line logs the full Socket Mode URL, which embeds a short-lived auth ticket. Out of scope here; follow-up only.
- **Rollout watch** _(infrabot)_ — first production exercise of these bounds after the 11h outage; worth a glance at reconnect-frequency metrics post-deploy, though 30/35/10s are generous vs. `PING_INTERVAL_SECS=30` / `IDLE_TIMEOUT_SECS=75`.

### Correctness questions — confirmed clean
1. **35s > 30s ordering / race?** No race. `SlackAdapter::new` sets `.timeout(30s)` covering the whole request cycle incl. `resp.json().await`; same tokio clock with a 5s margin → the inner reqwest error fires first. 35s is correctly scoped as a backstop for the fallback path where `.build()` fails to an unbounded `Client::new()`. The new HTTP test targets exactly that bare-client scenario.
2. **`break` on write timeout / in-flight state?** Safe — falls through to the same reconnect/backoff path as the pre-existing ping-failure break. Only real change is the ack-timeout-skips-routing behaviour noted above.
3. **10s shutdown-close truncating handshake?** No — `Message::Close(None)` is a local buffer write+flush; tungstenite doesn't wait for the peer's close-ack here, before or after. Worst-case shutdown latency goes unbounded → 10s, strictly better.

Also verified: all 4 previously-unbounded write sites now route through `send_with_timeout` (no unbounded write remains in the live loop); `Cargo.toml`'s new `[dev-dependencies] tokio test-util` is necessary (main dep uses `full`, which excludes `test-util`); `wait_backoff_or_shutdown` is a faithful, behaviour-preserving extraction across all 3 call sites.

chaodu-agent:
> [!NOTE]
> LGTM ✅ — Solid, well-tested fix for a real production outage. Addresses all prior review feedback from #1281.

## What This PR Does

Bounds all previously-unbounded network calls in the Slack Socket Mode reconnect loop and live socket write path with explicit timeouts, preventing stalled connections from permanently parking the adapter task (the root cause of an 11+ hour production outage on openab-viola, #1279).

## How It Works

1. **Reconnect path**: `connect_async` bounded by `RECONNECT_TIMEOUT_SECS` (30s), `get_socket_mode_url` bounded by `RECONNECT_HTTP_TIMEOUT_SECS` (35s — strictly above the inner client timeout for better diagnostics).
2. **Live socket writes**: new `send_with_timeout()` helper (10s) wraps all four write sites (envelope ack, pong, keepalive ping, shutdown close). Timeout or error → `break` to reconnect, consistent with the existing keepalive-ping error-handling pattern.
3. **Refactors**: `wait_backoff_or_shutdown()` deduplicates three copy-pasted backoff+shutdown blocks; `ConnectResult`/`SocketModeUrlResult` type aliases tame nested generics; `get_socket_mode_url` reuses the adapter's bounded `reqwest::Client` instead of constructing an unbounded one per call (folds in #386).
4. **Intentional behavior change**: ack-send failure now skips event dispatch (correct — Slack redelivers unacked envelopes, so this avoids double-processing).

## Findings

| 1 | 🟢 | Correct timeout layering: outer HTTP timeout (35s) > inner client timeout (30s), ensuring the diagnosable inner error fires first | `slack.rs:718-723` |
| 2 | 🟢 | Consistent 3-arm match pattern at all write sites prevents silently swallowing `tungstenite::Error` | `slack.rs` (all `send_with_timeout` call sites) |
| 3 | 🟢 | Excellent regression tests using `start_paused = true` + stalled local listeners — pins the actual production code path | `slack.rs` (test module) |
| 4 | 🟢 | Clean extraction of `wait_backoff_or_shutdown` eliminates copy-paste divergence risk at three retry points | `slack.rs:776-783` |
| 5 | 🟢 | Deliberate ack-skip-on-timeout documented in PR body and code comments, preventing future misidentification as a bug | `slack.rs:883-908` |

<details>

- **Production-motivated**: directly traces to a real 11-hour outage with clear repro scenario
- **Defense in depth**: bounds both the reconnect path *and* the live socket loop writes — same bug class, complete coverage
- **Testable design**: extracted helpers (`connect_with_reconnect_timeout`, `send_with_timeout`, `get_socket_mode_url_with_timeout`) are unit-testable in isolation
- **Minimal blast radius**: only touches `slack.rs` and its dev-dependencies; no cross-crate changes
- **Well-documented**: thorough PR description explains every decision, calls out the intentional behavior change, and references prior review rounds

</details>

<details>

- PR opened: 2026-07-08
- Main already has: idle-timeout watchdog (`IDLE_TIMEOUT_SECS`), backoff loop structure, `reqwest::Client` with 30s timeout in `SlackAdapter::new`
- Net-new value: timeout coverage for `connect_async`, `get_socket_mode_url`, and all four `write.send()` sites — none of which had any timeout on main

</details>

1. **Tokio-tungstenite or reqwest could add native timeouts upstream** — but they haven't in years, and a production outage already happened. Waiting is not an option.
2. **The idle-timeout watchdog *should* catch half-open sockets** — but as the PR correctly identifies, a stalled *write* in a `select!` branch prevents the tick from ever firing. The existing watchdog has a blind spot this PR closes.
3. **The original #1281 was already LGTM'd and could just be rebased** — this PR *is* that rebase, plus it addresses the one remaining finding (unbounded writes) that the last review round identified. Strictly better than a raw rebase.

chaodu-agent:
> [!NOTE]
> LGTM ✅ — Well-structured timeout + write-bounding fix for a real 11-hour production outage, with proper regression tests.

## What This PR Does

Fixes #1279: the Slack Socket Mode reconnect path (`get_socket_mode_url`, `connect_async`) and the live socket write calls (ack, pong, keepalive ping, shutdown close) had no timeout bounds. A stalled network path during reconnect or a half-open socket on the write side could park the adapter task forever with zero observability. This caused an 11+ hour outage on openab-viola.

## How It Works

1. **Reconnect-path timeouts:** `connect_with_reconnect_timeout()` wraps `connect_async` in a 30s timeout; `get_socket_mode_url_with_timeout()` wraps the HTTP call in a 35s timeout (set above the client's own 30s so the diagnosable inner error fires first).
2. **Write-path timeouts:** `send_with_timeout()` bounds all four outbound write sites with a 10s timeout (`WRITE_TIMEOUT_SECS`). Each site now `break`s to force reconnect on timeout or send error, matching the existing keepalive-ping error-handling pattern.
3. **Client reuse:** `get_socket_mode_url` now accepts the adapter's bounded `reqwest::Client` instead of constructing an unbounded one per call (folds in #386).
4. **Deduplication:** `wait_backoff_or_shutdown()` replaces three copy-pasted backoff+shutdown `select!` blocks.
5. **Deliberate behavior change:** ack-send timeout/error now skips event dispatch (breaks before routing) — correct because Slack redelivers unacked envelopes, avoiding double-processing.
6. **Type aliases:** `ConnectResult`, `SocketModeUrlResult`, `SlackWsSink` for readability.
7. **Tests:** 4 new regression tests covering both timeout legs, `wait_backoff_or_shutdown` happy path and shutdown path, all using `start_paused = true` for instant virtual-clock resolution.

## Findings

| 1 | 🟢 | Timeout hierarchy is well-designed: HTTP 35s > client 30s > WS handshake 30s > write 10s — each layer fires before its outer backstop | constants |
| 2 | 🟢 | Extracting `send_with_timeout` / `connect_with_reconnect_timeout` / `wait_backoff_or_shutdown` into named functions both improves readability and makes the exact production code path testable | `slack.rs` |
| 3 | 🟢 | The ack-skip-dispatch behavior change is explicitly documented in the PR description and is semantically correct (Slack's at-least-once delivery guarantees safe redelivery) | `slack.rs` ack site |
| 4 | 🟢 | 3-arm match pattern (`Ok(Err(e))` / `Err(_)` / `Ok(Ok(()))`) at all write sites ensures both completed-but-errored sends (e.g. broken pipe) and timeouts trigger reconnect, rather than being silently swallowed | all write sites |
| 5 | 🟢 | `dev-dependencies` addition of `tokio = { version = "1", features = ["test-util"] }` is properly scoped and doesn't affect production builds | `Cargo.toml` |

<details>

- PR opened: 2026-07-09 (rebases #1281, which was auto-closed by stale bot)
- Main already has: `IDLE_TIMEOUT_SECS` + `PING_INTERVAL_SECS` idle-watchdog for the *connected* read loop (added in #1120). No timeout protection in the *reconnect* path or on *outbound writes*.
- Net-new value: Bounds all previously-unbounded network calls (2 in reconnect path, 4 in write path), reuses the adapter client (#386), deduplicates backoff logic, adds 4 regression tests. Addresses the exact production outage scenario from #1279.

</details>

<details>

- Complete coverage of the unbounded-call surface: both the reconnect path (HTTP + WS handshake) and the live socket write path (ack, pong, ping, close) are now bounded.
- Tests exercise the actual extracted production functions rather than duplicating the timeout logic, making them true regression guards.
- The `wait_backoff_or_shutdown` extraction removes a common async footgun (copy-pasting `select!` blocks that could drift out of sync).
- PR description is exceptionally thorough — documents the intentional behavior change, links to prior art, and explains the timeout hierarchy.
- CI: `check` passes, all smoke tests pass, 66/66 unit tests pass including the 2 new ones.

</details>

1. **The original #1281 was already LGTM'd** — This is functionally a rebase + one additional fix (write-side timeouts). However, #1281 was auto-closed and could not be merged, so a new PR is the only path forward.
2. **Write-side stalls might be rare in practice** — The original incident was a reconnect-path stall, not a write-path stall. But the PR correctly identifies that write stalls are the same bug class and would produce the same symptom (task parked forever, idle watchdog unreachable), so defense-in-depth is warranted.
3. **A liveness probe would catch future variants** — True, but a liveness probe is a separate concern (detect + restart) vs. this fix (self-heal via reconnect). Both are needed; this PR addresses the root cause rather than papering over it with restarts.

thepagent (approved):

PR #1343, opened 2026-07-10
feat(media): make text file size limit configurable via [media] max_text_file_kb

slps970093:

Text file attachments larger than 512 KB are silently dropped before being passed to the agent. Users have no indication this is happening — the file uploads successfully, the bot receives the message, but the agent simply never sees the file content.

Observed on chisato-bot (2026-07-10): 9 occurrences of \	est-results.txt\ (535–563 KB) being discarded in under 40 minutes, causing the user to repeatedly re-upload the same file.

Discord Discussion URL: https://discordapp.com/channels/1491295327620169908/1491365158868619404/1525027437086380042

## At a Glance

\\\
User uploads test-results.txt (535 KB)
        ↓
Discord/Slack reports attachment size to OpenAB
        ↓
download_and_read_text_file(size, max_bytes)
        ↓
 size > max_bytes? ──yes──> return None  (silent drop ← bug)
        ↓ no
   HTTP download → inline into prompt → agent sees file
\\\

After this change, \max_bytes\ comes from \[media] max_text_file_kb\ in config.toml (default: 512, backward-compatible).

Not applicable — this is a local prompt-assembly parameter with no cross-system protocol implications.

## Proposed Solution

- Add \MediaConfig\ struct to \config.rs\ with a single \max_text_file_kb: u64\ field (default 512).
- Replace the hardcoded \const MAX_SIZE: u64 = 512 * 1024\ in \media.rs\ with a \max_bytes: u64\ parameter.
- Thread \media_config\ through \discord.rs\, \slack.rs\, and \main.rs\ following the same pattern as \SttConfig\.
- Add two regression tests confirming the size gate behaviour.

## Why this approach?

Keeps the change minimal and targeted. The \TEXT_TOTAL_CAP\ (1 MB aggregate) remains a hard ceiling, so raising the per-file limit cannot cause prompt bloat beyond what was already allowed. All other adapters (Google Chat, Feishu, etc.) have their own download logic and are unaffected.

## Alternatives Considered

- **Raise the hardcoded default to 1 MB** — affects all deployments without opt-in, violates backward-compat rule.
- **Notify the user when a file is skipped** — useful but orthogonal; can be a follow-up.

## Validation

- [x] \cargo fmt\ passes
- [x] \cargo clippy --workspace -- -D warnings\ passes
- [x] \cargo clippy --workspace --features unified -- -D warnings\ passes
- [x] \cargo test --workspace\ passes (916 tests)
- [x] New regression tests: \	ext_file_rejects_when_size_exceeds_max_bytes\, \	ext_file_size_gate_is_strict_greater_than\
- [x] Manual: adding \[media] max_text_file_kb = 1024\ to config.toml resolves the chisato issue

### DIFF
--- a/crates/openab-core/src/config.rs
+++ b/crates/openab-core/src/config.rs
@@ -140,6 +140,8 @@ pub struct Config {
     #[serde(default)]
     pub stt: SttConfig,
     #[serde(default)]
+    pub media: MediaConfig,
+    #[serde(default)]
     pub markdown: MarkdownConfig,
     #[serde(default)]
     pub cron: CronConfig,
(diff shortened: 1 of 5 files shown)


slps970093:
https://discordapp.com/channels/1491295327620169908/1491365158868619404/1525027437086380042

PR #1385, opened 2026-07-13
feat(feishu): full [feishu] section — all settings config-first

chaodu-agent:
## What problem does this solve?

Final platform slice of the config-first parity umbrella #1375: Feishu was the only platform with **no config section at all** — 21 settings were env-only. The new `[feishu]` section covers credentials, connection, behavior, and L3 identity trust, resolving each field **config → `FEISHU_*` env → default**.

## How It Works

**Single source of truth design**: the gateway adapter's parser keeps all parsing rules and defaults. `FeishuConfig::from_env` is refactored onto a `from_reader`, and the core section's `resolve_pairs()` renders typed TOML fields into the same string form the env vars use — no default value or enum rule is duplicated between crates (with 21 fields, duplicated defaults would drift).

| Layer | Change |
|-------|--------|
| core | `FeishuConfig` (21 typed fields) + `resolve_pairs()` + `trust_config()` view |
| gateway | `from_reader` refactor (`parse_csv` inlined through the reader); `GatewayFeishuConfig` bridge + `apply_feishu_config()` — `app_id`+`app_secret` mandatory, incomplete section disables the adapter |
| unified | Applies `[feishu]` before `warn_unenforceable_l1` (config-supplied `encrypt_key` satisfies the #1373 L1 check — test-pinned); **feishu joins the per-platform trust registry override** — it was the last platform riding the uniform `GATEWAY_*` seed (the registry half of #1357; gateway-side double-gate elimination stays there); webhook mount honors the resolved path |
| docs | example (full commented block), config-reference (21-field table), feishu.md |

## Tests

- `feishu_resolve_pairs_scenarios` — absent keys passthrough, config-wins-over-env, typed→string rendering (bool/u32/CSV), empty-string `${}` fallthrough, `trust_config()`
- `feishu_section_parses_from_toml` — TOML parse of typed fields
- `apply_feishu_config_requires_credentials` — adapter built via shared `from_reader` (connection_mode/encrypt_key verified), config-supplied `encrypt_key` clears the L1 flag, missing secret disables

## Validation

- clippy `-D warnings` clean (core+gateway all-features; gateway no-features); unified + default builds pass
- core 665 passed (1 known pre-existing macOS `secrets` failure); gateway 262/262
- rustfmt isolated: feishu.rs 93 vs baseline 94 (no new drift; one pre-existing diff incidentally fixed), other files 0→0

Closes #1377. Refs #1375 (umbrella — completes the platform matrix), #1357 (registry task), #1373.

PR #1381, opened 2026-07-12
feat(line): full [line] section — credentials + connection config-first

chaodu-agent:
## What problem does this solve?

First platform slice of the config-first parity umbrella #1375: LINE's channel credentials were env-only (`LINE_CHANNEL_SECRET` / `LINE_CHANNEL_ACCESS_TOKEN`), violating the invariant that every platform setting resolves **`[section]` config → `PLATFORM_*` env → default** with config always winning.

## Proposed Solution

Mirrors the `TelegramConfig` reference pattern end-to-end:

| Layer | Change |
|-------|--------|
| core `config.rs` | `LineConfig` gains `channel_secret`, `channel_access_token`, `webhook_path`; `resolve()` applies config → `LINE_*` env → default per field (empty-string `${}` expansion falls through to env) |
| gateway `lib.rs` | `GatewayLineConfig` bridge + `AppState::apply_line_config()` (keeps the crate free of openab-core); new `line_webhook_path` state; standalone gateway honors `LINE_WEBHOOK_PATH` and mounts at the resolved path |
| unified `main.rs` | Applies `[line]` **before** `warn_unenforceable_l1`, so a config-supplied `channel_secret` satisfies the #1373 L1 startup check; route mounts at the resolved path |
| docs | `config.toml.example`, `config-reference.md` field table, `line.md` (`[line]` primary, env fallback) |

Backward compatible: env-only deployments resolve identically; default mount path unchanged.

## Tests

- `line_section_parses_from_toml` — new fields parse
- `line_resolve_all_scenarios` — 3 new scenarios: config-wins-over-env for all three fields; empty-string expansion falls through to env; nothing set → defaults
- `apply_line_config_overrides_env_state_and_feeds_l1_warning` — pins the #1373 integration: env-derived flagged state clears once config supplies the secret

## Validation

At head (macOS arm64, main `9af9fcd`):
- `cargo clippy -p openab-core -p openab-gateway --all-features -- -D warnings` — clean; gateway no-features also clean
- `cargo check --features unified` and default — pass
- core line tests 16/16; gateway l1_audit tests 5/5
- `rustfmt --check` isolated per file: zero drift vs main (all three files 0→0)

Closes #1376. Refs #1375 (umbrella), #1373 (L1 warning integration).

Issue #1375, opened 2026-07-12
umbrella: config-first parity — every platform section resolves config → env → default for ALL fields


chaodu-agent:
> [!NOTE]
> **✅ COMPLETE (2026-07-13).** All six platform sections shipped (#1381–#1385), docs landed, config-supplied L1 secrets feed the #1373 startup warning (test-pinned per platform), and the conformance guard (#1387) prevents env-first regression workspace-wide. The invariant holds everywhere: **`[section]` config → `PLATFORM_*` env → default, config always wins.**

## Summary

Guarantee every platform has a **complete first-class config section** (`[telegram]`, `[line]`, `[feishu]`, `[googlechat]`, `[wecom]`, `[teams]`) covering **all** of its settings — credentials, connection, and behavior — with the invariant:

> **Resolution order: `[section]` config value → `PLATFORM_*` env var → default. Config always wins; env is a fallback only. Never consult env without checking the section first.**

Today only the trust fields (`allow_all_users` / `allowed_users`, from #1356's Phase 1 PRs) follow this. Platform credentials and connection settings are still env-only for five of six platforms.

## Parity Status (updated 2026-07-13 — ALL PLATFORMS DONE 🎉)

| Platform | Section | Config-first today | Env-only remaining |
|----------|---------|--------------------|--------------------|
| Telegram | `[telegram]` | ✅ ALL fields (`resolve()` → `ResolvedTelegram`, bridged via `apply_telegram_config`) | — (reference pattern) |
| LINE | `[line]` | ✅ ALL fields (#1381) | — |
| Feishu | `[feishu]` | ✅ ALL 21 fields (#1385) + joined the per-platform trust registry | — |
| WeCom | `[wecom]` | ✅ ALL fields (#1382) | — |
| Google Chat | `[googlechat]` | ✅ ALL fields (#1383, incl. `from_parts` DRY refactor) | — |
| MS Teams | `[teams]` | ✅ ALL fields (#1384) | — |

## Reference Pattern (Telegram — already compliant)

1. Core `TelegramConfig` with all fields `Option<T>` + `resolve()` applying config → `TELEGRAM_*` env → default per field (empty-string `${}` expansion treated as unset)
2. Bridge struct (`GatewayTelegramConfig`) keeps `openab-gateway` free of an `openab-core` dependency
3. `AppState::apply_telegram_config()` overrides env-derived state after `from_env`
[...]

PR #1387, opened 2026-07-13
test(gateway): config-first conformance guard (#1375)

chaodu-agent:

The final open item on umbrella #1375: with all six platform sections shipped (#1381–#1385), the only missing piece is a **guard preventing regression to env-first** — nothing stopped a future change from adding a new `FEISHU_NEW_KNOB` env read without a corresponding `[feishu]` field, silently violating the config → env → default invariant.

## How It Works

A source-scan conformance test in `openab-gateway/tests/` (same pattern as the `platform-schema` suite — deterministic, no network, reads the crate's own sources via `CARGO_MANIFEST_DIR`):

- **`every_platform_env_var_has_a_config_section_field`** — extracts platform-prefixed env-var string literals from all crate sources and asserts each is in the canonical covered set (the per-platform section↔env mapping). A violation fails with step-by-step fix instructions: add the section field → thread the bridge → document → extend the allowlist.
- **`covered_set_has_no_stale_platform_entries`** — symmetric hygiene check so the allowlist cannot accumulate dead entries (trust vars consumed by openab-core's registry path are exempt).

Runs in the normal `cargo test` flow — no new CI wiring needed, and unlike extending the `platform-schema` workflow (which only triggers on schema paths), this fires on any gateway-crate change.

## Self-review (round 2 commit)

- **Scope widened to the workspace**: the crate-local scan missed that env-first regression can also land in `openab-core` or the `openab` binary. The guard now walks all three source roots; negative-tested in each (rogue vars injected into `openab-core/src/gateway.rs` and `src/main.rs` are both caught with file locations).
- **Dynamic-key subtlety surfaced and documented**: trust vars are read via `format!("{prefix}_ALLOW_ALL_USERS")` in `PlatformTrustConfig::resolve_with_env` — no literal exists for a literal scan, so the stale-check exempts them by suffix, now with the rationale in-code (round 1 had the right exemption for an unstated reason).
- **Extractor heuristic honestly documented**: conservative fail-closed direction; prose fails the charset filter, comment mentions flag conservatively, escaped-quote mis-pairs produce garbage that gets filtered.

## Validation

- Current tree: 2/2 green — all platform env vars referenced across the workspace are covered
- **Negative tests ×3**: rogue env reads injected into the gateway crate, `openab-core`, and the binary each fail the guard, naming the var and file
- rustfmt clean (0 diffs); clippy `--tests -D warnings` clean

Refs #1375 — closes out the umbrella's final task.

### DIFF
--- a/crates/openab-gateway/tests/config_first_conformance.rs
+++ b/crates/openab-gateway/tests/config_first_conformance.rs
@@ -1,8 +1,9 @@
 //! Config-first conformance guard (#1375).
 //!
-//! Invariant: **every platform environment variable read anywhere in this
-//! crate must have a corresponding first-class `[section]` field in
-//! `openab-core`'s config**, so that the resolution order
+//! Invariant: **every platform environment variable referenced anywhere in
+//! the workspace (this crate, `openab-core`, and the `openab` binary) must
+//! have a corresponding first-class `[section]` field in `openab-core`'s
+//! config**, so that the resolution order
 //! `config → PLATFORM_* env → default` holds for every setting (config always
 //! wins; env is a fallback only — never the sole source).
 //!
@@ -115,6 +116,15 @@ const PLATFORM_PREFIXES: &[&str] = &[
 ];
 
 /// Extract platform-prefixed env-var string literals from Rust source text.
+///
+/// Deliberately conservative heuristic (not a Rust parser): any quoted
+/// literal that starts with a platform prefix and is entirely
+/// `[A-Z0-9_]` counts — including ones appearing in comments or test code.
+/// Failure direction is safe: prose in strings/comments fails the charset
+/// filter and is ignored, while a var name mentioned in a comment flags
+/// conservatively (fail-closed). Escaped quotes can mis-pair the scanner,
+/// but real env reads are always plain literals, and garbage extractions
+/// fail the charset filter.
 fn platform_env_literals(source: &str) -> BTreeSet<String> {
     let mut found = BTreeSet::new();
     let bytes = source.as_bytes();
@@ -150,14 +160,38 @@ fn collect_rs_files(dir: &Path, out: &mut Vec<std::path::PathBuf>) {
     }
 }
 
+/// Scan roots: this crate, openab-core, and the binary crate. Workspace-
+/// relative paths from CARGO_MANIFEST_DIR (crates/openab-gateway) — these
+/// crates are not published, so the layout is stable in CI and dev trees.
+fn workspace_scan_roots() -> Vec<std::path::PathBuf> {
+    let here = Path::new(env!("CARGO_MANIFEST_DIR"));
+    vec![
+        here.join("src"),
+        here.join("../openab-core/src"),
+        here.join("../../src"),
+    ]
+}
+
+fn collect_workspace_files() -> Vec<std::path::PathBuf> {
+    let mut files = Vec::new();
+    for root in workspace_scan_roots() {
+        assert!(
+            root.is_dir(),
+            "scan root missing ({}) — workspace layout changed? update \
+             workspace_scan_roots()",
+            root.display()
+        );
+        collect_rs_files(&root, &mut files);
+    }
+    files
+}
+
 #[test]
 fn every_platform_env_var_has_a_config_section_field() {
-    let src = Path::new(env!("CARGO_MANIFEST_DIR")).join("src");
-    let mut files = Vec::new();
-    collect_rs_files(&src, &mut files);
+    let files = collect_workspace_files();
     assert!(
         !files.is_empty(),
-        "no source files found — CARGO_MANIFEST_DIR layout changed?"
+        "no source files found — workspace layout changed?"
     );
 
     let covered: BTreeSet<&str> = COVERED.iter().copied().collect();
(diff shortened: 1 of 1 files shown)


chaodu-agent:
> [!NOTE]
> LGTM ✅ — Clean, self-contained conformance guard that prevents config-first regression.

## What This PR Does

Closes the last open item on umbrella #1375: adds a compile-time source-scan test that fails if any platform-prefixed env var is read in the gateway crate without a corresponding first-class config section field.

## How It Works

Two integration tests in a new `tests/config_first_conformance.rs`:

1. **`every_platform_env_var_has_a_config_section_field`** — scans all `src/**/*.rs` for `"PLATFORM_*"` string literals and asserts each is in the canonical `COVERED` allowlist.
2. **`covered_set_has_no_stale_platform_entries`** — symmetric check ensuring the allowlist does not accumulate dead entries (trust vars exempted).

Uses `CARGO_MANIFEST_DIR` to locate sources at test time — no network, no external state, runs in normal `cargo test`.

## Findings

| 1 | 🟢 | Correct scoping: scans only `src/`, avoiding self-reference from test files | `config_first_conformance.rs:141` |
| 2 | 🟢 | Good exemption for trust vars (`*_ALLOW_ALL_USERS`, `*_ALLOWED_USERS`) consumed by core registry | `config_first_conformance.rs:199` |
| 3 | 🟢 | Actionable error messages with step-by-step fix instructions on failure | `config_first_conformance.rs:160-175` |
| 4 | 🟢 | `BTreeSet` ensures deterministic output ordering in violation reports | `config_first_conformance.rs:90` |

<details>

- **Zero-config CI integration** — runs in `cargo test` without new workflow wiring
- **Symmetric design** — guards both directions (uncovered vars AND stale allowlist entries)
- **Developer UX** — failure messages include exact file + var name + numbered fix steps
- **Minimal surface** — single file, no new dependencies, purely additive
- **Sound parser design** — only matches double-quoted uppercase string literals with platform prefixes, avoiding false positives from comments or non-literal references

</details>

<details>

- PR opened: 2026-07-13
- Main already has: no `tests/` directory in `openab-gateway`, no conformance guard for config-first invariant
- Net-new value: automated regression prevention ensuring the config → env → default resolution order is maintained for all platform settings

</details>

1. **Clippy/CI could catch this differently** — a custom lint or `deny(unused)` on env reads could serve the same purpose. However, custom lints require nightly and are harder to maintain; a test is the pragmatic choice.
2. **The allowlist is manual maintenance** — every new env var requires updating `COVERED`. However, the staleness test ensures dead entries are caught, and the failure message guides developers through the exact steps.
3. **Source scanning is fragile against refactors** — if env reads move to a macro or proc-macro expansion, the string scanner won't catch them. However, the current codebase uses direct `std::env::var("...")` calls, and any macro adoption would be a deliberate architectural shift that would naturally prompt updating the test.

thepagent (approved):

chaodu-agent:
> [!NOTE]
> LGTM ✅ — Clean, well-documented conformance guard that fail-closes against env-first regression.

## What This PR Does

Adds a source-scan conformance test that prevents platform env-var reads from bypassing the config-first invariant (config → env → default). Closes the final task on umbrella #1375.

## How It Works

A 254-line integration test in `crates/openab-gateway/tests/config_first_conformance.rs`:

1. **Literal scanner** — extracts all `"PLATFORM_*"` string literals from workspace sources (gateway, openab-core, binary)
2. **Positive guard** — asserts every found var is in the `COVERED` allowlist (fails with step-by-step fix instructions)
3. **Symmetric hygiene** — asserts no stale entries remain in the allowlist (dynamic-key trust vars exempted by suffix)

No CI wiring needed — runs in the standard `cargo test` flow.

## Findings

| 1 | 🟢 | Fail-closed extraction heuristic: conservative charset filter means false-positives flag instead of silently missing vars | `config_first_conformance.rs:100-115` |
| 2 | 🟢 | Dynamic-key exemption is properly justified — `PlatformTrustConfig::resolve_with_env` uses `format!` for `_ALLOW_ALL_USERS`/`_ALLOWED_USERS`, making literal scan impossible | `config_first_conformance.rs:225-232` |
| 3 | 🟢 | Symmetric hygiene prevents allowlist rot without false-positives on dynamic keys | `config_first_conformance.rs:210-253` |
| 4 | 🟢 | Workspace-wide scanning (3 roots) with assert guards — catches regressions in any crate, not just gateway | `config_first_conformance.rs:138-150` |

<details>

- **Fail-closed direction**: the scanner is deliberately conservative — charset filter rejects prose but catches conservatively-mentioned vars in comments. This is the right failure direction for a guard.
- **Honest documentation of limitations**: the doc-comment on `platform_env_literals` openly states what can go wrong (escaped quotes, comment mentions) and why the failure direction is safe.
- **Dynamic-key subtlety**: `_ALLOW_ALL_USERS`/`_ALLOWED_USERS` vars are constructed via `format!()` at runtime — the stale-check exempts them by suffix with clear rationale in-code.
- **Self-contained**: no network, no external state, deterministic — same pattern as the existing `platform-schema` suite.

</details>

<details>

- PR opened: refs umbrella #1375, follows platform sections #1381–#1385
- Main already has: no `tests/` directory in `crates/openab-gateway/`
- Net-new value: first integration test for the gateway crate; compile-time guard against config-first regression across the entire workspace

</details>

1. **Literal scan is inherently incomplete** — dynamically constructed env-var names (beyond the trust vars) would escape detection. However, the current codebase uses only `format!` for trust vars, and the fail-closed direction means new patterns would likely surface as false positives rather than silent gaps.

2. **Clippy/lints could evolve to catch this** — a custom lint or workspace-level policy could enforce the invariant. But no such lint exists today, and this guard is zero-cost to maintain (only fires when someone adds a new env read).

3. **The allowlist is manual maintenance** — every new platform var requires updating `COVERED`. However, the test itself prints step-by-step instructions when violated, and the stale-check prevents accumulation — the maintenance burden is minimal and self-documenting.

PR #1382, opened 2026-07-13
feat(wecom): full [wecom] section — credentials + connection config-first

chaodu-agent:
> [!NOTE]
> **Stacked on #1381** (base branch is the LINE slice; will re-target to `main` after it merges). Net-new diff = the last commit.

## What problem does this solve?

Second platform slice of #1375: WeCom's credentials and connection settings were env-only. `[wecom]` graduates from the shared trust-only `PlatformTrustConfig` to a dedicated `WecomConfig` covering all 8 settings + trust, resolving each field **config → `WECOM_*` env → default**.

## How It Works

| Layer | Change |
|-------|--------|
| core | `WecomConfig`/`ResolvedWecom` + per-field `resolve()`; `trust_config()` view keeps the trust-registry override path unchanged |
| gateway | `GatewayWecomConfig` bridge + `AppState::apply_wecom_config()` — rebuilds the adapter through the **same `from_reader` validation** as env-only construction (5 mandatory credentials, numeric `agent_id`, 43-char AES key); incomplete section ⇒ adapter disabled, matching env-only semantics |
| unified | Applies `[wecom]` before `warn_unenforceable_l1` |
| docs | `config.toml.example`, `config-reference.md` (dedicated `[wecom]` section; combined section shrinks to googlechat/teams), `wecom.md` |

## Tests

- `wecom_resolve_all_scenarios` — defaults, config-wins-over-env, empty-string `${}` fallthrough, `trust_config()` semantics
- Existing `platform_trust` tests unchanged and passing (trust behavior preserved)
- `from_reader` reuse means adapter-construction validation is already covered by the adapter's own tests

## Validation

- clippy `-D warnings` clean (core+gateway all-features; gateway no-features); `cargo check --features unified` + default pass
- core suite 661 passed (1 known pre-existing macOS `secrets` failure); gateway 259 passed
- rustfmt isolated: 0 new drift (config.rs 0→0 after fixes, lib.rs 0→0)

Closes #1378. Refs #1375 (umbrella).

Issue #1356, opened 2026-07-11
umbrella: Phase 1 trust migration — per-platform [section] for all adapters


chaodu-agent:
## Summary

Track the migration of all platform trust configurations from legacy env vars / shared gateway config to first-class per-platform `[section]` entries in `config.toml`, as defined in the identity-trust-none ADR (#1291, `docs/adr/identity-trust-none.md`).

## Goal

Each platform gets its own config section with `allowed_users`, `allow_all_users`, and platform-specific trust extensions. The shared `PlatformTrustConfigs` registry routes trust decisions per-platform. Legacy env vars are deprecated → error → removed across Phase 1/2/3.

## Current Registry State (verified against `src/main.rs` on main, updated 2026-07-12 after #1363/#1365/#1366)

The `PlatformTrustConfigs` registry (`crates/openab-core/src/trust.rs`) is wired in `src/main.rs`:

- **Uniform seed**: all six gateway platforms are first seeded from the shared `GATEWAY_ALLOW_ALL_USERS` / `GATEWAY_ALLOWED_USERS` env vars — now only a **fallback**; per-platform sections override it.
- **Per-platform overrides in place**: `[discord]` (#1270), `[telegram]` (#1297), `[slack]` (#1363 — registry entry + redundant-but-matching `gate_incoming` in the adapter), `[line]` (#1365), `[wecom]`/`[googlechat]`/`[teams]` (#1366, shared `PlatformTrustConfig` + `platform_trust_override` helper). Each of line/wecom/googlechat/teams logs a Phase 1 deprecation warning when it is active but still trust-driven by the legacy `GATEWAY_*` env.
- ~~Feishu on uniform seed~~ — resolved by #1385: `[feishu]` joined the per-platform registry override. **All 8 platforms now have per-platform trust config.** The remaining Feishu item is the gateway-crate double-gate elimination (#1357).
- ~~**Mode scoping caveat**~~ resolved by #1391: per-platform sections now apply in BOTH modes — the WS path consults the shared registry with precedence `GATEWAY_*` env < `[gateway]` < `[<platform>]`.

## Platform Migration Status

| Platform | Current State | Remaining Work | Sub-issue | Status |
|----------|--------------|----------------|-----------|--------|
| Discord | `[discord]` section wired into registry (L3) | — | #1270 | ✅ Done |
| Telegram | `[telegram]` section wired into registry | Group policy | #1362 | ⚠️ Partial |
[...]

PR #1389, opened 2026-07-13
fix(wecom): enable config-only unified startup

### LINKED ISSUE #1388: fix(wecom): allow config-only unified startup without WECOM_CORP_ID env
## Summary

A complete first-class `[wecom]` section cannot start the unified WeCom adapter unless `WECOM_CORP_ID` is also present in the process environment.

This breaks the config-first contract introduced by #1382: individual fields correctly resolve `config -> WECOM_* env -> default`, but adapter detection still depends on the environment before the resolved WeCom config is applied.

## Reproduction

Use a config containing only a complete WeCom adapter section:

```toml
[wecom]
corp_id = "ww1234567890abcdef"
secret = "test-secret"
token = "test-token"
encoding_aes_key = "abcdefghijklmnopqrstuvwxyzABCDEFGH123456789"
agent_id = "1000002"
allow_all_users = true
```

Ensure `WECOM_CORP_ID` and the other `WECOM_*` variables are unset, then run:

```console
openab run -c config.toml
```

Current result:

```text
Error: no adapter configured — add [discord], [slack], [telegram], or [gateway] to config, or set platform env vars (TELEGRAM_BOT_TOKEN, etc.)
```

Expected result: the unified gateway starts and mounts the configured WeCom webhook route using the `[wecom]` credentials.

## Root cause

There are two config-blind startup gates in `src/main.rs`:

1. The no-adapter validation checks `cfg.telegram.is_none()` but does not consider `cfg.wecom`.
2. Embedded gateway initialization uses `has_unified_platform_env() || cfg.telegram.is_some()`. `has_unified_platform_env()` only checks `WECOM_CORP_ID`, not resolved `[wecom]` configuration.

The WeCom trust deprecation activity signal also uses only `WECOM_CORP_ID`; config-only activation should be considered there as well.

## Impact

- Existing env-only WeCom deployments are unaffected.
- `[wecom]` configurations using literal values or non-`WECOM_*` secret substitution cannot start without redundantly exporting `WECOM_CORP_ID`.
- The field-level config resolution tests pass, so the current config-first conformance guard does not catch this startup-path regression.

## Suggested acceptance criteria

- A complete `[wecom]` section starts the unified gateway with all `WECOM_*` variables unset.
- An incomplete `[wecom]` section continues to fail safely or leave the adapter disabled according to the existing validation contract.
- Adapter detection and trust migration warnings recognize config-only WeCom activation.
- Add a regression test covering config-only startup/adapter detection, not only `WecomConfig::resolve()`.

Related: #1375, #1382.

## Context

- [Discord discussion](https://discord.com/channels/1491295327620169908/1491365158868619404/1526200802824421486)

canyugs:
## Summary

- recognize a complete first-class `[wecom]` section during no-adapter startup validation
- start the embedded unified gateway when WeCom credentials resolve from config without `WECOM_CORP_ID` being exported
- evaluate activation after secret substitution and before config fields are moved into runtime components
- add a regression test for config-only WeCom activation

## Root cause

#1382 made every WeCom field resolve `config -> WECOM_* env -> default`, but the outer startup gates still relied on `has_unified_platform_env()`. A literal or secret-substituted `[wecom]` section was therefore rejected before `apply_wecom_config()` could construct the adapter.

The new helper requires all five credentials needed by the WeCom adapter after config/env resolution, so incomplete sections do not become an activation signal.

## Validation

- `cargo clippy -- -D warnings`
- `cargo clippy --features unified -- -D warnings`
- `cargo test`
- `cargo test --features unified`
- config-only startup smoke with all `WECOM_*` variables unset: adapter configured, `/webhook/wecom` mounted, server listening, graceful shutdown successful
- changed hunks are rustfmt-clean; repository-wide `cargo fmt --all -- --check` remains blocked by pre-existing formatting drift across unrelated files

Closes #1388.

## Context

- [Discord discussion](https://discord.com/channels/1491295327620169908/1491365158868619404/1526200802824421486)

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -129,6 +129,24 @@ fn has_unified_platform_env() -> bool {
                 .unwrap_or(false))
 }
 
+/// Returns true when the first-class `[wecom]` section resolves all credentials
+/// required to construct the embedded WeCom adapter. This must remain separate
+/// from [`has_unified_platform_env`]: config-first deployments may intentionally
+/// provide literal or secret-substituted values without exporting `WECOM_*`.
+fn has_unified_wecom_config(cfg: &config::Config) -> bool {
+    if !cfg!(feature = "wecom") {
+        return false;
+    }
+    cfg.wecom.as_ref().is_some_and(|wecom| {
+        let resolved = wecom.resolve();
+        resolved.corp_id.is_some()
+            && resolved.secret.is_some()
+            && resolved.token.is_some()
+            && resolved.encoding_aes_key.is_some()
+            && resolved.agent_id.is_some()
+    })
+}
+
 /// Apply a platform's first-class trust section to the registry, or — when the
 /// platform is active but still trust-driven by the deprecated uniform
 /// `GATEWAY_ALLOW_ALL_USERS`/`GATEWAY_ALLOWED_USERS` env — log the Phase 1
@@ -267,9 +285,10 @@ async fn main() -> anyhow::Result<()> {
         && cfg.gateway.is_none()
         && cfg.telegram.is_none()
         && !has_unified_platform_env()
+        && !has_unified_wecom_config(&cfg)
     {
         anyhow::bail!(
-            "no adapter configured — add [discord], [slack], [telegram], or [gateway] to config, or set platform env vars (TELEGRAM_BOT_TOKEN, etc.)"
+            "no adapter configured — add [discord], [slack], [telegram], [wecom], or [gateway] to config, or set platform env vars (TELEGRAM_BOT_TOKEN, etc.)"
         );
     }
 
@@ -299,6 +318,18 @@ async fn main() -> anyhow::Result<()> {
         let substituted = secrets::substitute(&raw_expanded, &resolved);
         cfg = config::parse_config_str(&substituted, &config_source)?;
     }
+
+    // Compute before individual config fields are moved into their runtime
+    // components below. Secret-backed values have been substituted by now.
+    #[cfg(any(
+        feature = "telegram",
+        feature = "line",
+        feature = "feishu",
+        feature = "googlechat",
+        feature = "wecom",
+        feature = "teams",
+    ))]
+    let unified_wecom_configured = has_unified_wecom_config(&cfg);
 
     let shutdown_hook = cfg.hooks.pre_shutdown.clone();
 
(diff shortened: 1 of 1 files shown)


thepagent (approved):
Maintainer review — **Approving** ✅

Root cause analysis is exactly right: #1382 made every `[wecom]` field resolve config-first, but both activation gates (`no adapter configured` bail + unified-server spawn) still keyed solely on `has_unified_platform_env()` — so a config-only section was rejected before `apply_wecom_config()` ever ran. Thanks for the precise diagnosis and the tight scope.

What I verified locally (macmini, PR head):

- `cargo clippy -- -D warnings` and `--features unified` — both clean
- `complete_wecom_section_enables_unified_startup_without_env` passes under default and unified feature sets — nice touch asserting `== cfg!(feature = "wecom")` so the test is correct in both matrices
- Traced the config-only path end-to-end: bail gate → `unified_wecom_configured` (computed post-substitution, pre-move, as your comment documents) → server spawn → `apply_wecom_config` (`if let Some(ref w) = cfg.wecom`, no env needed) → route mount keyed on `gw_state.wecom` — no remaining env dependency anywhere on the path
- The all-five-credentials requirement is a good anti-footgun: an incomplete section can't become an activation signal

Non-blocking findings, all fine to land as-is:

1. **Bail-gate runs pre-substitution** (line ~288 vs substitution at ~318): for a `secrets`-driven section, the first `has_unified_wecom_config` call sees unresolved placeholder strings — non-empty, so the gate passes and the authoritative post-substitution recompute takes over. Correct today, but subtly order-dependent; a one-line comment at the bail-gate call noting "placeholder values are fine here — only reachability matters, the post-substitution recompute is authoritative" would protect it from future refactors.
2. **Sibling platforms have the same gap**: `[line]`/`[feishu]`/`[googlechat]`/`[teams]` config-only startup still requires the env activity signal. Scoping this PR to WeCom (matching #1388) is right; I'll track the generalization (a `has_unified_platform_config()` covering all six, likely replacing the wecom-specific helper) as a follow-up on #1356.
3. **Trust-override activity signal** (line ~549) still keys on `WECOM_CORP_ID` env — in a config-only deployment the Phase-1 deprecation warning for uniform `GATEWAY_*` trust won't fire. Warn-only impact, and it folds naturally into follow-up (2).

CI matrix still finishing at review time; assuming green, this is ready to merge from my side.

thepagent (approved):

Issue #1357, opened 2026-07-11
feat(feishu): migrate Feishu trust config to first-class [feishu] section (Phase 1)


chaodu-agent:
## Summary

Migrate Feishu adapter trust configuration from gateway env vars to a first-class `[feishu]` section in `config.toml`.

## Current State

- `FEISHU_ALLOWED_USERS` / `FEISHU_ALLOWED_GROUPS` env vars in gateway crate
- Double-gating: gateway crate checks AND core checks (can diverge, core side fails open when list empty)
- `trusted_bot_ids` needed by Receiver for `is_bot` (Feishu marks bots as `sender_type="user"`)

## Target State

```toml
[feishu]
allowed_users = ["ou_xxxxxxxxxxxxxxxxxxxx"]
trusted_bot_ids = ["ou_bot1", "ou_bot2"]
# allow_all_users = true  # explicit opt-in only
```

## Tasks

- [ ] Add `[feishu]` section to config schema
- [ ] Wire Feishu trust through `PlatformTrustConfigs` registry
- [ ] Migrate `FEISHU_ALLOWED_USERS` / `FEISHU_ALLOWED_GROUPS` to core config
- [ ] `trusted_bot_ids` in core config (shared: Receiver reads for `is_bot`, Handler for admission)
- [ ] Eliminate double-gating (gateway = L1 only, trust = core Trust Gate)
- [ ] Deprecate gateway env vars (Phase 1 warn → Phase 2 error → Phase 3 remove)
- [ ] Update docs and `config.toml.example`

## Notes

- `open_id` is **per-app** — same user has different ID in different Feishu apps (document in gotcha)
- Feishu marks other bots as `sender_type="user"` — unreliable; must match against `trusted_bot_ids`

## Refs

- Umbrella: #1356
- ADR: #1291
- LINE equivalent: #1355

PR #1297, opened 2026-07-05
feat(telegram): add allow_all_users/allowed_users to [telegram] config

chaodu-agent:
## Summary

`[discord]`/`[slack]` have had `allowed_users` in their first-class config
sections for a while; `[telegram]` never did. The only way to restrict who
can message a Telegram bot today is the shared `GATEWAY_ALLOW_ALL_USERS`/
`GATEWAY_ALLOWED_USERS` env vars — which default to deny-all (per the
identity-trust-none ADR) and aren't Telegram-specific in name or intent.

Came up directly from deploying a real bot via `oabctl` — the manifest
schema has no plain (non-secret) env var field, only `spec.secrets`, so there
was no clean way to admit a single Telegram user ID without routing it
through Secrets Manager or opening the bot to everyone.

## Change

Adds `allow_all_users`/`allowed_users` to `[telegram]`, matching the exact
resolution convention already used for every other `[telegram]` field and
for `[discord].allowed_users`:

```toml
[telegram]
bot_token = "${TELEGRAM_BOT_TOKEN}"
allowed_users = ["176096071"]
```

- Config value wins → falls back to `TELEGRAM_ALLOW_ALL_USERS`/
  `TELEGRAM_ALLOWED_USERS` (comma-separated) env vars → auto-detects
  `allow_all_users` from whether the resolved list is empty (same convention
  as `[discord]`/`[slack]`: non-empty list defaults to deny-all-except-list,
  empty list defaults to allow-all).
- Wired into `main.rs`'s `PlatformTrustConfigs` registry as a
  Telegram-specific override, right after the existing Discord override —
  same pattern, same place.

```
Before:
  Telegram L3 identity ── only via shared GATEWAY_ALLOW_ALL_USERS /
                           GATEWAY_ALLOWED_USERS (deny-all by default,
                           not Telegram-specific)

After:
  [telegram].allowed_users (or TELEGRAM_ALLOWED_USERS env)
          │
          ▼
  TelegramConfig::resolve() → ResolvedTelegram.allowed_users/allow_all_users
          │
          ▼
  main.rs: reg.insert("telegram", TrustConfig::new(..., resolved values))
          │
          ▼
  overrides the shared GATEWAY_* L3 default for the "telegram" platform only
```

## Testing done

```
cargo build --features unified      # clean
cargo clippy --features unified -- -D warnings   # clean
cargo test --features unified       # openab-core: 48 passed; openab: 15 passed
```

[...]

PR #1391, opened 2026-07-13
feat(gateway)!: route standalone WS path through the shared trust gate

chaodu-agent:

The **standalone-gateway WebSocket path was the last ingress that bypassed the shared trust registry** (#1356 Phase 1c prerequisite). `run_gateway_adapter` enforced channel/user allowlists via its own inline filter fed only by `[gateway]` values — so the per-platform `[<platform>]` trust sections (shipped across #1297–#1385) had **no effect** on standalone deployments, and six doc callouts had to carry mode-scoping caveats explaining that gap.

## Before vs after

```text
BEFORE                                      AFTER
======                                      =====
unified path:                               unified path:
  should_skip_event (structural)              should_skip_event (structural)
  → gate_incoming (registry)                  → gate_gateway_event (shared helper)
                                                └ gate_incoming (registry)
WS path:                                    WS path:
  should_skip_event with [gateway]            should_skip_event (structural only)
  channel/user allowlists                     → gate_gateway_event (same helper,
  (registry NOT consulted;                      same registry, deny-echo included)
   [<platform>] sections ignored)
                                            registry precedence (per platform):
                                              GATEWAY_* env  <  [gateway]  <  [<platform>]
```

## How It Works

1. **`gate_gateway_event`** (new shared helper in `gateway.rs`): `gate_incoming` + `DenyIdentity` → throttled request-access echo / `DenyScope` → silent drop. Both `process_gateway_event` (unified) and the WS loop call it — single deny-handling implementation, no drift between modes. On the WS path it runs **before slash handling**, so untrusted senders cannot execute `/reset`/`/cancel`/`/config`.
2. **Registry seeding** (`main.rs`): the `[gateway]` section's trust values are inserted for its platform via `gateway_section_trust()` — same `resolve_allow_all` semantics the old inline filter used (behavior-preserving for the common case). Seeding order gives the documented precedence: `GATEWAY_*` env (uniform seed) < `[gateway]` < `[<platform>]` (`platform_trust_override`, which already runs in both modes).
3. **WS filter neutered to structural gating** (bot admission + @mention), exactly like the unified path since #1297.
4. **Dead code removed**: `GatewayParams`/`GatewayEventContext` channel/user fields (the unified ones were already unread; the WS ones moved to the registry) and the unified block's `GATEWAY_ALLOW_ALL_*` env reads (the registry's uniform seed already consumes those vars).
5. **Docs**: all six mode-scoping callouts (line/wecom/google-chat/msteams/config-reference ×2) rewritten — the sections now apply in **both** deployment modes with the precedence spelled out.

## ⚠️ Breaking Change

In standalone-gateway deployments that define **both** `[gateway]` allowlists **and** a first-class `[<platform>]` trust section with different values, the platform section now wins on the WS path (previously it was silently ignored there). This is the documented target state — config-first, most-specific section is authoritative. **Migration:** align the `[<platform>]` section values; they now govern that platform in both modes.

Known edge (documented, not changed): a standalone gateway forwarding events for platforms *other than* `[gateway].platform` — those events now resolve against the uniform `GATEWAY_*` env seed (or that platform's own section) instead of the `[gateway]` allowlists. Multi-platform standalone setups should use per-platform sections.

## Self-review (round 2 commit)

Adversarial pass over my own round-1 implementation caught a real bug:

- **WS-loop deadlock/DoS via deny-echo (fixed)**: round 1 awaited `send_message` inline inside the WS select arm. In streaming mode `send_gateway_reply` waits (5s timeout) on a `GatewayResponse` **dispatched by that same loop arm** — every `DenyIdentity` would stall all event processing for the full timeout, and distinct untrusted senders could chain stalls into a DoS. `gate_gateway_event` is now a sync decision helper returning `GateOutcome::{Allow, Deny{echo}}`; the WS path **spawns** the echo onto its JoinSet (matching the existing fire-and-forget pattern for slash responses), the unified path awaits directly (axum/bridge context, safe). Rationale documented at the enum and both call sites.
- Also verified: `[gateway]` registry seed runs **after** secret substitution (line 443 vs 347); registry insert/lookup both lowercase-normalize (no casing mismatch between `[gateway].platform` and event labels); the gate's spawned echo uses the same throttle (`echo_allowed`) as before, so amplification behavior is unchanged.

## Validation

- macmini: all-features core clippy, unified clippy, default clippy — all clean under `-D warnings`
- `cargo test -p openab-core --all-features` — 672 pass (1 pre-existing macOS-only failure: `secrets::tests::resolve_exec_nonzero_exit`)
- `cargo test --features unified --bin openab` — 18 pass (includes the 3 new tests)
- New tests: `gateway_section_trust_mirrors_ws_filter_semantics` (section → TrustConfig semantics), `gateway_trust_seed_precedence_env_lt_gateway_lt_platform` (pins the three-layer precedence), `ws_path_filter_is_structural_only` (unknown users/channels pass the structural filter; bot/@mention gating stays)
- rustfmt: zero new drift vs per-file baseline (repo has pre-existing drift)

Refs #1356 (Phase 1c prerequisite — unblocks #1357 double-gate elimination).

### DIFF
--- a/crates/openab-core/src/gateway.rs
+++ b/crates/openab-core/src/gateway.rs
@@ -889,9 +889,25 @@ pub async fn run_gateway_adapter(
                                     // identity) — same per-platform registry as
                                     // the unified path. Placed before slash
                                     // handling so untrusted senders cannot
-                                    // execute /reset|/cancel|/config.
-                                    if !gate_gateway_event(&router, adapter.as_ref(), &event).await {
-                                        continue;
+                                    // execute /reset|/cancel|/config. The echo
+                                    // is SPAWNED, never awaited here: streaming
+                                    // send_message waits for a GatewayResponse
+                                    // that only this loop can dispatch, so an
+                                    // inline await would stall all event
+                                    // processing for the reply timeout.
+                                    match gate_gateway_event(&router, &event) {
+                                        GateOutcome::Allow => {}
+                                        GateOutcome::Deny { echo } => {
+                                            if let Some((echo_channel, msg)) = echo {
+                                                let echo_adapter = adapter.clone();
+                                                tasks.spawn(async move {
+                                                    let _ = echo_adapter
+                                                        .send_message(&echo_channel, &msg)
+                                                        .await;
+                                                });
+                                            }
+                                            continue;
+                                        }
                                     }
 
                                     info!(
@@ -1259,30 +1275,39 @@ fn echo_allowed(key: &str) -> bool {
     }
 }
 
+/// Outcome of the shared ingress trust gate. `Deny { echo }` carries the
+/// throttled request-access echo payload (channel + message) when the deny is
+/// an identity deny and the per-sender throttle admits it; the CALLER decides
+/// how to deliver it. Delivery must NOT be awaited inline inside the WS event
+/// loop: `GatewayAdapter::send_message` (streaming mode) waits for a
+/// `GatewayResponse` that is dispatched by that same loop — awaiting there
+/// would stall all event processing for the reply timeout. The WS path spawns
+/// the echo; the unified path (axum/bridge task) awaits it directly.
+enum GateOutcome {
+    Allow,
+    Deny { echo: Option<(ChannelRef, String)> },
+}
+
 /// Shared ingress trust gate for gateway events — used by BOTH the standalone
 /// WebSocket path (`run_gateway_adapter`) and the unified path
 /// (`process_gateway_event`), so L2 (channel scope) and L3 (identity) are
 /// enforced by the same per-platform registry regardless of deployment mode
 /// (#1356 Phase 1c prerequisite).
 ///
-/// Returns `true` if the event may proceed. On `DenyIdentity`, sends the
-/// throttled request-access echo via `adapter`; on `DenyScope` (and any future
-/// variant), drops silently — scope is not a security boundary, so no echo.
+/// On `DenyIdentity`, returns the throttled request-access echo payload; on
+/// `DenyScope` (and any future variant), denies silently — scope is not a
+/// security boundary, so no echo.
 ///
 /// Phase 1: `is_dm = false` preserves today's behavior where gateway DMs are
 /// evaluated against the channel allowlist like any other channel (the
 /// `allow_dm` surface semantics arrive with the per-platform trust flip).
 /// TODO(phase-2): derive is_dm from the event/ChannelRef carrier so the
 /// `allow_dm` L2 surface can be enforced and tested for gateway platforms.
-async fn gate_gateway_event(
-    router: &crate::adapter::AdapterRouter,
-    adapter: &dyn ChatAdapter,
-    event: &GatewayEvent,
-) -> bool {
+fn gate_gateway_event(router: &crate::adapter::AdapterRouter, event: &GatewayEvent) -> GateOutcome {
     let decision =
         router.gate_incoming(&event.platform, &event.channel.id, false, &event.sender.id);
     match decision {
-        crate::trust::Decision::Allow => true,
+        crate::trust::Decision::Allow => GateOutcome::Allow,
         crate::trust::Decision::DenyIdentity => {
             // L3 identity deny → echo the sender their ID so they can request
             // access (throttled to avoid amplification). Bots never reach here
@@ -1294,7 +1319,7 @@ async fn gate_gateway_event(
                 "gateway event denied (identity); echoing request-access"
             );
             let throttle_key = format!("{}:{}", event.platform, event.sender.id);
-            if echo_allowed(&throttle_key) {
+            let echo = if echo_allowed(&throttle_key) {
                 let echo_channel = ChannelRef {
                     platform: event.platform.clone(),
                     channel_id: event.channel.id.clone(),
@@ -1302,13 +1327,15 @@ async fn gate_gateway_event(
                     parent_id: None,
                     origin_event_id: Some(event.event_id.clone()),
                 };
-                let echo = format!(
+                let msg = format!(
                     "⚠️ You are not on this bot's trusted list.\nYour ID: {}\nAsk the admin to add it to allowed_users.",
                     event.sender.id
                 );
-                let _ = adapter.send_message(&echo_channel, &echo).await;
-            }
-            false
+                Some((echo_channel, msg))
+            } else {
+                None
+            };
+            GateOutcome::Deny { echo }
         }
         // DenyScope (and any future variant) → silent drop (scope is not a
         // security boundary; no request-access echo).
@@ -1320,7 +1347,7 @@ async fn gate_gateway_event(
                 ?decision,
                 "gateway event denied (scope); silent"
             );
-            false
+            GateOutcome::Deny { echo: None }
         }
     }
 }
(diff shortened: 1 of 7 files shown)


chaodu-agent:
> [!NOTE]
> LGTM ✅ — Clean extraction unifying both ingress paths under the shared trust registry, with behavior-preserving semantics and good test coverage.

## What This PR Does

The standalone-gateway WebSocket path was the last ingress that bypassed the shared per-platform trust registry (#1356). This PR routes it through the same `gate_incoming` call as the unified path, eliminating drift between deployment modes and enabling the six "mode scoping" doc callouts to be rewritten as unconditional.

## How It Works

1. **`gate_gateway_event`** — new shared async helper in `gateway.rs` that wraps `router.gate_incoming()` with deny-handling (throttled echo on `DenyIdentity`, silent drop on `DenyScope`). Both the WS loop and `process_gateway_event` call it.
2. **Registry seeding** — `gateway_section_trust()` in `main.rs` converts the `[gateway]` section into a `TrustConfig` and inserts it into the registry with documented precedence: `GATEWAY_*` env < `[gateway]` < `[<platform>]`.
3. **WS filter neutered** — `should_skip_event` on the WS path now passes `allow_all_channels=true, allow_all_users=true` (structural gating only: bot filter + @mention). L2/L3 enforcement is delegated to the gate.
4. **Dead code removal** — channel/user fields stripped from `GatewayParams` and `GatewayEventContext`; inline `GATEWAY_ALLOW_ALL_*` env reads removed from the unified path.
5. **Docs** — five platform docs + config-reference rewritten from "Mode scoping" to "Trust resolution" with precedence spelled out.

## Findings

| 1 | 🟢 | Gate placement before slash handling prevents untrusted `/reset`/`/cancel`/`/config` execution — good security posture | `gateway.rs:885` |
| 2 | 🟢 | Three-layer precedence test (`env < [gateway] < [platform]`) pins the documented contract | `main.rs` tests |
| 3 | 🟢 | `ws_path_filter_is_structural_only` test locks the neutered-filter invariant so regressions are caught | `gateway.rs` tests |
| 4 | 🟢 | `is_dm = false` TODO is documented, scoped to Phase 2, and does not change behavior | `gateway.rs:1280` |

<details>

- **Single implementation, no drift**: both paths now call the same function — impossible for WS and unified to diverge on deny behavior.
- **Behavior-preserving for common case**: `resolve_allow_all` semantics are identical to the old inline filter; the breaking change only surfaces for the uncommon dual-section-with-conflicting-values scenario, which is the desired target state.
- **Gate before slash**: on the WS path the trust gate runs before the slash-command dispatcher, closing a privilege gap where an untrusted sender could have run `/config` if they passed the old structural filter.
- **Well-structured tests**: each new test targets a specific invariant (semantics mirror, precedence ordering, structural-only filter). No over-fitting to implementation details.
- **Doc consistency**: all six callouts updated atomically in the same commit — no "docs lie" window.

</details>

<details>

- PR opened: 2026-07-13
- Main already has: shared trust registry (`PlatformTrustConfigs`), `gate_incoming` on `AdapterRouter`, per-platform `[line]`/`[wecom]`/`[googlechat]`/`[teams]` trust sections, unified path gating
- Net-new value: standalone WS path now routes through the same registry (was the last bypass); `[gateway]` section seeded into registry; dead inline-filter code removed; docs updated to unconditional

</details>

1. **No production incidents from the gap** — the WS path's inline filter was functionally equivalent for single-platform standalone setups (the majority case). Unifying is architecturally correct but not urgency-driven.
2. **Breaking change with no auto-migration** — users with conflicting `[gateway]` + `[<platform>]` sections will silently change behavior on upgrade. A startup warning or config-check CLI subcommand could reduce surprise.
3. **Phase 2 dependency (`is_dm`) remains unresolved** — the gate hardcodes `is_dm = false`, meaning the `allow_dm` L2 surface is still unenforced for gateway platforms. This PR moves the architecture forward but does not close the full L2 story.

All three are acceptable for Phase 1c scope: the first two are acknowledged in the PR body with clear migration guidance, and the third is a documented TODO for the next phase.

chaodu-agent:
> [!NOTE]
> LGTM ✅ — Clean architecture consolidation that unifies trust enforcement across both deployment modes, with excellent self-review catching a real deadlock/DoS vector.

## What This PR Does

Routes the standalone-gateway WebSocket path through the same shared per-platform trust registry that the unified path already uses, eliminating the last ingress that bypassed it. This closes the gap where `[<platform>]` trust sections had no effect on standalone deployments.

## How It Works

1. **`gate_gateway_event`** — new sync helper returning `GateOutcome::{Allow, Deny{echo}}` that both WS and unified paths call, ensuring single deny-handling implementation with no drift.
2. **Registry seeding** — `gateway_section_trust()` seeds `[gateway]` values into the shared registry with documented precedence: `GATEWAY_*` env < `[gateway]` < `[<platform>]`.
3. **WS filter neutered** — `should_skip_event` now handles structural gating only (bot admission + @mention); L2/L3 delegated to `gate_gateway_event`.
4. **Dead code removed** — channel/user fields pruned from `GatewayParams`/`GatewayEventContext`; unified path env reads eliminated (registry already consumes them).
5. **WS-loop DoS fix** — deny-echo spawned (not awaited) in WS path, preventing deadlock from streaming `send_message` waiting on the same loop.

## Findings

| 1 | 🟢 | Excellent self-review catching WS-loop deadlock/DoS before merge | `gateway.rs:885-915` |
| 2 | 🟢 | Clean `GateOutcome` enum design — caller-decides-delivery pattern avoids tight coupling | `gateway.rs:1275-1280` |
| 3 | 🟢 | Three well-targeted tests pin the exact properties the refactor relies on | `gateway.rs:1687+`, `main.rs:1566+` |
| 4 | 🟢 | Docs updated consistently across all six platform pages | docs/* |
| 5 | 🟢 | Breaking change documented with clear migration path | PR description |

<details>

- **Deadlock prevention**: The round-2 self-review caught a genuine DoS vector — awaiting `send_message` inside the WS select arm would stall all event processing. The spawn-on-JoinSet fix is correct and matches the existing pattern for slash command responses.
- **Single source of truth**: Both paths now call the same `gate_gateway_event` → `gate_incoming` chain. No more drift risk between deployment modes.
- **Precedence correctness**: The seeding order in `main.rs` (env → `[gateway]` → `[<platform>]`) is enforced by insertion order with `reg.insert` overwriting, and pinned by a dedicated test.
- **Clean dead code removal**: 4 fields removed from `GatewayParams`, 4 from `GatewayEventContext`, and the 18-line unified-path env-reading block — all provably unreachable after the registry handles L2/L3.
- **`GateOutcome` API**: Making delivery the caller's responsibility is the right abstraction. The WS path spawns; the axum path awaits. Both are documented at the call site with rationale for why.
- **Test coverage**: All three new tests target real invariants rather than implementation details — they will break correctly if the contract changes.

</details>

<details>

- PR opened: 2026-07-13
- Main already has: shared trust registry (`PlatformTrustConfigs`), `gate_incoming`, per-platform sections, unified path using registry, WS path with inline allowlist filter
- Net-new value: WS path routed through the same registry; `[gateway]` section trust seeded into registry; dead inline filter code removed; six doc callouts updated; WS-loop deadlock vector eliminated

</details>

1. **Could wait for Phase 2** — The full trust flip (#1357) is the ultimate goal; this intermediate step adds a migration cliff for users with divergent `[gateway]`/`[<platform>]` configs. *Counter:* Phase 1c is a prerequisite for #1357 and the breaking change is the documented target state with clear migration guidance.

2. **Single-platform assumption** — `gateway_section_trust` seeds only `gw.platform`; multi-platform standalone gateways are an acknowledged edge case rather than first-class. *Counter:* Multi-platform standalone is rare and documented; per-platform sections already cover it.

3. **Spawned echo is fire-and-forget** — If the echo fails (WS disconnected mid-send), the denied user gets no feedback. *Counter:* This matches the existing slash-command pattern; the throttle already limits echo frequency; and a failed echo for an untrusted sender is an acceptable degradation.

All three are reasonably addressed — the PR is justified.

thepagent (approved):

PR #1393, opened 2026-07-13
fix(ci): harden release-pr workflow against expression injection

antigenius0910:
## Summary

Harden `.github/workflows/release-pr.yml` against **GitHub Actions expression injection** in the `Resolve version` step. The free-form `inputs.version` string was expanded into the `run:` bash script via `${{ inputs.version }}` before bash ever saw the script, so a crafted input such as `0.8.0"; malicious_command; echo "` would break the shell string context and execute arbitrary commands with the job's `contents: write` + `pull-requests: write` permissions and the generated App token.

Pass all workflow-context values (`inputs.*` and `steps.*.outputs.*`) through an `env:` block and reference them as `"$VAR"` inside the run scripts — the canonical GitHub-recommended pattern.

## Discussion

- Discord: https://discord.com/channels/1491295327620169908/1491365158868619404/1526331374158610495

## Threat model

- **Trigger**: `workflow_dispatch` → requires write access to the repo (mitigating), but still exposes insider-threat / compromised-credential paths.
- **Blast radius**: the job holds `contents: write`, `pull-requests: write`, plus an App token generated via `actions/create-github-app-token@v3`.
- **Exploit shape**: any `${{ inputs.version }}` expansion inside a `run:` block. `inputs.bump` (`type: choice`) is validated at dispatch time so it is not directly exploitable, but env-hardening it too keeps the pattern consistent and guards against future type changes.
- **Downstream flow**: `steps.version.outputs.version` and `.stable` are derived from the tainted input unchanged, so `Update version files` and `Create release PR` are also hardened.

## Files

| File | Change |
|------|--------|
| `.github/workflows/release-pr.yml` | Move every `${{ inputs.* }}` / `${{ steps.*.outputs.* }}` reference in a `run:` block into an `env:` block; use `"$VAR"` in shell |

## Test plan

- [x] `python3 -c "import yaml; yaml.safe_load(...)"` on the edited workflow — parses clean
- [x] Diff review: env vars added to `Resolve version` (INPUT_VERSION, INPUT_BUMP), `Update version files` (VERSION, STABLE), and `Create release PR` (VERSION alongside existing GH_TOKEN); all shell interpolations swapped to `"$VAR"`
- [ ] Maintainers: on next dispatch, verify the `Resolve version` step still resolves both auto-bump and explicit-version cases (behavior is unchanged — same variables, just passed through env)

## Reference

- Reported via OX Security (issue `019d8c18-6fe2-7bb9-8b99-73f3be56b6ca`)
[...]

PR #1395, opened 2026-07-13
fix(ci): eliminate script injection in pending-decision.yml

chaodu-agent:
## What

Fixes a script-injection vulnerability in `pending-decision.yml`, found during a follow-up audit after #1393.

The `github-script` step interpolated both `workflow_call` inputs directly into the JS source string:

```js
const prNumber = ${{ inputs.pr_number }};
const decision = '${{ inputs.decision }}';
```

`inputs.decision` is `type: string` with no `choice` constraint or runtime validation — only a comment documents the expected values (`"maintainer"` / `"contributor"`). A crafted `decision` value could break out of the single-quoted JS literal and execute arbitrary code inside the `github-script` sandbox, which runs with `pull-requests: write` + `issues: write`.

## Blast radius

This `workflow_call` reusable workflow currently has **no caller** in the repo — searched `.github/workflows/` for any `uses:` reference and found none. `pending-maintainer.yml` / `issue-pending-maintainer.yml` both implement the label-flip logic inline rather than calling this workflow. So there's no live exploit path today, but the footgun is real the moment something calls it with attacker-influenced input (e.g. a bot command parsed from a PR comment).

## Fix

Same pattern as the GitHub Actions security hardening guide (and the one already applied in #1393): pass values through `env:` and read them via `process.env` inside the script, never string-interpolated into the JS source.

- `PR_NUMBER` → `Number(process.env.PR_NUMBER)`
- `DECISION` → `process.env.DECISION`

`process.env` values are never parsed as code, so this closes the injection regardless of the input's declared type — not just for the current `string`/`number` types but for any future schema change too.

No behavior change: same label-flip logic, just a safe value-delivery mechanism.

## Verification

- YAML parses cleanly
- `actionlint` clean (`exit=0`)

### DIFF
--- a/.github/workflows/pending-decision.yml
+++ b/.github/workflows/pending-decision.yml
@@ -18,12 +18,15 @@ jobs:
       issues: write
     steps:
       - uses: actions/github-script@v7
+        env:
+          PR_NUMBER: ${{ inputs.pr_number }}
+          DECISION: ${{ inputs.decision }}
         with:
           script: |
             const MAINTAINER = 'pending-maintainer';
             const CONTRIBUTOR = 'pending-contributor';
-            const prNumber = ${{ inputs.pr_number }};
-            const decision = '${{ inputs.decision }}';
+            const prNumber = Number(process.env.PR_NUMBER);
+            const decision = process.env.DECISION;
 
             const addLabel = decision === 'maintainer' ? MAINTAINER : CONTRIBUTOR;
             const removeLabel = decision === 'maintainer' ? CONTRIBUTOR : MAINTAINER;


chaodu-agent:
> [!NOTE]
> **LGTM ✅** — Correct, minimal, idiomatic script-injection fix. No blocking findings.

## What This PR Does

`pending-decision.yml` was directly interpolating `${{ inputs.pr_number }}` and `${{ inputs.decision }}` into the JavaScript source string of a `github-script` step. A crafted `decision` value could escape the single-quoted literal and execute arbitrary JS inside the `github-script` sandbox (permissions: `pull-requests: write` + `issues: write`). This PR moves both values to `env:` and reads them via `process.env`, closing the injection regardless of input type or future schema changes.

## How It Works

```yaml
env:
  PR_NUMBER: ${{ inputs.pr_number }}
  DECISION: ${{ inputs.decision }}
```
```js
const prNumber = Number(process.env.PR_NUMBER);
const decision = process.env.DECISION;
```

`env:` values are delivered as data — never parsed as code. `Number(...)` preserves the correct type for the `type: number` input. Behavior is identical: same label-flip ternary, same API calls.

## Findings

| 1 | 🟢 | `env:` + `process.env` is the canonical pattern (same as #1393); fix is consistent with established repo convention | `.github/workflows/pending-decision.yml` |
| 2 | 🟢 | Honest blast-radius analysis in PR description — correctly identifies no live caller, no live exploit path today | PR description |
| 3 | 🟢 | Minimal scope: only the two interpolated values changed; no accidental behavior drift | diff |

<details>

### 🟢 F1: Canonical fix pattern
The `env:` → `process.env` pattern matches the fix already applied in #1393. Consistency across workflows is good defensive discipline — future reviewers and contributors will recognize the pattern immediately.

### 🟢 F2: Clear blast-radius documentation
The PR body accurately identifies that `pending-decision.yml` has no callers on `main` today, so there is no live exploit path. It also calls out the precise risk scenario (bot command parsed from a PR comment) that would activate the vulnerability. This kind of honest threat modeling is exactly what good security PRs should include.

### 🟢 F3: Zero behavior drift
Only the value-delivery mechanism changed. The label-flip logic, ternary conditions, API calls, and console output are untouched. `Number(process.env.PR_NUMBER)` correctly coerces the env string back to the number type that the rest of the script expects.

</details>

<details>

- PR opened against: `main`
- Main already has: `pending-decision.yml` with the vulnerable inline interpolation (`${{ inputs.pr_number }}` / `${{ inputs.decision }}` direct in JS source)
- Net-new value: safe value-delivery via `env:` + `process.env`; closes injection vector proactively before any caller is wired up

</details>

<details>
<summary>5️⃣ Three Reasons We Might Not Need This PR</summary>

1. **No caller, no risk today** — `pending-decision.yml` has zero `uses:` references in the repo. The vulnerability cannot currently be triggered. Merging adds complexity for a threat that does not exist yet.
2. **Delete instead of fix** — If this workflow has no callers and no planned callers, the lower-maintenance option is to delete the file entirely rather than maintain a hardened but unused workflow.
3. **`type: number` input partially mitigates `pr_number`** — `inputs.pr_number` is declared as `type: number`, which limits its injection surface to numeric values. The original code (`const prNumber = ${{ inputs.pr_number }};`) was arguably already safe for that specific field. The real risk was only `inputs.decision`. A narrower fix scoped to `decision` alone would have been equally correct.

> These are not blockers — they are decision checkpoints. The fix is correct and the approach is sound.

</details>

thepagent (approved):

PR #1398, opened 2026-07-14
fix(discord): truncate select menu option text to Discord's 100-char cap

chaodu-agent:
## Problem

`/models` on B0 fails with Discord's "The application did not respond" — the interaction response is rejected before it ever reaches the user.

B0's logs show the actual error:

```
ERROR openab_core::discord: failed to respond to slash command
error=Invalid Form Body (data.components.0.components.0.options.12.description:
Must be 100 or fewer in length.) category="model"
```

One over-long option invalidates the entire form body, so the whole menu fails — not just that option. Any backend can trigger this at any time by shipping a long model/agent description, so this is a defensive gap on our side rather than a data problem.

## Fix

- `truncate_for_discord(s, max)`: character-based (not byte-based — slicing on a byte boundary would panic on multi-byte UTF-8) truncation with a trailing `…`
- Applied to both option **label** and **description** in `build_config_select_menu`, capped at `SELECT_OPTION_TEXT_MAX = 100`

Pagination already handles the 25-option limit (`SELECT_MENU_PAGE_SIZE`), so no change needed there even though the model list has grown past 25.

## Tests

- exact in-the-wild case: the ~140-char `claude-fable-5` description truncates to exactly 100 chars ending in `…`
- boundary: 100-char string passes through unchanged
- CJK safety: 120 multi-byte chars truncate by character count without panicking

## Verification

- `cargo clippy -p openab-core --all-features` — clean
- `cargo test -p openab-core --all-features truncate_for_discord` — 4/4 pass
- rustfmt isolated per-file check vs origin/main baseline — no new drift (884 = 884 lines, all pre-existing)

### DIFF
--- a/crates/openab-core/src/discord.rs
+++ b/crates/openab-core/src/discord.rs
@@ -37,6 +37,26 @@ const PARTICIPATION_CACHE_MAX: usize = 1000;
 /// Discord StringSelectMenu hard limit on options.
 const SELECT_MENU_PAGE_SIZE: usize = 25;
 
+/// Discord caps select menu option labels and descriptions at 100
+/// characters; anything longer makes the entire interaction response fail
+/// with "Invalid Form Body", which surfaces to users as "The application
+/// did not respond". (Hit in the wild when a backend model description
+/// exceeded the cap.)
+const SELECT_OPTION_TEXT_MAX: usize = 100;
+
+/// Truncate to at most `max` characters (not bytes — Discord counts
+/// characters, and slicing on a byte boundary would panic on multi-byte
+/// UTF-8). Appends '…' when truncated.
+fn truncate_for_discord(s: &str, max: usize) -> String {
+    if s.chars().count() <= max {
+        s.to_string()
+    } else {
+        let mut out: String = s.chars().take(max.saturating_sub(1)).collect();
+        out.push('…');
+        out
+    }
+}
+
 /// Avoid unbounded Discord history exports from very large threads.
 const THREAD_EXPORT_MESSAGE_LIMIT: usize = 5000;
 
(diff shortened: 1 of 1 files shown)


chaodu-agent:
> [!NOTE]
> LGTM ✅ — Well-scoped defensive fix for Discord's 100-char select menu option limit.

## What This PR Does

The `/models` slash command fails with "The application did not respond" when any backend model has a description exceeding Discord's 100-character limit for select menu option text. This PR adds character-based truncation (with trailing `…`) to both option labels and descriptions, preventing the entire interaction from being rejected.

## How It Works

- `truncate_for_discord(s, max)` counts Unicode characters (not bytes — slicing on a byte boundary would panic on multi-byte UTF-8), and appends `…` when truncating.
- Applied at the single `CreateSelectMenuOption` call site in `build_config_select_menu`.
- Uses `saturating_sub(1)` to leave room for the ellipsis character within the 100-char budget.

## Findings

| 1 | 🟢 | Correct character-based (not byte-based) truncation prevents panic on multi-byte boundaries | `discord.rs:48-54` |
| 2 | 🟢 | Defensive at the right layer — applied at the view boundary, not data layer | `discord.rs:1557-1562` |
| 3 | 🟢 | Good test coverage: boundary, over-limit with real data, and CJK/multibyte safety | `discord.rs:3273-3307` |

<details>

### 🟢 F1: Character-based truncation is correct
Discord counts characters, not bytes. Using `.chars().count()` and `.chars().take(n)` correctly handles multi-byte UTF-8 sequences (CJK, emoji, etc.) without risking panics from slicing at byte boundaries.

### 🟢 F2: Truncation at the view boundary
Truncation is applied at the select menu construction site rather than at data ingestion. This is the right choice — the raw data is preserved for other uses (logging, API responses), and the Discord-specific formatting constraint is handled at the Discord-specific layer.

### 🟢 F3: Comprehensive test coverage
Tests cover: short strings (pass-through), exact boundary (100 chars unchanged), real-world over-limit case (the actual `claude-fable-5` description), and CJK safety (120 multi-byte characters). All four verify char count, not byte length.

</details>

<details>

- PR opened: 2026-07-14
- Main already has: Truncation logic for channel titles (100-char) and message content (2000-char UTF-16), but **no** truncation for select menu options
- Net-new value: Defensive cap on select menu option label/description text, preventing the entire interaction from failing when any backend ships a long model description

</details>

<details>

- Single-responsibility: one function, one constant, one call site
- Tests use real-world data that triggered the bug
- Correct handling of the `…` character (1 char, 3 bytes) within the budget
- PR description clearly documents root cause, fix approach, and verification steps

</details>

1. **Fix the data source instead** — Could cap model descriptions at the backend/registry level before they reach Discord. However, this is fragile (any new backend can violate it) and the defensive layer here is the right approach.
2. **Discord might increase the limit** — Possible, but the 100-char limit has been stable for years and the fix is a no-op when strings are within bounds.
3. **Truncation loses information** — Users see `…` instead of the full description. However, the alternative is a complete command failure with no response at all, which is strictly worse.

All three are easily countered — this PR is clearly necessary.

thepagent (approved):

PR #1411, opened 2026-07-16
ci(pr-review): add repository_dispatch trigger for targeted PR review

chaodu-agent:
## Summary

The `pull_request` event path of `pr-bot-review.yml` is intentionally skipped for fork PRs (read-only `GITHUB_TOKEN` → status write 403s), leaving fork PRs dependent on the `*/5` cron schedule. GitHub Actions cron is best-effort — runs can be delayed or dropped entirely under load — so fork PR reviews have no reliable trigger.

This adds a `repository_dispatch` (`event_type: pr-review`) trigger for immediate, targeted review of a single PR:

```bash
gh api repos/openabdev/openab/dispatches \
  -f event_type=pr-review \
  -F 'client_payload[pr_number]=<N>'
```

## How it works

- **Full-permission token**: `repository_dispatch` runs with a normal repo-scoped `GITHUB_TOKEN`, so status writes work for fork PRs
- **Trusted sender**: firing `repository_dispatch` requires repo write access, so the `author_association` / `safe-to-review` gate is bypassed on this path only
- **Force semantics**: an explicit dispatch bypasses the already-reviewed-on-HEAD skip (a maintainer asking again means re-review); the pending-recent guard (<30 min) and the 30-cycle circuit breaker still apply
- **Eligibility still enforced**: closed, draft, or `review-limit-reached` PRs are refused with a log message
- **Concurrency isolation**: dispatches use a per-PR concurrency group (`pr-review-poller-<N>`) so they neither cancel nor get cancelled by the scheduled poller
- **Payload validation**: non-numeric / missing `pr_number` fails fast with a clear error
- Poll/cron/workflow_dispatch paths are byte-for-byte unchanged in behavior

## Testing

- YAML validated (`yaml.safe_load`)
- Both embedded shell scripts pass `bash -n`
- End-to-end dispatch can only be verified after merge (`repository_dispatch` executes the workflow from the default branch) — will verify with a live dispatch against an open PR post-merge

## Notes

The cron schedule stays as a catch-all backstop; this does not replace it.

### DIFF
--- a/.github/workflows/pr-bot-review.yml
+++ b/.github/workflows/pr-bot-review.yml
@@ -6,9 +6,19 @@ on:
   schedule:
     - cron: '*/5 * * * *'
   workflow_dispatch: {}
+  # Targeted review trigger: a maintainer (or maintainer agent) fires
+  #   gh api repos/<repo>/dispatches -f event_type=pr-review \
+  #     -F 'client_payload[pr_number]=<N>'
+  # repository_dispatch requires repo write access, so the sender is trusted.
+  # Unlike the cron path this is immediate and reliable, and unlike the
+  # pull_request path it carries a full-permission GITHUB_TOKEN for fork PRs.
+  repository_dispatch:
+    types: [pr-review]
 
 concurrency:
-  group: pr-review-poller
+  # Targeted dispatches get their own group per PR so they neither cancel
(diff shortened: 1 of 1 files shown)


chaodu-agent:
Superseded — going with an issue_comment /review trigger instead of repository_dispatch.

PR #1412, opened 2026-07-16
ci(pr-review): add /review comment command for targeted PR review

chaodu-agent:
## Summary

The `pull_request` event path of `pr-bot-review.yml` is intentionally skipped for fork PRs (read-only `GITHUB_TOKEN` → status write 403s), leaving fork PRs dependent on the `*/5` cron schedule. GitHub Actions cron is best-effort — runs can be delayed or dropped — so fork PR reviews have no reliable trigger.

This adds a **`/review` slash command**: a maintainer comments `/review` on any PR and the review pipeline kicks off immediately.

## How it works

- **`issue_comment` event**: always runs in the base repo with a full-permission `GITHUB_TOKEN`, so fork PRs work (unlike the `pull_request` path)
- **Job-level gate**: PR comments only (`github.event.issue.pull_request`), body starts with `/review`, author_association ∈ OWNER/MEMBER/COLLABORATOR — non-matching comments never start the job
- **Ack**: reacts 👀 on the triggering comment so the requester knows it was picked up
- **Force semantics**: an explicit `/review` bypasses the already-reviewed-on-HEAD skip (asking again = re-review); the <30-min pending guard and 30-cycle circuit breaker still apply
- **Eligibility still enforced**: closed, draft, or `review-limit-reached` PRs are refused with a log message
- **Concurrency isolation**: commands use a per-PR concurrency group (`pr-review-poller-<N>`) so they neither cancel nor get cancelled by the scheduled poller
- **Injection-safe**: the comment body is only evaluated in the workflow `if` expression, never interpolated into shell; the script uses the numeric issue number only
- Poll/cron/workflow_dispatch paths are unchanged in behavior

## Testing

- YAML validated (`yaml.safe_load`); both embedded shell scripts pass `bash -n`
- End-to-end can only be verified after merge (`issue_comment` executes the workflow from the default branch) — will verify with a live `/review` comment on an open PR post-merge

## Notes

Supersedes #1411 (repository_dispatch approach) — a comment command needs no CLI and works from the GitHub UI. The cron schedule stays as a catch-all backstop.

### DIFF
--- a/.github/workflows/pr-bot-review.yml
+++ b/.github/workflows/pr-bot-review.yml
@@ -6,9 +6,17 @@ on:
   schedule:
     - cron: '*/5 * * * *'
   workflow_dispatch: {}
+  # Slash command: a maintainer comments "/review" on a PR to trigger an
+  # immediate targeted review. issue_comment always runs in the base repo
+  # with a full-permission GITHUB_TOKEN, so fork PRs work too (unlike the
+  # pull_request path) and there is no dependency on the best-effort cron.
+  issue_comment:
+    types: [created]
 
 concurrency:
-  group: pr-review-poller
+  # /review commands get their own group per PR so they neither cancel
+  # nor get cancelled by the scheduled poller (or other PRs' commands).
+  group: pr-review-poller-${{ github.event_name == 'issue_comment' && github.event.issue.number || 'poll' }}
   cancel-in-progress: true
 
 permissions:
@@ -24,9 +32,17 @@ jobs:
     # the commit status 403s and the run fails. Skip the event-driven path for
     # fork PRs — the scheduled cron (full permissions) still reviews them.
     # schedule/workflow_dispatch always run; same-repo PRs run on the event.
+    # issue_comment runs only for "/review" comments on PRs from maintainers
+    # (OWNER/MEMBER/COLLABORATOR) — comment body is never interpolated into
+    # shell, only the numeric issue number is used.
     if: >-
-      github.event_name != 'pull_request' ||
-      github.event.pull_request.head.repo.full_name == github.repository
+      (github.event_name == 'issue_comment' &&
+       github.event.issue.pull_request &&
+       startsWith(github.event.comment.body, '/review') &&
+       contains(fromJSON('["OWNER","MEMBER","COLLABORATOR"]'), github.event.comment.author_association)) ||
+      (github.event_name != 'issue_comment' &&
+       (github.event_name != 'pull_request' ||
+        github.event.pull_request.head.repo.full_name == github.repository))
     steps:
       - name: Find PRs needing review
         id: poll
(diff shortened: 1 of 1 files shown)


thepagent (approved):

chaodu-agent:
> [!IMPORTANT]
> **CHANGES REQUESTED â ï¸ â Workflow-level concurrency lets any comment on a PR cancel a valid `/review` run, which can leave the required commit status stuck at `pending` for up to 30 minutes.**

## What This PR Does

This PR adds an immediate `/review` comment command so maintainers can trigger the review pipeline for a specific PR, including fork PRs that cannot use the write-enabled `pull_request` path.

## How It Works

The workflow subscribes to `issue_comment.created`, gates the job by PR/comment/author association, acknowledges accepted commands with an eyes reaction, fetches the targeted PR, bypasses an already-reviewed status for explicit requests, and dispatches the existing Discord webhook. It also assigns issue-comment runs a per-PR concurrency group.

### @thepagent

> Approved without written findings.

â¹ï¸ **Additional review finding:** The PR was merged while this review was in progress. The successful syntax/CI checks do not exercise the interaction between workflow-level concurrency and the job-level command gate; the blocking cancellation path below therefore still requires a follow-up fix on `main`.

## Review Coverage

| Reviewer | Focus | Outcome |
|---|---|---|
| Reviewer A | Correctness and race conditions | Changes requested: cancellation and duplicate-dispatch races |
| Reviewer B | Security and trust boundaries | Changes requested: ordinary comments can cancel privileged command runs |
| Reviewer C | GitHub Actions and operability | Changes requested: workflow-level gating mismatch and command parsing |

## Findings

| 1 | ð´ Critical | Every `issue_comment` run enters the per-PR workflow concurrency group before the job gate is evaluated, so any ordinary comment can cancel an active valid `/review` run. | `.github/workflows/pr-bot-review.yml:16-20` |
| 2 | ð¡ Important | Command runs and poller runs use different concurrency groups, allowing both to observe the same pre-pending state and dispatch duplicate reviews. | `.github/workflows/pr-bot-review.yml:19`, `.github/workflows/pr-bot-review.yml:93-168` |
| 3 | ð¡ Important | `startsWith(..., '/review')` also accepts unintended commands such as `/reviewer` and `/reviews`. | `.github/workflows/pr-bot-review.yml:41` |
| 4 | ð¢ Praise | The implementation keeps the untrusted comment body out of shell, uses a numeric issue number, enforces PR eligibility, and preserves the stale-pending/circuit-breaker protections. | `.github/workflows/pr-bot-review.yml:38-84` |

<details>

### ð´ F1: Non-command comments can cancel an accepted review command

`concurrency` is defined at workflow scope, while the `/review`, PR, and author-association checks are defined in `jobs.poll-and-review.if`. GitHub creates a workflow run for every new issue comment and assigns it to `pr-review-poller-<PR number>` before deciding that the job should be skipped. GitHub's documented behavior for `cancel-in-progress: true` is to cancel any running workflow in the same group.

A concrete failure sequence is:

1. A trusted `/review` run passes the job gate.
2. It writes `OpenAB PR Review = pending` at lines 163-168.
3. Any user posts an ordinary comment on that PR.
4. The new workflow run cancels the valid run, then skips its own job because the comment is not `/review`.
5. No webhook is delivered and no error status is written; the commit remains pending until the 30-minute stale timeout.

Move concurrency below the command gate and avoid canceling an accepted run after it has claimed the status. For example, use job-level concurrency with queuing, or give non-matching events a unique `github.run_id` fallback and add cancellation cleanup. The fix must also cover repeated valid commands, which can trigger the same pending-status failure.

### ð¡ F2: Poll and command paths do not share a serialization boundary

Scheduled, dispatch, and pull-request runs use `pr-review-poller-poll`, while a command uses `pr-review-poller-<PR>`. A poll run and a command run can therefore both read the previous status before either writes `pending`, add the same PR to `TARGETS`, and each invoke the webhook. This can produce duplicate public reviews and, for `auto-fix`, conflicting side effects.

Serialize dispatch claims across both paths or perform a shared atomic claim immediately before webhook delivery. A single queued dispatcher group is the simplest safe option; a matrix/per-PR design would preserve more parallelism.

### ð¡ F3: The command boundary is too broad

The documented command is `/review`, but a prefix check also treats `/reviewer`, `/reviews`, and `/review-anything` as commands. Use exact matching if arguments are not supported, or explicitly accept only end-of-input/whitespace after `/review` and cover the grammar with tests.

### ð¢ F4: Trust-boundary handling is otherwise careful

The comment body is evaluated only by the Actions expression engine and is never interpolated into shell. The targeted path uses GitHub-provided numeric IDs, filters closed/draft/limited PRs, and retains existing review-loop protections.

</details>

<details>

- **PR opened:** 2026-07-16 02:10:50 UTC, against `main` at `9940337c18311e15bec55477fabd59ededde7b70`.
- **Opening main already had:** scheduled polling plus same-repository `pull_request` triggering, pending-status recovery, and the circuit breaker; it did not have an `issue_comment` `/review` path.
- **Net-new value:** immediate targeted review triggering for fork and same-repository PRs, with acknowledgement and force-re-review semantics.
- **Current state:** the PR was merged during this review, so `main` now contains the change and needs a follow-up correction for F1-F3.
- **Validation:** one workflow file changed (`+54/-16`); both embedded shell scripts pass `bash -n`; `git diff --check` passes; the GitHub checks and the changed workflow's pull-request run completed successfully. These checks do not exercise `issue_comment` concurrency behavior.

</details>

<details>
<summary>What's Good (ð¢)</summary>

- Solves a real reliability gap for fork PRs without evaluating contributor code in a privileged context.
- Reuses the existing eligibility, status, webhook, auto-fix, and circuit-breaker paths rather than creating a second review implementation.
- Provides immediate acknowledgement to the requester and keeps the comment body out of shell.

</details>

### 5ï¸â£ Three Reasons We Might Not Need This PR

1. **The cron path already provides eventual coverage** â if occasional delay is acceptable, a privileged comment-trigger path may add more state-machine complexity than value.
2. **A global dispatcher is easier to reason about** â mixing a batch poller with independently keyed command runs introduces cancellation and duplicate-claim races unless dispatch is redesigned.
3. **`workflow_dispatch` can cover rare urgent cases** â a manual targeted input could be safer operationally until command authorization, parsing, and concurrency have dedicated regression tests.

PR #1404, opened 2026-07-15
refactor(oabctl): split operator into library + thin CLI binary

chaodu-agent:
## Summary

Split `operator/` into a library crate plus a thin `oabctl` binary so downstream control planes can reuse typed manifest validation and ECS apply reconciliation in-process.

- Exposes a deliberately narrow public facade: manifest models plus `apply_manifests`
- Adds explicit `ApplyOptions` for the ECS cluster, optional control-plane bucket, and stabilization wait behavior
- Returns structured `ApplyReport` / `AppliedService` results with actions, webhook URLs, and best-effort warnings
- Returns structured `ApplyError` values with failure phase, failed service identity, completed batch results, and a preserved error source chain
- Keeps CLI/resource-management modules private; the binary delegates through the library entry point without widening the supported programmatic lifecycle
- Uses task-local progress suppression so concurrent library calls do not mutate process-global output behavior

## Motivation

Project NEST provisions one OAB service per tenant. Its manager currently shells out to `oabctl apply`; this facade lets it construct `OABServiceManifest` values and receive typed apply results without YAML templating, subprocess management, or stdout parsing.

This PR intentionally exposes **apply only**. Delete, scale, bootstrap, and other resource-management commands remain CLI implementation details until their programmatic contracts are designed independently.

## Behavior and IAM

Both CLI and programmatic apply validate the configured ECS cluster before mutation. The caller requires `ecs:DescribeClusters`; AWS does not support resource-level permissions for this action, so the IAM statement uses `Resource: "*"` even though the request targets only the configured cluster.

Programmatic apply does not read `~/.oabctl/config.toml`. The control-plane bucket is supplied through `ApplyOptions`, `OAB_CONTROL_PLANE_BUCKET`, or account-based resolution.

## Review fixes

- Reject empty manifest sets and implicit/blank cluster targets before mutation
- Propagate `DescribeServices` failures instead of treating them as an absent service
- Preserve immediate `anyhow` context through `ApplyError::source()`
- Return ingress cleanup, duplicate-resource, and configuration-mismatch diagnostics through `AppliedService.warnings`
- Document the `ecs:DescribeClusters` caller permission in crate and user documentation
- Make delete cleanup resumable when ECS is already `DRAINING`/`INACTIVE`, attempt independent S3 cleanup steps, and aggregate retry-safe failures

## Validation

The `operator` CI job runs:

- `cargo check`
- `cargo clippy -- -D warnings`
- `cargo test`
- `cargo build --release`

[...]

PR #1415, opened 2026-07-16
feat(oabctl): programmatic delete API with explicit target contract

chaodu-agent:
#1415 feat(oabctl): programmatic delete API with explicit target contract
State: open | Author: chaodu-agent | Branch: refactor/oabctl-delete-api

## Summary

Designs and implements the **programmatic delete contract** that the #1404 review intentionally deferred (&#34;until their programmatic contracts are designed independently&#34;). This PR is stacked on `refactor/oabctl-lib`.

```rust
let report = oabctl::delete_services(
    &amp;aws,
    &amp;[DeleteTarget::new(&#34;prod&#34;, &#34;nest-my-oab&#34;)],
    &amp;DeleteOptions::new(&#34;oab&#34;),
).await?;
```

The API is symmetric with `apply_manifests`:

- `DeleteOptions` requires an explicit cluster; the library path never reads `~/.oabctl/config.toml`.
- Bucket resolution shares apply&#39;s chain: explicit option ÃÂ¢ÃÂÃÂ `OAB_CONTROL_PLANE_BUCKET` ÃÂ¢ÃÂÃÂ `oab-control-plane-{account}`.
- Contract guards run before AWS calls, including rejection of delimiter-ambiguous namespaces.
- `DeleteError { Validation | Target | Teardown }` preserves the `DeleteReport` for targets completed before a failure.
- Exact AWS identities are checkpointed before mutation so teardown failures can be retried without name-only cleanup.
- Programmatic calls suppress human-readable progress while the CLI retains its existing rendering.

## Review Contract

Contract revision: **RC-1**

### Goal

Provide a public, in-process Rust API that safely deletes one or more explicitly identified OAB ECS deployments without reading CLI home configuration. A successful result must mean that the checkpointed ECS incarnation and its exact ingress dependencies are gone and the target&#39;s control-plane manifest/artifacts are removed. Ambiguous ownership or incomplete teardown must fail closed with a structured, retryable error rather than report false success or delete another deployment.

### Non-goals

- Providing linearizable coordination between concurrent `apply_manifests` and `delete_services` calls for the same logical target; callers must serialize same-target mutations as described under Accepted Residual Risks.
- Performing name-only orphan cleanup when neither a live ECS service nor a matching durable delete checkpoint can establish ownership.
- Removing platform-side webhooks, Telegram disclosure state, Secrets Manager secrets, shared bootstrap infrastructure, logs, or unrelated task-definition revisions.
- Redesigning the stacked `apply_manifests` API, changing existing CLI UX, or introducing a general distributed workflow engine.
- Parallelizing multi-target deletion or continuing later targets after the first target failure.

### Accepted Residual Risks

- **Same-target apply/delete concurrency is unsupported in RC-1.** The caller must serialize mutations for the same `(AWS account, Region, control-plane bucket, cluster, namespace, name)` tuple. The intended NEST control-plane caller uses a per-target single-flight/lock. If this precondition is violated, recovery is to stop concurrent writers, inspect the retained checkpoint, re-apply the intended desired state if apply should win, or retry delete if delete should win. A shared generation fence remains a follow-up.
- **AWS control-plane eventual consistency can require retries.** Ambiguous or not-yet-converged ECS responses fail closed, dependent cleanup does not start, and the durable checkpoint remains for retry.
- **An already-orphaned deployment without a checkpoint is not auto-cleaned.** This can leave resources behind, but avoids destructive name-only guessing; recovery is explicit operator inspection/remediation.
- **Batch deletion is sequential and stops at the first failed target.** The structured error retains prior completed targets and identifies the failed target, so the caller can retry deterministically.

### Acceptance Criteria

1. The crate publicly exports `delete_services`, `DeleteTarget`, `DeleteOptions`, `DeletedService`, `DeleteReport`, `DeleteError`, and `DeleteErrorKind` without changing the existing CLI command surface.
2. Before any AWS call, the API rejects an empty target set, a blank cluster, blank namespace/name components, and any logical target that cannot map unambiguously to one physical delete identity. Distinct accepted `(namespace, name)` pairs must not resolve to the same ECS/API/Cloud Map identity.
3. The library requires an explicit cluster, never reads `~/.oabctl/config.toml`, and resolves the control-plane bucket using the same explicit option ÃÂ¢ÃÂÃÂ environment ÃÂ¢ÃÂÃÂ caller-account chain as apply.
4. Before mutating ECS, delete persists a checkpoint bound to the exact target and caller boundary: namespace/name, bucket, partition/account/Region, requested and canonical cluster identity, ECS service ARN and `createdAt` incarnation, and exact registry/API identifiers when ingress exists.
5. Initial missing, empty-success, mixed-failure, duplicate, or otherwise ambiguous AWS identity responses fail closed. Retries may continue only from a matching checkpoint, and dependent ingress/S3 cleanup must not begin until the checkpointed ECS incarnation is unambiguously absent.
6. Delete has no first-name-match or name-only fallback for ECS, Cloud Map, or API Gateway. It must not remove resources belonging to a different cluster, account, Region, logical target, or recreated ECS incarnation.
7. A successful `DeleteReport` contains only fully completed targets. `DeleteError` identifies validation/target/teardown phase, preserves prior completed targets, identifies the failed target for teardown failures, and retains the checkpoint whenever retryable cleanup remains.
8. Programmatic apply/delete calls emit no human-readable progress to process-global stdout/stderr; existing CLI output and aggregate CLI failure behavior remain unchanged.
9. Public Rustdocs or user documentation explicitly state the same-target serialization precondition and the no-orphan-cleanup behavior so callers can satisfy the accepted-risk mitigations.
10. Focused tests cover pre-AWS validation, logical-identity collision prevention, checkpoint matching and fail-closed AWS response classification, incarnation mismatch, drain timeout, partial-report preservation, and retry behavior. `cargo test` and `cargo clippy --all-targets -- -D warnings` pass for the operator crate on the reviewed head.

### Follow-ups

- Add a shared target generation/lease fence plus conditional S3 writes so apply and delete can safely arbitrate concurrent same-target mutations instead of requiring caller serialization.
- Introduce a durable opaque deployment ID or reversible canonical resource-name encoding across apply/delete, beyond the minimum collision prevention required by RC-1.
- Design an explicit, auditable orphan-recovery API that can discover and remove resources without weakening the default fail-closed path.
- Add optional bounded parallelism and a collect-all-results batch mode if control-plane scale requires it.
- Decide separately whether secret deletion, webhook removal, disclosure cleanup, and broader tenant deprovisioning belong in this crate or in the control plane.

## CLI

No intended behavior change: `oabctl delete ` and `oabctl delete -f ` retain config-file cluster resolution, printed progress, and aggregate failure reporting.

## Motivation

NEST&#39;s manager needs `/delete` for tenant deprovisioning. Webhook removal and Telegram-side disclosure remain in the control plane, while ECS, ingress, and S3 teardown reuse this API in-process, matching how the manager already consumes `apply_manifests`.

## Validation Evidence

- Follow-up implementation head: `18322e16f2b63e5484b800507fe23451bbac1751`.
- RC-1 fixes reject delimiter-ambiguous delete namespaces before AWS, preserve partial reports, and fail closed before destructive mutation when an apply ingress-teardown checkpoint is pending.
- Prior exact-head `Review Contract` validation and `operator` CI passed before this follow-up; the current head is tracked below.
- Author-reported real-AWS E2E on the prior exact-identity head covered initial fail-closed drain behavior, checkpointed retry to full teardown, already-gone refusal without a checkpoint, and post-success checkpoint removal.
- Focused re-review of `1bd61e8..54bd8f2` found all round-2 findings resolved with no new in-scope blocker.

## Follow-up validation

- Follow-up head `18322e16f2b63e5484b800507fe23451bbac1751` adds duplicate-target validation, a bounded Telegram webhook client timeout, alias-equivalent physical-identity serialization guidance, bounded ownership probes, CLI cleanup warnings, and S3 path redaction.
- B4 follow-up preserves CLI best-effort behavior after drain timeout, continues S3 cleanup after individual failures, and treats an absent exact apply service as detached; the strict API remains fail-closed.
- B7-B2-B16 follow-up enforces the documented identity charset (with a legacy CLI namespace-hyphen exception), aborts ownership probes before admitting replacements after the first error, and treats ECS INACTIVE apply responses as absent.
- B14 fixes resumable DRAINING ECS deletion by reissuing `delete_service(force=true)` before polling, extends Cloud Map deletion to 12 attempts with capped 5/10/20/30-second backoff, and documents that manifest-first apply failures are recovered by re-running apply because no apply rollback checkpoint exists.
- Local `git diff --check`, Rustfmt parser checks, and focused B14 static assertions passed. Local `cargo check` was attempted but is blocked by the environment (`linker \`cc\` not found`); GitHub CI for this exact head is the authoritative Rust validation.

- This head also bounds identity ownership probes to 8 concurrent S3 requests and adds a 201-character/100-alias regression test; GitHub CI for this exact head completed with `changes`, `check`, `validate`, and `poll-and-review` successful, but `operator` failed at `cargo check` (job `89065627641`).

### DIFF
--- a/operator/src/delete.rs
+++ b/operator/src/delete.rs
@@ -262,6 +262,40 @@ fn ecs_delete_phase(status: Option<&str>) -> Result<EcsDeletePhase> {
     }
 }
 
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+enum DrainPollAction {
+    Complete,
+    Retry,
+    TimedOut,
+}
+
+fn drain_poll_action(is_gone: bool, attempt: u32, max_attempts: u32) -> DrainPollAction {
+    debug_assert!(max_attempts > 0);
+    debug_assert!(attempt < max_attempts);
+    if is_gone {
+        DrainPollAction::Complete
+    } else if attempt + 1 == max_attempts {
+        DrainPollAction::TimedOut
+    } else {
+        DrainPollAction::Retry
+    }
+}
+
+fn collect_best_effort_warnings(
+    warnings: &mut Vec<String>,
+    result: Result<Vec<String>>,
+    failure_context: &str,
+) {
+    match result {
+        Ok(step_warnings) => warnings.extend(step_warnings),
+        Err(error) => {
+            let warning = format!("{failure_context}: {error}");
+            eprintln!("  ⚠ {warning}");
+            warnings.push(warning);
+        }
+    }
+}
+
 /// Delete every OABService defined in a manifest file or directory.
 pub(crate) async fn run_from_file(
     aws_config: &aws_config::SdkConfig,
@@ -422,36 +456,39 @@ async fn run_with_bucket(
                     false
                 }
             };
-            if is_gone {
-                if attempt == 0 {
-                    eprintln!(" done (immediate)");
-                } else {
-                    let elapsed = u64::from(attempt) * DRAIN_POLL_INTERVAL.as_secs();
-                    eprintln!(" done ({elapsed}s)");
+            match drain_poll_action(is_gone, attempt, DRAIN_POLL_ATTEMPTS) {
+                DrainPollAction::Complete => {
+                    if attempt == 0 {
+                        eprintln!(" done (immediate)");
+                    } else {
+                        let elapsed = u64::from(attempt) * DRAIN_POLL_INTERVAL.as_secs();
+                        eprintln!(" done ({elapsed}s)");
+                    }
+                    break;
                 }
-                break;
-            }
-            if attempt == DRAIN_POLL_ATTEMPTS - 1 {
-                eprintln!(" timed out (service may still be draining)");
-            } else {
-                eprint!(".");
-                tokio::time::sleep(DRAIN_POLL_INTERVAL).await;
+                DrainPollAction::Retry => {
+                    eprint!(".");
+                    tokio::time::sleep(DRAIN_POLL_INTERVAL).await;
+                }
+                DrainPollAction::TimedOut => {
+                    let elapsed = u64::from(attempt) * DRAIN_POLL_INTERVAL.as_secs();
+                    eprintln!(" timed out ({elapsed}s)");
+                    anyhow::bail!(
+                        "ECS service {service_name} is still draining after {elapsed}s; dependent cleanup was not started (safe to retry)"
+                    );
+                }
             }
         }
     }
 
-    if let Err(error) =
-        crate::ingress::teardown(aws_config, namespace, name, registry_arn.as_deref()).await
-    {
-        let w = format!("ingress teardown skipped: {error}");
-        eprintln!("  ⚠ {w}");
-        warnings.push(w);
-    }
-    if let Err(error) = crate::ingress::delete_api(aws_config, namespace, name).await {
-        let w = format!("HTTP API cleanup skipped: {error}");
-        eprintln!("  ⚠ {w}");
-        warnings.push(w);
-    }
+    let ingress_result =
+        crate::ingress::teardown(aws_config, namespace, name, registry_arn.as_deref()).await;
+    collect_best_effort_warnings(&mut warnings, ingress_result, "ingress teardown skipped");
+
+    let delete_api_result = crate::ingress::delete_api(aws_config, namespace, name)
+        .await
+        .map(|()| Vec::new());
+    collect_best_effort_warnings(&mut warnings, delete_api_result, "HTTP API cleanup skipped");
 
     let mut cleanup_failures = Vec::new();
     let manifest_key = format!("manifests/{namespace}/{name}.yaml");
@@ -565,6 +602,36 @@ mod tests {
         assert_eq!(
             DeleteTarget::new("prod", "nest-my-oab").ecs_service_name(),
             "oab-prod-nest-my-oab"
+        );
+    }
+
+    #[test]
+    fn drain_poll_action_requires_completion_before_cleanup() {
+        assert_eq!(drain_poll_action(true, 0, 12), DrainPollAction::Complete);
+        assert_eq!(drain_poll_action(false, 0, 12), DrainPollAction::Retry);
+        assert_eq!(drain_poll_action(false, 11, 12), DrainPollAction::TimedOut);
+    }
+
+    #[test]
+    fn best_effort_warnings_preserve_step_warnings_and_errors() {
+        let mut warnings = Vec::new();
+        collect_best_effort_warnings(
+            &mut warnings,
+            Ok(vec!["route cleanup incomplete".to_string()]),
+            "ingress teardown skipped",
+        );
+        collect_best_effort_warnings(
+            &mut warnings,
+            Err(anyhow::anyhow!("access denied")),
+            "HTTP API cleanup skipped",
+        );
+
+        assert_eq!(
+            warnings,
+            vec![
+                "route cleanup incomplete",
+                "HTTP API cleanup skipped: access denied",
+            ]
         );
     }
 
--- a/operator/src/delete.rs
+++ b/operator/src/delete.rs
@@ -21,11 +21,11 @@ macro_rules! eprint {
 ///
 /// Deletion does not require the original manifest: the control-plane copy
 /// under `manifests/{namespace}/{name}.yaml` (and every other resource) is
-/// addressed by `namespace` + `name` alone. Because physical resource names
-/// use `-` as the component delimiter, [`delete_services`] accepts namespaces
-/// without `-`; names may still contain it. This keeps every accepted logical
-/// target's physical delete identity injective without renaming existing
-/// deployments.
+/// addressed by `namespace` + `name` alone. Targets are subject to the shared
+/// injective identity rule (see [`crate::identity`]): namespaces must not
+/// contain `-`, names may. Before any destructive mutation, delete also
+/// verifies in the control plane that no other recorded logical pair claims
+/// the same physical `oab-{namespace}-{name}` identity.
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub struct DeleteTarget {
     pub namespace: String,
@@ -42,7 +42,7 @@ impl DeleteTarget {
 
     /// ECS service name derived from the target (`oab-{namespace}-{name}`).
     pub fn ecs_service_name(&self) -> String {
-        format!("oab-{}-{}", self.namespace, self.name)
+        crate::identity::physical_service_name(&self.namespace, &self.name)
     }
 }
 
@@ -342,7 +342,7 @@ fn classify_service_identity(
 }
 
 fn checkpoint_key(namespace: &str, name: &str) -> String {
-    format!("delete-checkpoints/{namespace}/{name}.json")
+    crate::identity::delete_checkpoint_key(namespace, name)
 }
 
 #[allow(clippy::too_many_arguments)]
@@ -386,7 +386,7 @@ fn validate_checkpoint(
     }
     validate_checkpoint_arns(
         checkpoint,
-        &format!("oab-{namespace}-{name}"),
+        &crate::identity::physical_service_name(namespace, name),
         partition,
         account,
         region,
@@ -416,19 +416,11 @@ fn validate_delete_request(
         )));
     }
     for target in targets {
-        if target.namespace.trim().is_empty() || target.name.trim().is_empty() {
-            return Err(DeleteError::validation(anyhow::anyhow!(
-                "delete target namespace and name must not be empty (got '{}'/'{}')",
-                target.namespace,
-                target.name
-            )));
-        }
-        if target.namespace.contains('-') {
-            return Err(DeleteError::validation(anyhow::anyhow!(
-                "delete target namespace must not contain '-' because it is the physical identity delimiter (got '{}')",
-                target.namespace
-            )));
-        }
+        // Shared injective identity rule (see `crate::identity`): non-empty
+        // components and a hyphen-free namespace, matching what apply-side
+        // manifest validation accepts.
+        crate::identity::validate_injective_identity(&target.namespace, &target.name)
+            .map_err(DeleteError::validation)?;
     }
     Ok(())
 }
@@ -459,10 +451,16 @@ fn record_delete_result(
 /// - the target set must be non-empty
 /// - [`DeleteOptions::cluster`] must be non-empty; its optional bucket override
 ///   follows the shared control-plane resolver
-/// - every target's `namespace`/`name` must be non-empty
-/// - target namespaces must not contain `-`; that delimiter may appear in names,
-///   so this restriction makes `oab-{namespace}-{name}` injective for every
-///   accepted target
+/// - every target must satisfy the shared injective identity rule (see
+///   [`crate::identity`]): non-empty `namespace`/`name` and a hyphen-free
+///   namespace, the same rule apply-side manifest validation enforces. Legacy
+///   deployments created under a hyphenated namespace are not accepted here;
+///   remove them with the `oabctl delete` CLI, which performs the
+///   control-plane ownership check for such targets.
+///
+/// Additionally, before any destructive mutation, delete verifies in the
+/// control plane that no other recorded logical pair claims the target's
+/// physical `oab-{namespace}-{name}` identity, and fails closed if one does.
 ///
 /// # Concurrency
 ///
@@ -740,7 +738,7 @@ async fn prepare_identity(
         return Ok(PreparedIdentity::Exact(Box::new((checkpoint, state))));
     }
 
-    let service_name = format!("oab-{namespace}-{name}");
+    let service_name = crate::identity::physical_service_name(namespace, name);
     let response = ecs.describe_services().cluster(cluster).services(&service_name)
         .send().await.context("failed to describe ECS service before delete")?;
     let service = single_described_service(&response)?;
@@ -1012,7 +1010,7 @@ async fn delete_checkpointed_ecs(
 async fn cleanup_s3(
     s3: &aws_sdk_s3::Client, bucket: &str, namespace: &str, name: &str,
 ) -> Result<()> {
-    let manifest_key = format!("manifests/{namespace}/{name}.yaml");
+    let manifest_key = crate::identity::manifest_key(namespace, name);
     s3.delete_object().bucket(bucket).key(&manifest_key).send().await
         .with_context(|| format!("failed to delete s3://{bucket}/{manifest_key}"))?;
     println!("  ✓ Manifest removed from S3");
@@ -1049,8 +1047,28 @@ async fn run_with_bucket(
     if resource != "oabservice" {
         anyhow::bail!("unknown resource type: {resource}. Use 'oabservice'");
     }
+    if namespace.trim().is_empty() || name.trim().is_empty() {
+        anyhow::bail!(
+            "delete requires a non-empty namespace and name (got '{namespace}'/'{name}')"
+        );
+    }
     let ecs = aws_sdk_ecs::Client::new(aws_config);
     let s3 = aws_sdk_s3::Client::new(aws_config);
+    if namespace.contains('-') {
+        // Legacy-deployment policy: hyphenated namespaces predate the shared
+        // injective identity rule (see `crate::identity`). Programmatic
+        // targets are rejected earlier; the CLI keeps them deletable, guarded
+        // by the exclusive-ownership check below.
+        eprintln!(
+            "  ⚠ Namespace '{namespace}' contains '-', which predates the hyphen-free \
+             namespace rule; continuing only if the control plane records no colliding \
+             deployment for '{}'",
+            crate::identity::physical_service_name(namespace, name)
+        );
+    }
+    // Shared ownership rule: refuse to touch a physical identity that another
+    // recorded logical pair (e.g. a legacy hyphenated namespace) claims.
+    crate::identity::ensure_exclusive_physical_identity(&s3, bucket, namespace, name).await?;
     crate::apply::ensure_no_pending_ingress_teardown(&s3, bucket, namespace, name)
         .await?;
     println!("Deleting {name}...");
@@ -1429,7 +1447,41 @@ mod tests {
             .await
             .unwrap_err();
         assert_eq!(error.kind, DeleteErrorKind::Validation);
-        assert!(error.to_string().contains("must not contain '-'"));
+        assert!(error.to_string().contains("must not contain '-'"), "{error}");
+    }
+
+    #[test]
+    fn colliding_logical_pairs_guard_every_entry_point() {
+        // `prod/team-bot` and `prod-team/bot` share the physical identity
+        // `oab-prod-team-bot`. Cross-entry-point regression for the shared
+        // rule in `crate::identity`:
+        //
+        // 1. Apply cannot create the hyphenated-namespace side (domain rule in
+        //    manifest validation, exercised in `manifest`/`apply` tests) and
+        //    programmatic delete cannot target it
+        //    (`ambiguous_namespace_delimiter_is_rejected_before_aws` above).
+        assert!(
+            crate::identity::validate_injective_identity("prod-team", "bot").is_err(),
+            "hyphenated namespace must be outside the accepted identity domain"
+        );
+        // 2. For the accepted side, the pre-mutation ownership probe shared by
+        //    apply (`apply_ecs`) and every delete entry point
+        //    (`run_with_bucket`) checks exactly the legacy pair's manifest and
+        //    checkpoint keys, so a recorded `prod-team/bot` blocks overwrite,
+        //    checkpointing, and deletion of `oab-prod-team-bot` via
+        //    `prod/team-bot` — and vice versa.
+        assert_eq!(
+            crate::identity::collision_aliases("prod", "team-bot"),
+            vec![("prod-team".to_string(), "bot".to_string())]
+        );
+        assert!(crate::identity::collision_aliases("prod-team", "bot")
+            .contains(&("prod".to_string(), "team-bot".to_string())));
+        // 3. The probed ownership keys are derived from the same helpers the
+        //    mutation paths use, so probe and mutation cannot drift apart.
+        assert_eq!(
+            crate::identity::ownership_keys("prod-team", "bot")[1],
+            checkpoint_key("prod-team", "bot"),
+        );
     }
 
     #[test]
--- a/operator/src/delete.rs
+++ b/operator/src/delete.rs
@@ -415,12 +415,21 @@ fn validate_delete_request(
             "DeleteOptions.cluster must not be empty"
         )));
     }
+
+    let mut seen = std::collections::HashSet::with_capacity(targets.len());
     for target in targets {
         // Shared injective identity rule (see `crate::identity`): non-empty
         // components and a hyphen-free namespace, matching what apply-side
         // manifest validation accepts.
         crate::identity::validate_injective_identity(&target.namespace, &target.name)
             .map_err(DeleteError::validation)?;
+        if !seen.insert((&target.namespace, &target.name)) {
+            return Err(DeleteError::validation(anyhow::anyhow!(
+                "duplicate delete target '{}/{}'",
+                target.namespace,
+                target.name
+            )));
+        }
     }
     Ok(())
 }
@@ -466,9 +475,12 @@ fn record_delete_result(
 ///
 /// This function is not serialized with [`crate::apply::apply_manifests`].
 /// Callers must serialize mutations for the same AWS account, Region,
-/// control-plane bucket, ECS cluster, namespace, and name. If that precondition
-/// is violated, stop concurrent writers, inspect the retained checkpoint, then
-/// either re-apply the intended desired state or retry delete.
+/// control-plane bucket, ECS cluster, and physical service identity. This
+/// includes every alias-equivalent logical pair that could map to the same
+/// `oab-{namespace}-{name}` value (for example `prod/team-bot` and
+/// `prod-team/bot`). If that precondition is violated, stop concurrent
+/// writers, inspect the retained checkpoint, then either re-apply the intended
+/// desired state or retry delete.
 ///
 /// Before ECS mutation, delete persists an exact-identity checkpoint containing
 /// the caller account/region, bucket, canonical cluster and service ARNs, and
@@ -1181,6 +1193,18 @@ mod tests {
         .await
         .unwrap_err();
         assert_eq!(err.kind, DeleteErrorKind::Validation);
+    }
+
+    #[test]
+    fn delete_request_rejects_duplicate_targets_before_aws() {
+        let targets = [
+            DeleteTarget::new("prod", "bot"),
+            DeleteTarget::new("prod", "bot"),
+        ];
+        let error = validate_delete_request(&targets, &DeleteOptions::new("cluster"))
+            .expect_err("duplicate targets must fail validation");
+        assert_eq!(error.kind, DeleteErrorKind::Validation);
+        assert!(error.to_string().contains("duplicate delete target"), "{error}");
     }
 
     #[test]
--- a/operator/src/delete.rs
+++ b/operator/src/delete.rs
@@ -928,7 +928,7 @@ async fn delete_checkpointed_ecs(
     )
     .map_err(anyhow::Error::msg)?;
     if state == ServiceIdentityState::RetryGone {
-        return Ok(());
+        return Ok(false);
     }
     let service = service.context("checkpointed ECS service disappeared ambiguously")?;
     let delete_phase = ecs_delete_phase(service.status())?;
@@ -984,8 +984,23 @@ async fn delete_checkpointed_ecs(
             }
             Err(error) => return Err(error).context("failed to scale ECS service to zero"),
         }
+    } else if delete_phase == EcsDeletePhase::Drain {
+        match ecs
+            .delete_service()
+            .cluster(&checkpoint.cluster_arn)
+            .service(&checkpoint.service_arn)
+            .force(true)
+            .send()
+            .await
+        {
+            Ok(_) => println!("  ✓ ECS draining service deletion resumed"),
+            Err(error) if error.code() == Some("ServiceNotFoundException") => {
+                println!("  ✓ ECS draining service disappeared while resuming delete")
+            }
+            Err(error) => return Err(error).context("failed to resume ECS service deletion"),
+        }
     } else {
-        println!("  ✓ ECS service is already draining; resuming delete cleanup");
+        println!("  ✓ ECS service is already inactive; resuming delete cleanup");
     }
 
     const DRAIN_POLL_ATTEMPTS: u32 = 12;
--- a/operator/src/delete.rs
+++ b/operator/src/delete.rs
@@ -324,6 +324,7 @@ fn checkpoint_key(namespace: &str, name: &str) -> String {
     format!("delete-checkpoints/{namespace}/{name}.json")
 }
 
+#[allow(clippy::too_many_arguments)]
 fn validate_checkpoint(
     checkpoint: &DeleteCheckpoint,
     namespace: &str,
@@ -649,10 +650,11 @@ fn single_described_service(
     }
 }
 enum PreparedIdentity {
-    Exact(DeleteCheckpoint, ServiceIdentityState),
+    Exact(Box<(DeleteCheckpoint, ServiceIdentityState)>),
     LegacyOrphan,
 }
 
+#[allow(clippy::too_many_arguments)]
 async fn prepare_identity(
     aws_config: &aws_config::SdkConfig,
     ecs: &aws_sdk_ecs::Client,
@@ -689,7 +691,7 @@ async fn prepare_identity(
         if let Some(service) = service {
             validate_returned_service(&checkpoint, service)?;
         }
-        return Ok(PreparedIdentity::Exact(checkpoint, state));
+        return Ok(PreparedIdentity::Exact(Box::new((checkpoint, state))));
     }
 
     let service_name = format!("oab-{namespace}-{name}");
@@ -761,7 +763,7 @@ async fn prepare_identity(
         &checkpoint.region,
     )?;
     save_checkpoint(s3, &checkpoint).await?;
-    Ok(PreparedIdentity::Exact(checkpoint, state))
+    Ok(PreparedIdentity::Exact(Box::new((checkpoint, state))))
 }
 
 fn validate_returned_service(
@@ -929,7 +931,8 @@ async fn run_with_bucket(
     match prepare_identity(
         aws_config, &ecs, &s3, namespace, name, cluster, bucket, policy,
     ).await? {
-        PreparedIdentity::Exact(checkpoint, state) => {
+        PreparedIdentity::Exact(boxed) => {
+            let (checkpoint, state) = *boxed;
             delete_checkpointed_ecs(&ecs, &checkpoint, state).await?;
             crate::ingress::delete_exact(
                 aws_config,
--- a/operator/src/delete.rs
+++ b/operator/src/delete.rs
@@ -492,8 +492,10 @@ fn record_delete_result(
 /// An absent ECS service or ambiguous `DescribeServices` response is rejected
 /// unless that matching durable checkpoint already exists. Cleanup uses only IDs
 /// from the checkpoint and removes the checkpoint last, making partial failures
-/// safe to retry. The legacy CLI uses best-effort warnings for dependent/S3
-/// cleanup, while this programmatic API keeps those failures fatal.
+/// safe to retry. The legacy CLI warns on a drain timeout, continues dependent/S3
+/// cleanup, retains the checkpoint for retry, and continues after individual S3
+/// object failures, while this programmatic API keeps exact-identity dependent
+/// and S3 cleanup failures fatal.
 pub async fn delete_services(
     aws_config: &aws_config::SdkConfig,
     targets: &[DeleteTarget],
@@ -899,10 +901,11 @@ async fn delete_checkpointed_ecs(
     ecs: &aws_sdk_ecs::Client,
     checkpoint: &DeleteCheckpoint,
     state: ServiceIdentityState,
-) -> Result<()> {
+    cleanup_mode: CleanupMode,
+) -> Result<bool> {
     if state == ServiceIdentityState::RetryGone {
         println!("  ✓ ECS service already absent; resuming exact dependent cleanup");
-        return Ok(());
+        return Ok(false);
     }
 
     let response = ecs
@@ -1018,44 +1021,93 @@ async fn delete_checkpointed_ecs(
             }
         };
         match drain_poll_action(complete, attempt, DRAIN_POLL_ATTEMPTS) {
-            DrainPollAction::Complete => { eprintln!(" done"); return Ok(()); }
+            DrainPollAction::Complete => {
+                eprintln!(" done");
+                return Ok(false);
+            }
             DrainPollAction::Retry => {
                 eprint!(".");
                 tokio::time::sleep(DRAIN_POLL_INTERVAL).await;
             }
-            DrainPollAction::TimedOut => anyhow::bail!(
-                "checkpointed ECS service did not reach an unambiguous absent state; dependent cleanup was not started (safe to retry)"
-            ),
+            DrainPollAction::TimedOut => {
+                let message =
+                    "checkpointed ECS service did not reach an unambiguous absent state";
+                match cleanup_mode {
+                    CleanupMode::Strict => anyhow::bail!(
+                        "{message}; dependent cleanup was not started (safe to retry)"
+                    ),
+                    CleanupMode::BestEffort => {
+                        eprintln!(
+                            "\n  ⚠ {message}; continuing dependent cleanup and retaining checkpoint"
+                        );
+                        return Ok(true);
+                    }
+                }
+            }
         }
     }
-    unreachable!()
+    anyhow::bail!("ECS drain polling ended without a terminal result")
 }
 
 async fn cleanup_s3(
     s3: &aws_sdk_s3::Client, bucket: &str, namespace: &str, name: &str,
 ) -> Result<()> {
+    let mut failures = Vec::new();
     let manifest_key = crate::identity::manifest_key(namespace, name);
-    s3.delete_object().bucket(bucket).key(&manifest_key).send().await
-        .context("failed to delete control-plane manifest")?;
-    println!("  ✓ Manifest removed from S3");
+    match s3
+        .delete_object()
+        .bucket(bucket)
+        .key(&manifest_key)
+        .send()
+        .await
+    {
+        Ok(_) => println!("  ✓ Manifest removed from S3"),
+        Err(_) => failures.push("failed to delete control-plane manifest"),
+    }
 
     let artifact_prefix = format!("artifacts/{namespace}/{name}/");
     let mut continuation_token = None;
     loop {
-        let response = s3.list_objects_v2().bucket(bucket).prefix(&artifact_prefix)
-            .set_continuation_token(continuation_token).send().await
-            .context("failed to list control-plane config artifacts")?;
+        let response = match s3
+            .list_objects_v2()
+            .bucket(bucket)
+            .prefix(&artifact_prefix)
+            .set_continuation_token(continuation_token)
+            .send()
+            .await
+        {
+            Ok(response) => response,
+            Err(_) => {
+                failures.push("failed to list control-plane config artifacts");
+                break;
+            }
+        };
         for object in response.contents() {
             if let Some(key) = object.key() {
-                s3.delete_object().bucket(bucket).key(key).send().await
-                    .context("failed to delete control-plane config artifact")?;
+                if s3
+                    .delete_object()
+                    .bucket(bucket)
+                    .key(key)
+                    .send()
+                    .await
+                    .is_err()
+                {
+                    failures.push("failed to delete control-plane config artifact");
+                }
             }
         }
         continuation_token = response.next_continuation_token().map(str::to_owned);
-        if continuation_token.is_none() { break; }
+        if continuation_token.is_none() {
+            break;
+        }
     }
-    println!("  ✓ Config artifacts removed from S3");
-    Ok(())
+
+    if failures.is_empty() {
+        println!("  ✓ Config artifacts removed from S3");
+        Ok(())
+    } else {
+        anyhow::bail!("control-plane S3 cleanup incomplete: {}", failures.join("; "))
+    }
 }
 
 async fn run_with_bucket(
@@ -1114,7 +1166,14 @@ async fn run_with_bucket(
                         .to_string(),
                 );
             }
-            delete_checkpointed_ecs(&ecs, &checkpoint, state).await?;
+            let drain_incomplete =
+                delete_checkpointed_ecs(&ecs, &checkpoint, state, cleanup_mode).await?;
+            if drain_incomplete {
+                warnings.push(
+                    "ECS drain timed out; dependent cleanup continued and delete checkpoint was retained"
+                        .to_string(),
+                );
+            }
 
             match cleanup_mode {
                 CleanupMode::Strict => {
@@ -1136,7 +1195,7 @@ async fn run_with_bucket(
                     println!("  ✓ Delete checkpoint removed");
                 }
                 CleanupMode::BestEffort => {
-                    let mut cleanup_failed = false;
+                    let mut cleanup_failed = drain_incomplete;
                     if let Err(error) = crate::ingress::delete_exact(
                         aws_config,
                         namespace,
--- a/operator/src/delete.rs
+++ b/operator/src/delete.rs
@@ -1127,6 +1127,7 @@ async fn run_with_bucket(
             "delete requires a non-empty namespace and name (got '{namespace}'/'{name}')"
         );
     }
+    crate::identity::validate_legacy_delete_identity(namespace, name)?;
     let ecs = aws_sdk_ecs::Client::new(aws_config);
     let s3 = aws_sdk_s3::Client::new(aws_config);
     if namespace.contains('-') {
--- a/operator/src/delete.rs
+++ b/operator/src/delete.rs
@@ -42,23 +42,30 @@ impl DeleteTarget {
     }
 }
 
-/// Target options for [`delete_services`]. The library never reads CLI
-/// configuration or guesses which control-plane bucket owns a deployment.
+/// Target options for [`delete_services`]. Mirrors
+/// [`ApplyOptions`](crate::apply::ApplyOptions): the cluster is required and
+/// the library never reads `~/.oabctl/config.toml`.
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub struct DeleteOptions {
     /// ECS cluster name or ARN the services were deployed into.
     pub cluster: String,
-    /// Exact S3 bucket that owns the deployment and its delete checkpoint.
-    pub control_plane_bucket: String,
+    /// Optional control-plane bucket override. When absent, resolution uses
+    /// `OAB_CONTROL_PLANE_BUCKET`, then `oab-control-plane-{account}` from
+    /// the caller's AWS identity, matching apply's shared resolver.
+    pub control_plane_bucket: Option<String>,
 }
 
 impl DeleteOptions {
-    /// Bind a delete request to both its ECS cluster and control-plane bucket.
-    pub fn new(cluster: impl Into<String>, control_plane_bucket: impl Into<String>) -> Self {
+    pub fn new(cluster: impl Into<String>) -> Self {
         Self {
             cluster: cluster.into(),
-            control_plane_bucket: control_plane_bucket.into(),
+            control_plane_bucket: None,
         }
+    }
+
+    pub fn with_control_plane_bucket(mut self, bucket: impl Into<String>) -> Self {
+        self.control_plane_bucket = Some(bucket.into());
+        self
     }
 }
 
@@ -85,6 +92,8 @@ pub struct DeleteReport {
 pub enum DeleteErrorKind {
     /// Contract violation caught before any AWS call.
     Validation,
+    /// The deployment target (bucket resolution) could not be established.
+    Target,
     /// Teardown of a specific service failed. Cleanup is resumable: calling
     /// [`delete_services`] again with the same target continues from the
     /// remaining resources.
@@ -105,6 +114,15 @@ impl DeleteError {
     fn validation(source: impl Into<anyhow::Error>) -> Self {
         Self {
             kind: DeleteErrorKind::Validation,
+            failed_service: None,
+            completed: DeleteReport::default(),
+            source: source.into(),
+        }
+    }
+
+    fn target(source: impl Into<anyhow::Error>) -> Self {
+        Self {
+            kind: DeleteErrorKind::Target,
             failed_service: None,
             completed: DeleteReport::default(),
             source: source.into(),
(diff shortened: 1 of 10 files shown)


chaodu-agent:
> [!IMPORTANT]
> CHANGES REQUESTED ⚠️ — The public delete contract can report success while ECS teardown is still incomplete, and it drops ingress cleanup warnings that callers need for recovery.

## What This PR Does

This PR adds a public, in-process `delete_services` API for deleting OAB services by explicit `namespace` + `name`, with an explicit cluster, structured partial results, and no dependency on CLI home configuration. It is intended to give control-plane callers the delete counterpart to `apply_manifests`.

## How It Works

`DeleteOptions` resolves the control-plane bucket through the same explicit option → environment → AWS-account chain as apply. Each `DeleteTarget` is processed sequentially through the existing ECS, ingress, API Gateway, and S3 teardown path; completed targets are retained in `DeleteReport`, and a failing target is exposed through `DeleteError`. A shared task-local gate suppresses CLI progress output for library callers.

## Findings

| 1 | 🔴 Critical | Exhausting the ECS drain poll still falls through to `Ok`, so `DeleteReport` can claim a completed deletion while the service remains `DRAINING`. | `operator/src/delete.rs:404-440` |
| 2 | 🟡 Important | Successful ingress teardown warnings and API Gateway delete failures are discarded, contradicting the structured warning contract. | `operator/src/delete.rs:443-454` |
| 3 | 🟢 Praise | The explicit options, partial-result error type, pre-AWS validation, and task-local progress suppression form a clean programmatic API foundation. | — |

<details>

### 🔴 F1: A drain timeout is returned as a successful deletion

The polling loop prints `timed out (service may still be draining)` after its final attempt, but it neither records an incomplete state nor returns an error. Execution proceeds through ingress/S3 cleanup and ultimately returns `Ok(warnings)` at `operator/src/delete.rs:519`; `delete_services` then adds the target to `DeleteReport.services`.

That violates the new public contract documented on `DeletedService`: core ECS/S3 teardown should either complete or fail the target. It also lets an in-process caller treat tenant deprovisioning as complete while ECS may still be draining.

Return a teardown error when the polling budget is exhausted, or model an explicit typed pending/incomplete outcome that cannot be confused with successful completion. Add a regression test for the timeout transition.

### 🟡 F2: Best-effort ingress failures do not reach `DeletedService.warnings`

`ingress::teardown` accumulates best-effort failures and returns them as `Ok(Vec<String>)` (`operator/src/ingress.rs:299-466`), but the new call site handles only `Err` and drops the `Ok(warnings)` value. In addition, `ingress::delete_api` logs an SDK deletion failure and still returns `Ok(())` (`operator/src/ingress.rs:475-484`), so `delete.rs` cannot surface that failure either. Because programmatic calls suppress those log messages, the caller receives neither output nor structured warnings.

Extend the target's warning list with warnings returned by `ingress::teardown`, and make `delete_api` return its SDK error so the existing warning conversion can record it. Add regression coverage proving both warning paths appear in `DeletedService.warnings`.

</details>

<details>

- PR opened: 2026-07-16
- Base branch: `refactor/oabctl-lib` at `6dd5285`; main checked at `5a6bc8f`
- Main/base already have: the structured `apply_manifests` API and the existing CLI delete implementation
- Net-new value: a public structured delete API with explicit target/options, partial reports, shared bucket resolution, and silent library execution
- CI at review time: the `operator` job is still in progress; other reported checks are successful or skipped. Local Rust validation could not run because this review environment has no Cargo toolchain.

</details>

<details>

- Request validation runs before bucket resolution or other AWS calls.
- `DeleteError` preserves both the failed target and successfully completed prior targets.
- Reusing task-local progress suppression avoids changing existing CLI rendering.
- Bucket resolution is intentionally aligned with the apply path.

</details>

No external review comments or inline findings were present when this review was performed.

1. **The existing CLI already performs teardown** — if all callers can safely execute the CLI, a second public contract increases API surface and maintenance cost.
2. **The inherited teardown path was designed around human-readable best effort** — exposing it as a structured library API requires stronger, machine-verifiable completion semantics, as the findings above demonstrate.
3. **`namespace` + `name` may be weaker than a durable deployment identity** — long-term callers may be safer deleting from persisted apply output or a control-plane record that also binds cluster and bucket.

chaodu-agent:
> [!IMPORTANT]
> CHANGES REQUESTED ⚠️ — The previous drain/warning blockers are fixed, but the new public delete contract can still remove dependent resources for the wrong deployment and does not fully honor its silent-library-call guarantee.

## What This PR Does

This PR adds a public, in-process `delete_services` API for deleting OAB services by explicit `namespace` + `name`, with an explicit cluster, structured partial results, shared control-plane bucket resolution, and no CLI home-configuration dependency.

## How It Works

`DeleteOptions` resolves the control-plane bucket through the same explicit option → environment → AWS-account chain as apply. Each `DeleteTarget` runs sequentially through the existing ECS, ingress, API Gateway, and S3 teardown path; successful targets are recorded in `DeleteReport`, while `DeleteError` preserves completed prior targets. A task-local progress gate is intended to keep programmatic calls silent.

## Findings

| 1 | 🔴 Critical | An initial ECS `DescribeServices` response with no service is treated as proof that ECS is already gone, so a wrong-but-valid cluster can leave the real service running while the code deletes ingress and S3 resources; the Cloud Map fallback can also delete the first account/Region-wide name match. | `operator/src/delete.rs:387-405`, `operator/src/delete.rs:484-552`, `operator/src/ingress.rs:419-433` |
| 2 | 🟡 Important | `delete_services` suppresses only macros in `apply.rs`/`delete.rs`; `ingress::teardown` and `delete_api` still write directly to stderr, violating the documented silent programmatic contract. | `operator/src/delete.rs:207-247`, `operator/src/ingress.rs:339-466`, `operator/src/ingress.rs:475-487` |
| 3 | 🟢 Praise | The follow-up correctly turns drain exhaustion into a teardown error before dependent cleanup and now returns successful ingress/API cleanup warnings to callers. | `operator/src/delete.rs:437-491` |

<details>

### 🔴 F1: “Service not returned” is not sufficient ownership proof for destructive dependent cleanup

The initial `DescribeServices` call reads only `services().first()` and ignores `describe_response.failures()`. The official ECS `DescribeServices` contract returns `failures` and `services` as separate fields in an HTTP 200 response, and AWS documents a missing service as potentially indicating the wrong cluster or Region. Here, an empty service list becomes `EcsDeletePhase::Cleanup`, after which the code deletes API Gateway and S3 resources addressed only by `namespace` + `name`.

This means a caller can provide a valid but incorrect cluster, receive no ECS service, and still destroy the actual deployment's dependent resources while its ECS service remains active elsewhere. The Cloud Map path compounds this: without the ECS registry ARN it scans all services in the account/Region and deletes the first matching name, despite the adjacent comment acknowledging same-name services in different VPC namespaces.

Fail closed on an initially missing ECS service unless orphan cleanup is explicitly and safely requested. Inspect the response `failures`, and bind dependent cleanup to persisted or caller-supplied deployment identity (for example exact ECS/Cloud Map/API resource identifiers). At minimum, remove the ambiguous first-name-match Cloud Map fallback. Add regression coverage for a missing service in a wrong valid cluster and for duplicate Cloud Map names.

### 🟡 F2: Programmatic delete still emits ingress progress to stderr

`with_progress_suppressed` changes a task-local value consumed by macros defined in `apply.rs` and `delete.rs`, but Rust macros are lexical. The direct `eprintln!` calls inside `ingress.rs` do not consult that task-local gate. Any target with ingress cleanup, Cloud Map cleanup, or API deletion therefore writes human-readable output during `delete_services`, contrary to its public documentation.

Route ingress rendering through the shared progress gate (or return structured events and render only in CLI entry points), while continuing to return warnings structurally. Add a regression test that exercises the programmatic path with ingress activity and verifies no stdout/stderr output.

</details>

<details>

- PR opened: 2026-07-16
- Base branch: `refactor/oabctl-lib` at `6dd5285`; main checked at `5a6bc8f`
- Main already has: no programmatic delete API; the stacked base has the structured apply API and existing CLI delete implementation
- Net-new value: a public structured delete API with explicit target/options, partial reports, shared bucket resolution, and resumable teardown
- CI: the `operator` check passed at `e0f5333`; other relevant checks passed or were skipped. Local `cargo test`/`cargo clippy` could not run because this review environment has no Cargo toolchain.

</details>

## Previous Review Feedback

- ✅ **Drain timeout falsely succeeded — addressed in `e0f5333`**: the final poll now returns an error before ingress/S3 cleanup, with focused state-transition coverage.
- ✅ **Ingress/API warnings were dropped on successful targets — addressed in `e0f5333`**: returned teardown warnings are accumulated, and API deletion failures now propagate into `DeletedService.warnings`.

No external review comments, submitted reviews, or inline findings were present at this commit.

<details>
<summary>Review Coverage</summary>

| Reviewer | Focus | Result |
|---|---|---|
| Reviewer A | Correctness and teardown state transitions | F1; prior drain blocker verified fixed |
| Reviewer B | Public API and architecture | Explicit options/report design is clean; F1 blocks safe identity semantics |
| Reviewer C | Testing | Existing focused tests are useful; F1/F2 need regression coverage |
| Reviewer D | Safety and operability | F1 |
| Reviewer E | Independent audit | CHANGES REQUESTED |

</details>

<details>

- Contract validation runs before bucket resolution or other AWS calls.
- `DeleteError` preserves the failed target and successfully completed prior targets.
- Bucket resolution intentionally matches the apply path.
- The drain state helper makes the timeout behavior explicit and testable.
- Successful best-effort ingress failures now reach structured warnings.

</details>

1. **The existing CLI already performs teardown** — if all callers can safely execute the CLI, a second public contract increases API surface and maintenance cost.
2. **The current resource identity is weaker than the destructive operation requires** — `namespace` + `name` + caller-supplied cluster cannot safely identify orphaned cross-system resources in every retry scenario.
3. **The inherited teardown path was designed around CLI best effort** — fully converting it into a silent, machine-verifiable library contract requires stricter ownership and output boundaries than this version currently provides.

chaodu-agent:
## Pre-merge verification — code review + real-AWS E2E at 1bd61e8

**Code review of the identity-checkpoint path** (`prepare_identity` / `classify_service_identity` / `validate_checkpoint`):

- The checkpoint is minted **from the authoritative DescribeServices response** (service ARN, canonical cluster ARN, createdAt incarnation, exact registry ARN, resolved API id) and persisted **before any teardown starts** — dependent cleanup is never addressed by name alone. The old first-name-match Cloud Map fallback is gone (round-2 F1 resolved at the root).
- `classify_service_identity` is fail-closed on every ambiguous case: empty response without an explicit `MISSING` failure, INACTIVE without a checkpoint, status without identity, multiple services, multiple registries. A checkpoint only authorizes "gone" for exactly zero services + one MISSING failure, or INACTIVE after full ARN/incarnation validation.
- Checkpoint validation binds version + namespace/name + requested cluster + bucket + partition/account/region; it is deleted **only after** S3 cleanup completes, so a same-name recreate cannot inherit a stale identity.

**E2E against real AWS** (synthetic tenant `nest-e2e-1415`, account 903779448426, cluster `oab`; harness calls `apply_manifests`/`delete_services` in-process):

| Scenario | Result |
|---|---|
| Apply (no checkpoint — same as any pre-checkpoint deployment) then delete while the service was still provisioning | ✅ Refused dependent cleanup after the drain wait: `checkpointed ECS service did not reach an unambiguous absent state; dependent cleanup was not started (safe to retry)` — checkpoint retained for resume |
| Retry delete (checkpoint resume path) | ✅ Full teardown (ECS + Cloud Map + API GW + S3), zero warnings, **checkpoint removed after success** — verified via S3/ECS/API-GW state, not just the report |
| Delete the now-gone tenant again (INACTIVE visible, no checkpoint) | ✅ Failed closed: `ECS returned INACTIVE without a matching delete checkpoint` — no false success, no cleanup of unowned resources |
| Bonus: apply with an invalid Telegram token | ✅ webhook registration failure surfaced as a structured `AppliedService.warnings` entry, apply succeeded |

Also verified earlier on this branch: CI operator check green (clippy 1.97 `-D warnings`, 96 unit tests + doc-test).

**Verdict: good to merge.** The delete contract does what the two review rounds demanded — checkpointed exact identity, fail-closed ambiguity handling, resumable teardown with no false success — and it holds up against real AWS, including the failure paths.

operator/src/delete.rs
```diff
@@ -1,7 +1,485 @@
...
+            "no targets to delete (empty target set)"
+        )));
+    }
+    if opts.cluster.trim().is_empty() {
+        return Err(DeleteError::validation(anyhow::anyhow!(
+            "DeleteOptions.cluster must not be empty"
+        )));
+    }
+    for target in targets {
+        if target.namespace.trim().is_empty() || target.name.trim().is_empty() {
```
chaodu-agent: **🔴 F1 — Make the target-to-resource identity injective**

Blank-only validation allows distinct targets such as `prod/team-bot` and `prod-team/bot` to derive the same ECS, API Gateway, and Cloud Map names. On the first delete, `prepare_identity` can therefore resolve the other logical target’s live ECS service and mint a fully valid exact-resource checkpoint for the wrong pair.

**Requested change:** use a reversible/injective physical identity (or immutable logical-identity metadata verified before checkpoint creation) and add a regression proving distinct namespace/name pairs cannot collide.

operator/src/delete.rs
```diff
@@ -81,215 +578,823 @@ pub(crate) async fn run(
...
+    let s3 = aws_sdk_s3::Client::new(aws_config);
+    println!("Deleting {name}...");
 
+    match prepare_identity(
+        aws_config, &ecs, &s3, namespace, name, cluster, bucket,
+    )
+    .await? {
+        PreparedIdentity::Exact(boxed) => {
+            let (checkpoint, state) = *boxed;
+            delete_checkpointed_ecs(&ecs, &checkpoint, state).await?;
```
chaodu-agent: **🔴 F2 — Fence post-ECS cleanup against concurrent apply**

Once this call observes the old incarnation gone, the code immediately deletes checkpointed API/Cloud Map resources and name-keyed S3 state. `apply` does not read `delete-checkpoints` and can reuse those dependency IDs or write a new manifest in this window, after which this delete removes the recreated deployment’s resources/state.

**Requested change:** introduce a target-scoped generation/lease/tombstone protocol shared by apply and delete, with conditional S3 mutation, and test an apply interleaving after ECS drain completion.

chaodu-agent:
> [!IMPORTANT]
> **CHANGES REQUESTED ⚠️** — Exact checkpoints now fail closed on ambiguous AWS responses, but the logical target is still non-unique and apply can race the post-ECS cleanup, allowing deletion of another or newly recreated deployment.

## What This PR Does

This PR adds a public in-process `delete_services` API for deleting OAB services by explicit `namespace` + `name`, with an explicit ECS cluster, structured partial results, shared control-plane bucket resolution, silent library execution, and durable exact-resource checkpoints for resumable teardown.

## How It Works

Each target is validated, resolved to its live ECS service, and checkpointed in S3 with canonical cluster/service ARNs, the ECS `createdAt` incarnation, registry ARN, API ID, caller boundary, and bucket. Delete then drains the checkpointed ECS incarnation, removes checkpointed ingress dependencies, deletes the target's manifest/artifacts, and removes the checkpoint last. Initial missing or ambiguous ECS responses fail closed; retries continue only from a matching checkpoint.

## Findings

| 1 | 🔴 Critical | `namespace` and `name` are concatenated with `-`, so distinct logical targets can resolve to the same ECS/API/Cloud Map identity; the initial lookup can checkpoint and delete the other target's live resources. | `operator/src/delete.rs:415` |
| 2 | 🔴 Critical | After the old ECS incarnation is observed gone, delete performs dependency and name-keyed S3 cleanup without a fence shared with apply; a concurrent recreate can reuse those resources and then have them deleted. | `operator/src/delete.rs:1032` |
| 3 | 🟢 Praise | The exact ARN/incarnation checkpoint, explicit-MISSING retry rule, fatal dependent cleanup, and structured partial report substantially improve fail-closed and resumable teardown behavior. | — |

<details>

### 🔴 F1: The logical target does not map injectively to its physical identity

Validation rejects only blank components, while ECS and Cloud Map use `oab-{namespace}-{name}` and API Gateway uses `oab-webhook-{namespace}-{name}`. For example, `prod/team-bot` and `prod-team/bot` both resolve to `oab-prod-team-bot` and `oab-webhook-prod-team-bot`.

The checkpoint does not close this initial ambiguity. If `prod/team-bot` is deployed and a caller requests deletion of `prod-team/bot`, `prepare_identity` describes the same physical ECS service. Its ARN and `createdAt` are valid for the derived name, so the code mints a new checkpoint carrying the requested—but wrong—logical pair and proceeds to delete that service and its exact dependencies. Separate S3 paths only mean the actual deployment's manifest may remain after its ECS/API/Cloud Map resources are removed.

**Requested change:** define one injective canonical resource identity and verify the logical identity before minting a checkpoint. A reversible encoding or immutable namespace/name tags/metadata can preserve both components; validation alone is sufficient only if it proves the delimiter cannot occur. Add a regression proving that two distinct component pairs can never resolve to the same delete identity.

### 🔴 F2: Delete and apply have no shared target fence

`delete_checkpointed_ecs` protects mutations of the checkpointed ECS incarnation, but once it returns `RetryGone`, `run_with_bucket` immediately deletes the checkpointed API/Cloud Map identities and then unconditionally deletes `manifests/{namespace}/{name}.yaml` plus `artifacts/{namespace}/{name}/`. There is no final incarnation check, conditional S3 generation/ETag, or target lease, and `apply.rs` never reads `delete-checkpoints`.

A same-target apply can therefore start after the old ECS incarnation is observed gone but before cleanup completes. If it starts before dependency cleanup, `ensure_api` and `ensure_cloud_map` reuse the still-existing API ID and registry ARN that delete is about to remove. Regardless of ingress timing, apply can write a new manifest/artifact generation that `cleanup_s3` then deletes by name. Keeping the checkpoint until the end preserves retry identity but does not fence a concurrent writer.

**Requested change:** add a target-scoped fencing protocol shared by apply and delete—such as a lease/tombstone with a generation token that apply must reject or supersede explicitly, plus conditional S3 mutation. Revalidation alone only narrows the race. Add an interleaving test where apply recreates the target after ECS drain completion and prove delete cannot remove the new incarnation or its desired state.

### 🟢 F3: Exact checkpointing closes the earlier ambiguous-cleanup paths

The current code now requires one identified ECS service before creating a checkpoint, binds retries to canonical ARNs plus `createdAt`, treats drain exhaustion as failure, and preserves the checkpoint on dependent/S3 errors. These are strong foundations once logical identity and cross-operation serialization are fixed.

</details>

<details>

- PR opened: 2026-07-16
- Reviewed head: `1bd61e825e3f2e1308bde3be6294e10b0e3bead6`
- Base branch: `refactor/oabctl-lib` at `6dd5285`; `main` checked at `30bc145`
- Main already has: no programmatic delete API matching this contract
- Net-new value: explicit typed delete targets/options/results plus exact resumable teardown identity
- Validation: `git diff --check` passed; the exact-head `operator` CI job passed. Local `cargo test`/`cargo clippy` could not run because this review environment has no Cargo toolchain.

</details>

## Previous Review Feedback

- ✅ **Drain timeout could report success** — fixed: exhaustion now returns an error before dependent cleanup.
- ✅ **Ingress failures/warnings were lost and library calls emitted progress** — fixed: dependent failures are fatal or structured, and ingress rendering now uses the shared progress gate.
- ✅ **Initial missing ECS responses and name-only Cloud Map fallback were unsafe** — fixed: initial ambiguity fails closed and dependent IDs come from the checkpointed ECS identity.

### @chaodu-agent (pre-merge verification at `1bd61e8`)

> “The checkpoint is minted from the authoritative DescribeServices response” and the real-AWS retry completed exact teardown with the checkpoint removed last.

✅ **Confirmed for the exercised scenarios:** the current code and reported E2E demonstrate fail-closed missing/ambiguous responses, safe retry of the same checkpointed incarnation, and successful complete teardown.

⚠️ **Two cases remain unexercised:** the E2E used one canonical namespace/name pair, so it did not test a different pair with the same derived physical name (F1). It also did not interleave a new apply after the old ECS incarnation became absent but before dependency/S3 cleanup (F2). Removing the checkpoint last supports delete retries, but apply does not consult that checkpoint, so it is not a cross-operation fence.

<details>
<summary>Review Coverage</summary>

| Reviewer | Focus | Result |
|---|---|---|
| Reviewer A | Correctness / target identity | F1 |
| Reviewer B | Architecture / API contract | F1, F2 |
| Reviewer C | Safety / operability | F2 |
| Reviewer D | Testing / regression coverage | Collision and apply/delete interleaving tests required |
| Reviewer E | Independent exact-SHA audit | Confirmed the missing apply/delete fence; final aggregation strengthened the initial-lookup collision analysis |

</details>

<details>

- Contract validation and bucket selection are explicit and library-friendly.
- `DeleteError` retains both the failed target and successfully completed prior targets.
- Exact ARN, caller-boundary, registry, and `createdAt` checks eliminate the earlier wrong-cluster/name-only cleanup behavior.
- Dependent and S3 failures retain the durable checkpoint for safe retry.
- The author supplied useful real-AWS evidence for initial delete, retry, and already-gone failure paths.

</details>

1. **The CLI already owns teardown orchestration** — if control-plane callers can safely invoke it, a second public contract adds a long-lived compatibility surface.
2. **Logical identity is not yet durable enough for a destructive API** — a persisted deployment ID or immutable tagged identity may be safer than reconstructing ownership from two free-form strings.
3. **Apply/delete serialization may belong in the control plane** — if this crate cannot provide a shared fence, exposing raw concurrent operations could create more risk than an orchestrated deprovision workflow.

chaodu-agent (commented):
> [!IMPORTANT]
> **CHANGES REQUESTED ⚠️** — The logical delete target is non-unique, and post-ECS cleanup is not fenced against a concurrent apply.

Consolidated review: https://github.com/openabdev/openab/pull/1415#issuecomment-4998788402

GitHub event: COMMENT — self-review delivery only; this is not an approval.

operator/src/delete.rs
chaodu-agent: **🔴 F1 — Make logical target identity injective**

Validation accepts both `(prod, team-bot)` and `(prod-team, bot)`, although both derive `oab-prod-team-bot`; an initial lookup can therefore checkpoint and delete the other logical target. This violates RC-1 AC-2 before exact checkpoint validation can help.

**Requested change:** use one injective mapping shared by apply/delete, or reject every ambiguous component shape before AWS calls. A batch-only duplicate check is insufficient because separate calls must also be safe. Add the colliding-pair regression.

operator/src/delete.rs
```diff
@@ -81,215 +578,823 @@ pub(crate) async fn run(
...
+                namespace,
+                name,
+                checkpoint.api_id.as_deref(),
+                checkpoint.registry_arn.as_deref(),
+                &checkpoint.partition,
+                &checkpoint.account,
+                &checkpoint.region,
+            )
+            .await?;
+            cleanup_s3(&s3, bucket, namespace, name).await?;
```
chaodu-agent: **🟡 F2 — Remove stale apply ingress-teardown state on successful delete**

This success path removes the manifest/artifacts and delete checkpoint, but not `ingress-teardown-checkpoints/{namespace}/{name}.json`. If a prior apply was interrupted after writing that key, delete can report success while the stale key deterministically blocks re-apply.

**Requested change:** after exact dependent cleanup succeeds, remove the target apply-side teardown checkpoint without weakening retry safety, and test interrupted ingress teardown → delete → re-apply eligibility.

operator/src/delete.rs
```diff
@@ -1,7 +1,485 @@
...
+                "delete target namespace and name must not be empty (got '{}'/'{}')",
+                target.namespace,
+                target.name
+            )));
+        }
+    }
+    Ok(())
+}
+
+/// Tear down OAB services programmatically, without reading CLI home
```
chaodu-agent, line 458: **🟡 F3 — Document the required same-target serialization boundary**

RC-1 accepts the lack of an internal apply/delete fence only when callers serialize the full account/Region/bucket/cluster/namespace/name target. The public Rustdocs and user docs do not state this mandatory safety precondition.

**Requested change:** document the serialization and recovery rule for both `apply_manifests` and `delete_services`, while retaining the already-documented no-orphan behavior.

operator/src/delete.rs
```diff
@@ -81,215 +578,823 @@ pub(crate) async fn run(
...
+            service_created_at: 1_700_000_000,
+            registry_arn: Some(
+                "arn:aws:servicediscovery:us-east-1:123456789012:service/srv-123".to_string(),
+            ),
+            api_id: Some("api-123".to_string()),
+        }
+    }
+
+    #[tokio::test]
+    async fn delete_services_rejects_empty_target_set() {
```
chaodu-agent, line 1303: **🟡 F4 — Add the contract-required regression evidence**

The current tests cover fail-closed identity responses and checkpoint boundaries, but not logical-name collision prevention or the public partial-report promise when a later target fails.

**Requested change:** add a colliding-pair test plus a two-target failure test proving the first completion remains in `DeleteError.completed` and processing stops deterministically; include the F2 checkpoint cleanup sequence.

chaodu-agent:
> [!IMPORTANT]
> **CHANGES REQUESTED ⚠️** — RC-1 accepts caller-serialized same-target mutations, but the implementation still permits colliding logical targets and leaves contract-required cleanup, documentation, and regression coverage incomplete.

## What This PR Does

This PR adds a public in-process `delete_services` API for explicit OAB `namespace` + `name` targets. It resolves the same control-plane bucket chain as apply, records exact ECS and ingress identity before mutation, reports structured partial results, suppresses library progress output, and keeps the existing CLI flow.

## How It Works

Each target is validated and resolved to one live ECS service. Before mutation, delete stores a durable S3 checkpoint containing the caller boundary, bucket, canonical cluster/service ARNs, ECS `createdAt` incarnation, registry ARN, and API ID. It then drains that exact ECS incarnation, removes exact ingress dependencies, deletes the target manifest/artifacts, and removes the delete checkpoint last. Matching retries resume from the checkpoint; initial missing or ambiguous identity fails closed.

## Review Contract Freeze (RC-1)

- Contract revision: `RC-1`
- Reviewed head: `1bd61e825e3f2e1308bde3be6294e10b0e3bead6`
- Contract SHA-256: `a2377a772fc9ba7f4856cea6af2cc6f6596dba919f12256295fd65516de3c674`
- Hash input: the exact PR-description text from `## Review Contract` through the end of `### Follow-ups`, with trailing whitespace removed and one final newline.
- Frozen decision: same-target apply/delete concurrency is an accepted residual risk only when callers serialize the full target boundary described in RC-1. A shared generation/lease fence is a non-blocking follow-up. Logical target collisions, wrong-resource cleanup, false success, undocumented safety preconditions, and missing required evidence are not accepted.

## Findings

| 1 | 🔴 Critical | Distinct accepted `(namespace, name)` pairs still collapse to the same ECS, Cloud Map, and API names; delete can checkpoint and remove the other logical target. | `operator/src/delete.rs:39-41,400-423` |
| 2 | 🟡 Important | A successful delete does not remove apply's per-target `ingress-teardown-checkpoints/...` state, so a prior interrupted ingress removal can deterministically block a later re-apply. | `operator/src/delete.rs:982-1050` |
| 3 | 🟡 Important | The mandatory external serialization precondition is absent from public Rustdocs and user documentation. | `operator/src/delete.rs:426-440` |
| 4 | 🟡 Important | RC-1-required regressions for logical-identity collision and partial-report preservation are missing. | `operator/src/delete.rs:1057-1400` |
| 5 | 🟢 Praise | Exact ARN/incarnation checkpoints, fail-closed AWS response classification, drain-timeout handling, and checkpoint-last retry semantics substantially reduce destructive ambiguity. | — |

<details>

### 🔴 F1 — Logical target identity is not injective

**Lineage:** `ORIGINAL`  
**Affected contract clause:** RC-1 Acceptance Criterion 2 (and Criterion 6).

`DeleteTarget::ecs_service_name` produces `oab-{namespace}-{name}`, while apply and ingress use the same delimiter-only construction for ECS/Cloud Map and `oab-webhook-{namespace}-{name}` for API Gateway. Validation checks only that each component is non-blank. Therefore `("prod", "team-bot")` and `("prod-team", "bot")` both resolve to the same physical identities.

This is unsafe across separate calls, not merely when both targets appear in one batch. If the first pair is deployed and the second pair is passed to delete, the initial ECS lookup returns the first deployment's legitimate service; the code can then mint an exact checkpoint labelled with the wrong logical pair and delete that service. The later exact checkpoint checks cannot detect an ambiguity introduced before checkpoint creation.

**Requested change:** define one injective identity rule shared by manifest/apply/delete. Use reversible encoding or durable logical metadata, or reject every component shape that makes the existing delimiter mapping ambiguous before any AWS call. A batch-only duplicate `HashSet` is insufficient because separate calls must also be safe. Add a regression proving the two pairs above cannot both be accepted and resolve to one delete identity.

### 🟡 F2 — Successful delete leaves apply's teardown checkpoint behind

**Lineage:** `NEW EVIDENCE`  
**Affected contract clause:** RC-1 Goal and Acceptance Criterion 7.

Apply persists `ingress-teardown-checkpoints/{namespace}/{name}.json` before removing ingress and deliberately blocks future apply when that checkpoint remains. Delete removes the manifest, artifacts, exact ingress resources, and its own `delete-checkpoints/...` key, but never removes this apply-side per-target checkpoint.

A sequential, non-concurrent failure path therefore remains: an apply is interrupted after saving its ingress teardown checkpoint; delete then completes successfully for the exact deployment; a later apply with ingress fails with `a previous ingress teardown is still pending`, while an ingress-free apply can fail because the ECS service is now absent. Manual S3 surgery is required even though delete reported completion.

**Requested change:** after exact dependent cleanup succeeds, remove the target's apply-side ingress teardown checkpoint as part of successful delete (while retaining failure/retry safety). Add a regression for interrupted ingress teardown → successful delete → clean re-apply eligibility.

### 🟡 F3 — The accepted concurrency risk is not documented for API consumers

**Lineage:** `NEW EVIDENCE`  
**Affected contract clause:** RC-1 Acceptance Criterion 9.

RC-1 permits the absence of an internal apply/delete fence only because callers must serialize the same account/Region/bucket/cluster/namespace/name target. Neither `apply_manifests` nor `delete_services` Rustdocs, the crate-level docs, `operator/README.md`, or `docs/oabctl.md` states this mandatory safety precondition.

**Requested change:** document the full same-target serialization boundary and recovery rule in the public apply/delete Rustdocs and user-facing library documentation. Keep the no-orphan-cleanup behavior already documented.

### 🟡 F4 — Required contract regressions are missing

**Lineage:** `NEW EVIDENCE`  
**Affected contract clause:** RC-1 Acceptance Criterion 10.

The delete tests cover empty inputs, checkpoint boundary validation, fail-closed response classification, incarnation mismatch, and drain timeout. They do not cover collision prevention, nor do they exercise the public partial-result promise that a later target failure preserves earlier completed targets and stops the batch deterministically.

**Requested change:** add focused tests for the F1 collision rule and for a two-target sequence where the first completion remains in `DeleteError.completed` when the second target fails. Add the F2 sequential checkpoint cleanup regression as well.

### 🟢 F5 — Exact identity and retry handling are strong foundations

The reviewed head validates caller/account/Region/cluster/service ARN boundaries, binds retries to ECS `createdAt`, rejects empty-success and mixed failure responses, delays dependent cleanup until exact ECS absence, and retains the delete checkpoint on cleanup failure. These directly satisfy the core fail-closed portions of RC-1.

</details>

<details>

- PR opened: 2026-07-16
- Reviewed head: `1bd61e825e3f2e1308bde3be6294e10b0e3bead6`
- Base branch: `refactor/oabctl-lib` at `6dd5285af10a237bef54f7ebeb797dc3b7450d1f`
- Main checked at: `f8e91565a4e044892515067532be9ad3100dd2b6`
- Main/base already have: no public `delete_services`, `DeleteTarget`, or `DeleteOptions` API
- Net-new value: a typed in-process delete API with exact durable teardown identity and structured partial results
- Validation: `git diff --check` passed; the exact-head `operator` CI check passed. This review environment has no Cargo executable, so local `cargo test`/`cargo clippy` could not be rerun.

</details>

## Previous Review and External Feedback

### @chaodu-agent (round 1)

> The logical delete target is non-unique, and post-ECS cleanup is not fenced against a concurrent apply.

- ⚠️ **Logical target collision remains open as F1** and directly violates frozen RC-1 Acceptance Criterion 2.
- ℹ️ **The concurrent apply/delete fence is now an explicitly accepted RC-1 residual risk**, conditional on caller serialization. It is no longer a blocker for this PR; only the missing public precondition documentation remains as F3.
- ✅ Earlier drain false-success, warning propagation, progress-output, missing-service ambiguity, and name-only dependency cleanup findings remain resolved on this head.

### @chaodu-agent (pre-merge real-AWS verification)

> The exact checkpoint retry completed ECS + Cloud Map + API Gateway + S3 teardown and removed the checkpoint last.

✅ **Confirmed for the exercised canonical target:** the code and E2E evidence support exact checkpoint retry and fail-closed absence handling. The E2E did not exercise a colliding logical pair or an interrupted apply-side ingress teardown checkpoint, so it does not resolve F1 or F2.

No other submitted reviewer findings were present at this head. The automated stale-contributor notice is operational rather than implementation feedback; this new canonical comment supplies the requested PR activity.

<details>
<summary>Review Coverage</summary>

| Reviewer | Focus | Result |
|---|---|---|
| Reviewer A | Destructive correctness and identity | F1; exact checkpoint paths confirmed |
| Reviewer B | API architecture and contract scope | F2, F3; shared fence correctly treated as accepted follow-up |
| Reviewer C | Tests and docs | F3, F4 |
| Reviewer D | Security and caller boundaries | No additional in-scope blocker beyond F1 |
| Reviewer E | Independent audit | Deduplicated findings and confirmed changed-line eligibility |

</details>

<details>

- Initial missing, empty-success, mixed-failure, and duplicate service responses fail closed.
- Exact cluster/service/registry boundaries and service incarnation are revalidated before destructive steps.
- Dependent cleanup uses checkpointed IDs instead of first-name matches.
- Drain exhaustion cannot become a successful report and retains retry identity.
- Library progress suppression and CLI rendering separation are cleanly structured.
- Structured errors preserve the failed target and prior completed report by design.

</details>

1. **The CLI already performs teardown** — if every control-plane caller can safely execute it, the public library contract adds compatibility and testing surface.
2. **Logical names may be the wrong destructive identity** — a durable opaque deployment ID could be safer long-term than deriving ownership from two free-form strings.
3. **Tenant deprovisioning spans more than ECS/ingress/S3** — orchestration may ultimately belong in the control plane, which also owns webhook, disclosure, secret, and serialization policy.

chaodu-agent (commented):
> [!IMPORTANT]
> **CHANGES REQUESTED ⚠️** — RC-1 is frozen; logical identity collision plus cleanup, documentation, and test gaps remain.

Consolidated review: https://github.com/openabdev/openab/pull/1415#issuecomment-5012152804

Contract freeze: `RC-1` at `1bd61e825e3f2e1308bde3be6294e10b0e3bead6`; SHA-256 `a2377a772fc9ba7f4856cea6af2cc6f6596dba919f12256295fd65516de3c674`.

GitHub event: COMMENT — self-review delivery only; this is not an approval.

chaodu-agent:
> [!IMPORTANT]
> **CHANGES REQUESTED ⚠️** — The new delete-only namespace guard does not make the shared apply/delete resource identity injective, so a valid delete target can still remove a deployment created under a colliding logical pair.

## What This PR Does

This PR adds a public in-process `delete_services` API for explicit OAB `namespace` + `name` targets. It resolves the shared control-plane bucket chain, records exact ECS and ingress identity before mutation, reports structured partial results, suppresses library progress output, and preserves the existing CLI flow.

## How It Works

Each target is validated and resolved to one live ECS service. Before mutation, delete stores a durable S3 checkpoint containing the caller boundary, bucket, canonical cluster/service ARNs, ECS `createdAt` incarnation, registry ARN, and API ID. It then drains that exact ECS incarnation, removes exact ingress dependencies, deletes the target manifest/artifacts, and removes the delete checkpoint last. Matching retries resume from the checkpoint; initial missing or ambiguous identity fails closed. This revision also blocks delete while apply has a pending ingress-teardown checkpoint and documents the caller-side same-target serialization precondition.

## Findings

| 1 | 🔴 Critical | The collision guard exists only in programmatic delete. Apply still accepts hyphenated namespaces and every CLI delete path bypasses the guard, so `prod/team-bot` can still target the deployment created as `prod-team/bot`; both resolve to `oab-prod-team-bot`. | `operator/src/delete.rs:426-431` |
| 2 | 🟢 Praise | The pending ingress-checkpoint guard runs before identity preparation, concurrency recovery is now documented across the public API, and the partial-report helper preserves completed targets on a later failure. | — |

<details>

### 🔴 F1 — Apply and delete still operate over different logical identity domains

The new validation rejects `-` only in `DeleteTarget.namespace`, and only through `delete_services`. It does not establish ownership of the physical resource against identities that apply can create:

- `OABServiceManifest::validate` still checks only that `metadata.namespace` is non-empty (`operator/src/manifest.rs:303-305`).
- The manifest schema explicitly accepts hyphenated namespaces with `^[a-z0-9][a-z0-9-]*$` (`operator/schema/oabservice-v2.json:38`).
- Apply derives ECS and Cloud Map identities with the same delimiter-only `oab-{namespace}-{name}` mapping (`operator/src/manifest.rs:354-361`).
- CLI delete dispatches through `run` / `run_from_file`, which call `run_with_bucket` directly rather than `validate_delete_request` (`operator/src/cli.rs:188-197`, `operator/src/delete.rs:573-618`).

Consequently, both `prod/team-bot` and `prod-team/bot` remain valid apply identities and both map to `oab-prod-team-bot`. The former remains a valid programmatic delete target, so it can resolve, checkpoint, and remove the latter's live service. The pending ingress-teardown checkpoint does not close this path because that checkpoint is keyed by the original logical `namespace/name`; an alias uses a different S3 key.

**Requested change:** enforce one shared injective identity/ownership rule before AWS calls across manifest/fleet apply and every programmatic and CLI delete entry point. Include an explicit migration or legacy-deployment policy rather than silently making existing hyphenated namespaces undeletable. Extend the regression to exercise both apply and delete entry points and prove that colliding logical pairs cannot overwrite, checkpoint, or delete one another.

### 🟢 F2 — The other RC-1 follow-ups are materially improved

Delete now checks the apply-side ingress-teardown checkpoint before identity preparation and destructive mutation. The public apply/delete Rustdocs, crate docs, README, and CLI guide state the full caller serialization boundary and recovery action. The extracted result-folding helper also preserves prior completed targets when a later target fails.

</details>

<details>

- PR opened: 2026-07-16
- Reviewed head: `54bd8f28eda19c3bd209789be89c44740d3a55f1`
- Previous reviewed head: `1bd61e825e3f2e1308bde3be6294e10b0e3bead6`
- Main checked at: `f8e91565a4e044892515067532be9ad3100dd2b6`
- Main already has: no public `delete_services`, `DeleteTarget`, or `DeleteOptions` API
- Net-new value: a typed in-process delete API with exact durable teardown identity and structured partial results
- Validation: `git diff --check` passed; the exact-head `operator` check passed. Cargo is unavailable in the review environment, so local Rust tests and clippy could not be rerun.

</details>

## Previous Review and External Feedback

### Previous consolidated review (RC-1 round 2)

> Logical target collisions, pending apply-side cleanup state, missing concurrency documentation, and required regressions remained open.

- ⚠️ **Logical target collision remains open as F1.** The new test demonstrates the collision but only rejects one alias at the programmatic-delete boundary; it does not restrict apply's identity domain or CLI delete.
- ✅ **Pending apply-side ingress teardown is addressed safely.** Delete now fails before destructive mutation and directs the caller to finish the ingress-free apply first.
- ✅ **The caller serialization precondition is documented.** Public apply/delete Rustdocs and user documentation include the full boundary and recovery rule.
- ℹ️ **Partial-report coverage was added through the extracted result-folding helper.** The collision regression still needs cross-entry-point coverage as part of F1.

### Prior real-AWS verification

> Exact-checkpoint retry completed ECS, Cloud Map, API Gateway, and S3 teardown and removed the checkpoint last.

✅ This evidence remains valid for the exercised canonical target and supports the checkpoint/retry design. It did not exercise two colliding logical pairs, so it cannot establish ownership before the initial exact checkpoint is minted.

### Automated stale-contributor notice

The notice is operational rather than implementation feedback. This new exact-head review supplies current PR activity; it does not affect the technical verdict.

No other external reviewer findings were present on the reviewed head.

<details>
<summary>Review Coverage</summary>

| Reviewer | Focus | Result |
|---|---|---|
| Reviewer A | Destructive correctness and identity | F1 reproduced across apply and delete domains |
| Reviewer B | API architecture and RC-1 contract | F1; checkpoint and documentation findings closed |
| Reviewer C | Tests and changed-line eligibility | Collision test needs cross-entry-point coverage |
| Reviewer D | Independent exact-SHA audit | Confirmed CLI bypass and alias-keyed checkpoint gap |

</details>

<details>

- Initial missing, empty-success, mixed-failure, and duplicate service responses remain fail-closed.
- Exact cluster/service/registry boundaries and service incarnation are revalidated before destructive steps.
- The pending apply-side teardown checkpoint is checked before delete identity preparation.
- Concurrency and recovery responsibilities are now explicit in public and user-facing documentation.
- Structured teardown errors preserve the failed target and prior completed report.
- Exact-head operator CI passed.

</details>

1. **The CLI already performs teardown** — if every control-plane caller can safely execute it, the public library contract adds compatibility and testing surface.
2. **Logical names may be the wrong destructive identity** — a durable opaque deployment ID or immutable ownership metadata would be safer than reconstructing ownership from two free-form strings.
3. **Tenant deprovisioning spans more than ECS/ingress/S3** — orchestration may ultimately belong in the control plane, which also owns webhook, disclosure, secret, and serialization policy.

operator/src/delete.rs
```diff
@@ -1,7 +1,515 @@
...
+    }
+    for target in targets {
+        if target.namespace.trim().is_empty() || target.name.trim().is_empty() {
+            return Err(DeleteError::validation(anyhow::anyhow!(
+                "delete target namespace and name must not be empty (got '{}'/'{}')",
+                target.namespace,
+                target.name
+            )));
+        }
+        if target.namespace.contains('-') {
```
chaodu-agent: **🔴 F1 — Apply and delete still operate over different logical identity domains**

This guard rejects hyphenated namespaces only for programmatic delete, while manifest/apply validation still accepts them and CLI delete bypasses this function. Therefore `prod/team-bot` remains an accepted delete alias for a deployment created as `prod-team/bot`; both derive `oab-prod-team-bot`, and their logical checkpoint keys differ.

**Requested change:** enforce one shared injective identity/ownership rule before AWS calls across manifest/fleet apply and every programmatic and CLI delete entry point, with an explicit legacy migration policy and a cross-entry-point regression.

chaodu-agent (commented):
> [!IMPORTANT]
> **CHANGES REQUESTED ⚠️** — The delete-only namespace guard does not close the shared apply/CLI identity collision.

Consolidated review: https://github.com/openabdev/openab/pull/1415#issuecomment-5012696527

GitHub event: COMMENT — self-review delivery only; this is not an approval.

chaodu-agent:
> [!NOTE]
> **LGTM ✅** — The exact requested head closes the prior identity-collision and teardown-safety findings; no new in-scope blockers were found.

## What This PR Does

This PR adds a public, in-process `delete_services` API for explicitly identified OAB deployments, with structured partial results, exact resumable teardown checkpoints, and no dependency on CLI home configuration. It also makes the apply/delete physical identity rule shared and injective while preserving a guarded CLI migration path for legacy hyphenated namespaces.

## How It Works

`identity.rs` centralizes physical resource names and control-plane keys. Apply and programmatic delete reject hyphenated namespaces before AWS calls; the CLI can still delete legacy namespaces, but checks all alternative logical splits in the control plane and fails closed when an alias owns the same physical identity. Apply uses the same ownership check before reconciliation.

Delete resolves the explicit cluster and shared bucket chain, records exact ECS cluster/service ARNs, the `createdAt` incarnation, and ingress IDs before mutation, then drains and removes only checkpointed resources. Ambiguous ECS responses, mismatched incarnations, and cleanup failures retain retry state. Existing CLI progress remains enabled while programmatic apply/delete use task-local suppression.

## Findings

| 1 | 🟢 | Shared identity derivation and symmetric alias ownership checks prevent distinct logical pairs from silently targeting the same physical deployment. | `operator/src/identity.rs:38-167` |
| 2 | 🟢 | Exact checkpoint/incarnation validation, fail-closed ECS classification, and checkpoint-last cleanup preserve safe retry semantics. | `operator/src/delete.rs:706-1085` |
| 3 | 🟢 | The public contract, legacy migration behavior, caller serialization precondition, and focused collision/partial-report tests are documented and covered across apply, delete, and CLI paths. | `operator/src/lib.rs:1-92`, `operator/src/delete.rs:1415-1487` |

<details>

### 🟢 F1: Shared physical identity is guarded consistently

The new identity module gives apply and delete one source of truth for ECS/Cloud Map names and control-plane keys. It rejects ambiguous hyphenated namespaces for the normal apply/programmatic-delete domain and probes every alternate split (manifest plus both checkpoint types) before mutation. Legacy CLI deletion remains available but is subject to the same exclusive-ownership check.

### 🟢 F2: Destructive teardown remains fail-closed and resumable

The reviewed path persists the exact caller boundary, canonical cluster/service ARNs, service incarnation, and ingress identifiers before ECS mutation. It rejects ambiguous or mismatched responses, refuses dependent cleanup until the checkpointed incarnation is unambiguously absent, retains the checkpoint on failure, and removes it only after dependent and S3 cleanup succeeds.

### 🟢 F3: Contract and regression coverage are aligned

The public Rustdocs and user documentation state the no-orphan behavior and same-target serialization requirement. Focused tests cover the shared namespace rule, collision alias enumeration, cross-entry-point identity assertions, partial-report preservation, pending ingress teardown, and drain timeout handling.

</details>

<details>

- PR opened: 2026-07-16
- Reviewed head: `be006ec67f5329df90423785e417e0ffb4671d85`
- Base branch: `refactor/oabctl-lib` at `6dd5285af10a237bef54f7ebeb797dc3b7450d1f`
- Main already has: no public `delete_services`, `DeleteTarget`, or `DeleteOptions` API
- Net-new value: a typed programmatic delete contract with exact resumable identity and shared apply/delete collision protection
- Validation: `git diff --check` passed; `validate`, `changes`, and `poll-and-review` checks passed; the exact-head `operator` check was still in progress when reviewed. Local Cargo/Rust tooling is unavailable in this environment, so local `cargo test` and `cargo clippy` could not be rerun.

</details>

## Previous Review Feedback

- ✅ Drain timeout cannot report a successful deletion; dependent cleanup is blocked until exact ECS absence.
- ✅ Ingress/API cleanup failures and warnings are no longer silently discarded.
- ✅ Initial missing/ambiguous ECS responses and name-only dependent cleanup paths fail closed.
- ✅ The accepted caller-side apply/delete serialization precondition, pending apply-side ingress checkpoint behavior, and partial-report semantics are documented and covered.
- ✅ The remaining logical-identity collision concern is addressed in this head by applying one domain rule to apply and programmatic delete plus symmetric control-plane ownership checks for legacy CLI deletion.

No external submitted review or implementation comment raised feedback on the reviewed head. The automated stale-contributor notice is operational rather than code-review feedback. Prior review concerns listed above were rechecked against the exact requested SHA.

<details>
<summary>Review Coverage</summary>

| Reviewer | Focus | Result |
|---|---|---|
| Reviewer A | Destructive correctness and identity | No new blocker; shared collision guard verified |
| Reviewer B | API architecture and contract scope | Explicit options, checkpoints, and caller preconditions verified |
| Reviewer C | Testing and documentation | Focused identity, timeout, and partial-report coverage verified |
| Reviewer D | Safety and operability | Fail-closed response handling and retry retention verified |
| Reviewer E | Independent exact-SHA audit | No blocker found |

</details>

<details>

- Physical identity and control-plane key construction are centralized instead of duplicated.
- Legacy namespace deletion is preserved without weakening the normal accepted identity domain.
- Exact ECS incarnation and caller-boundary checks prevent stale checkpoints from authorizing another deployment.
- Programmatic output suppression is separated from existing CLI rendering.
- The structured report/error types preserve successful prior targets when a later target fails.

</details>

1. **The existing CLI already performs teardown** — if every control-plane caller can safely invoke it, a second public contract adds compatibility and maintenance surface.
2. **Logical names remain a long-term identity choice** — a durable opaque deployment ID could eventually be safer than deriving ownership from namespace and name.
3. **Tenant deprovisioning spans more than this crate** — webhook, disclosure, secret, and concurrency orchestration may ultimately belong in the control plane.

chaodu-agent (commented):
> [!NOTE]
> **LGTM ✅** — The exact requested head closes the prior identity-collision and teardown-safety findings; no new in-scope blockers were found.

Consolidated review: https://github.com/openabdev/openab/pull/1415#issuecomment-5029395797

GitHub event: COMMENT — self-review delivery only; this is not an approval.

chaodu-agent:
> [!NOTE]
> **LGTM ✅** — The round-3 blocker is closed: apply and every delete entry point now share one injective identity rule plus a pre-mutation control-plane ownership check, with an explicit legacy-namespace policy.

## What This PR Does

Adds a public in-process `delete_services` API for explicit OAB `namespace`/`name` targets, symmetric with `apply_manifests`: explicit cluster, shared bucket-resolution chain, no `~/.oabctl/config.toml` reads, durable exact-identity checkpoints before mutation, structured partial results, and suppressed progress output for programmatic callers. The CLI keeps its existing flow.

## How It Works

Each target resolves to one live ECS service. Before mutation, delete stores a durable S3 checkpoint (caller boundary, bucket, canonical cluster/service ARNs, ECS `createdAt` incarnation, registry ARN, API ID), then drains that exact incarnation, removes exact ingress dependencies, deletes the manifest/artifacts, and removes the checkpoint last. Ambiguous AWS identity responses fail closed; matching retries resume from the checkpoint.

New in this revision, a shared `operator/src/identity.rs` module owns logical→physical identity:

- **Domain rule** — `validate_injective_identity` (non-empty components, hyphen-free namespace) runs before any AWS call in manifest validation (CLI apply, fleet expansion, programmatic apply) and in programmatic delete, so `oab-{namespace}-{name}` parses back to exactly one accepted pair.
- **Ownership rule** — `ensure_exclusive_physical_identity` runs pre-mutation in `apply_ecs` and in the shared delete choke point `run_with_bucket` (all CLI and programmatic delete paths). It probes the exact control-plane keys (manifest, delete checkpoint, ingress-teardown checkpoint) of every colliding alias split and fails closed naming the contested pair. Non-NotFound probe errors also fail closed.
- **No drift** — all ECS/Cloud Map names and control-plane keys (manifest, delete-checkpoint, ingress-teardown-checkpoint) derive from the shared module; a regression asserts probe keys equal mutation keys.
- **Legacy policy** — hyphenated-namespace manifests are rejected at apply with a migration message; existing legacy deployments remain deletable via `oabctl delete` (warning + ownership check). Schema `metadata.namespace` tightened to `^[a-z0-9]+$`.

## Findings

| 1 | 🟢 Praise | Round-3 F1 is closed at every entry point: `prod/team-bot` vs `prod-team/bot` can no longer overwrite, checkpoint, or delete one another — the hyphenated side is outside the accepted domain, and the accepted side is blocked by the ownership probe whenever the legacy pair is recorded. | `operator/src/identity.rs` |
| 2 | 🟢 Praise | The legacy-deployment policy is explicit rather than silent: apply fails with a migration hint, CLI delete warns and proceeds only under the ownership check. | `operator/src/delete.rs`, `operator/src/manifest.rs` |
| 3 | 🟢 Praise | Cross-entry-point regressions cover apply-side rejection/acceptance, programmatic-delete rejection, alias symmetry, and probe/mutation key alignment. | `operator/src/{identity,manifest,apply,delete}.rs` tests |

<details>

### 🟢 F1 (round-3) — Shared injective identity/ownership rule, verified per entry point

Independent exact-SHA audit of `54bd8f2..be006ec` traced every entry point:

- **Apply:** CLI `apply -f` and programmatic `apply_manifests` both reach `validate_apply_request` → `OABServiceManifest::validate` → `validate_injective_identity`; fleet manifests are validated post-expansion. `apply_ecs` runs the ownership probe before reading generation state or mutating.
- **Delete:** CLI `delete <name>`, CLI `delete -f`, and programmatic `delete_services` all funnel through `run_with_bucket`, which runs the ownership probe before the pending-ingress-teardown guard, identity preparation, and any checkpoint/ECS mutation. Programmatic targets are additionally domain-rule-rejected pre-AWS in `validate_delete_request`.
- **Probe correctness:** `collision_aliases` enumerates every alternative split point of `{namespace}-{name}`; within the hyphen-free-namespace domain the mapping is injective by construction, so collisions can only involve recorded legacy pairs — exactly what the probe checks. Probed keys are derived from the same helpers the mutation paths use.

### ℹ️ Residual notes (within RC-1's accepted risks, documented)

- The ownership check treats the control plane as the source of truth. A legacy deployment whose stored manifest was manually removed is invisible to the probe. Apply always records manifests and delete removes them only after full teardown (checkpoint last), so system-created states are covered; manual S3 tampering stays out of scope, consistent with RC-1's no-name-only-guessing stance. The existing "durable opaque deployment ID" follow-up would close this fully.
- The probe-then-mutate sequence is not atomic. Same-physical-identity concurrency is governed by the same caller-serialization precondition RC-1 already documents; the generation/lease-fence follow-up remains the structural fix.

</details>

<details>

- PR opened: 2026-07-16
- Reviewed head: `be006ec67f5329df90423785e417e0ffb4671d85`
- Previous reviewed head: `54bd8f28eda19c3bd209789be89c44740d3a55f1`
- Main checked at: `d8d4bf15ca9a6e7380a5683a87527b93f527badb`
- Main already has: no public `delete_services`/`DeleteTarget`/`DeleteOptions` API and no shared identity module
- Net-new value since round 3: `operator/src/identity.rs` (domain rule, alias enumeration, ownership probe), wiring into manifest validation / `apply_ecs` / `validate_delete_request` / `run_with_bucket`, schema namespace pattern `^[a-z0-9]+$`, legacy policy docs, and cross-entry-point regressions
- `git diff --check` clean on the delta

</details>

<details>
<summary>Verified</summary>

On the exact reviewed head `be006ec`:

- `cargo clippy --all-targets -- -D warnings` — clean
- `cargo test` — 113 unit tests + 1 doctest passed, 0 failed
- Exact-head CI: `operator` job (check, clippy, test, build) ✅, `Review Contract` ✅, all other required checks ✅
- Independent exact-SHA audit (entry-point tracing, probe-key/mutation-key alignment, legacy policy, offline-safety of new tests) — passed
- Note: the crate is intentionally not blanket-`rustfmt`ed in this commit; the pre-existing head was not fmt-clean and the operator CI job has no fmt gate, so the delta stays minimal for review.

</details>

## Previous Review and External Feedback

### Round-3 consolidated review

> 🔴 F1 — the collision guard exists only in programmatic delete; apply still accepts hyphenated namespaces and every CLI delete path bypasses the guard. Requested: one shared injective identity/ownership rule across all entry points, an explicit legacy-deployment policy, and cross-entry-point regressions.

✅ **Addressed in `be006ec`.** Shared rule in `operator/src/identity.rs`, enforced at every apply and delete entry point (domain rule pre-AWS, ownership probe pre-mutation); legacy hyphenated namespaces are apply-rejected with a migration message but remain CLI-deletable under the ownership check; regressions exercise both apply and delete sides of the colliding pair, including the alias-keyed checkpoint gap called out in round 3 (the probe checks the alias's manifest, delete-checkpoint, and ingress-teardown-checkpoint keys).

### Prior real-AWS verification

> Exact-checkpoint retry completed ECS, Cloud Map, API Gateway, and S3 teardown and removed the checkpoint last.

✅ Still valid for the canonical path; this revision adds only pre-mutation validation/probing before that path, plus read-only S3 probes that fail closed.

### Automated stale-contributor notice

Operational only; superseded by current PR activity.

No other external reviewer findings were present on the reviewed head.

<details>

- One identity module now owns every physical-name and control-plane-key derivation — probe and mutation cannot drift.
- Fail-closed posture is preserved end-to-end: ambiguous ECS responses, contested identities, and probe errors all stop before destructive mutation.
- The breaking change is explicit: `!` commit, migration message in the error itself, README/CLI-guide policy sections, and a tightened published schema.
- Programmatic progress suppression and structured partial reports are unchanged and still covered by tests.

</details>

1. **The CLI already performs teardown** — if every control-plane caller could shell out safely, the public library contract adds compatibility and testing surface.
2. **Logical names may be the wrong destructive identity** — a durable opaque deployment ID would eliminate the alias problem structurally instead of validating around a string encoding; it remains the listed follow-up.
3. **Tenant deprovisioning spans more than ECS/ingress/S3** — webhook, disclosure, secret, and serialization policy still live in the control plane, which could argue for orchestrating deletion there entirely.

chaodu-agent (commented):
> [!NOTE]
> **LGTM ✅** — Round-3 F1 (non-injective apply/delete identity) is closed on head `be006ec` by a shared identity module enforced at every apply and delete entry point, with an explicit legacy-namespace policy and cross-entry-point regressions.

Consolidated review: https://github.com/openabdev/openab/pull/1415#issuecomment-5029564668

Issue #703, opened 2026-05-02
feat(gateway): add Feishu/Lark adapter with WebSocket long-connection


wangyuyan-agent:
### Description

Add a Feishu/Lark adapter to the Custom Gateway, enabling OpenAB to receive and reply to messages from Feishu (China) and Lark (international).

The adapter supports two connection modes:
- **WebSocket long-connection** (default) — gateway connects outbound to Feishu, no public URL required
- **HTTP Webhook** (fallback) — for environments where outbound WebSocket is blocked

This is the first gateway adapter that uses an outbound long-connection rather than inbound webhooks.

```
                     External                          Internal

Telegram  --POST-->  ┌─────────────────┐
LINE      --POST-->  │  OpenAB Gateway │ <--WS-- OAB Pod
Teams     --POST-->  │    :8080        │
                     │  ┌───────────┐  │
Feishu    <--WS----  │  │ Feishu WS │  │
(default)            │  │ Client    │  │
                     │  └───────────┘  │
Feishu    --POST-->  │  (fallback)     │
                     └─────────────────┘
```

### Use Case

1. **Chinese enterprise teams** — Feishu is where most Chinese tech teams collaborate daily. Without Feishu support, OpenAB cannot serve this market.
2. **Private deployment without public IP** — WebSocket mode lets operators run behind NAT/firewall without HTTPS endpoints or Cloudflare Tunnel.
3. **Multi-platform agent** — Teams using both Discord and Feishu can run the same OAB agent on both platforms with per-platform @mention gating.
4. **Lark international** — Same adapter supports Lark by setting `FEISHU_DOMAIN=lark`.

### Proposed Solution

Add `gateway/src/adapters/feishu.rs` following the existing adapter pattern:

```
Feishu Cloud <--WS--> Gateway (feishu.rs) <--WS--> OAB Core --> Agent
```

Key components:
- **WebSocket**: POST /callback/ws/endpoint (AppID+AppSecret auth), protobuf frame decoding (prost), JSON event extraction
- **Webhook fallback**: URL challenge, verification token, encrypt key (AES-256-CBC), signature verification, rate limiting
- **Token management**: tenant_access_token with auto-refresh
- **Reactions**: Map OAB emoji to Feishu reaction API on original messages
- **Mention gating**: Gateway-side filtering (FEISHU_REQUIRE_MENTION) + OAB-side (bot_username = bot open_id)
- **Deduplication**: event_id + message_id TTL cache

```
Inbound:
[...]

PR #1440, opened 2026-07-23
fix(feishu): start websocket in unified mode

SunnyYYLin:

Unified mode registers the Feishu webhook route but does not resolve the bot identity or start the Feishu WebSocket client. Since WebSocket is the default connection mode, a configured Feishu bot can start without receiving inbound events. This is tracked under #1356.

Discord Discussion URL: https://discord.com/channels/1491295327620169908/1529851383321329765

This is a minimal rebase of @angmeng (Jaz Lim)'s prior closed implementation in #1293.

### Goal

Make unified mode receive Feishu events in its default WebSocket configuration by matching the standalone gateway startup sequence: resolve bot identity, start the card idle reaper when applicable, and start the WebSocket client.

### Non-goals

This PR does not change Feishu webhook routing, configuration precedence, CardKit rendering, card streaming behavior, proxy/TLS handling, or any non-Feishu unified adapter.

### Accepted Residual Risks

The WebSocket shutdown sender is intentionally retained for the process lifetime because dropping it terminates the reconnect loop. As with the standalone gateway, shutdown is process-lifecycle based; operators recover from a failed service by restarting OpenAB.

### Acceptance Criteria

- [x] With `connection_mode=websocket`, unified mode resolves the Feishu bot identity and starts a WebSocket task.
- [x] The WebSocket client connects and Feishu direct messages reach the agent.
- [x] Card streaming idle reaping remains enabled for non-Post modes, matching the standalone gateway.
- [x] The change is limited to unified Feishu startup.

### Follow-ups

Config-only unified Feishu activation and the broader webhook/long-connection mode alignment remain tracked under #1356. Card streaming behavior is intentionally deferred to a separate change.

## At a Glance

```text
Feishu Cloud -- WebSocket --> unified OpenAB -- ACP --> agent
                               | resolve bot identity
                               | start idle reaper
                               | retain reconnect task
```

**OpenClaw:** #703 documents the Feishu/Lark adapter's WebSocket long-connection as the default transport: https://github.com/openabdev/openab/issues/703

**Hermes Agent:** #703 records Hermes Agent as prior art for a Feishu adapter supporting WebSocket plus webhook fallback; this PR preserves that outbound-long-connection model in unified mode.

**Other references:** Adapted from @angmeng's verified unified-mode fix in #1293: https://github.com/openabdev/openab/pull/1293

## Proposed Solution

After unified mode registers the Feishu webhook route, mirror the standalone gateway startup path: call `resolve_bot_identity`, spawn `run_idle_reaper` when CardKit streaming is active, and call `start_websocket` for WebSocket connection mode. Keep the watch sender alive so the reconnect loop is not terminated by a dropped sender.

## Why this approach?

It reuses the already-established standalone gateway lifecycle and the prior #1293 implementation instead of introducing a separate unified-only receiver. The change is small, preserves webhook support, and restores compatibility with Feishu's default WebSocket mode.

## Alternatives Considered

- Require operators to use public HTTPS webhooks: rejected because it contradicts Feishu's default long-connection mode and adds unnecessary deployment infrastructure.
- Move all startup lifecycle code into a new abstraction: deferred because it expands the scope beyond restoring missing parity.
- Change CardKit streaming in this PR: deferred to keep runtime delivery behavior isolated from transport startup.

## Validation

- [x] `cargo test --features unified --bin openab has_unified_platform_checks_config_and_env`
- [x] `cargo build --release --features unified`
- [x] `cargo clippy --features unified -- -D warnings`
- [x] Manual Windows validation: deployed the release build behind the configured HTTP proxy; logs confirmed bot identity resolution and a connected Feishu WebSocket; an OpenCode Feishu direct message round-tripped successfully.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -1081,6 +1081,36 @@ async fn main() -> anyhow::Result<()> {
                 );
             }
 
+            // Feishu bot identity + WebSocket long-connection (unified mode).
+            // This mirrors the standalone gateway: the default websocket mode
+            // otherwise mounts only the webhook route and receives no events.
+            #[cfg(feature = "feishu")]
+            if let Some(ref f) = gw_state.feishu {
+                use openab_gateway::adapters::feishu;
+                f.resolve_bot_identity().await;
+                if f.config.streaming_mode != feishu::StreamingMode::Post {
+                    let idle_ms = f.config.card_idle_finalize_ms;
+                    tokio::spawn(feishu::run_idle_reaper(
+                        f.stream_sessions.clone(),
+                        f.token_cache.clone(),
+                        f.client.clone(),
+                        f.config.api_base(),
+                        idle_ms,
+                    ));
+                    info!(idle_ms, "unified: feishu card-streaming idle reaper started");
+                }
+                if f.config.connection_mode == feishu::ConnectionMode::Websocket {
+                    let (shutdown_tx, shutdown_rx) = tokio::sync::watch::channel(false);
+                    match feishu::start_websocket(f, event_tx.clone(), shutdown_rx).await {
+                        Ok(_handle) => info!("unified: feishu websocket task spawned"),
+                        Err(e) => error!(err = %e, "unified: feishu websocket startup failed"),
+                    }
+                    // Keep the sender alive so dropping it cannot terminate the
+                    // reconnect loop through shutdown_rx.changed().
+                    std::mem::forget(shutdown_tx);
+                }
+            }
+
             #[cfg(feature = "wecom")]
             if let Some(ref w) = gw_state.wecom {
                 info!(path = %w.config.webhook_path, "unified: wecom adapter enabled");


wangyuyan-agent (approved):
Reviewed against the current `main` tip (clean rebase, 0 commits behind) with a full read of the changed startup path and the standalone `run_gateway` it mirrors. This is a correct, well-scoped fix — approving. The two items below are non-blocking suggestions, not merge blockers.

**What it fixes.** The added block faithfully reproduces the standalone Feishu startup — `resolve_bot_identity()`, the non-Post idle reaper, and `start_websocket()` under `ConnectionMode::Websocket`. It closes the real gap where a default (websocket) unified deployment mounts the webhook route but never connects, so no inbound events arrive and `@mention` gating can't resolve the bot's own open_id. No behavioral impact on the other adapters — the block is `#[cfg(feature = "feishu")]` + `if let Some(f) = gw_state.feishu` and only reads `f.*` / clones `event_tx`.

**Non-blocking suggestions:**

1. `std::mem::forget(shutdown_tx)` is load-bearing here — dropping the sender at block scope would fire `shutdown_rx.changed()` and kill the reconnect loop, and the leak is a single `watch::Sender<bool>`, so it is harmless. It is, however, the one spot that diverges from the standalone path you are mirroring, where `feishu_shutdown_tx` lives in an outer scope and is `drop`ped after `serve` returns. Consider threading the sender out of the unified setup block so it lives for the process lifetime the same way — that restores true parity and keeps a graceful-shutdown path instead of a permanent leak.
2. The WS client (a producer on the `event_tx` broadcast) is spawned before the bridge task calls `subscribe()`. In practice the WS handshake is far slower than the synchronous setup in between, so nothing is lost — and this is actually tighter than the standalone path. To close the theoretical window entirely, subscribe the bridge before `start_websocket()`.

**Pre-existing (not introduced here — belongs to #1356):** unified passes `warn_unenforceable_l1(true)` because the webhook route is mounted unconditionally, so a websocket-mode deployment still emits the missing-L1-secret warning even though L1 does not apply on the WS path. Now that this PR makes WS mode functional, that false positive is more visible — worth folding into the mount-by-mode alignment already tracked on #1356.

Scope looks right: leaving the unconditional webhook mount alone and deferring mount-by-mode to #1356 keeps this minimal and additive. No double-delivery concern — Feishu delivers over a single connection mode, and the webhook/WS paths share the same dedupe cache regardless.

PR #1457, opened 2026-07-25
fix(pool): prevent panic on Instant subtraction underflow

### LINKED ISSUE #1459: bug(pool): panic on Instant subtraction underflow when uptime < session TTL
## Description

`Instant::now() - ttl` panics with `overflow when subtracting duration` when the monotonic clock has not yet elapsed `ttl` worth of time. This is a logic error, not a platform-specific bug.

## Root cause

`pool.rs` computes `let cutoff = Instant::now() - ttl`. Rust's `Instant - Duration` calls `checked_sub().expect()` - it panics on underflow. The monotonic clock starts at boot on all platforms, so this panics whenever **uptime < TTL** (e.g. `session_ttl_hours=48` on a machine with < 48h uptime).

## Why was this never reported?

1. Linux servers typically have uptimes of weeks/months, far exceeding the 48h default TTL
2. The panic only triggers on the reaper tick (every ~60s), not at startup
3. The bug is platform-agnostic - a freshly booted Linux machine would also panic

## Expected behavior

The reaper should classify sessions as fresh (not idle) when the elapsed time cannot be computed, rather than panicking.

## Reproduction

1. Set `session_ttl_hours = 48` in openab config
2. Boot any machine (Linux or Windows)
3. Start openab before uptime reaches 48h
4. Wait for first reaper tick (~60s)
5. Observe panic: `overflow when subtracting duration from instant`

## Proposed fix

Replace `Instant::now() - ttl` with `now.saturating_duration_since(last_active) > ttl`. Returns `Duration::ZERO` on underflow instead of panicking.

## Relevant code

- `crates/openab-core/src/acp/pool.rs` - `classify_idle` function and reaper loop

SunnyYYLin:

`Instant::now() - ttl` panics when the monotonic clock has not yet elapsed `ttl` worth of time. This is a logic error in the original code, not a platform-specific bug.

Closes #1459

Discord Discussion URL: https://discordapp.com/channels/1491295327620169908/1530676804996825088

### Goal
Fix the panic caused by `Instant::now() - ttl` when the elapsed time since the monotonic clock epoch is less than `ttl` (e.g. `session_ttl_hours=48` on a machine with < 48h uptime).

### Non-goals
No behavioral change. No change to TTL semantics.

### Accepted Residual Risks
None - `saturating_duration_since` returns `Duration::ZERO` on underflow, which correctly classifies the session as fresh (not idle).

### Acceptance Criteria
- [x] `cargo test -p openab-core classify_idle --lib` passes (3 tests)
- [x] `cargo clippy --workspace --features unified -- -D warnings` clean
- [x] No behavioral change when `now > last_active + ttl` (normal case)

### Follow-ups
No deferred work - the fix is complete and self-contained.

## Why was this never reported?

`Instant - Duration` in Rust calls `checked_sub().expect()` - it panics on underflow. The original code computes `Instant::now() - ttl` to get a cutoff point, then compares `last_active < cutoff`.

This panics whenever the monotonic clock has not elapsed `ttl` since its epoch. On all platforms the monotonic clock's epoch is boot, so the trigger condition is **uptime < TTL**. The default is `session_ttl_hours = 4` (`config.rs`, `default_ttl_hours()`), so with defaults the window is a host's first 4 hours of uptime; a deployment that raises the TTL widens it proportionally.

It went unreported because **the panic is silent**, not because the window is narrow:

1. The reaper is a bare `tokio::spawn` loop (`main.rs`). Its `JoinHandle` is only `.abort()`ed at shutdown - never awaited, never monitored - and the build uses the default unwind panic strategy with no panic hook. A panic there kills only that task.
2. So the process does not crash. The reaper simply dies on its first tick (60s after startup) and never runs again, leaving one panic line in the log.
3. The visible symptom is therefore indirect and slow: idle sessions stop expiring and accumulate toward `max_sessions`, and hung-session force-eviction - which also lives in `cleanup_idle` - stops running too.

The bug is platform-agnostic; the `Windows` framing in the first commit's subject is incidental to where it was first hit. Any freshly booted host, container, or CI runner is affected, since a container observes the host kernel's `CLOCK_MONOTONIC` rather than its own start time.

## Proposed Solution

Replace `let cutoff = Instant::now() - ttl; last_active < cutoff` with `now.saturating_duration_since(last_active) > ttl`. `last_active` is always <= `now`, so `saturating_duration_since` never underflows.

## At a Glance

```
Before (panics when uptime < TTL):
  cutoff = Instant::now() - ttl        ← underflow
  last_active < cutoff

After (always safe):
  now.saturating_duration_since(last_active) > ttl
```

Not applicable - trivial arithmetic safety fix.

## Alternatives Considered

- `checked_sub` + unwrap_or: more verbose, same effect.
- `Instant::now().checked_duration_since(last_active)`: returns Option, requires extra branching.

## Validation

- [x] `cargo test -p openab-core classify_idle --lib` - 3 passed
- [x] `cargo clippy --workspace --features unified -- -D warnings` - clean

### DIFF
--- a/crates/openab-core/src/acp/pool.rs
+++ b/crates/openab-core/src/acp/pool.rs
@@ -862,8 +862,12 @@ mod tests {
     #[test]
     fn classify_idle_marks_stale_by_time() {
         let ttl = std::time::Duration::from_secs(60);
-        let now = Instant::now();
-        let last_active = now - std::time::Duration::from_secs(120);
+        // Build fixtures forward from `last_active`, never backward from `now`:
+        // `Instant::now() - d` carries the very underflow panic this PR removes
+        // from production, and would fire on a host whose monotonic clock is
+        // younger than `d`.
+        let last_active = Instant::now();
+        let now = last_active + ttl + std::time::Duration::from_secs(60);
         assert!(classify_idle(last_active, true, now, ttl));
     }
 
@@ -887,8 +891,9 @@ mod tests {
     #[test]
     fn classify_idle_keeps_sessions_at_the_exact_ttl_boundary() {
         let ttl = std::time::Duration::from_secs(60);
-        let now = Instant::now();
-        assert!(!classify_idle(now - ttl, true, now, ttl));
+        let last_active = Instant::now();
+        let now = last_active + ttl;
+        assert!(!classify_idle(last_active, true, now, ttl));
     }
 
     /// `now` is sampled once at the top of `cleanup_idle`, but a turn can finish
--- a/crates/openab-core/src/acp/pool.rs
+++ b/crates/openab-core/src/acp/pool.rs
@@ -816,7 +816,7 @@ impl SessionPool {
 mod tests {
     use super::{
         better_candidate, classify_hung, classify_idle, get_or_insert_gate, purge_session_entries,
-        remove_if_same_handle, PoolState,
+        remove_if_same_handle, PoolState, SessionPool,
     };
     use crate::acp::connection::SessionActivity;
     use std::collections::HashMap;
@@ -861,21 +861,72 @@ mod tests {
 
     #[test]
     fn classify_idle_marks_stale_by_time() {
+        let ttl = std::time::Duration::from_secs(60);
         let now = Instant::now();
         let last_active = now - std::time::Duration::from_secs(120);
-        assert!(classify_idle(last_active, true, now, std::time::Duration::from_secs(60)));
+        assert!(classify_idle(last_active, true, now, ttl));
     }
 
     #[test]
     fn classify_idle_marks_stale_by_death() {
+        let ttl = std::time::Duration::from_secs(60);
         let now = Instant::now();
-        assert!(classify_idle(now, false, now, std::time::Duration::from_secs(60)));
+        assert!(classify_idle(now, false, now, ttl));
     }
 
     #[test]
     fn classify_idle_keeps_fresh_alive_sessions() {
+        let ttl = std::time::Duration::from_secs(60);
         let now = Instant::now();
-        assert!(!classify_idle(now, true, now, std::time::Duration::from_secs(60)));
+        assert!(!classify_idle(now, true, now, ttl));
+    }
+
+    /// The expiry boundary is strict: elapsed == ttl is still fresh. Pins that
+    /// the switch from `last_active < now - ttl` to `elapsed > ttl` did not
+    /// shift the comparison by one tick.
+    #[test]
+    fn classify_idle_keeps_sessions_at_the_exact_ttl_boundary() {
+        let ttl = std::time::Duration::from_secs(60);
+        let now = Instant::now();
+        assert!(!classify_idle(now - ttl, true, now, ttl));
+    }
+
+    /// `now` is sampled once at the top of `cleanup_idle`, but a turn can finish
+    /// and touch `last_active` before this connection's `try_lock` is reached —
+    /// so `last_active > now` is reachable in production. `saturating_duration_since`
+    /// yields ZERO there, which must classify the session as fresh rather than
+    /// underflowing.
+    #[test]
+    fn classify_idle_keeps_sessions_touched_after_now() {
+        let ttl = std::time::Duration::from_secs(60);
+        let now = Instant::now();
+        assert!(!classify_idle(now + ttl, true, now, ttl));
+    }
+
+    /// Regression for the panic this PR fixes: `cleanup_idle` used to compute
+    /// `Instant::now() - ttl` before touching the pool, which panics whenever the
+    /// TTL exceeds the monotonic clock's age (uptime). An oversized TTL reproduces
+    /// that unconditionally on every platform — this test panics on the old
+    /// implementation and completes on the current one. The helper tests above
+    /// cannot catch it: the panic happened before `classify_idle` was called.
+    #[tokio::test]
+    async fn cleanup_idle_survives_ttl_larger_than_monotonic_clock() {
+        let pool = SessionPool::new(
+            crate::config::AgentConfig {
+                command: "/bin/true".into(),
+                args: vec![],
+                working_dir: "/tmp".into(),
+                env: HashMap::new(),
+                inherit_env: vec![],
+                command_explicit: true,
+            },
+            1,
+            crate::config::default_prompt_hard_timeout_secs(),
+            HashMap::new(),
+        );
+
+        // No sessions are needed: the panicking subtraction ran unconditionally.
+        pool.cleanup_idle(u64::MAX).await;
     }
 
     #[test]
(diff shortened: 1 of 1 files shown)


chaodu-agent:
> [!NOTE]
> **LGTM &#x2705;** - Both round-3 findings and the round-4 fixture finding are closed in code; the production fix is minimal and now covered by a regression that provably fails without it.

**Delivery note:** this round's last three commits were authored by the reviewer, so this is a self-review. It records the internal verdict only and is **not** an approval; it does not satisfy required-review protection. An independent approving review is still needed.

## What This PR Does

`cleanup_idle` computed its expiry threshold as `Instant::now() - ttl`. Rust's `Sub<Duration> for Instant` is `checked_sub().expect(...)`, so that subtraction panics whenever the monotonic clock's age is smaller than the TTL. The fix inverts the comparison to `now.saturating_duration_since(last_active) > ttl`, which cannot underflow and keeps the strict expiry boundary.

## How It Works

`cleanup_idle` samples `now` once and converts `ttl_secs` to a `Duration`, then passes both into `classify_idle`. Dead-connection eviction is unchanged. Because `saturating_duration_since` yields `Duration::ZERO` when `last_active > now`, the predicate is also safe for a connection touched after `now` was sampled - which is reachable, since `now` is taken before the per-connection `try_lock` loop.

## Findings

| 1 | &#x1F7E2; | Round-3 F1 closed: a production-path regression now covers the panic, and it is verified to fail without the fix. | `crates/openab-core/src/acp/pool.rs:906-931` |
| 2 | &#x1F7E2; | Round-3 F2 closed: `pool.rs` is rustfmt-clean, achieved without reformatting the rest of the crate. | `crates/openab-core/src/acp/pool.rs` |
| 3 | &#x1F7E2; | Round-4 F1 closed: both stale/boundary fixtures now build forward from `last_active`, so no test depends on host uptime. | `crates/openab-core/src/acp/pool.rs:862-895` |
| 4 | &#x1F7E2; | Predicate coverage is now complete: stale, dead, fresh, exact boundary, and future-timestamp. | `crates/openab-core/src/acp/pool.rs:862-903` |
| 5 | &#x1F7E2; | Description corrected: the stated default TTL was wrong and the root-cause explanation was inaccurate. | PR body |

<details>

### &#x1F7E2; F1: The regression exercises the real failure, and is proven to

The helper tests alone could not close round-3 F1, because the panic happened in `cleanup_idle` *before* `classify_idle` was reached. `cleanup_idle_survives_ttl_larger_than_monotonic_clock` calls the real `cleanup_idle(u64::MAX)` on an empty pool, which is sufficient precisely because the bad subtraction ran before the pool was touched.

Verified by mutation rather than by inspection: reintroducing `let _cutoff = Instant::now() - ttl;` into `cleanup_idle` makes the test fail with `overflow when subtracting duration from instant`, and removing it makes the test pass.

### &#x1F7E2; F2: rustfmt-clean without collateral reformatting

Binding `ttl` to a local keeps each `classify_idle` call on one line. `cargo fmt -p openab-core -- --check` reports zero diffs for `pool.rs`. The rest of `openab-core` has pre-existing formatting drift across roughly twenty files and there is no `cargo fmt` gate in CI, so a crate-wide `cargo fmt` would have buried a three-line production fix in hundreds of unrelated hunks. That was deliberately avoided.

### &#x1F7E2; F3: Fixtures no longer reproduce the bug they test

Round-4 F1 was correct and worth holding the PR over: the stale-by-time fixture used `Instant::now() - 120s` and the newly added boundary fixture used `now - ttl`, both of which carry the very underflow being fixed. Both now construct timestamps forward - `now = last_active + ttl` for the exact boundary, `now = last_active + ttl + 60s` for the stale case - preserving the assertions while removing host-uptime dependence.

### &#x1F7E2; F4: Boundary and future-timestamp cases are meaningful, not decorative

The boundary case pins that `elapsed > ttl` preserved the strict semantics of `last_active < now - ttl`, so the refactor cannot have shifted expiry by one tick. The future-timestamp case is load-bearing rather than theoretical: `now` is sampled at the top of `cleanup_idle`, so a turn can finish and touch a connection before that connection's `try_lock` is reached, making `last_active > now` reachable in production.

### &#x1F7E2; F5: Description accuracy

Two claims in the PR body were wrong and have been corrected in place:

- It stated the default was `session_ttl_hours=48`. The actual default is `4` (`default_ttl_hours()` in `config.rs`), so with defaults the trigger window is a host's first 4 hours of uptime, not 48.
- It attributed the lack of reports to long server uptimes. The verified reason is that the panic is silent: the reaper is a bare `tokio::spawn` loop whose `JoinHandle` is only aborted at shutdown, the build uses the default unwind strategy, and there is no panic hook. The process does not crash; the reaper dies on its first tick and never runs again. The observable consequence is indirect - idle sessions stop expiring toward `max_sessions`, and hung-session force-eviction, which also lives in `cleanup_idle`, stops with it.

This makes the change more valuable than the original description suggested, not less.

</details>

<details>

- PR opened: 2026-07-25T20:24:46Z
- Declared base: `main`
- Resolved base: `origin/main`
- Reviewed head: `803f7ac7397d50e25e6a56a21ad1d6a2d6702ebc`
- Merge-base: `685973307bf79d6848c21b33a5bd55b6aea1c433`
- Main already has: the panic-prone cutoff subtraction in `cleanup_idle`.
- Net-new value: a saturating predicate that removes the panic without changing expiry semantics, plus coverage that fails without it.

</details>

<details>
<summary>Verified</summary>

Run locally at `803f7ac`:

- `cargo test -p openab-core --lib idle` - 9 passed, 0 failed, including all five `classify_idle` cases and the `cleanup_idle` regression.
- Mutation check - reintroducing the pre-fix subtraction fails `cleanup_idle_survives_ttl_larger_than_monotonic_clock` with `overflow when subtracting duration from instant`.
- `cargo fmt -p openab-core -- --check` - zero diffs for `pool.rs`.
- `cargo clippy --workspace --features unified -- -D warnings` - clean.
- Exact-head CI: 40 checks pass, 0 fail, 2 skipped. One earlier smoke-test failure on the previous head was a Docker Hub manifest timeout (`dial tcp ...:443: i/o timeout` for `node:22-trixie-slim`), unrelated to a change confined to a `#[cfg(test)]` module, and passed on re-run.

</details>

## Follow-ups (non-blocking, out of scope here)

1. **Remaining uptime-dependent fixtures in the same module.** `better_candidate_*` and the `purge_session_entries` fixture still use `Instant::now() - duration`. They are outside this diff and round 3 explicitly deferred them to a separate test-hardening change, so they were left alone rather than expanding a frozen scope.
2. **Background tasks have no panic supervision.** This bug's real severity came from the reaper's panic being invisible. The spawned cleanup loop, and other `tokio::spawn` loops like it, have no supervision, restart, or alerting. That is the root cause worth its own issue; this PR only removes the first known way to trip it.

### 5. Three Reasons We Might Not Need This PR

1. **The trigger window is bounded** - only hosts younger than the TTL are affected, so a long-lived server never hits it. But containers, CI runners, and freshly provisioned instances live entirely inside that window, and a cleanup task dying silently is not an acceptable outcome at any uptime.
2. **A checked cutoff would also work** - `checked_sub` with a fallback avoids the panic too, but it is more code for the same result and leaves the awkward cutoff framing in place.
3. **The blast radius looks cosmetic** - it is not: `cleanup_idle` also owns hung-session force-eviction, so the silent task death disables session expiry and hung recovery together.

PR #1159, opened 2026-06-20
feat(feishu): card streaming (no 20-edit cap, native tables)

wangyuyan-agent:
## What problem does this solve?

Feishu streaming edits a `post` message in place via `PATCH /open-apis/im/v1/messages/{id}`, which hits two walls:

- **20-edit hard cap** (errcode `230072`): a long streamed reply is edited dozens of times and rejected after 20. PR #1122 added cap-recovery to stop the silent truncation, but it does not escape the cap itself.
- **`post` cannot render tables**: `markdown_to_post` drops GFM tables — the `|` characters surface as plain text.

Relates to #1124 (cards render tables/markdown natively; the post-path renderer is left unchanged).

Discord Discussion URL: https://discord.com/channels/1491295327620169908/1500160821567684660

## At a Glance

```
core: edit_message(full text)  ──▶  gateway: handle_card_edit
                                      │
        short / plain text  ──────────┤─▶  post   (PATCH in place, native reply UI, #1122 recovery)
                                      │
   long / code fence / table  ────────┤─▶  promote ─▶ CardKit card
                                      │              create → send → delete post placeholder
                                      │              update (FULL text + monotonic seq)  ··· no edit cap
                                      │
              idle ~3s (reaper)  ─────┴─▶  finish: rebuild as static card → table re-renders
```

## Prior Art & Industry Research

**OpenClaw:** the official Feishu plugin `larksuite/openclaw-lark` drives CardKit + streaming (`im/v1/cards` / `cardkit/v1`). We borrow its API call sequence and rate-limit/error-code handling. openclaw-lark issue #565 (the stream-text body must be the **full accumulated** value, not a delta) is encoded as a wire-level guard test.

**Hermes Agent:** the community `lark-streaming` work takes the same CardKit route; its issue #63 (card-split retry storm → duplicate sends) informs the failure-cleanup path and its test.

**Other references:** Feishu CardKit v1 docs (`open.feishu.cn/document/cardkit-v1/*`) — `card/create`, `card-element/content` (streaming text), `card/update` (full static rebuild), verified against a live Feishu probe.

Not a fork: reimplemented as a lean Rust module (~700 lines of logic + tests).

## Proposed Solution

New `gateway/src/adapters/feishu_card.rs`:

- **CardKit client**: `create_streaming_card` → `send_card_message` → `update_card_stream` (full content + strictly increasing `sequence`) → `finish_card_stream` (`PUT card/update` rebuilds a **static** card so markdown re-renders).
[...]

PR #1122, opened 2026-06-16
fix(streaming): recover from Feishu 20-edit cap (errcode 230072)

wangyuyan-agent:
## Summary

Long Feishu replies that triggered streaming edits beyond 20 cycles got truncated mid-content while the bot still showed 🆗 — users saw a half-baked reply and a "success" reaction at the same time. This PR makes streaming edit-cap aware: detect Feishu's `230072` cap, abort cosmetic edits cleanly, and recover by deleting the half-edited placeholder and posting the full content as fresh messages.

Discord Discussion URL: https://discord.com/channels/1491295327620169908/1500160821567684660

## Root cause

Feishu's `PATCH /open-apis/im/v1/messages/{id}` enforces a hard cap of **20 edits per message**, returning `{"code":230072, "msg":"The message has reached the number of times it can be edited."}` ([official help center](https://www.feishu.cn/hc/en-US/articles/219464343884)). OAB's streaming layer flushes cosmetic edits every ~1.5s, so any LLM reply that runs longer than ~30s exhausts the quota and every subsequent edit is rejected.

Four overlapping defects let this fail silently:

1. **`src/adapter.rs` streaming finalize swallowed every error** — every delivery-path `let _ = adapter.edit_message(...)` / `let _ = adapter.send_message(...)` site in the streaming closure returned `Ok(())` regardless of outcome, so `dispatch.rs` always saw `result.is_ok()` and triggered `set_done()` (🆗) even when delivery never completed.
2. **`gateway/src/adapters/feishu.rs::edit_feishu_message` returned `()`** — failures only landed in `tracing::error!`, never propagated to the core.
3. **`src/gateway.rs::edit_message` was fire-and-forget** (`request_id: None`) — even if the gateway emitted a `GatewayResponse{success:false}`, the core dropped it on the floor.
4. **No edit-cap detection or recovery path** — once Feishu started rejecting, OAB just kept hammering on every cosmetic flush.

In production this surfaced as: placeholder visible with the cosmetic-truncated tail of the reply, then 🆗 attached to the trigger message, with the rest of the LLM output silently lost. One observed trigger logged **88 `edit_message` commands → 68 rejections with `230072` → exactly 20 successful edits before the cap kicked in**, matching the documented limit precisely.

[...]

PR #1458, opened 2026-07-25
fix(gateway/feishu): enable CardKit typewriter streaming in card mode

### LINKED ISSUE #1367: bug(gateway/feishu): CardKit streaming shows no typewriter effect in card mode
## Describe the bug

When `FEISHU_CARD_STREAMING_MODE=card`, messages are delivered as CardKit interactive cards but with **no visible typewriter/streaming effect**. The full reply appears all at once after the agent finishes generating.

## Root cause

In `card` mode, the core creates a **draft placeholder** (`message_id="draft"`) because `GatewayAdapter::show_streaming_placeholder()` returns `false`. All incremental `edit_message` commands for the draft are **skipped** by the Feishu adapter dispatch seam because `"draft"` fails `is_valid_feishu_message_id`.

The only delivery happens at turn end via `try_send_initial_card`, which creates a CardKit card with the **full accumulated text** at once — no incremental updates.

## Expected behavior

CardKit cards should display a typewriter effect as the agent streams content, matching how `auto` mode behaves after post-to-card promotion.

## Reproduction

1. Set `FEISHU_CARD_STREAMING_MODE=card`
2. Send a long prompt to trigger multi-second agent response
3. Observe: card appears with complete text at end, no incremental updates

## Proposed fix

Make `GatewayAdapter::show_streaming_placeholder()` return `true`, so the core sends a real Feishu post first. The existing post-to-card promotion path in `handle_card_edit`/`promote_and_respond` takes over on the first edit, and subsequent edits stream through `update_card_stream`.

## Relevant code

- `adapter.rs:751` — draft placeholder when `show_streaming_placeholder()` is false
- `gateway.rs:733` — `show_streaming_placeholder` returns `false` by default
- `feishu.rs:2538` — draft skip seam (`is_valid_feishu_message_id`)
- `feishu.rs:2642` — `try_send_initial_card` one-shot delivery
- `feishu.rs:2806` — `handle_card_edit` post-to-card promotion

SunnyYYLin:

When `FEISHU_CARD_STREAMING_MODE=card`, messages are delivered as CardKit cards but with **no visible typewriter effect** — the full reply appears all at once after the agent finishes. Root cause: the unified adapter's `show_streaming_placeholder()` returns false, so the core creates a draft placeholder (`message_id="draft"`), and all incremental `edit_message` commands are skipped by the `is_valid_feishu_message_id` seam.

This bug first appeared in v0.9.0, caused by the unified adapter replacing the legacy Feishu adapter's streaming path.

Closes #1367
Supersedes #1367 scope — see #1464 for the full v0.9.0 Feishu streaming regression analysis.

Discord Discussion URL: https://discordapp.com/channels/1491295327620169908/1530153804333449226

### Goal
Restore CardKit typewriter streaming for Feishu in unified mode by giving the streaming placeholder a real `om_` message ID, so incremental edits flow through `handle_card_edit` → `update_card_stream` instead of being dropped at the draft seam.

### Non-goals
- No changes to Telegram streaming (retains Draft strategy).
- No changes to the default Feishu post-edit path (mode unset or `post`). Post/auto keeps `Draft` strategy to preserve v0.9.0 behavior (streaming is a no-op but `keep_full_text` stays true so send-once includes inter-tool narration).
- No multi-card splitting for very long responses (follow-up).
- No changes to the ACP platform streaming path (append-only deltas, unrelated).

### Accepted Residual Risks
- Feishu CardKit streaming mode has a server-side idle timeout. If the agent is idle longer than `FEISHU_CARD_IDLE_FINALIZE_MS`, the idle reaper finalizes the card. Late edits after finalize now report `success: false` (`session_finalized`) so core's delete-and-resend fallback triggers — this requires the `delete_message` override (added in this PR) to actually remove the stale card. Without the override, the fallback would leave a stale card plus a duplicate plain-text reply.
- Ordinary `send_message` remains fire-and-forget for non-streaming paths. Both `send_streaming_placeholder` and card-mode `edit_message` wait for the platform response (via `dispatch_with_response`), because they need the real `om_` ID and edit-cap/finalize signals respectively. This is the intended scope: the wait is isolated to the card streaming path.

### Acceptance Criteria
- [x] Without `FEISHU_CARD_STREAMING_MODE=card`, behavior is identical to before (post-edit path, Draft strategy).
- [x] With `FEISHU_CARD_STREAMING_MODE=card`, real CardKit updates stream during generation.
- [x] `card stream updated` with `seq=1` captured in production logs after model config fix.
- [x] `cargo test --features unified` passes.
- [x] `cargo clippy --workspace --features unified -- -D warnings` clean.
- [x] No trait-breaking changes for external/custom adapters.
- [x] `delete_message` override verified: stale card deleted after `streaming mode closed` (300309), fresh card sent via delete-and-resend recovery.
- [x] `Existing::Finalized` reports `success: false` — regression test `s5_edit_finalized_session_reports_failure`.

### Follow-ups
- Multi-card splitting for responses exceeding Feishu's per-card content limit.
- Configurable streaming flush interval (currently tied to dispatch batch cadence).
- F1 root cause: core→gateway card lifecycle lacks an explicit finish signal; the `stream_finish` trait seam (`adapter.rs:429`) is a candidate so the idle reaper does not finalize mid-turn. Tracked in #1464.

## At a Glance

```
Before (broken since v0.9.0):
  core → draft placeholder ("draft") → edit_message("draft", text)
       → is_valid_feishu_message_id("draft") == false → SKIPPED
       → try_send_initial_card at turn end → full text, no streaming

After (restores pre-v0.9.0 behavior):
  core → EditablePlaceholder → send_streaming_placeholder()
       → create Feishu post → dispatch_with_response() → real om_ ID
       → edit_message(om_xxx, text) → handle_card_edit
       → update_card_stream(card_id, seq++) → typewriter effect
       → idle > FEISHU_CARD_IDLE_FINALIZE_MS → finish_card_stream
       → late edit → Existing::Finalized → success: false
       → core delete-and-resend → delete_message override → DELETE /im/v1/messages/{id}
```

**OpenClaw:**
Uses draft-message editing for Telegram streaming (edit_message every N tokens). Feishu path uses post-edit with a 20-edit cap before promoting to card.

**Hermes Agent:**
N/A — Hermes does not support Feishu.

**Other references:**
- Feishu CardKit streaming API: PUT /cardkit/v1/cards/:id/elements/:eid/content with strictly-increasing sequence numbers.
- Feishu errcode 300317: sequence must be strictly increasing.
- Feishu errcode 300309: streaming mode closed after server-side idle timeout.

## Proposed Solution

The fix restores the pre-v0.9.0 streaming path through the unified adapter architecture:

1. **`adapter.rs`**: Additive `StreamingStrategy` enum (Disabled / Draft / EditablePlaceholder) + `streaming_strategy()` method with default impl. Legacy `use_streaming()` / `show_streaming_placeholder()` retained — no trait break.

2. **`unified_adapter.rs`**: Selects `EditablePlaceholder` for Feishu when `FEISHU_CARD_STREAMING_MODE=card`. Post/auto keeps `Draft` to preserve v0.9.0 behavior. Adds `dispatch_with_response()` — a request-response channel that waits for the Feishu adapter to return the real `om_` message ID. The streaming placeholder path and card-mode edits both use this; normal `send_message` remains fire-and-forget. Adds `delete_message` override mirroring the standalone `GatewayAdapter` override so the edit-cap / session-finalized recovery path can perform a real `DELETE /im/v1/messages/{id}`.

3. **`main.rs`**: The `streaming_strategy` threading happens entirely inside `AdapterRouter`; no `main.rs` change was needed. The only `main.rs` change is filtering `GatewayResponse` events out of the unified inbound bridge (necessary companion to introducing `request_id` on this path).

4. **`feishu.rs`**: `Existing::Finalized` arm now reports `success: false` (`session_finalized`) so core's delete-and-resend fallback triggers. +1 line observability log on `CardOutcome::Updated` (card_id, msg_id, seq).

### Why a response channel instead of modifying `send_message`?

The unified adapter's `send_message` is fire-and-forget by design. Changing it to always wait for a platform response would:
- Add ~500ms latency to every message send (not just streaming)
- Require all platform adapters (Telegram, Discord, Line) to return message IDs
- Break the fire-and-forget contract that non-streaming paths rely on

The response channel isolates the wait to `send_streaming_placeholder` and card-mode `edit_message` only.

## Alternatives Considered

1. **Modify `send_message` to return real IDs**: Fewer lines (~50 vs ~140) but changes behavior for all platforms and all message paths. Rejected — too broad for an opt-in feature.
2. **One-line fix (`show_streaming_placeholder() → true`)**: Works but leaks Feishu-specific semantics into the generic adapter trait. No mechanism for adapters to customize placeholder behavior.
3. **Delta-based updates**: Rejected — Feishu CardKit requires full content on each PUT, not deltas.

## Validation

- [x] `cargo test --features unified` — all pass
- [x] `cargo clippy --workspace --features unified -- -D warnings` — clean
- [x] `cargo build --release --features unified` — success
- [x] Manual testing: deployed to local service, observed `card stream updated seq=1` on card_id after agent reply. `delete_message` override verified: 300309 → `feishu message deleted` → fresh card sent via recovery path.

## Review Findings Addressed

This commit addresses all four findings from the round-1 review by @wangyuyan-agent:

| Finding | Fix |
|---|---|
| **Blocking 1** (PR description inaccuracies) | Corrected: main.rs threading claim, send_message fire-and-forget scope, idle-finalize recovery path. |
| **Blocking 2** (missing `delete_message` override) | Added `delete_message` override on `UnifiedGatewayAdapter` mirroring standalone `gateway.rs:707`. |
| **F1** (idle reaper drops reply silently) | `Existing::Finalized` now reports `success: false`; combined with Blocking 2, core's delete-and-resend triggers correctly. Root cause (missing turn-finish signal) tracked as follow-up in #1464. |
| **F2** (post/auto default changed) | Reverted to `Draft` strategy for post/auto to preserve v0.9.0 `keep_full_text` behavior. |

### DIFF
--- a/src/unified_adapter.rs
+++ b/src/unified_adapter.rs
@@ -369,6 +369,14 @@ impl ChatAdapter for UnifiedGatewayAdapter {
             "telegram" => StreamingStrategy::Disabled,
             #[cfg(feature = "feishu")]
             "feishu" => {
+                // Honor the streaming enablement decision before selecting a
+                // strategy. When streaming is explicitly disabled (e.g.
+                // TELEGRAM_STREAMING=false), Feishu must also be Disabled -
+                // otherwise Draft would set streaming=true, run a useless edit
+                // loop, and force keep_full_text=true (narration replay).
+                if !self.use_streaming(false) {
+                    return StreamingStrategy::Disabled;
+                }
                 match self
                     .gw_state
                     .feishu
@@ -444,6 +452,7 @@ mod tests {
     fn feishu_uses_real_editable_placeholder() {
         let (event_tx, _) = tokio::sync::broadcast::channel(16);
         let mut state = AppState::test_default(event_tx);
+        state.telegram_rich_messages = true;
         let mut pairs = std::collections::HashMap::new();
         pairs.insert("FEISHU_APP_ID".into(), "app".into());
         pairs.insert("FEISHU_APP_SECRET".into(), "secret".into());
@@ -466,6 +475,7 @@ mod tests {
     fn feishu_post_mode_keeps_draft_strategy() {
         let (event_tx, _) = tokio::sync::broadcast::channel(16);
         let mut state = AppState::test_default(event_tx);
+        state.telegram_rich_messages = true;
         let mut pairs = std::collections::HashMap::new();
         pairs.insert("FEISHU_APP_ID".into(), "app".into());
         pairs.insert("FEISHU_APP_SECRET".into(), "secret".into());
@@ -481,6 +491,26 @@ mod tests {
         );
         assert_eq!(
             adapter.streaming_strategy(&channel("feishu"), true),
+            StreamingStrategy::Disabled,
+        );
+    }
+
+    #[cfg(feature = "feishu")]
+    #[test]
+    fn feishu_disabled_when_streaming_off() {
+        let (event_tx, _) = tokio::sync::broadcast::channel(16);
+        let mut state = AppState::test_default(event_tx);
+        // telegram_rich_messages defaults to false -> use_streaming == false.
+        let mut pairs = std::collections::HashMap::new();
+        pairs.insert("FEISHU_APP_ID".into(), "app".into());
+        pairs.insert("FEISHU_APP_SECRET".into(), "secret".into());
+        pairs.insert("FEISHU_CARD_STREAMING_MODE".into(), "card".into());
+        state.apply_feishu_config(openab_gateway::GatewayFeishuConfig { pairs });
+        let adapter = UnifiedGatewayAdapter::new(Arc::new(state));
+
+        // Even in card mode, streaming must be Disabled when use_streaming is false.
+        assert_eq!(
+            adapter.streaming_strategy(&channel("feishu"), false),
             StreamingStrategy::Disabled,
         );
     }
--- a/src/unified_adapter.rs
+++ b/src/unified_adapter.rs
@@ -3,12 +3,28 @@
 
 use anyhow::Result;
 use async_trait::async_trait;
-use openab_core::adapter::{ChannelRef, ChatAdapter, MessageRef};
-use openab_gateway::schema::{Content, GatewayReply, ReplyChannel};
+use openab_core::adapter::{ChannelRef, ChatAdapter, MessageRef, StreamingStrategy};
+use openab_gateway::schema::{Content, GatewayReply, GatewayResponse, ReplyChannel};
 use openab_gateway::AppState;
 use std::collections::HashMap;
+use std::sync::atomic::{AtomicU64, Ordering};
 use std::sync::Arc;
 use tokio::sync::Mutex;
+
+const GATEWAY_RESPONSE_TIMEOUT: std::time::Duration = std::time::Duration::from_secs(5);
+static REQUEST_SEQUENCE: AtomicU64 = AtomicU64::new(1);
+
+fn next_request_id() -> String {
+    format!(
+        "unified_{}",
+        std::time::SystemTime::now()
+            .duration_since(std::time::UNIX_EPOCH)
+            .unwrap_or_default()
+            .as_nanos()
+            .saturating_mul(1_000_000)
+            .saturating_add(REQUEST_SEQUENCE.fetch_add(1, Ordering::Relaxed) as u128)
+    )
+}
 
 pub struct UnifiedGatewayAdapter {
     pub gw_state: Arc<AppState>,
@@ -22,6 +38,23 @@ impl UnifiedGatewayAdapter {
             gw_state,
             telegram_reaction_state: Arc::new(Mutex::new(HashMap::new())),
         }
+    }
+
+    fn uses_feishu_card_streaming(&self, channel: &ChannelRef) -> bool {
+        if channel.platform != "feishu" {
+            return false;
+        }
+
+        #[cfg(feature = "feishu")]
+        {
+            self.gw_state.feishu.as_ref().is_some_and(|feishu| {
+                feishu.config.streaming_mode
+                    == openab_gateway::adapters::feishu::StreamingMode::Card
+            })
+        }
+
+        #[cfg(not(feature = "feishu"))]
+        false
     }
 
     /// Dispatch a GatewayReply to the correct platform adapter.
@@ -96,8 +129,62 @@ impl UnifiedGatewayAdapter {
                 }
             }
             other => {
-                tracing::warn!(platform = other, "unified adapter: unknown platform, cannot route reply");
+                tracing::warn!(
+                    platform = other,
+                    "unified adapter: unknown platform, cannot route reply"
+                );
             }
+        }
+    }
+
+    /// Dispatch a reply and, when requested, wait for the platform response.
+    /// Feishu returns the native om_... message ID; unified mode must preserve
+    /// that ID because synthetic IDs cannot be edited by the Feishu API.
+    async fn dispatch_with_response(
+        &self,
+        reply: &GatewayReply,
+    ) -> Result<Option<GatewayResponse>> {
+        let Some(request_id) = reply.request_id.as_deref() else {
+            self.dispatch_reply(reply).await;
+            return Ok(None);
+        };
+        if reply.platform != "feishu" {
+            self.dispatch_reply(reply).await;
+            return Ok(None);
+        }
+
+        let mut rx = self.gw_state.event_tx.subscribe();
+        self.dispatch_reply(reply).await;
+        let request_id = request_id.to_string();
+        let wait = async {
+            loop {
+                match rx.recv().await {
+                    Ok(json) => {
+                        let Ok(response) = serde_json::from_str::<GatewayResponse>(&json) else {
+                            continue;
+                        };
+                        if response.schema == "openab.gateway.response.v1"
+                            && response.request_id == request_id
+                        {
+                            return Ok(response);
+                        }
+                    }
+                    Err(tokio::sync::broadcast::error::RecvError::Lagged(_)) => continue,
+                    Err(tokio::sync::broadcast::error::RecvError::Closed) => {
+                        anyhow::bail!("gateway response channel closed")
+                    }
+                }
+            }
+        };
+        match tokio::time::timeout(GATEWAY_RESPONSE_TIMEOUT, wait).await {
+            Ok(Ok(response)) if response.success => Ok(Some(response)),
+            Ok(Ok(response)) => Err(anyhow::anyhow!(response
+                .error
+                .unwrap_or_else(|| "gateway reported failure".into()))),
+            Ok(Err(error)) => Err(error),
+            Err(_) => Err(anyhow::anyhow!(
+                "timed out waiting for gateway response to {request_id}"
+            )),
         }
     }
 
@@ -144,8 +231,29 @@ impl ChatAdapter for UnifiedGatewayAdapter {
         self.dispatch_reply(&reply).await;
         Ok(MessageRef {
             channel: channel.clone(),
-            message_id: format!("unified_{:x}", std::time::SystemTime::now()
-                .duration_since(std::time::UNIX_EPOCH).unwrap_or_default().as_nanos()),
+            message_id: next_request_id(),
+        })
+    }
+
+    async fn send_streaming_placeholder(
+        &self,
+        channel: &ChannelRef,
+        content: &str,
+    ) -> Result<MessageRef> {
+        if !self.uses_feishu_card_streaming(channel) {
+            return self.send_message(channel, content).await;
+        }
+
+        let mut reply = self.build_reply(channel, content, None, None);
+        reply.request_id = Some(next_request_id());
+        let message_id = self
+            .dispatch_with_response(&reply)
+            .await?
+            .and_then(|response| response.message_id)
+            .ok_or_else(|| anyhow::anyhow!("Feishu placeholder response omitted message_id"))?;
+        Ok(MessageRef {
+            channel: channel.clone(),
+            message_id,
         })
     }
 
@@ -187,7 +295,12 @@ impl ChatAdapter for UnifiedGatewayAdapter {
         let mut reply = self.build_reply(&msg.channel, content, Some("edit_message"), None);
         // Use the actual platform message_id (e.g. "draft" for streaming, or numeric for edits)
         reply.reply_to = msg.message_id.clone();
-        self.dispatch_reply(&reply).await;
+        if self.uses_feishu_card_streaming(&msg.channel) {
+            reply.request_id = Some(next_request_id());
+            self.dispatch_with_response(&reply).await?;
+        } else {
+            self.dispatch_reply(&reply).await;
+        }
         Ok(())
     }
 
@@ -201,8 +314,7 @@ impl ChatAdapter for UnifiedGatewayAdapter {
         self.dispatch_reply(&reply).await;
         Ok(MessageRef {
             channel: channel.clone(),
-            message_id: format!("unified_{:x}", std::time::SystemTime::now()
-                .duration_since(std::time::UNIX_EPOCH).unwrap_or_default().as_nanos()),
+            message_id: next_request_id(),
         })
     }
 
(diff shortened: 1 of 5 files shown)


src/unified_adapter.rs
```diff
@@ -144,8 +231,29 @@ impl ChatAdapter for UnifiedGatewayAdapter {
...
+    ) -> Result<MessageRef> {
+        if !self.uses_feishu_card_streaming(channel) {
+            return self.send_message(channel, content).await;
+        }
+
+        let mut reply = self.build_reply(channel, content, None, None);
+        reply.request_id = Some(next_request_id());
+        let message_id = self
+            .dispatch_with_response(&reply)
+            .await?
```
wangyuyan-agent, line 251: This path inverts three deliberate choices in the standalone equivalent. `GatewayAdapter::edit_message` uses `EDIT_RESPONSE_TIMEOUT_MS = 800`, returns `Ok(())` on timeout — with the comment "Treat as success to avoid false-positive ❌; the cap-reached path already short-circuits much faster" — and returns `Ok(())` when the response channel closes. `dispatch_with_response` uses 5000ms and turns both timeout and closed-channel into `Err`.

Consequences, in order of likelihood:

- A merely *slow* tenant (not a failing one) now produces an error at end-of-turn, which drives the delete-and-resend path: the card is dropped and the reply is re-sent as plain text. That is the false positive the standalone comment was written to avoid, and it is worse here because it discards a rendered card.
- On the placeholder specifically, `?` lets a cosmetic element abort the whole turn: the caller surfaces an error to the user instead of degrading. `send_streaming_placeholder` failing is a good reason to fall back to `StreamingStrategy::Disabled` (send-once), not to fail the turn.
- A timeout on the placeholder also leaves an orphan: the card exists with `…` and its session sits at `sequence == 0`, which `idle_keys` deliberately skips, so the reaper will never finalize it. It stays until FIFO eviction.
- `RecvError::Lagged(_) => continue` drops frames silently; if the matching response is in a dropped frame the wait necessarily runs the full 5s and then reports failure. Worth aligning the timeout with the standalone value, or at least documenting why card mode needs 5s where post mode needs 800ms.

Measured for margin: creating and sending the initial card took ~1.7s of the 5s budget in a healthy local run (two REST round-trips).

src/unified_adapter.rs
```diff
@@ -3,13 +3,29 @@
...
+static REQUEST_SEQUENCE: AtomicU64 = AtomicU64::new(1);
+
+fn next_request_id() -> String {
+    format!(
+        "unified_{}",
+        std::time::SystemTime::now()
+            .duration_since(std::time::UNIX_EPOCH)
+            .unwrap_or_default()
+            .as_nanos()
+            .saturating_mul(1_000_000)
```
wangyuyan-agent, line 24: Two small things.

The name covers only one of the two uses — `next_request_id()` also produces the synthetic `message_id` returned by `send_message` and `send_message_with_reply`.

And `saturating_mul` is unreachable defense here: nanos-since-epoch is ~1.8e18, so the product is ~1.8e24 against a u128 ceiling of ~3.4e38. If it ever did saturate, every id would collapse to one value — worse than overflowing. Uniqueness, incidentally, does not come from the `1e6` slot: `REQUEST_SEQUENCE` never resets, so it will eventually exceed 1e6 and spill into the next timestamp's band. It holds because the timestamp is non-decreasing while the counter strictly increases, so `t*1e6 + s` is strictly increasing overall. A comment recording that reasoning, or `checked_mul` with an explicit branch, would be clearer than a saturating op that cannot fire.

wangyuyan-agent (changes requested):
Reviewed at head `55df8b6d`. Verdict: **request changes** — not because the approach is wrong, but because three statements in the PR description do not match the code, and the recovery path this change depends on is not wired in unified mode.

Direction first, because it is sound. Issue #1367 laid out three options and recommended A (add `platform: &str` to the streaming trait methods, accepting a breaking change). What ships here is more accurate to describe as **A's contract shell plus B's message-ID passthrough**: `streaming_strategy()` gives per-channel strategy selection without the breaking parameter, while `send_streaming_placeholder()` + `dispatch_with_response()` — core waiting for a `GatewayResponse` before entering the edit loop — is precisely what the issue defined as Option B's mechanism, minus B's per-platform inner adapter. That combination is a good trade, and the default impls mean no existing `ChatAdapter` implementation changes behavior. `dispatch_with_response` subscribes before dispatching, so there is no lost-response race. Every `CardOutcome` arm in `handle_card_edit` emits a response, so the 5s wait cannot stall the loop on a rate-limited frame. `try_send_initial_card` falling back to the post path still yields a real `om_` id.

Two things need to change before merge, plus two findings that need an answer.

## Blocking 1 — three claims in the PR description are not true of the code

Each is independently checkable without a Feishu tenant:

1. **"Late edits after finalize fall back to delete-and-resend (existing #1122 recovery path)."** `handle_card_edit`'s `Existing::Finalized` arm calls `emit_response(..., true, Some(om_post), None)` — `success: true`. Core's `edit_message` therefore returns `Ok`, and the finalization branch in `AdapterRouter` only deletes-and-resends inside `if let Err(e) = adapter.edit_message(...)`. For the reaper's own finalize, that recovery never runs. (It does run for `CardOutcome::Failed`, so the sentence is right for hard API failures and wrong for idle finalize — which is the case the paragraph is actually about.)

2. **"`main.rs`: Passes `streaming_strategy` into the ACP turn loop."** `streaming_strategy` has zero occurrences in `src/main.rs` at this head. The file's only changes are a comment relocation and the `GatewayResponse` filter in the event bridge. The threading happens entirely inside `AdapterRouter`; no `main.rs` change was needed. Please drop or correct the claim.

3. **"Only `send_streaming_placeholder` waits for the platform response."** `edit_message` also waits, on every frame, whenever `uses_feishu_card_streaming(&msg.channel)` is true — that is the point of the change and it is correct behavior, but it makes the "Accepted Residual Risks" statement about `send_message` remaining fire-and-forget read as broader than it is. Worth restating as "ordinary `send_message` remains fire-and-forget; the streaming placeholder and card-mode edits both wait."

## Blocking 2 — unified never overrides `delete_message`, so the recovery this PR depends on cannot complete

`UnifiedGatewayAdapter` has no `delete_message` implementation, so it inherits the trait default at `adapter.rs:396`: `self.edit_message(msg, "\u{200b}")`. In card mode that edit is routed straight back into `handle_card_edit` — it does not delete anything. Depending on session state it either overwrites the card with a zero-width space or reports success and does nothing.

The standalone adapter already solved exactly this, and its own comment states why:

> Override default `delete_message` (which falls back to edit-to-zero-width) so platforms with native delete APIs (e.g. Feishu `DELETE /im/v1/messages/{id}`) can perform real deletions. **Critical for the streaming-edit-cap recovery path** … The default zero-width-edit fallback would itself fail on a cap-reached message, leaving the placeholder visible.

Since this PR is what makes core's edit errors reachable in unified mode at all, it is also what makes the missing override matter. Without it, every `Err` path that ends in delete-then-resend leaves the original card in place and appends a second copy of the reply. Mirroring the standalone override on `UnifiedGatewayAdapter` closes it.

This also constrains the fix for F1 below: reporting failure from a finalized session **without** a working delete would produce a stale card sitting above a fresh full reply. The two have to land together.

## F1 — the idle reaper finalizes mid-turn and the reply is silently dropped

Mechanism, all three links checkable in-tree:

- Core's cosmetic edit loop issues an edit only when the rendered text actually changes (`buf_rx.has_changed()` then `content != last`). Nothing is sent while a tool runs and produces no display change.
- `FeishuStreamRegistry::idle_keys` selects sessions with `sequence > 0 && is_idle(idle_ms)`, and `card_idle_finalize_ms` defaults to **3000**. `run_idle_reaper` then rebuilds the card as static and calls `mark_finalized`.
- Every later edit lands on `Existing::Finalized`, which reports success (Blocking 1, item 1). Core sees `Ok`, so the end-of-turn delivery does not fall back either.

So one tool call longer than 3s closes the stream mid-turn, and the answer that follows it is discarded with no signal in the client or the logs.

Reproduced on PR head `55df8b6d` merged with `main` at `53061d69` — the four files carrying this change (`adapter.rs`, `unified_adapter.rs`, `feishu.rs`, `feishu_card.rs`) are byte-identical to the PR head in that tree — with `FEISHU_CARD_STREAMING_MODE=card` on a real Feishu tenant over WebSocket. Prompt: run `sleep 15` via shell, then write ~400 words plus a markdown table.

```
01:57:30  feishu first reply sent directly as card   card_msg_id=om_x100…c0a
01:57:33  feishu card stream updated                 card_id=7668505399948233666 seq=1
01:57:37  feishu card stream finalized (idle)        <- reaper; sleep 15 still running
01:57:37–01:58:10  agent emits 315 agent_message_chunk updates
01:58:10  turn ends
```

No further `card stream updated`, no warning, no error, no delete, no fresh send. The agent produced 892 characters including the complete table; the card, read back from the API, contains exactly `🔧 Running: sleep 15...`. Precisely: what is lost is the first delivery chunk, i.e. the whole reply whenever it fits in one message. Content beyond `message_limit` still reaches the user, because the finalization path sends `chunks.iter().skip(1)` via `send_message`.

**Scope.** The `Existing::Finalized` arm and the reaper are pre-existing #1159 design, **not introduced by this PR**. Standalone gateway card mode can reach the same state, though only when `[gateway] streaming = true` — which is not the default (`GatewayConfig.streaming` defaults to `false`). What belongs to this PR is turning the path on for unified mode, which is the point of the change, and describing the risk backwards.

**On the fix, and the order matters.** Reporting `success: false` from `Existing::Finalized` is a safety net, not the fix — and on its own it converts a truncated card into a stale card plus a duplicate plain-text reply (see Blocking 2). The root cause is that the reaper infers turn boundaries from idleness, while the fact of a turn ending is known only to core, and the core→gateway command vocabulary has no terminal signal for the card lifecycle. This PR extends that lifecycle to *start* and *update*; the missing third state is *finish*. Note the trait already has a `stream_finish` seam (`adapter.rs:429`, currently used only on the native-streaming path) — an explicit finish command, or simply not finalizing while a turn is still open, addresses the cause. Then `success: false` plus the `delete_message` override from Blocking 2 becomes a correct backstop rather than a new failure mode.

## F2 — `post` / `auto` defaults do change, contrary to the description

The body states default behavior is unchanged and that card mode is the only opt-in. For `post` and `auto` in unified mode that does not hold:

- `unified = ["telegram", "line", "feishu", …]`, so a unified binary always carries the `telegram` feature, and `TELEGRAM_RICH_MESSAGES` resolves via `unwrap_or(true)`. With neither that variable nor `TELEGRAM_STREAMING` set, `UnifiedGatewayAdapter::use_streaming` returns `true`.
- Before this PR that made Feishu take the streaming path with `show_streaming_placeholder() == false`, i.e. `StreamingStrategy::Draft`. After it, `streaming_strategy` returns `Disabled` for any mode other than `card`.
- `streaming` does not only gate the placeholder: `keep_full_text = streaming || narration_display`. So the delivered message changes from the full turn buffer to the final answer block only — inter-tool narration disappears for `post` and `auto` users who never opted into anything.

I think the new behavior is the better one — send-once should not replay narration the user never saw stream. The issue is that it is a user-visible change presented as a no-op, and the new tests cannot catch it because `AppState::test_default` sets `telegram_rich_messages: false`, which is the opposite of the `from_env` default. Suggest saying so in the body and adding a case with `telegram_rich_messages: true` asserting the intended `post`/`auto` strategy.

## What's good

- Additive trait extension with defaults; `streaming_strategy_preserves_legacy_placeholder_contract` pins the legacy mapping down.
- Subscribe-before-dispatch, correlation by exact `request_id`, ordinary `send_message` untouched.
- Filtering `openab.gateway.response.v1` out of the unified inbound bridge is the right seam, and it is a necessary companion to introducing `request_id` on this path — before this change unified produced no responses at all. `GatewayResponse` cannot swallow inbound events either, since `request_id` and `success` are required fields absent from the event schema.
- Deriving `streaming` from a single strategy value instead of two independent predicates removes a real footgun for the next multiplexed adapter.
- The `CardOutcome::Updated` log is what made the timeline above legible.

SunnyYYLin:
### F3: Unified test CI coverage

Agreed that `src/unified_adapter.rs` tests require `--features unified` and are not exercised by the current `cargo test --workspace` (default features) step in CI.

However, `.github/workflows/ci.yml` is not modifiable from this fork PR (workflow files require maintainer approval). Could a maintainer add the following step to the CI test job?

```yaml
- name: cargo test (unified)
  run: cargo test --features unified
```

This would cover the `streaming_strategy`, `dispatch_with_response`, and `delete_message` tests that gate the unified Feishu card streaming path.

The existing `cargo clippy --workspace --features unified` and `cargo build --features unified` steps already compile these modules, so the added test step is a natural companion.

wangyuyan-agent (commented):
Re-reviewed at head `53026045`. All four findings from my round-1 review are closed, and the two blocking ones are closed in code rather than in prose — the ordering discipline on F1 and Blocking 2, with the commit message stating why they had to land together, is exactly right.

One item keeps my request-changes in place, and it is small: the acceptance claim offered as proof of the `delete_message` override describes a path that does not reach the override. That is the same class of issue as round-1 Blocking 1, so having just held the PR over it once, I cannot wave it through now. Reword that item (or produce the run) and correct the description points below, and I approve — no new conditions. Everything else here is either a suggestion you may defer to #1464 or a note requiring nothing.

I have read the round-3 automated review on this same head. Its F1 (orphan card after an ambiguous placeholder result) and my "residual" section below are the same defect reached independently; its remedy is the more complete one and I defer to it. Its F2 (CI) needs a maintainer. Nothing below contradicts its assessment.

## Round-1 findings — verified closed

- **Blocking 2.** `UnifiedGatewayAdapter::delete_message` now dispatches `command = Some("delete_message")` with `reply_to` set to the target. The gateway seam accepts it (the command is in the `interpolates_message_id` set, so the id shape is validated), and the `"delete_message"` arm resolves the session's `card_message_id` before falling back to `reply_to`, then calls native `delete_feishu_message`. The trait default's zero-width edit is unreachable on this path. Correct.
- **F1.** `Existing::Finalized` now emits `success: false` with `error = "session_finalized"`. Consumer side traced: the cosmetic loop's `consecutive_failures` break stops the futile edits, and the end-of-turn path takes `if let Err(e) = adapter.edit_message(...)` → `delete_message` → `send_message(first)` → `chunks.iter().skip(1)`. The chain completes.
- **F2.** Reverted to `Draft`, then refined. Coverage now spans both sides of the `use_streaming` decision: `feishu_post_mode_keeps_draft_strategy` (new) and `feishu_uses_real_editable_placeholder` (existing, amended) set `telegram_rich_messages = true`, which is the `from_env` default; `feishu_disabled_when_streaming_off` (new) leaves it at the `test_default` value of `false`. That closes the gap I flagged — the earlier tests could only reach the non-default side.
- **Blocking 1.** The `main.rs` threading claim and the `send_message` fire-and-forget scope are corrected. Two description points remain; see below.

F1's root cause is untouched, as expected: `idle_keys` still selects on `sequence > 0 && is_idle` and `card_idle_finalize_ms` still defaults to 3000. Shipping the backstop here and tracking the turn-finish signal in #1464 is the right split.

## The one hard item — the evidence for the `delete_message` override does not exercise it

The acceptance list reads: "`delete_message` override verified: stale card deleted after `streaming mode closed` (300309), fresh card sent via delete-and-resend recovery."

300309 is streaming-mode-closed, not the edit cap (that is 230072), and it arrives as `CardOutcome::Failed`, which the gateway resolves internally before core hears about it: the `Failed` arm removes the session and, when `card_fallback_to_post` is set, calls `delete_feishu_message` on the card itself, and only then reports failure. On that path the card is already gone when core's recovery runs; core's subsequent `delete_message` finds no session, targets an id that no longer exists, and — being fire-and-forget — reports nothing either way. The `feishu message deleted` line in that run is equally consistent with the gateway's own cleanup, and the same log string is emitted from more than one call site.

The path the override exists for is `session_finalized`, where the session is kept and the card is still on screen. That one has a unit test proving the gateway reports failure, and no end-to-end evidence.

Two ways to close this, either is fine:

- Show core's own `final streaming edit failed; deleting placeholder and sending fresh` warning in a run, followed by the delete. That warning is emitted only by the core path, so it disambiguates.
- Or reword the item to state what the run actually demonstrated (gateway-side 300309 recovery), and let the cross-layer test the round-3 review asks for cover the override: drive a finalized session through `handle_reply` with a `delete_message` command and assert the session is removed from the registry. That needs no tenant and no HTTP mock.

I am raising this rather than letting it go because a claim that reads as proof of a mechanism the run did not touch is what round-1 Blocking 1 was about.

## Residual, and a correction in this PR's favour

The description undersells what this change fixes, and my round-1 F1 write-up did too. Worth stating precisely, because it changes who owns what.

On the merge base, unified card mode never finalized anything. `show_streaming_placeholder()` returns false, so core takes the Draft branch, every `edit_message` carries the `draft` sentinel and is rejected at the id-shape seam, and the whole reply is delivered at end of turn through `send_message`. In card mode each of those sends reaches `try_send_initial_card`, which registers a fresh session via `promote`, and `FeishuStreamSession::new` starts at `sequence: 0`. `idle_keys` filters on `sequence > 0`, so the reaper never touched those sessions: every reply sat in a card left in streaming state — cursor unsettled, and `finish_card_stream`'s full static replace never applied, which is the step that re-renders a GFM table as a table instead of a code block. Base `edit_message` was also fire-and-forget returning `Ok(())`, so core could not see a failure at all and the delete-and-resend path was unreachable regardless of what the `Finalized` arm reported.

By giving card mode a real `om_` id, this PR makes the main path advance `next_sequence`, which brings those sessions above the `sequence > 0` guard and lets the reaper finalize them. Tables render correctly at the end of a card-mode turn for the first time since v0.9.0. That is a real gain and it is not claimed anywhere in the description — worth adding, since it is more than "typewriter effect".

What remains are the paths that still deliver through a plain `send_message` and therefore still land at `sequence == 0` with nothing following to advance them:

- The recovery resend itself (`delete_message` then `send_message(first)`), which this PR makes reachable for the first time in unified mode.
- Overflow chunks in card mode (`chunks.iter().skip(1)`), on the ordinary success path as well as in recovery — each becomes its own card and only the last one, if any, gets edits.
- A placeholder whose card was sent but whose response was lost, which leaves a card showing `…`. This is the round-3 review's F1; its remedy — retaining the native card id or retiring the card rather than falling back to the synthetic one — is the right one, because here the problem is a duplicate bubble, not an unsettled cursor. Finalizing it would not remove it.
- `promote_and_respond`, when promotion happens on what turns out to be the final edit of a turn.

All four sit on the same `sequence > 0` guard, and all four are pre-existing behavior rather than something you introduced — the first is old behavior arriving through a newly reachable path, not a new end state. So I am not asking for them here. If you want a cheap partial: a longer idle window for `sequence == 0` sessions would let the reaper settle cases 1, 2 and 4 without weakening what the guard protects against (a session finalized while core is still thinking), and needs no new signal from core. The general fix is the turn-finish signal in #1464. Either way, please record the residual there, because as written the description implies the backstop leaves a finished card.

## F-a — card mode is now gated behind a Telegram-named flag (optional, not blocking)

`53026045` places `if !self.use_streaming(false) { return Disabled; }` ahead of the mode match, and `use_streaming` on this adapter reads only Telegram state: `telegram_streaming` when set, otherwise `telegram_rich_messages`. So `TELEGRAM_STREAMING=false`, `TELEGRAM_RICH_MESSAGES=false`, or `[telegram] streaming = false` in config (which `apply_telegram_config` copies over the env-derived value) silently overrides `FEISHU_CARD_STREAMING_MODE=card`, with no warning. `feishu_disabled_when_streaming_off` pins that behavior down.

Severity, so this is not read as a regression:

- Defaults are safe. `TELEGRAM_RICH_MESSAGES` resolves through `unwrap_or(true)` in both `AppState::from_env` and `TelegramConfig::resolve`, so *removing* Telegram variables from a Feishu-only deployment leaves card streaming working. It takes an explicit `false` to trip.
- The resulting behavior is send-once, which is what an explicitly-streaming-disabled deployment already got. Nothing regresses; what is new is only that a per-platform opt-in can be overridden by another platform's switch.
- It is not really this PR's issue. The same coupling already governs post/auto on main, and the generic arm ignores WeCom's own `streaming_enabled` (`WECOM_STREAMING_ENABLED`) for the same structural reason: `use_streaming` and `show_streaming_placeholder` take no channel, so only `streaming_strategy` can see the platform. Round-2's F1 asked for the guard "before selecting Draft or EditablePlaceholder", and round 3 lists the result under What's Good.

So I am not asking you to change the strategy logic — reversing it would put you between reviewers with no maintainer to arbitrate. Optional, if you happen to touch this file again: a startup warning when `streaming_mode == Card` and `use_streaming()` is false, so an operator who opted in and sees nothing has something to find. The real fix — a channel on the enablement predicate, or a `[feishu] streaming` knob — belongs with the trait work in #1464 and is worth noting there as a general gap rather than a Feishu one.

## Description corrections

- The At a Glance diagram shows `send_streaming_placeholder() → create Feishu post → … → real om_ ID`. In card mode no post is created: the handler takes the `streaming_mode == Card` branch into `try_send_initial_card`, which creates and sends the card, keys the session by the card's own message id, and responds with that id — no placeholder, no revoke notice, which is the #1159 property that function's own comment names. The post path is only the fallback when card create or send fails. The real flow is a better story for this PR than the drawn one.
- Non-goals and Proposed Solution §2 both still say "Post/auto keeps `Draft` strategy"; after `53026045` that holds only when streaming is enabled.
- The first Accepted Residual Risk reads as though the idle finalize were Feishu's. Both exist: Feishu closes a stream server-side (that is the 300309 signal), and separately `run_idle_reaper` finalizes on our own `FEISHU_CARD_IDLE_FINALIZE_MS` default of 3000. The paragraph describes the second while attributing it to the first, which matters mainly because it suggests the 3s window is not ours to change.
- "Review Findings Addressed" covers round 1 only; the two round-2 fixes are not recorded.
- Minor: the Prior Art line "Feishu path uses post-edit with a 20-edit cap before promoting to card" describes this project's own #1122 / #1159 behavior. I cannot verify OpenClaw's Feishu path myself — if the attribution is right, a link would settle it.
- Minor: the acceptance item "`cargo test --features unified` passes" is a local-only claim until CI runs that target. Worth marking as such; it is the round-3 review's F2 and needs a maintainer to add the step, which is outside a fork's reach.

## Notes, no action needed

- The override is fire-and-forget and always returns `Ok(())`, so core's `"delete placeholder failed; user will see overlap"` warning is unreachable in unified mode: a failed native delete leaves the stale card and records nothing. The comment states the trade-off and I agree with the call — the same one was made deliberately in the #1122 work. Noting it so the next person debugging a duplicated reply knows the log will be silent.
- `GATEWAY_RESPONSE_TIMEOUT` is 5s and turns both timeout and a closed channel into `Err`, where standalone's `edit_message` uses 800ms and returns `Ok(())` on timeout with a comment explaining that it avoids a false-positive failure marker. Inverted semantics on the card-mode edit path: a merely slow PUT now drives delete-and-resend against a card that may already have been updated. The window is narrow and the round-2 fallback removed the damaging half of this on the placeholder, so it is not worth blocking — but it is the remaining piece of my round-1 inline, and 5s also does not bound the dispatch itself, only the wait for a reply.
- Verified, not assumed: the recovery's delete-then-send cannot reorder, since both are awaited in sequence on the same task. And the fallback to the `"draft"` sentinel is quiet — the gateway skips draft-shaped ids at debug level and the cosmetic loop breaks after its failure threshold, so there is no warn storm.
- `feishu_post_mode_keeps_draft_strategy` also asserts the `other_bot_present = true` arm, which both gateway-path call sites currently pass as `false`. Harmless, and reasonable to pin for when it becomes reachable.

## What's good

- Round-1 items were fixed in the order the dependency required, and the commit message records why. That is usually the part a reviewer has to re-explain on the next round.
- `try_send_initial_card` keying the session by the card's own message id means the id core holds is the card id, so the recovery delete lands on the visible message with no translation. Worth carrying into #1464.
- Not aborting the turn when the placeholder fails is the right instinct — a cosmetic element should never cost the answer. The round-3 review is right that the current form gives up the card's identity while doing so; the instinct is sound and only the fallback target needs work.
- Collapsing two independent predicates into one strategy value is what made the round-2 finding visible at all. The abstraction is earning its keep even where it is still incomplete.

PR #1465, opened 2026-08-07
docs(adr): agent control plane for direct inter-agent delegation

chaodu-agent:
Proposes the **Agent Control Plane (CP)** — a hub-and-spoke registration and routing service so agents delegate to each other directly over ACP/WS instead of round-tripping through Discord/Telegram (rate limits, latency, 2000-char unstructured payloads, orchestration noise in human channels).

## Architecture

```mermaid
flowchart TB
    subgraph CP["openab-cp (standalone binary, stateless)"]
        REG["Registry<br/>who is alive, labels, replicas"]
        RTR["Router<br/>cp/delegate by name / label / ns"]
        POL["Policy<br/>depth · cycles · namespace · allowlist"]
    end

    subgraph OABA["OAB runtime: koudu — type=primary, ns=prod"]
        DA["Discord / Slack adapters"]
        FAC["MCP facade<br/>spawn_agent · check_delegation<br/>list_agents · cancel_delegation"]
        UDS["local API (UDS)<br/>/run/openab/agent.sock"]
        CPCA["CP client (WS, dial-out)"]
        AGA["Agent A (ACP stdio)"]
        DA --- AGA
        AGA -- "MCP tool call" --> FAC
        FAC --> UDS
        UDS --> CPCA
    end

    subgraph OABB["OAB runtime: worker-1 — type=worker (headless, no platform adapters)"]
        CPCB["CP client (WS, dial-out)"]
        POOL["pool.get_or_create"]
        AGB["Agent B (ACP stdio)"]
        CPCB --> POOL --> AGB
    end

    CLI["openab agent &lt;verb&gt;<br/>(ops / hooks / cron)"] --> UDS

    CPCA -- "register + cp/delegate" --> CP
    CPCB -- "register" --> CP
    CP -- "route delegation" --> CPCB

    USER(("Discord user")) --> DA
```

## Delegation flow

```mermaid
sequenceDiagram
    participant A as Agent A (LLM)
    participant OA as OAB-A (facade + CP client)
    participant CP as openab-cp
    participant OB as OAB-B (worker runtime)
    participant B as Agent B (LLM)

    Note over OA,CP: both runtimes already registered<br/>(namespace, name, type, labels, instance_id)

    A->>OA: MCP: spawn_agent{target: labels{backend=kiro}, prompt, deadline}
    OA->>OA: facade: validate schema, enforce chain/depth,<br/>clamp deadline, audit log
    OA->>CP: cp/delegate{delegation_id, target, chain:[koudu], deadline}
[...]

PR #1469, opened 2026-08-11
feat(cp): openab-cp control plane — registry, router, policy (PR 1/4)

chaodu-agent:

First implementation slice (PR 1 of 4) of the Agent Control Plane ADR (#1465): the standalone `openab-cp` binary — wire protocol, identity-bound registry, CP-authoritative policy engine, and delegation router. Agents delegate to each other over ACP/WS instead of round-tripping through Discord/Telegram.

Planned follow-up PRs: (2) OAB-side `[control_plane]` config + registration client + worker pool wiring, (3) MCP facade over UDS + `openab agent <verb>` CLI, (4) streaming + observer endpoint.

## What's in this PR

- **`crates/openab-cp`** (new workspace member, standalone binary per the `openab-gateway` precedent):
  - `proto.rs` — versioned JSON-RPC 2.0 envelopes; `cp/register`(+ack), `cp/heartbeat`, `cp/delegate`, `cp/delegate_result`, `cp/cancel`; distinct machine-actionable error codes (`IDENTITY_MISMATCH`, `POLICY_DENIED`, `NO_TARGET`, `SATURATED`, `DEADLINE_EXCEEDED`, `TARGET_DISCONNECTED`, `DUPLICATE_DELEGATION`, `UNSUPPORTED_VERSION`)
  - `config.rs` — CP-owned identity table (auth key → immutable `namespace`/`name`/`type` claims, `${ENV_VAR}` expansion, constant-time key lookup, duplicate-key rejection), per-namespace policy overrides
  - `registry.rs` — instance registry with replica semantics (rolling deploys route new work to the newest healthy instance), lease/heartbeat expiry, saturation-aware selection distinguishing `NO_TARGET` from `SATURATED`
  - `policy.rs` — CP-authoritative checks: initiator role, depth, cycle (incl. self), namespace boundary, deadline sanity/cap/parent-budget
  - `router.rs` — in-flight table; **CP-constructed chain** (callers pass only `parent_delegation_id`; ancestry derives from authenticated identities — unforgeable); deadline sweep (synthesized `timeout` + best-effort `cp/cancel`); disconnect handling both directions; result size truncation; idempotency
  - `server.rs` — axum WS server; auth via `Authorization: Bearer` at upgrade (keys never in URLs); mandatory `cp/register` first frame verified against the key's bound identity
- **ADR amendments** (`docs/adr/agent-control-plane.md`): new "v1 contract amendments" section freezing the #1465 review findings (identity binding, CP-authoritative policy, restart/saturation/timeout semantics, result cap, idempotency); §11 Q1 streaming resolved as committed fast-follow with the observer endpoint sketch.

### Goal
Ship a correct, tested control-plane core (registry + router + policy + wire contract) that later PRs (runtime client, facade, streaming) build on without protocol changes, implementing the frozen ADR baseline including the #1465 review checklist items.

### Non-goals
No OAB-runtime-side code: no `[control_plane]` config section, no registration client, no MCP facade/UDS/CLI, no streaming/observer endpoint (PRs 2–4). No CP HA (single instance; restart semantics are defined and implemented). No durable state, queueing, or delegation history. No Docker/Helm packaging.

### Accepted Residual Risks
- Wire contract is exercised by unit tests, not yet by a real OAB runtime client; PR 2 integration may surface adjustments. Mitigation: `protocol_version` field + `UNSUPPORTED_VERSION` rejection gives a compatibility gate.
- Registered-runtime JSON-RPC responses to CP-forwarded frames are logged and dropped (correlation is by `delegation_id`, not rpc id). Acceptable for v1; revisit if per-frame acks become necessary.
- `cp/cancel` to a serving runtime is best-effort; a runtime that ignores it burns tokens until its own deadline handling stops it. The propagated deadline is the hard upper bound.

### Acceptance Criteria
- Registration rejects: wrong first frame, unsupported protocol version, identity claim mismatch (namespace/name/type each), empty instance_id — all tested
- Policy denies: worker initiation (default), depth > namespace max, cycles incl. self-delegation, cross-namespace, past/over-cap/over-parent deadlines — all tested
- Router: happy-path roundtrip with CP-stamped chain; duplicate id rejection; saturation fast-fail; deadline sweep synthesizing timeout+cancel; worker/initiator disconnect handling; spoofed-completion rejection; late-result drop (CP restart case); oversized-result truncation — all tested
- `cargo fmt --check`, `cargo clippy --all-targets -- -D warnings`, `cargo test -p openab-cp` (42/42), `cargo check --workspace` all pass
- No mockall, no network/fs in unit tests, tests inline `#[cfg(test)]`

### Follow-ups
- PR 2: `[control_plane]` OAB config + outbound registration client (gateway.rs backoff shape) + worker pool wiring
- PR 3: MCP facade (`spawn_agent`, `check_delegation`, `list_agents`, `cancel_delegation`) over UDS + CLI
- PR 4: streaming frames + read-only `/observe` endpoint + `openab-cp tail`
- Packaging (Dockerfile/Helm) once PR 2 makes the binary deployable end-to-end

## Verified

On macmini at the squashed head (content-identical to ac9c023):
- `cargo fmt -p openab-cp -- --check` — clean
- `cargo clippy -p openab-cp --all-targets -- -D warnings` — clean
- `cargo test -p openab-cp` — 42 passed, 0 failed
- `cargo check --workspace` — passes with the new member

## ⚠️ Breaking Change — `admission` is required on `cp/delegate_result` and `cp/cancel`

The CP wire protocol gains a required `admission` token on two frames:
`cp/delegate_result` (round-8 revision) and `cp/cancel` (this revision, both
directions). Both are **required** fields, so this is a wire-breaking protocol
change: a runtime built against the pre-token contract has **every** result and
**every** cancel refused with `INVALID_PARAMS` once the CP is upgraded.

There is no backward-compatible spelling. An optional token would mean "apply
this frame to whatever holds the `delegation_id` right now", which is precisely
the misdelivery the token exists to close: `delegation_id` is client-supplied
and cancel-then-retry legitimately reuses it, so with a single replica the retry
re-admits the same id to the same worker. Without a token on the wire, a late
result for the cancelled admission is delivered to the initiator as the retry's
terminal frame and commits the retry — releasing capacity it still occupies —
and a stale or CP-synthesized cancel for the ended admission aborts the retry's
work at the worker.

### Migration for runtime authors

- **Serving runtimes:** copy `admission` from the forwarded `cp/delegate` into
  the matching `cp/delegate_result`. Match incoming `cp/cancel` on the token,
  not on `delegation_id`: a CP-synthesized cancel names the admission it ends
  and can arrive *after* a forward that reused the same id.
- **Initiators:** read `admission` from the `cp/delegate` ack. Send it on
  `cp/cancel` to name the admission you mean to abort — a cancel naming a
  superseded admission is refused, which is what prevents a retried cancel from
  killing the re-admission that replaced its target. Correlate terminal frames
  on `admission`, not on `delegation_id` ("first terminal frame per admission
  wins").
- **Refusals are indistinguishable by design.** An unknown id, another
  instance's live id, and your own id under a superseded admission all return
  the same byte-identical `POLICY_DENIED`. Treat a refusal as "not mine / not
  live" and reconcile against your own state; the CP deliberately reports
  nothing about whether an id is currently re-admitted.

No shipped clients are affected at this point in the stack — every serving
runtime in PRs 2–4 learns the token from the forwarded `cp/delegate` — so the
migration is mechanical. It is not, however, silent or optional.

Contract details: `docs/adr/agent-control-plane.md` §4 (v1 contract
amendments, "Admissions carry a protocol-visible token"); runtime-author
summary in `docs/control-plane.md` ("Client behavior to expect").

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2593,7 +2593,6 @@ dependencies = [
  "toml",
  "tracing",
  "tracing-subscriber",
- "uuid",
 ]
 
 [[package]]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2576,6 +2576,26 @@ dependencies = [
 ]
 
 [[package]]
+name = "openab-cp"
+version = "0.1.0"
+dependencies = [
+ "anyhow",
+ "axum",
+ "chrono",
+ "clap",
+ "futures-util",
+ "parking_lot",
+ "serde",
+ "serde_json",
+ "subtle",
+ "tokio",
+ "toml",
+ "tracing",
+ "tracing-subscriber",
+ "uuid",
+]
+
+[[package]]
 name = "openab-gateway"
 version = "0.5.4"
 dependencies = [
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2589,6 +2589,7 @@ dependencies = [
  "serde_json",
  "subtle",
  "tokio",
+ "tokio-tungstenite 0.29.0",
  "toml",
  "tracing",
  "tracing-subscriber",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,5 +1,5 @@
 [workspace]
-members = ["crates/openab-core", "crates/openab-gateway", "crates/openab-mcp"]
+members = ["crates/openab-core", "crates/openab-gateway", "crates/openab-mcp", "crates/openab-cp"]
 exclude = ["openab-agent", "crates/platform-schema"]
 
 [package]
--- a/Dockerfile
+++ b/Dockerfile
@@ -11,16 +11,18 @@ COPY Cargo.toml Cargo.lock ./
 COPY crates/openab-core/Cargo.toml crates/openab-core/Cargo.toml
 COPY crates/openab-gateway/Cargo.toml crates/openab-gateway/Cargo.toml
 COPY crates/openab-mcp/Cargo.toml crates/openab-mcp/Cargo.toml
-RUN mkdir -p src crates/openab-core/src crates/openab-gateway/src crates/openab-mcp/src \
+COPY crates/openab-cp/Cargo.toml crates/openab-cp/Cargo.toml
+RUN mkdir -p src crates/openab-core/src crates/openab-gateway/src crates/openab-mcp/src crates/openab-cp/src \
     && echo 'fn main() {}' > src/main.rs \
     && echo '' > crates/openab-core/src/lib.rs \
     && echo '' > crates/openab-gateway/src/lib.rs \
     && echo '' > crates/openab-mcp/src/lib.rs \
+    && echo '' > crates/openab-cp/src/lib.rs \
     && cargo build --release \
-    && rm -rf src crates/openab-core/src crates/openab-gateway/src crates/openab-mcp/src
+    && rm -rf src crates/openab-core/src crates/openab-gateway/src crates/openab-mcp/src crates/openab-cp/src
 COPY crates/ crates/
 COPY src/ src/
-RUN touch src/main.rs crates/openab-core/src/lib.rs crates/openab-gateway/src/lib.rs crates/openab-mcp/src/lib.rs && \
+RUN touch src/main.rs crates/openab-core/src/lib.rs crates/openab-gateway/src/lib.rs crates/openab-mcp/src/lib.rs crates/openab-cp/src/lib.rs && \
     if [ "$BUILD_MODE" = "unified" ]; then \
       cargo build --release --features unified; \
     elif [ -n "$FEATURES" ]; then \
--- a/Dockerfile.agentcore
+++ b/Dockerfile.agentcore
@@ -9,16 +9,18 @@ COPY Cargo.toml Cargo.lock ./
 COPY crates/openab-core/Cargo.toml crates/openab-core/Cargo.toml
 COPY crates/openab-gateway/Cargo.toml crates/openab-gateway/Cargo.toml
 COPY crates/openab-mcp/Cargo.toml crates/openab-mcp/Cargo.toml
-RUN mkdir -p src crates/openab-core/src crates/openab-gateway/src crates/openab-mcp/src \
+COPY crates/openab-cp/Cargo.toml crates/openab-cp/Cargo.toml
+RUN mkdir -p src crates/openab-core/src crates/openab-gateway/src crates/openab-mcp/src crates/openab-cp/src \
     && echo 'fn main() {}' > src/main.rs \
     && echo '' > crates/openab-core/src/lib.rs \
     && echo '' > crates/openab-gateway/src/lib.rs \
     && echo '' > crates/openab-mcp/src/lib.rs \
+    && echo '' > crates/openab-cp/src/lib.rs \
     && cargo build --release --features agentcore \
-    && rm -rf src crates/openab-core/src crates/openab-gateway/src crates/openab-mcp/src
+    && rm -rf src crates/openab-core/src crates/openab-gateway/src crates/openab-mcp/src crates/openab-cp/src
 COPY crates/ crates/
 COPY src/ src/
-RUN touch src/main.rs crates/openab-core/src/lib.rs crates/openab-gateway/src/lib.rs crates/openab-mcp/src/lib.rs && cargo build --release --features agentcore
+RUN touch src/main.rs crates/openab-core/src/lib.rs crates/openab-gateway/src/lib.rs crates/openab-mcp/src/lib.rs crates/openab-cp/src/lib.rs && cargo build --release --features agentcore
 
 # --- Runtime stage ---
 FROM debian:trixie-slim
--- a/Dockerfile.antigravity
+++ b/Dockerfile.antigravity
@@ -5,16 +5,18 @@ COPY Cargo.toml Cargo.lock ./
 COPY crates/openab-core/Cargo.toml crates/openab-core/Cargo.toml
 COPY crates/openab-gateway/Cargo.toml crates/openab-gateway/Cargo.toml
 COPY crates/openab-mcp/Cargo.toml crates/openab-mcp/Cargo.toml
-RUN mkdir -p src crates/openab-core/src crates/openab-gateway/src crates/openab-mcp/src \
+COPY crates/openab-cp/Cargo.toml crates/openab-cp/Cargo.toml
+RUN mkdir -p src crates/openab-core/src crates/openab-gateway/src crates/openab-mcp/src crates/openab-cp/src \
     && echo 'fn main() {}' > src/main.rs \
     && echo '' > crates/openab-core/src/lib.rs \
     && echo '' > crates/openab-gateway/src/lib.rs \
     && echo '' > crates/openab-mcp/src/lib.rs \
+    && echo '' > crates/openab-cp/src/lib.rs \
     && cargo build --release \
-    && rm -rf src crates/openab-core/src crates/openab-gateway/src crates/openab-mcp/src
+    && rm -rf src crates/openab-core/src crates/openab-gateway/src crates/openab-mcp/src crates/openab-cp/src
 COPY crates/ crates/
 COPY src/ src/
-RUN touch src/main.rs crates/openab-core/src/lib.rs crates/openab-gateway/src/lib.rs crates/openab-mcp/src/lib.rs && cargo build --release
+RUN touch src/main.rs crates/openab-core/src/lib.rs crates/openab-gateway/src/lib.rs crates/openab-mcp/src/lib.rs crates/openab-cp/src/lib.rs && cargo build --release
 
 # --- Build agy-acp adapter ---
 FROM rust:1-bookworm AS adapter-builder
--- a/Dockerfile.builder
+++ b/Dockerfile.builder
@@ -23,20 +23,22 @@ COPY Cargo.toml Cargo.lock ./
 COPY crates/openab-core/Cargo.toml crates/openab-core/Cargo.toml
 COPY crates/openab-gateway/Cargo.toml crates/openab-gateway/Cargo.toml
 COPY crates/openab-mcp/Cargo.toml crates/openab-mcp/Cargo.toml
+COPY crates/openab-cp/Cargo.toml crates/openab-cp/Cargo.toml
 
 # 2. Dummy sources for dep-only build
-RUN mkdir -p src crates/openab-core/src crates/openab-gateway/src crates/openab-mcp/src \
+RUN mkdir -p src crates/openab-core/src crates/openab-gateway/src crates/openab-mcp/src crates/openab-cp/src \
     && echo 'fn main() {}' > src/main.rs \
     && echo '' > crates/openab-core/src/lib.rs \
     && echo '' > crates/openab-gateway/src/lib.rs \
     && echo '' > crates/openab-mcp/src/lib.rs \
+    && echo '' > crates/openab-cp/src/lib.rs \
     && cargo build --release \
-    && rm -rf src crates/openab-core/src crates/openab-gateway/src crates/openab-mcp/src
+    && rm -rf src crates/openab-core/src crates/openab-gateway/src crates/openab-mcp/src crates/openab-cp/src
 
 # 3. Copy real sources and build
 COPY crates/ crates/
 COPY src/ src/
-RUN touch src/main.rs crates/openab-core/src/lib.rs crates/openab-gateway/src/lib.rs crates/openab-mcp/src/lib.rs && \
+RUN touch src/main.rs crates/openab-core/src/lib.rs crates/openab-gateway/src/lib.rs crates/openab-mcp/src/lib.rs crates/openab-cp/src/lib.rs && \
     if [ "$BUILD_MODE" = "unified" ]; then \
       cargo build --release --features unified; \
     elif [ -n "$FEATURES" ]; then \
--- a/Dockerfile.claude
+++ b/Dockerfile.claude
@@ -5,16 +5,18 @@ COPY Cargo.toml Cargo.lock ./
 COPY crates/openab-core/Cargo.toml crates/openab-core/Cargo.toml
 COPY crates/openab-gateway/Cargo.toml crates/openab-gateway/Cargo.toml
 COPY crates/openab-mcp/Cargo.toml crates/openab-mcp/Cargo.toml
-RUN mkdir -p src crates/openab-core/src crates/openab-gateway/src crates/openab-mcp/src \
+COPY crates/openab-cp/Cargo.toml crates/openab-cp/Cargo.toml
+RUN mkdir -p src crates/openab-core/src crates/openab-gateway/src crates/openab-mcp/src crates/openab-cp/src \
     && echo 'fn main() {}' > src/main.rs \
     && echo '' > crates/openab-core/src/lib.rs \
     && echo '' > crates/openab-gateway/src/lib.rs \
     && echo '' > crates/openab-mcp/src/lib.rs \
+    && echo '' > crates/openab-cp/src/lib.rs \
     && cargo build --release \
-    && rm -rf src crates/openab-core/src crates/openab-gateway/src crates/openab-mcp/src
+    && rm -rf src crates/openab-core/src crates/openab-gateway/src crates/openab-mcp/src crates/openab-cp/src
 COPY crates/ crates/
 COPY src/ src/
-RUN touch src/main.rs crates/openab-core/src/lib.rs crates/openab-gateway/src/lib.rs crates/openab-mcp/src/lib.rs && cargo build --release
+RUN touch src/main.rs crates/openab-core/src/lib.rs crates/openab-gateway/src/lib.rs crates/openab-mcp/src/lib.rs crates/openab-cp/src/lib.rs && cargo build --release
 
 # --- Runtime stage ---
 FROM node:22-trixie-slim
--- a/Dockerfile.codex
+++ b/Dockerfile.codex
@@ -5,16 +5,18 @@ COPY Cargo.toml Cargo.lock ./
 COPY crates/openab-core/Cargo.toml crates/openab-core/Cargo.toml
 COPY crates/openab-gateway/Cargo.toml crates/openab-gateway/Cargo.toml
 COPY crates/openab-mcp/Cargo.toml crates/openab-mcp/Cargo.toml
-RUN mkdir -p src crates/openab-core/src crates/openab-gateway/src crates/openab-mcp/src \
+COPY crates/openab-cp/Cargo.toml crates/openab-cp/Cargo.toml
+RUN mkdir -p src crates/openab-core/src crates/openab-gateway/src crates/openab-mcp/src crates/openab-cp/src \
     && echo 'fn main() {}' > src/main.rs \
     && echo '' > crates/openab-core/src/lib.rs \
     && echo '' > crates/openab-gateway/src/lib.rs \
     && echo '' > crates/openab-mcp/src/lib.rs \
+    && echo '' > crates/openab-cp/src/lib.rs \
     && cargo build --release \
-    && rm -rf src crates/openab-core/src crates/openab-gateway/src crates/openab-mcp/src
+    && rm -rf src crates/openab-core/src crates/openab-gateway/src crates/openab-mcp/src crates/openab-cp/src
 COPY crates/ crates/
 COPY src/ src/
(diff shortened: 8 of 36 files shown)


chaodu-agent:
All round-1 findings addressed in e2853b4 (branch rewritten; the unrelated artifacts are gone from the diff entirely).

| # | Finding | Resolution |
|---|---------|------------|
| F1 🔴 | Client-selected `instance_id` as registry/ownership key | ✅ Registrations now keyed by a **CP-generated handle** (`registry.rs`); all ownership checks (completion, cancel, parent linkage, teardown) compare handles. A colliding `instance_id` cannot replace or tear down another connection's entry — regression tests `colliding_instance_id_cannot_replace_other_registration`, `result_from_wrong_handle_dropped_and_restored` |
| F2 🔴 | Check-then-act admission; send-before-insert | ✅ Admission (duplicate check → parent lookup → selection → capacity reservation → in-flight insert) is one atomic sequence under an admission lock; the in-flight entry exists **before** the forward frame is sent, with rollback on send failure — tests `inflight_exists_before_target_receives_frame`, `send_failure_rolls_back_reservation` |
| F3 🔴 | Any live `parent_delegation_id` accepted | ✅ The caller must be the instance **serving** the parent (`p.to_handle == from_handle`); unknown and unauthorized parents return the same non-enumerating error — test `chain_extends_through_parent_and_foreign_parent_rejected` (foreign-parent negative case included) |
| F4 🟡 | JSON-RPC 2.0 envelope not validated | ✅ `jsonrpc` field added to the incoming type; `require_request_envelope()` enforces `"2.0"` + request id for all `cp/*` methods, returning `INVALID_REQUEST` (-32600) — tests `request_envelope_validation`, `register_invalid_envelope_rejected` |
| F5 🟡 | No pre-allocation resource bounds | ✅ WS transport enforces `max_frame_bytes` (default 1 MiB) via `max_message_size`/`max_frame_size` before parsing; `max_prompt_bytes` rejects oversized prompts; per-connection outbound queues are bounded (256) with disconnect-on-overflow (`try_send` everywhere) |
| F6 🟡 | Heartbeat count merge; recency-over-load label scheduling | ✅ Heartbeats refresh the lease only — CP-owned counts are authoritative (test `heartbeat_does_not_mutate_session_count`); label selection is least-loaded-first with recency tie-break, exact-name keeps the newest-replica rule (inverse recency/load test `label_selection_least_loaded_first`) |
| F7 🟡 | Unrelated artifacts in diff | ✅ Branch rewritten from main; diff now contains only `crates/openab-cp`, workspace `Cargo.toml`/`Cargo.lock`, the ADR amendment, and the Dockerfile stub-layer updates required for the new workspace member |

ADR's "v1 contract amendments" section updated to record the handle-based ownership, envelope validation, and resource-bound semantics.

Verified at e2853b4 on the build host: `cargo fmt --check` OK, `cargo clippy --all-targets -- -D warnings` clean, `cargo test -p openab-cp` 48/48 (up from 42 — six new regression tests for F1/F2/F3/F4/F6), `cargo check --workspace` passes.

chaodu-agent:
Round-2 response — note the round-2 review examined **stale head `2842d264`** (pre-rewrite); findings F1/F2/F3/F6/F7 were already resolved in the branch rewrite (`e2853b4`, see the previous response comment for the per-finding test evidence). The two findings that applied to current code are now fixed in `33ed835`:

| # | Finding | Resolution |
|---|---------|------------|
| F4 🔴 | Bearer credentials over cleartext non-loopback TCP | ✅ Default bind is now `127.0.0.1:9800`; `validate()` rejects any non-loopback bind unless `allow_insecure_bind = true` is set explicitly, documented as requiring a TLS-terminating proxy (wss://) or a private network (tailnet) in front. `cp.toml.example` updated. Test: `non_loopback_bind_requires_override` (covers 0.0.0.0 rejection, override acceptance, and loopback variants incl. `[::1]`) |
| F5 🟡 (marker overflow) | Truncation marker pushed results past `max_result_bytes` | ✅ The marker now counts against the cap (`floor_char_boundary` budget math); degenerate tiny caps also never exceed the limit. Test: `oversized_result_truncated` asserts `result.len() <= cap` for both normal and degenerate caps |

The remainder of F5 (bounded outbound queues, `max_frame_bytes` WS transport limit, `max_prompt_bytes` rejection) was already in `e2853b4`.

Per the review contract's stopping rule, round 3 should verify these fixes against head `33ed835` and check for regressions only.

Verified at `33ed835` on the build host: fmt clean, clippy `--all-targets -D warnings` clean, 49/49 tests, `cargo check --workspace` passes.

chaodu-agent:
Round-4 response -- all fifteen findings addressed at `09968d1` (nine fixed in code/docs, five filed as follow-up issues, one classified for the maintainer).

## Fixed in `09968d1`

| # | Resolution |
|---|-----------|
| F2 | `uuid` dependency removed from `crates/openab-cp/Cargo.toml`; `Cargo.lock` regenerated (the crate loses the dep edge; the package entry remains for other workspace crates). |
| F3 | `codes::DEADLINE_EXCEEDED` **REMOVED** rather than reserved: admission already rejects expired deadlines via `PolicyDenial::DeadlinePast` -> `POLICY_DENIED`, and in-flight expiry is a synthesized `timeout` result, not an error code -- so no plausible future consumer exists. `-32006` is documented as unassigned to keep published numbering stable. |
| F4 | The duplicated ownership-check-and-remove now lives in one private `Router::claim(registry, handle, delegation_id, Owner) -> Claim {Owned, WrongOwner, NotFound, Unregistered}`. The single-lock-acquisition atomicity (round-3 F2) holds inside the helper; `complete`/`cancel` keep their distinct logging and return shapes. Zero changes inside `mod tests` -- the pre-existing race/ownership tests pass unmodified, which is the behavior-preservation proof. |
| F5 | Every `(review round-N FX)` tag rewritten into self-standing rationale; `grep -rniE 'round-[0-9]' crates/openab-cp/` is empty. The rationale prose was kept, only the external index references went. |
| F6 | ADR amended to match the enforced contract: registration deadline (pings do not extend it, 1008 close), lease expiry actively closes (recovery = reconnect + re-register, new handle), `max_connections_per_identity` counted from the upgrade with RAII release, `(namespace, delegation_id)` scoping with indistinguishable cancel refusals, and corrected CP-restart semantics. One further correction surfaced while verifying against the code: `fail_instance` outcomes are **side-specific and mutually exclusive** (serving-side death -> `target_disconnected` to the initiator; initiator-side death -> best-effort `cp/cancel` downstream) -- the ADR previously implied both frames per dead instance; only the deadline sweep emits both, because there both peers are still connected. |
| F7 | CP-initiated closes now carry meaning: WS close code **1008** with reason `"registration timeout"` / `"lease expired"`; the over-quota 503 carries a body naming `max_connections_per_identity`. The e2e lifecycle assertions were extended (not weakened) to check code, reason, and body -- and mutation-checked: altering the code or body fails the suite. |
| F13 | Frames arriving in the sweep->close window are now answered with a correlated `NOT_REGISTERED` JSON-RPC error instead of being dropped; a new test drives heartbeat-then-sweep-then-replay and asserts the error, and that a swept `cp/delegate` never reaches the router. |
| F15 | `router.rs` module doc now states the lock hierarchy: `admission -> inflight`, never the reverse; the registry lock is independent and never held together with `inflight`. |

## F1 -- answered rather than applied

`InFlight.namespace` is unread **within this PR**, but its consumer is one PR up the stack: #1470's observer event layer reads `e.namespace` in exactly the two paths that end a delegation without a client request (`fail_instance`, `sweep_deadlines`), where no request-scoped namespace exists. Removing it here would only force the stack to re-add it. The field now carries a doc comment saying precisely that, so it no longer reads as a remnant.

## F14 -- classification (maintainer-confirmed)

The admission bounds are **intentional hardening, not a breaking change**: this crate is pre-merge and has never shipped, so no deployment has ever run the unbounded behavior -- there is no prior contract to break and no migration to document. The bounds and their defaults (8 connections / 10s) are recorded in the ADR (per F6) and in `cp.toml.example`. Should a deployment legitimately need more than 8 replicas of one identity, the knob is per-identity configurable.

## F8–F12 -- follow-up issues filed

- #1473 -- scale hardening: indexed sweep failure paths (F8), sharded admission + serialization outside the lock (F9), borrowed lookup keys (F10). F9 remains an author-documented v1 tradeoff; the issue records the ceiling and the load-test requirement.
- #1474 -- sweeper panic supervision + liveness (F11). Same failure class as the session-pool reaper fixed in #1457; filed as a bug because silent sweeper death disables lease and deadline enforcement while `/health` stays green.
- #1475 -- container-grade shutdown: SIGTERM/SIGHUP handling, connection drain with Close frames, then sweeper stop (F12).

## Non-blocking observations

The idempotent double-cleanup note, `conn_count` visibility, and the `register_timeout_secs` cross-continent-TLS caveat are acknowledged; the ADR now records the assumption behind the 10s default. The pre-1.0 protocol framing suggestion is adopted implicitly: PR 3/4 (#1471) already exercises this wire contract end-to-end with a real runtime client against the in-process server, and no impedance mismatch has surfaced through four review rounds.

## Verification

`cargo test -p openab-cp` at `09968d1`: **68 passed** (baseline 67 + the F13 regression); clippy `--all-targets -D warnings` clean; rustfmt clean. The stacked branches were rebased onto this head and re-verified: #1470 at 100 tests (95 lib + 5 e2e), #1471 with its full suite (95 cp + 37 control-plane lib + 7 real-server integration + 25 bin) at exact clippy/fmt baselines.

chaodu-agent:
Round-6 response -- both critical findings and all four important findings addressed at `bde414a`.

## Fixed

| # | Resolution |
|---|-----------|
| F1 (critical) | Every admission is stamped with a CP-generated, never-reused **generation** (minted via `fetch_update` -- an exhausted counter is never written, so it can neither wrap nor re-issue; admission fails closed at the ceiling, with a boundary test pinning the refusal and that no capacity is reserved). The commit phase is an extracted, directly-callable step that removes the entry under one in-flight lock acquisition only when key AND serving handle AND generation all match; any mismatch leaves the entry -- and its capacity -- untouched. `CompleteOutcome::Delivered` became `Delivered { committed: bool }`, so wire delivery and state commit are reported distinctly instead of an unconditional `Delivered`. Deterministic regression drives the exact ABA sequence: peek A -> remove A -> re-admit B with the same id and worker -> commit A => `Superseded`, B still in flight with capacity intact, B then completes normally. |
| F2 (critical) | Every outbound write goes through a shutdown-aware, timeout-bounded send helper (new `write_timeout_secs` config, default 30, validated > 0, documented). Timeout or shutdown is a disconnect: teardown runs, the `ConnPermit` drops, downstream cancellation fires. Coverage includes the registration-phase writes and the post-loop close. Real-socket regression: a peer that stops reading has its task terminated and its quota slot freed **within the configured bound plus scheduling allowance** (the assertion is bounded by the timeout, not an unrelated poll window), and a new connection for the same identity then succeeds. Negative control: restoring the unbounded send fails the test at 30s+. |
| F3 | Documented first-terminal-wins in the ADR wire contract (initiators treat the first terminal frame per `delegation_id` as authoritative and ignore later ones); the router docs no longer describe the peek-send window as benign duplicate delivery. No suppression machinery this round -- CP state is exact under the generation rule; the wire contract handles the rest. |
| F4 | `docs/control-plane.md`: `[[identity]]` -> `[[agents]]`, cross-checked against `cp.toml.example` and `config.rs`. |
| F5 | `README.zh-TW.md` gained the control-plane feature entry mirroring `README.md`'s placement and content. |
| F6 | Deterministic interleaving tests drive the extracted commit step directly: commit-vs-cancel and commit-vs-sweep (both => `Vanished`, capacity released exactly once, verified by re-saturation), stale-commit-vs-re-admission (=> `Superseded`), and `delivered_reports_whether_it_committed`. No barrier timing. |

## Verification

`cargo test -p openab-cp` at `bde414a`: **71 lib + 6 e2e passed** (baseline 66+5); clippy `--all-targets -D warnings` clean; rustfmt clean. Negative controls confirmed both critical regressions bite: removing the generation match fails the ABA test with `left: Claimed, right: Superseded`; restoring the unbounded send fails the stalled-peer test after 30s.

Assertion changes are extensions only: the nine existing `Delivered` assertions now pin `committed: true` (strictly stronger), and the admission-bounds config test extends to `write_timeout_secs`.

The stacked branches were rebased and re-verified: #1470 at 103 lib + 6 e2e (zero test functions dropped from either side, verified by name-level parity audit), #1471 with its full suite (103 cp + 37 control-plane lib + 7 real-server integration + 25 bin) at exact clippy/fmt baselines.

chaodu-agent:
Round-8 response -- the critical finding and all four important findings addressed at `d1b5104`.

## Fixed

| # | Resolution |
|---|-----------|
| F1 (critical) | The admission identity is now protocol-visible. `DelegateAck`, `DelegateForward`, and `DelegateResultParams` carry an `admission` token; the echo is REQUIRED (absent = `INVALID_PARAMS`, never a wildcard); CP-synthesized terminals (`timeout`, `target_disconnected`) build from the entry so they carry the ending admission's token; `complete()`'s peek matches the echoed token BEFORE the initiator frame is built, so a late result for a superseded admission is dropped without touching the live entry. Token minting moved from one global counter to **per-namespace** counters (a global monotonic value on the wire would be a cross-namespace volume oracle of exactly the class round-3 F3 eliminated); the fail-closed exhaustion behavior is retained per namespace, with the boundary test. First-terminal-wins is re-keyed **per admission token** in the ADR and quickstart; `cp/cancel` token intentionally deferred to the runtime-client slice as the review suggested. Wire regressions: (a) A cancelled, B re-admitted same id + same worker, late result echoing A's token => dropped, B untouched with capacity intact, B completes normally with its own token; (b) A's and B's frames carry distinct tokens; plus real-socket e2e for the late-result drop and the missing-token rejection. |
| F2 | Outbound queues are byte-bounded: `FrameTx` reserves bytes before enqueue and releases on dequeue AND on teardown (`FrameRx::drop` drains); new `max_outbound_queue_bytes` (default 16 MiB, validated > `max_frame_bytes`). Regressions: refusal fires on the byte budget well before entry-count exhaustion; teardown releases the whole reservation. The pre-existing stalled-writer e2e sets its budget above its payload so it still proves the WRITE-TIMEOUT path (budget refusal has its own dedicated regressions). |
| F3 | Global `max_inflight_delegations` (default 4096, validated > 0) enforced at admission before capacity reservation => `SATURATED` naming the bound. The bound reads the in-flight table's own length, so it cannot drift from reality; the regression exercises release via all five removal paths (commit, cancel, sweep, fail_instance, rollback), re-admitting after each. |
| F4 | New `default_max_delegated_sessions_cap` (default 16, validated > 0) clamps any identity without a per-identity cap; the per-identity knob overrides in either direction. Tests: `u32::MAX` advertised => effective 16; e2e ack reports the clamped value. |
| F5 | Teardown now runs from a guard scoped to the registered lifetime, so normal return and panic unwind take the identical path; idempotency with the sweeper verified. Regressions: a panicking registered task still deregisters, clears in-flight rows, releases capacity, and queues downstream `cp/cancel`; double teardown does not double-release. |

## Stack propagation

The echo requirement lands with the stack: #1471's executor now carries `forward.admission` into all five terminal frame kinds, and its real-server integration suite passes against the token-enforcing CP -- the first live exercise of the new wire rule. #1470 rebased cleanly (114 lib + 9 e2e).

## Verification

`cargo test -p openab-cp` at `d1b5104`: **82 lib + 9 e2e passed** (baseline 71+6); clippy `--all-targets -D warnings` clean; rustfmt clean. Stack: #1470 at 114+9; #1471 full suite (37 control-plane lib + 7 integration + 25 bin) at clippy/fmt baselines.

On the accepted residual risks: agreed on all three; the key-length timing normalization is noted for the F1-adjacent follow-up. The docs cross-reference follow-ups are noted for a docs pass after the stack merges.

chaodu-agent:
Round-9 response -- the critical finding and both important findings addressed at `fbc8f96`.

## Fixed

| # | Resolution |
|---|-----------|
| F1 (critical) | The admission token now covers cancellation symmetrically. `CancelParams.admission` is REQUIRED; initiator cancels match `(from_handle, key, admission)` under the one lock acquisition that removes the entry, and a stale token is refused byte-identically with the unknown/wrong-owner refusals (extended serialization-equality test). Every CP-synthesized cancel -- deadline sweep and `fail_instance` (initiator disconnect; the stalled-initiator path routes through it, verified by grepping every frame producer) -- stamps the ended admission's generation from the entry it removed, so a cancel that races a same-id re-admission still names the admission that is over. `Claim`'s doc no longer implies the cancel path needs no admission identity. Both decisive regressions: sweep-expire A -> re-admit B same id+worker -> A's cancel frame carries A's token and B stays in flight with capacity intact; initiator retry with A's token -> refused, B untouched, B completes with its own token. All four new/extended tests negative-controlled. |
| F2 | The ADR `cp/delegate` example now shows what an initiator sends (`parent_delegation_id`, no `chain`/`admission`); the CP-forwarded frame is a separate block; a `cp/cancel` example was added. Bonus landed: `adr_wire_examples_round_trip_through_the_serde_structs` pulls the ADR in via `include_str!` and round-trips every JSON example through the structs -- the drift class this finding caught is now a compile-time-checked contract (negative control: reintroducing `"chain"` fails with the intended message). |
| F3 | The wire-breaking contract is called out in the ADR amendment block and the quickstart, and the PR description gains a compatibility section: runtimes built against the pre-token contract have every result and cancel refused `INVALID_PARAMS` after upgrade. |

## Stack propagation

#1471's runtime is now token-complete on both directions: the executor echoes `forward.admission` in all five terminal frames, and the worker-side cancel handler matches `(delegation_id, admission)` -- a stale cancel naming a superseded admission is ignored instead of aborting the live delegation (the worker-side half of the misdelivery this round closed, with its own regression). The initiator-side integration test echoes the ack's admission through `cp/cancel` against the real CP.

## Verification

`cargo test -p openab-cp` at `fbc8f96`: **85 lib + 9 e2e passed** (baseline 82+9); clippy `--all-targets -D warnings` clean; rustfmt clean. Stack: #1470 at 117+9; #1471 at 38 control-plane lib + 7 integration + 25 bin, clippy/fmt at baselines.

chaodu-agent:
Round-10 response -- the critical finding addressed at `cadca62`.

## Fixed

| # | Resolution |
|---|-----------|
| F1 (critical) | Parent linkage is now admission-exact -- the third and final reusable-id surface joins result (r8) and cancel (r9). `DelegateParams` gains `parent_admission`, presence-coupled with `parent_delegation_id` (id without token = `INVALID_PARAMS` naming the missing half; token without id likewise REJECTED -- silently treating it as a root would hide a client bug and mint a delegation with no ancestry). The parent branch matches id + serving handle + generation under the same single lock; stale admissions reuse the pre-existing refusal verbatim, so unknown/foreign/stale are byte-identical (three-state serialization-equality test -- no re-admission oracle). Chain and deadline derivation verified to read from the validated entry (no re-lookup existed; a comment now forbids introducing one). |

The module-level contract summary now states the general invariant the three rounds converged on: **every surface that references an existing delegation names it by the (id, admission) pair; a bare reusable id is never accepted as a reference.** A future fourth surface has a stated rule to violate rather than an enumeration to fall outside of.

## Regressions (all negative-controlled)

- Cancel-path and sweep-path ABA: end parent A, re-admit same-id B on the same worker, submit a child naming A's admission -> refused; B's chain, deadline, generation, endpoints, and both workers' capacity asserted untouched; no child admitted.
- Positive control sharpened beyond the request: parent admitted at 120s, ended, re-admitted at 60s -- a 90s child (fits A's budget, not B's) is DENIED, proving the clamp derives from the live admission rather than being merely consistent with it.
- Reverting only the generation comparison fails both ABA regressions and the oracle test (88/3) while the positive control passes.
- The ADR's parented example now carries `parent_admission`, and the include_str round-trip test gained an explicit pair-coupling assertion -- the original key-set check would NOT have caught a missing optional field (a gap found while wiring the requested enforcement, now closed).

## Docs

ADR section 4: "currently serving that parent" redefined as serving handle + specific forwarded admission; wire-breaking callout extended to parented delegations (roots unchanged). Quickstart mirrors both.

## Verification

`cargo test -p openab-cp` at `cadca62`: **91 lib + 9 e2e passed** (baseline 85+9); clippy `--all-targets -D warnings` clean; rustfmt clean. Stack rebased and green: #1470 at 123+9; #1471 at 38+7+25 (its integration suite issues root delegations only, so the parented-reference contract required no client change); clippy/fmt at exact baselines.

thepagent (approved):

PR #1533, opened 2026-09-14
feat(discord): support ACP form elicitation

gandazgul:
# Discord ACP form elicitation

## Summary

- Add official ACP v1 `elicitation/create` form support for Discord-backed sessions.
- Advertise `clientCapabilities.elicitation.form` only when a Discord form presenter is available.
- Add a platform-neutral elicitation coordinator, Discord progressive form presentation, text-only fallback commands, lifecycle invalidation, and bounded request handling.
- Add deterministic subprocess and behavior tests for the ACP wire contract, authorization, overload, lifecycle, and Discord presentation repair paths.
- Update the Discord guide, canary guide, domain language, README, and plan/work-record artifacts.

## Validation

Automated checks reported by the saved work record and prior execution:

- `cargo check --workspace` — passed.
- `cargo clippy --workspace -- -D warnings` — passed.
- `cargo test --workspace` — passed.
- `cargo clippy --workspace --features unified -- -D warnings` — passed.
- `cargo test --features acp` — passed.
- `cargo build --features unified` — passed.
- `cargo fmt --check` — did not pass cleanly. The saved work record says repository-wide formatting diffs outside the feature path remained.
- `cargo check --target x86_64-pc-windows-gnu` — not run successfully because the Windows target standard library was not installed.

No live Discord manual canary was run in this preparation step.

## Review notes

- Prepared with [RunWield](https://github.com/gandazgul/runwield).
- Branch: `gandazgul:discord-acp-form-elicitation`
- Base: `openabdev:main`
- Head commit: `ccc334e1993980b1c88097809c158699a78af6e9`
- Diff size: 49 files changed, 9228 insertions, 1898 deletions.

### Goal

This PR lets OpenAB act as an ACP Client that can receive official ACP v1 form elicitation requests from a downstream Agent Runtime during Discord turns. It must show bounded, progressive forms to the verified human users who contributed to the active turn, return exactly one ACP result, and leave unsupported adapters unchanged.

### Non-goals

- No URL elicitation support.
- No Slack, gateway, or other non-Discord form UI.
- No migration of the legacy `session/request_input` extension in `openab-agent`.
- No nested JSON Schema or array support beyond string-array multi-select enums.
- No secret-collection flow. The Discord form warns users not to enter credentials or other secrets.

### Accepted Residual Risks

- Discord cleanup after a terminal Submit, Decline, Cancel, expiry, or teardown is best effort. A stale visible message can remain if Discord edit/delete fails, but server-side nonce, revision, message, generation, and coordinator checks must prevent a second ACP response or state change.
- Live Discord manual evidence is deferred to canary validation because this preparation used the deterministic test harness and no Discord token.
- `cargo fmt --check` is not clean in the saved validation record because repository-wide formatting diffs remained outside the feature path. Reviewers should decide whether to require a separate formatting cleanup before merge.
- Windows cross-check evidence is absent because the Windows target standard library was not installed in the execution environment.

### Acceptance Criteria

- Discord-backed ACP sessions advertise exactly `clientCapabilities.elicitation.form`; unsupported sessions do not advertise `clientCapabilities.elicitation` and never advertise URL elicitation.
- Agent-to-client JSON-RPC requests are classified by `method` plus ID before outbound response correlation. Numeric and string request IDs are echoed unchanged, including same-number collisions with active OpenAB requests.
- Valid flat ACP form schemas can be answered through Discord progressive rich controls or through text-only fallback commands. Defaults, validation, review, edit, Submit, Decline, and Cancel all work.
- Agent-controlled form work is bounded to one pending form per connection, 64 KiB elicitation request envelope, 50 fields, 100 choices per field, and a 1 MiB inbound reader frame.
- Only verified human senders from the exact active dispatch batch can answer. Unauthorized, wrong-channel, wrong-message, stale-generation, stale-revision, duplicate, and post-terminal inputs are harmless.
- Pool cancel, cancel-all, reset, prompt deadline, prompt completion, process EOF, writer failure, stale replacement, idle/hung eviction, and shutdown invalidate the exact connection generation before old Discord inputs can write to a child process or successor session.
- Long form content, long choices, Unicode, defaults, validation errors, and review values are reachable through bounded Discord pages without silent truncation or attachment-only substitution.
- Tests cover R1-15 complete pagination, R1-16 full text-only recovery and stale-input protection, and R1-26 deterministic wire, dispatch, pool lifecycle, overload, liveness, and exact-boundary behavior through production paths with fake external services only.
- The Discord guide, canary guide, and domain-language glossary describe the delivered behavior and do not claim URL elicitation or legacy `session/request_input` compatibility.

### Follow-ups

- Run a live Discord canary with a disposable bot and an elicitation-capable Agent Runtime before release.
- Install the Windows target and run `cargo check --target x86_64-pc-windows-gnu` if cross-platform evidence is required before merge.
- Decide whether repository-wide formatting drift should be cleaned in this PR or a separate maintenance PR.
- Consider Slack or gateway form elicitation in a separate design if users need it.

https://discord.com/channels/1491295327620169908/1491365158868619404/1549133818089316454


gandazgul:
An agent can request a form, receive its answer, and immediately request a follow-up form based on that answer.

Live Discord testing exposed a sequencing bug: 

OpenAB sent the first answer before releasing the completed form’s pending slot. While Discord was still updating the first message, the follow-up request was rejected with “elicitation already pending,” even though the first form had already been answered.

The fix releases the completed form’s slot before sending its response. Discord message cleanup happens afterward, so it cannot block the next question.

A regression test deliberately delays Discord cleanup and verifies that a follow-up form opens and returns its answer successfully.

---

New fix: Replies after form elicitation were appended to an earlier progress message, placing the answer above the form. OpenAB now sends the completed reply below the form as a new message, while preserving tool status updates in the earlier message. Applies to any ACP agent using forms.

chaodu-agent:
Thanks for the substantial work here, and for running the live Claude canary — that closed the one question that was blocking this.

We support the direction. Before we take on a change of this size, though, we want to be able to measure who actually benefits. The premise of the PR is that downstream agents will honor `clientCapabilities.elicitation.form` by sending `elicitation/create`. We spent today verifying that on the wire with a mock ACP client that advertises `elicitation: { form: {}, url: {} }` and logs every agent→client request. Results:

| Agent | Emits `elicitation/create`? | Evidence |
|---|---|---|
| Claude (`claude-agent-acp`, claude 2.1.273) | **Yes** — one `mode: "form"` request with `toolCallId`; we answered `accept` and the model read it back. Full loop closes. | live wire, independently reproduced |
| Kiro CLI (`kiro-cli acp`, OpenAB's default) | **No** — `agentCapabilities` does not advertise elicitation; forced prompts end the turn with zero reverse requests. | live wire |
| Codex (`codex-acp` 1.12.0, gpt-6-astra) | **Not observed** — the adapter *does* implement the bridge (`shouldUseAcpElicitation` → `buildElicitationRequest`, gated on `clientSupportsFormElicitation`), but across three forcing prompts the model answered in prose and never produced a form. Note also that codex's own `openai/form` mode explicitly returns `false` and is not bridged. | source read + 3 live runs |
| Gemini, OpenCode, MiMo, agy-acp, pi-acp | No emission path in current source | static |
| Kimi | Appears to have one in source | static, not run |
| Copilot, Cursor, Grok, Devin | Unknown | — |

So today: one confirmed producer (Claude), one adapter that can but whose model didn't (Codex), and the default agent that cannot (Kiro). That is a thin base for a change of this footprint. Roughly 4.5k lines of production code, most of it a Discord-only presentation layer, currently serving one backend.

We'd like to proceed in a way that captures the value without committing to the whole surface on one data point:

1. **Split out the protocol fix and merge it now.** The reader currently classifies ID-bearing frames as responses before checking `method`, so an agent→client request can collide with an OpenAB request ID and remove the wrong pending response. That is a correctness bug independent of elicitation, and it's small. Please open it as its own PR — we'll take it quickly.

2. **For the elicitation feature itself, please add live-wire evidence beyond Claude**, ideally Codex and Kimi, showing an actual `elicitation/create` on the wire when the capability is advertised. If Codex can be made to produce one (a different model tier or a natural clarifying-question task), that materially changes the ROI. If only Claude produces it, we'd want to scope and document this explicitly as Claude Code support rather than general ACP elicitation.

3. **Docs must match the evidence.** README/discord.md should name the agents verified to trigger it and state that Kiro does not, rather than describing it as generic ACP support — otherwise users on the default agent will file "form never appears" issues.

4. One presentation finding from the real payload: Claude pairs every select with a `question_<n>_custom` free-text field marked `_meta._askUserQuestionCustomAnswer`. The PR renders that as a standalone string field rather than the select's "Other" box. Not a correctness issue, but worth handling since it appears on every Claude form.

The probe we used is a small stdio JSON-RPC client; happy to share it so the same runs can be reproduced against other adapters.

gandazgul:
new PR https://github.com/openabdev/openab/pull/1540 I cant produce this evidence unfortunately, RunWield will find a different path. I'll continue using OpenAB anyways if I find other feature gaps I'll bring them up in Discord first.

PR #1540, opened 2026-09-20
fix(acp): classify reverse requests before responses

gandazgul:
## Summary

- Classify agent-to-client JSON-RPC requests by `method` and ID before outbound response correlation.
- Return JSON-RPC `-32601` for unsupported reverse request methods without removing a same-ID pending OpenAB request.
- Preserve numeric, negative, and string request IDs when replying.
- Add two focused regression tests for same-number collisions and non-`u64` request IDs.

This extracts the protocol correctness fix requested in #1533. It does not include elicitation or presentation functionality.

## Validation

- `cargo check --workspace` — passed.
- `cargo clippy --workspace -- -D warnings` — passed.
- ACP reader-loop tests — 6 passed; test count increased by 2, with no tests removed or replaced.
- `cargo test --workspace -- --skip secrets::tests::resolve_exec_nonzero_exit` — passed.
- `cargo test --workspace` — one failure in the unchanged `secrets::tests::resolve_exec_nonzero_exit`; the same test fails on a clean `upstream/main` checkout in this environment.
- `cargo fmt --all -- --check` — reports existing formatting differences; the same command fails on a clean `upstream/main` checkout.
- `git diff --check` — passed.

Discord discussion: https://discord.com/channels/1491295327620169908/1491365158868619404/1549133818089316454

### Goal

Prevent an agent-to-client JSON-RPC request from being mistaken for an OpenAB request response and consuming the wrong pending response when their IDs collide.

### Non-goals

- Add ACP elicitation support.
- Add any Discord form or presentation behavior.
- Handle new agent-to-client methods beyond the existing permission request.

### Accepted Residual Risks

Unsupported agent-to-client methods receive JSON-RPC `-32601` until OpenAB implements them. This is explicit protocol behavior and does not affect pending outbound requests.

### Acceptance Criteria

- A frame with both `method` and ID is classified as an agent-to-client request before response correlation.
- A reverse request whose numeric ID matches an active OpenAB request does not remove or resolve that pending request.
- Existing `session/request_permission` requests continue to receive their current automatic response.
- Unsupported reverse requests receive `-32601` with the original numeric, negative, or string ID.
- Ordinary numeric response frames continue to resolve and forward the matching pending response.

### Follow-ups

Elicitation support and its presentation design remain separate work from #1533 and require the live-wire evidence requested there.


gandazgul:
Addressed both round-one findings in `c600812a`:

- **F1:** `JsonRpcId` now preserves signed negative IDs, the full `u64` range, and string IDs without narrowing casts. Matched responses forward the original ID unchanged.
- **F2:** Unsupported reverse requests now use a typed `JsonRpcErrorResponse`. Permission and error responses share one serialization/write/flush helper.

Added 3 regression tests, with no tests removed or replaced:

1. Full signed/unsigned/string ID round-tripping and fractional-ID rejection.
2. A `u64::MAX` reverse-request collision followed by successful pending-response correlation and subscriber forwarding.
3. Permission responses above `i64::MAX`.

Validation:

- `cargo check --workspace` — passed.
- `cargo clippy --workspace -- -D warnings` — passed.
- Focused new regression tests — passed. Mutation checks confirmed the ID-conversion and subscriber-forwarding assertions fail on incorrect implementations.
- `cargo test --workspace` — one unchanged failure: `secrets::tests::resolve_exec_nonzero_exit`; reproduced on untouched `main`.
- `cargo fmt --all -- --check` — still reports repository-wide formatting differences; reproduced on untouched `main`.
- `git diff --check` — passed.

thepagent (approved):

PR #1542, opened 2026-09-21
feat: auto-upload agent-generated files to Slack/Discord thread

### LINKED ISSUE #1541: Agent-generated files (image_gen) are not delivered to users on Slack/Discord
## Problem

When an agent (e.g. codex-acp with GPT) uses `image_gen` to generate an image, the image is saved locally on the pod but **never reaches the user on Slack or Discord**. OpenAB streams text only via ACP -- inline images (including base64 output from image_gen) are silently dropped.

The user sees only the text reply (e.g. "已產生圖片"), with no image attached. The agent believes the image was "rendered inline" and marks the task complete.

## Expected behavior

When an agent generates a file the user requested (image, PDF, etc.), OpenAB should either:
1. Auto-detect locally generated files and upload them to the originating thread, OR
2. Provide a first-class tool/API for agents to upload files to the thread (not requiring raw `curl` + bot token)

## Current workaround

Manual Slack v2 external upload via shell `curl` + `$SLACK_BOT_TOKEN` works when the user **explicitly** asks the agent to upload. But the agent (GPT via codex-acp) does not do this automatically -- the `imagegen` system skill treats inline render as task completion.

We attempted five approaches to make the agent auto-upload via steering/instructions:
1. CRITICAL warning in steering ("inline images are not forwarded") -- did not work
2. Full curl template in steering -- did not work
3. Definition of Done rewrite ("upload to Slack = complete") -- did not work
4. `~/.codex/instructions.md` with short DoD rule -- did not work
5. User explicitly says "generate image AND upload to this thread" -- works every time

None of the passive steering approaches override the system skill's built-in "inline render = done" behavior.

## Environment

- OpenAB: `0.10.0-beta.3`
- codex-acp: `@agentclientprotocol/codex-acp` (beta.10+ adapter)
- Platform: Slack (Socket Mode)
- Cluster: GKE Autopilot
- Bot token has `files:write` scope (verified working for manual upload)
- `$SLACK_BOT_TOKEN` available to subprocess via `[agent].inherit_env`

## Suggestion

A platform-level solution would be more reliable than steering:
- Option A: OpenAB intercepts file paths in agent output and auto-uploads
- Option B: OpenAB exposes a `upload_file_to_thread` tool via ACP that agents can call natively (no shell curl needed)
- Option C: The `imagegen` skill's workflow step 14 ("render inline for preview") should be configurable -- headless deployments need a different completion path

## References

- `docs/sendfiles.md` in this repo documents the text-only limitation
- heyu-ai/openab-workspace PR #288 (files:write scope), #290 (curl flow), #291 (CRITICAL warning), #292 (DoD rewrite) -- the investigation trail

howie:
Discord Discussion URL: https://discord.com/channels/1491295327620169908/1491365158868619404/1551566947429253121
and https://discord.com/channels/1491295327620169908/1551552347090714627/1551565829219418184

Agent-generated files (images from `image_gen`, PDFs, reports) are saved locally but **never reach Slack/Discord users**. OpenAB streams text only via ACP -- inline images are silently dropped. Users see "image generated" with no image attached.

We tested five approaches to make agents auto-upload via steering/instructions -- all failed because the agent's `imagegen` system skill treats inline render as task completion, overriding any steering rules. The only reliable path is a platform-level solution.

Closes #1541

## At a Glance

```
                      Before                              After

Agent: image_gen ──► base64 inline ──► ACP      Agent: image_gen ──► file on disk
                          │                                              │
                     (dropped)                   OpenAB: scan auto_upload_dirs
                          │                              │
                     User sees:                     Upload via Slack API
                     "已產生圖片"                         │
                     (no image)                     User sees:
                                                    text + image attachment
```

- **OpenAB `docs/sendfiles.md`**: documents the text-only limitation and recommends agents call platform APIs directly. This PR extends the same principle but moves execution into OpenAB core so agents don't need to know platform-specific upload APIs.
- **Slack v2 External Upload API** (`files.getUploadURLExternal` + `files.completeUploadExternal`): the current recommended file upload flow replacing the deprecated `files.upload`. Used for the Slack implementation.
- **Discord multipart upload**: the existing documented approach for Discord file delivery. Discord implementation is left as a follow-up (default no-op).

## Proposed Solution

New opt-in config: `[agent].auto_upload_dirs` -- a list of directories OpenAB watches during ACP turns.

### Files changed

| File | Change |
|------|--------|
| `crates/openab-core/src/config.rs` | Add `auto_upload_dirs: Vec<String>` to `AgentConfig` |
| `crates/openab-core/src/adapter.rs` | Add `ChatAdapter::upload_file()` trait method (default no-op); add `AdapterRouter::with_auto_upload_dirs()`; add post-turn `scan_new_files()` + upload loop |
| `crates/openab-core/src/slack.rs` | Implement Slack v2 external upload in `SlackAdapter::upload_file()` |
| `src/main.rs` | Wire `auto_upload_dirs` from config to `AdapterRouter` |
| `docs/sendfiles.md` | Document auto-upload feature, update Slack FAQ |

### How it works

1. Before dispatching the ACP prompt, record `SystemTime::now()`.
2. After text delivery completes (all chunks sent), if `auto_upload_dirs` is non-empty, call `scan_new_files()` -- walks each directory recursively (1 level) and collects files whose mtime > turn start.
3. For each new file, call `adapter.upload_file(channel, path, title)`.
4. Upload failures are logged as warnings but non-fatal -- text delivery always completes first.

### Config example

```toml
[agent]
auto_upload_dirs = ["/home/node/.codex/generated_images"]
```

## Why This Approach

- **Platform-level, not steering-level**: Steering/instructions cannot override an agent's system skill completion behavior (verified with 5 failed experiments). The platform is the reliable interception point.
- **Opt-in**: Empty by default -- zero impact on existing deployments.
- **Best-effort**: Upload failures don't break text delivery.
- **Agent-agnostic**: Works for any agent that produces local files, not just codex's `image_gen`.

## Alternatives Considered

1. **Agent-side upload (status quo)**: Agent calls Slack API via shell curl. Works when user explicitly requests it; fails when the agent's system skill bypasses steering. Rejected as unreliable for the common case.
2. **ACP protocol extension**: Add a `File` event type to ACP. Cleaner but much larger scope -- requires coordinating with all ACP adapter implementations. Could be a future improvement.
3. **Post-processing hook in OpenAB**: Detect file paths in agent text output and upload. Rejected because agents often don't include file paths in their response text.

## Validation

- `cargo check` passes (0 errors, 0 warnings on the changed crates)
- Slack upload code follows the same `api_post` pattern as existing `chat.postMessage`
- Config deserialization is backward-compatible (`#[serde(default)]` on the new field)
- Manual verification: the same three-step Slack API flow was tested from inside a live codex pod -- `files.getUploadURLExternal` + POST binary + `files.completeUploadExternal` with `$SLACK_BOT_TOKEN` successfully uploaded a 2.2MB image to a Slack thread (PR heyu-ai/openab-workspace#288 verified `files:write` scope)

### Goal

Auto-upload agent-generated files (images, PDFs, reports) to the originating Slack/Discord thread after each ACP turn completes, so users receive the actual files instead of text-only "image generated" messages with no attachment.

### Non-goals

- Discord `upload_file()` implementation (left as follow-up; trait method defaults to no-op)
- ACP protocol-level file event extension
- Modifying agent-side `image_gen` or other system skill behavior
- Supporting nested directory scanning beyond 1 level

### Accepted Residual Risks

- Upload failures are logged as warnings but non-fatal -- a transient Slack API error means the user does not receive the file for that turn, though text delivery always completes
- mtime-based file detection could miss files if the system clock is adjusted backward during a turn (extremely unlikely in practice)
- Shallow (1-level) directory scan means deeply nested generated files are not detected

### Acceptance Criteria

- [ ] `auto_upload_dirs: Vec<String>` added to `AgentConfig` with `#[serde(default)]` (backward-compatible)
- [ ] `AdapterRouter` records turn start time and scans configured dirs after text delivery
- [ ] `scan_new_files()` collects files with mtime > turn start from each configured directory (1 level)
- [ ] `SlackAdapter::upload_file()` implements Slack v2 external upload (`getUploadURLExternal` + PUT + `completeUploadExternal`)
- [ ] Empty `auto_upload_dirs` (default) causes zero behavior change to existing deployments
- [ ] Upload failures do not block, delay, or discard text delivery
- [ ] `docs/sendfiles.md` updated with auto-upload documentation

### Follow-ups

- [ ] Implement `DiscordAdapter::upload_file()` using multipart upload
- [ ] Add configurable file size cap and per-turn upload count limit
- [ ] Consider ACP protocol-level `File` event type as a cleaner long-term solution
- [ ] Add integration test for the scan + upload pipeline


howie:
# Final Aggregated Review -- PR #1542

## Mode
Claude-only review (1/3 voices active). Codex and Gemini voices marked unavailable: both review scripts resolved CWD to openab-workspace instead of the target openab repo, reviewing stale diff from a prior PR.

## Contract compliance
- AC-1: PASS -- `auto_upload_dirs: Vec<String>` added with `#[serde(default)]`
- AC-2: PASS -- `turn_start = SystemTime::now()` recorded at adapter.rs:790
- AC-3: PARTIAL -- `scan_new_files` exists but has unbounded recursion (see Critical #1) and no tests (see Important #1)
- AC-4: PASS -- `SlackAdapter::upload_file()` implements v2 external upload flow
- AC-5: PASS -- `auto_upload_dirs` defaults to empty `Vec::new()`, guarded by `!auto_upload_dirs.is_empty()`
- AC-6: PARTIAL -- upload failures are non-fatal (warn log), but no file size guard means a large file can OOM-kill the entire process (see NIT #2)
- AC-7: PARTIAL -- new auto-upload section added but file opening still says "does not relay file attachments"
- Scope drift: none
- Residual risks: within accepted boundaries

## Consensus Critical (must fix)

1. **scan_new_files has unbounded recursion, contradicting Non-goals and docs**
   - File: `crates/openab-core/src/adapter.rs:2649-2677`
   - The function recurses into all subdirectories without a depth limit. Comment says "Recurse one level" but there is no depth parameter. The Review Contract Non-goal explicitly excludes "Supporting nested directory scanning beyond 1 level". `docs/sendfiles.md` states "recursively, one level deep".
   - Contract mapping: unaccepted risk: unbounded directory traversal that the contract explicitly excluded
   - Evidence: `scan_new_files` at line 2662 calls `Box::pin(scan_new_files(&[path], since))` on every subdirectory entry without a depth check. Given `auto_upload_dirs = ["/a"]` where `/a/b/c/d/file.txt` exists, it recurses `/a` -> `/a/b` -> `/a/b/c` -> `/a/b/c/d`.
   - Fix: Add a `depth: usize` parameter. Pass `depth=1` from the call site. At depth 0, skip subdirectories.

## Consensus Important (must fix)

1. **Zero test coverage for scan_new_files and the upload integration path**
   - File: `crates/openab-core/src/adapter.rs:2649-2677`
   - 32 lines of async filesystem logic with recursion and 34 lines of dispatch integration have no tests. A mutation removing `mtime > since` would survive all tests.
   - Contract mapping: AC-3
   - Fix: Add `#[tokio::test]` for `scan_new_files`: files before `since` excluded, files after included, depth limit respected.

2. **docs/sendfiles.md opening statement contradicts its own auto-upload section**
   - File: `docs/sendfiles.md:14,53`
   - Line 14: "does not relay file attachments" and line 53: "only streams text via ACP" both contradict the auto-upload section added at line 303+.
   - Contract mapping: AC-7
   - Fix: Amend lines 14 and 53 to mention auto-upload as an alternative path.

## Actionable NIT (deferred -- never blocks; fix opportunistically)

1. **docs/sendimages.md not updated** -- sibling doc line 14 still says "does not relay images"; same inconsistency as sendfiles.md but in an unchanged file.

2. **No file size guard before fs::read** -- `fs::read(file_path)` at slack.rs:740 loads the entire file into memory. A hardcoded `MAX_AUTO_UPLOAD_SIZE` (e.g. 50MB) would prevent OOM from unexpectedly large files. This is in the Follow-ups scope ("Add configurable file size cap") so not blocking, but a cheap defensive addition.

## Deferred for lack of evidence (not blocking)
(none)

## Outside contract (not blocking)
(none)

## Disputed (user decides)
(none)

## Voices unavailable
- Codex: review script resolved CWD to openab-workspace (session launch dir), reviewed stale diff from a prior PR; all findings about gateway.rs/media.rs are not in this PR.
- Gemini (agy): same CWD issue; reviewed stale diff about gemini-auth-config/trustedFolders.json.

howie:
@chaodu-obk Your premise check is spot-on. Let me address it directly.

**You're right — the root cause is agent-specific, not a platform gap.** The five-experiment framing in #1541 overstates the failure scope. Experiment 5 proves the agent-side upload path works perfectly — the problem is narrowly that codex's `image_gen` system skill treats inline render as task completion and bypasses any steering/instructions we put in the system prompt.

**What we actually shipped (and what's working in production right now):** a file-watcher CronJob sidecar — exactly the pattern documented in `docs/sendimages.md` "Automated Sidecar Pattern". It polls `~/.codex/generated_images/` every minute, reverse-lookups the originating thread via `thread_map.json`, and uploads via Slack v2 external upload. This solves the per-session attribution problem cleanly (each session writes to its own subdirectory keyed by session ID) and doesn't touch OpenAB core at all. Verified working in production on GKE as of 2026-09-22.

**So where does that leave this PR?** I think you've identified the right conclusion:

1. **The sidecar pattern (already documented, already working) is the correct layer** for operators who hit this today. No core changes needed.
2. **The real fix belongs upstream in the agent/skill layer** — your Option C from #1541 (making `image_gen`'s inline-render completion step configurable for headless deployments) addresses the actual root cause.
3. **This PR's `auto_upload_dirs` could still have long-term value** as a convenience for operators who don't want to run a separate sidecar — but the unresolved F1 (cross-thread race) and F2 (symlink traversal) from your round-1 review are real blockers, and fixing them properly would require per-session directory tracking that's arguably more complexity than the sidecar pattern.

**My recommendation:** close this PR, keep #1541 open as the tracking issue, and add a note pointing to the sidecar pattern in `docs/sendimages.md` as the current recommended workaround. If multiple community members report the same `image_gen` completion behavior, that strengthens the case for Option C upstream in the agent/skill.

Thank you for the thorough review — the round-1 findings (especially F1 and F2) would have been real production issues had this merged as-is. The premise check saved us from solving the problem at the wrong layer.

PR #1545, opened 2026-09-22
fix(pool): reapply default config options on resume, extend idle TTL

yuaanlin:
## Summary
- Product direction change: every runtime (managed or self-hosted) is now treated as a persistent, long-lived instance rather than an ephemeral pod. This is the first of two changes needed on the openab side.
- `SessionPool::get_or_create` always spawns a brand-new OS process, whether it takes the `session_new` or `session_load` (resume) path — but `default_config_options` (e.g. `mode = "agent-full-access"`) was only reapplied on the `session_new` branch. Any pool-level mode/model override was silently dropped whenever a session got resumed after idle-TTL eviction or pool-full suspension. This is also believed to be the root cause of the reported Codex full-access-mode-lost bug. Now reapplied on every spawn.
- Default `session_ttl_hours` raised from 4h to 30 days (720h), and the setup wizard's suggested/example default updated to match, so idle sessions aren't routinely evicted as part of normal operation.

## Not changed here
- `max_sessions` (pool-full suspension threshold, default 10) is left as-is — it's a concurrency cap, not idle ephemerality, and is out of scope for this change. Note the config-reapply fix above also covers sessions resumed via pool-full suspension, not just TTL eviction.
- The hung-session force-eviction path (`prompt_hard_timeout_secs` + `hung_grace_secs`) is untouched — that's fault tolerance for a stuck prompt, not casual ephemerality.

## Test plan
- [x] `cargo build -p openab-core`
- [x] `cargo test -p openab-core acp::pool` (16/16 pass)
- [x] `cargo test -p openab-core` (677/678 pass; the one failure, `secrets::tests::resolve_exec_nonzero_exit`, reproduces identically on a clean `main` checkout — pre-existing, unrelated to this change)


yuaanlin:
Closing — this is an internal Zeabur/Nuphos infra change, applying the same fix directly against zeabur/openab (our production fork) instead of upstream. Not meant as a public community RFC.

